PR #32, opened 2024-12-27
Hide Updates section when no updates

mdroidian:
@sid597  As we are unsure when the first update will come, I think we should hide the section until we get the first update.

## Summary by CodeRabbit

- **New Features**
	- Conditional rendering for blog updates in the "Latest Updates" section and the navigation menu.
	- Enhanced user feedback with a message when no blog updates are available.

- **Bug Fixes**
	- Improved rendering logic to prevent empty sections from displaying.

- **Chores**
	- Cleaned up code by removing unnecessary commented-out elements.

### DIFF
--- a/apps/website/app/blog/page.tsx
+++ b/apps/website/app/blog/page.tsx
@@ -12,27 +12,33 @@ export default async function BlogIndex() {
           </div>
           <div>
             <ul className="space-y-6">
-              {blogs.map((blog) => (
-                <li
-                  key={blog.slug}
-                  className="flex items-start justify-between border-b border-gray-200 pb-4 last:border-b-0"
-                >
-                  <div className="w-4/5">
-                    <Link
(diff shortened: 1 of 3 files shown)


apps/website/app/blog/page.tsx
```diff
@@ -12,27 +12,33 @@ export default async function BlogIndex() {
...
-                    </p>
-                  </div>
-                  <div className="w-1/5 text-right text-gray-600">
-                    by {blog.author}
-                  </div>
-                </li>
-              ))}
+              {blogs.length === 0 ? (
+                <p className="text-left text-lg text-gray-600">
+                  No updates yet! Check back soon. 😊
```
mdroidian, line 17: I moved this here because this page is still accessible directly and will likely get indexed by the search engines

sid597 (approved):
Agree

Issue #1, opened 2024-10-18
AI Code Reviewer


mdroidian:
Install/test [CodeRabbit](https://coderabbit.ai/)
Or look into other options

sid597:
done

PR #37, opened 2024-12-27
Roam analytics

### LINKED ISSUE #11: Add Analytics
https://posthog.com/

sid597:
Initially
This will likely evolve over time, but initially basic usage would be good:

For the following we have to configure the dashboard, based on what we are collecting in the PII we would be able to query for some and maybe not for others.

- total unique users
- total unique graphs
- daily active users
- daily active graphs

Note: we'll have to be careful with [Personally Identifiable Information](https://posthog.com/blog/what-is-personal-data-pii).

Going forward
Then for some interesting usage stats, things like:

- [x] new Discourse Node Type created (Question, Evidence, etc)
- [x] new Discourse Node created (An individual instance of a Discourse Node Type)
- [x] Discourse Relation created
- [ ] Instantiated Discourse Relation
- [x] canvas created
- [x] canvas node created
- [x] canvas relation created
- [x] discourse context viewed
- [x] query builder page created/viewed
- [x] query builder block created/fired
- [x] query drawer opened
- [x] query drawer query created/viewed
- [x] canvas drawer opened
- [x] discourse graph exported
- [x] query block results shared
- [x] query block view types used
- [x] kanban specifically

At a high level, if it's a feature we are investing time into, we should have some insight into it.

Eventually, creating user-funnels from emerging patterns could inform new features, dev paths for new plugins, ux optimizations, etc.

## Summary by CodeRabbit

- **New Features**
	- Integrated PostHog analytics into various components for enhanced user interaction tracking.
	- Added functionality to log events for actions such as creating discourse nodes, viewing queries, and rendering components.
	- Enhanced tracking for user interactions in the query drawer, canvas, and results view.
	- Improved loading state management with a new debounce hook for a smoother user experience.

- **Chores**
	- Updated project dependencies to include `posthog-js` version `^1.203.2`.


sid597:
Took too many hours to figure out the properties, and how to turn them off, turns out we can see everything and cutomise many things in the ui itself.

sid597:
added capture event for query drawer but could not test them in resultsgraph because the page is not present. Tested other events

mdroidian:
So I think the `deploy` task is failing due Repository Forks not having access to GitHub org secrets.  I believe this is intentional [[1](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow), [2](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/)].

A couple possible solutions:
1) instead of forking, clone the repo directly
2) add the secret to your forked repo (you can snag it from [Vercel](https://vercel.com/discourse-graphs/discourse-graph/settings/environment-variables))


`#1` would probably be better long term, as you wouldn't need to add more secrets over time

sid597:
will do so after this pr is done

sid597:
@mdroidian  how is "Instantiated Discourse Relation" different from "Discourse Relation created"?

sid597:
@mdroidian  Not sure about the PII, most of the analytics data does not seem without any block uid, source and target relation, new node title etc.

mdroidian:
> @mdroidian Not sure about the PII, most of the analytics data does not seem without any block uid, source and target relation, new node title etc.

Sorry, could you clarify?  I don't believe block UID, source/target or node title would be PII inheritly.  Block UID definitely not and node title possibly, depending on content, but highly unlikely.

> Examples of PII
> Again, we're leaning on the NIST here, which provides numerous examples of PII:
>
> Names, including an individual's full name, maiden name, mother's maiden name, or known aliases
> 
> Personal identification numbers, such as such a social security number (SSN), passport number, driver‘s license, taxpayer identification number, patient identification number, and financial accounts or credit card numbers
> 
> Digital identifiers, which includes but is not limited to static IP addresses, MAC addresses, and unique device identifiers (UDI)
> 
> Addresses and contact information, including home addresses, email addresses, and any type of telephone number
> 
> Personal characteristics, such as photos of an individual's face or other distinguishing characteristics, x-rays, fingerprints, and biometric data
> 
> Any other linked of linkable information, including date of birth, place of birth, race, religion, weight, employment information, medical data, education information, financial information, and most kinds of geographical identifie

mdroidian:
> @mdroidian how is "Instantiated Discourse Relation" different from "Discourse Relation created"?

"Discourse Relation created" would be when a relation is defined

image

"Instantiated Discourse Relation"

Would be when two nodes form a relation.
image

I don't think there is an efficient way to track this one right now, so I'd skip it.

apps/roam/src/index.ts
```diff
@@ -24,10 +24,35 @@ import { initObservers } from "./utils/initializeObserversAndListeners";
 import { addGraphViewNodeStyling } from "./utils/graphViewNodeStyling";
 import { setQueryPages } from "./utils/setQueryPages";
 import initializeDiscourseNodes from "./utils/initializeDiscourseNodes";
+import posthog from "posthog-js";
+
+const initPostHog = () => {
```
mdroidian, line 30: Does this take care of 

> total unique users
> total unique graphs
> daily active users
> daily active graphs

I'm guessing it inherently might take care of unique/active users.  But what about graphs?

mdroidian:
@sid597 could you share where to access this information to verify it is recording?

sid597:
For custom events we will see them under "Data management" 
image


Here you can click on any event and then you will be taken to the event page, scroll to the bottom, then you can use filter under "Matching events" to set time for last 14days etc. 


> I'm guess it inherently might take car of unique/active users. But what about graphs?

 😅  Before I was thinking graphs as in graphs of total unique users not roam graphs.  It currently does not track unique graphs but it should be as simple as adding the graphid on instantiation. Once we have the graph id then we can track daily active graphs. We would not be tracking how many lab users are using the graph only if the graph is opened by any person on the team. 
 
Re. PII, I think the goal of it is to track the usage data but the viewer of data should not be able to identify the user. Since ours is a collaborative tool so I think it extends to not be able to identify the team.


I am not sure how much of the traditional PII stuff is applicable in our context since it is going to be private beta and the goal (from my pov) is to help the labs and lab members onboard the graph smoothly. We are also looking for insights into how different lab members are using the graph, what are the common hurdles, onboarding dropoff etc. I think most of it is going to be understood by talking to the users directly but having a dashboard bases things in data. It might also be helpful for the lab manager themselves to look at the data and see for themselves if DGs are working for them.

 I think currently events tell us how many nodes, relations, query pages, canvas nodes and relations etc. were created giving us a high level view but not enough we might need ask "how much time it takes from a node being created to it being first related to another existing node", for query builder I think we should record the whole session, from its instantiation to how the user filters, see if its confusing for them, are they trying but could not complete it etc. if we do session recording then we can already see whats in the graph and can guess who the user is because we are going to onboard the labs ourselves. 
 
 I am not saying we need to look over the shoulders of every user but I think we might need to track: graph-name, username, page name (if relevant) and block uid to get a richer analytics.
 
What are the privacy T&C for labs signing up? I think based on that we can make a better decision what we should track. We can make it explicit what we are tracking and the purpose for which we will use the data.

apps/roam/src/utils/triplesToBlocks.ts
```diff
@@ -25,6 +26,12 @@ const triplesToBlocks =
     }[],
   ) =>
   () => {
+    triples.forEach((t) => {
+      posthog.capture("New Relation: Created", {
+        relation: t.relation,
+      });
+    });
+
```
sid597: @mdroidian Is this code related to "Discourse Relation created" or "Instantiated Discourse Relation"?

sid597:
Now we should see the property "Initial person info" 
image


Using this property we can create a dashboard to check for unique and active url

apps/roam/src/utils/triplesToBlocks.ts
mdroidian: Neither, specifically.

Technically it is related to "Instantiated Discourse Relation" via the Canvas.
But it is also used in the the Relation Config to create previews.
And could be used elsewhere in the future for something unrelated.

image

So this `posthog.caputre` would create false positives.

sid597: got it, can you point me to where I can find code relate to "Discourse Relation created"? I was not able to find it, there was a file related to getDiscourseRelation but could not find code related to creation, then I searched for page  and block creation then found this triplesToBlocks and from my naive pov it seemed to be the place where relations were being created.

mdroidian: In `DiscourseRelationConfigPanel.tsx`

https://github.com/DiscourseGraphs/discourse-graph/blob/cb84f3b6bd427a5af906558070e548dc50478699/apps/roam/src/components/settings/DiscourseRelationConfigPanel.tsx#L982-L1000

sid597:
Testing: 


- [x] new Discourse Node Type created (Question, Evidence, etc)
- [x]  new Discourse Node created (An individual instance of a Discourse Node Type)
- [x]  Discourse Relation created
- [x]  canvas mounted
- [x]  canvas node created
- [x]  canvas relation created
- [x]  discourse context viewed
- [x]  query builder page created/viewed
- [x]  query builder block created/fired
- [x]  query drawer opened
- [x]  query drawer query created/viewed
- [x]  canvas drawer opened
- [x]  discourse graph exported
- [x]  query block results shared
- [x]  query block view types used
- [x]  kanban specifically

See testing results under "Data management".

apps/roam/src/components/DiscourseContext.tsx
```diff
@@ -344,6 +345,9 @@ export const ContextContent = ({ uid, results }: Props) => {
             },
           },
         }));
+        posthog.capture("Discourse context: Show result", {
```
mdroidian: Let's move this to the when the caret is clicked to open.  But only on open, not close.


discourse-context

https://github.com/DiscourseGraph/discourse-graph/blob/c21cc0bdd04813bedeb809177f99f2c06c2b8f45/apps/roam/src/components/DiscourseContext.tsx#L439

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -66,6 +67,11 @@ const NodeMenu = ({ onClose, textarea }: { onClose: () => void } & Props) => {
         )}[[${pageName}]]${currentBlockText.substring(textarea.selectionEnd)}`;
 
         updateBlock({ text: newText, uid: blockUid });
+        posthog.capture("New Discourse node type: created", {
```
mdroidian: This would be the same as "New discourse node: created" in `createDiscourseNode`.  Right after this `createDiscourseNode` is called, actually.  What is happen here is a user is clicking on a Discourse Node via the Node menu:

image

If we want, we could keep this capture and name it something like "Discourse Node: Created via Node Menu".  Then in our analytics/reports we would just have to remember this would be double counted with "New discourse node: created" ... which is file I think.

### New Discourse node type: created

"New Discourse node type: created" would be referring to when a new type is created in settings:
(Let's change the naming to "Discourse Node: Type Created")

image

https://github.com/sid597/discourse-graph/blob/analytics-roam/apps/roam/src/components/settings/DiscourseNodeConfigPanel.tsx#L56-L89

apps/roam/src/components/ExportDiscourseContext.tsx
```diff
@@ -553,6 +554,11 @@ const GraphExportButton = () => {
       icon="export"
       onClick={() => {
         const { degreesIn, degreesOut } = getGraphOverviewDepthValues();
+        posthog.capture("Discourse graph: export button clicked", {
```
mdroidian: When I wrote `discourse graph exported` I didn't take into account all of the things that could mean. 😅

image
image
image

This code  is explicitly for the Graph Overview Export button, so let's clarify that and table the rest of the Export options for later.

apps/roam/src/components/QueryDrawer.tsx
```diff
@@ -362,8 +376,12 @@ const QueryDrawer = ({
   </ResizableDrawer>
 );
 
-export const openQueryDrawer = (onloadArgs: OnloadArgs) =>
-  Promise.resolve(
+export const openQueryDrawer = (onloadArgs: OnloadArgs) => {
+  posthog.capture("Query drawer: opened", {
+    source: "drawer_open",
```
mdroidian: What purpose is `source: "drawer_open"` serving?

apps/roam/src/components/QueryDrawer.tsx
```diff
@@ -374,7 +392,12 @@ export const openQueryDrawer = (onloadArgs: OnloadArgs) =>
...
   );
-export const render = (props: Props) =>
-  renderOverlay({ Overlay: QueryDrawer, props });
+};
 
-export default QueryDrawer;
+export const render = (props: Props) => {
+  posthog.capture("Query drawer: rendered", {
+    blockUid: props.blockUid,
+  });
```
mdroidian: Remove this, redundant with "Query drawer: opened"

apps/roam/src/components/ResultsView/views/Kanban.tsx
```diff
@@ -279,6 +280,8 @@ const Kanban = ({
   const [isDragging, setIsDragging] = useState(false);
   const [cursorPosition, setCursorPosition] = useState({ x: 0, y: 0 });
 
+  posthog.capture("Kanban: created", { page: page });
```
mdroidian: Let's remove this, it is a duplicate because "Query view: type changed" captured the layout change.

apps/roam/src/components/canvas/DiscourseNodeUtil.tsx
```diff
@@ -386,6 +387,12 @@ export class DiscourseNodeUtil extends TLBoxUtil<DiscourseNodeShape> {
                   });
                 else await updateBlock({ uid: shape.props.uid, text });
               }
+              posthog.capture("Canvas node: created", {
```
mdroidian: This would fire on editing a canvas node as well.

apps/roam/src/components/settings/QueryPagesPanel.tsx
```diff
@@ -46,6 +47,7 @@ const QueryPagesPanel = ({
             setTexts(newTexts);
             extensionAPI.settings.set("query-pages", newTexts);
             setValue("");
+            posthog.capture("Query builder Page: Created", { newType: value });
```
mdroidian: Just FYI, this is recording a user adding another path to match which `[[query/pages]]` will render a query page.  

I'll be renaming it to `Query Page: Page Format Added`

image

apps/roam/src/utils/triplesToBlocks.ts
```diff
@@ -25,6 +26,7 @@ const triplesToBlocks =
     }[],
   ) =>
   () => {
+
```
mdroidian: .

apps/roam/src/utils/triplesToBlocks.ts
```diff
@@ -1,5 +1,6 @@
 import type { InputTextNode } from "roamjs-components/types";
 import { Condition } from "./types";
+import posthog from "posthog-js";
```
mdroidian: .

apps/roam/src/components/DiscourseContext.tsx
mdroidian: nit: Also rename to "Discourse Context: Show Results" (Pascal Case)

mdroidian:
@sid597 my OCD required me to rename the capture events 😅

Just pull these changes in before doing the final edits.

apps/roam/src/components/ExportDiscourseContext.tsx
mdroidian: I went ahead and renamed it, I'll leave this comment unresolved so you can see it.

apps/roam/src/index.ts
mdroidian: What do you feel about adding something like this?

```javascript
posthog.capture("Extension Loaded", {
  graphName: window.roamAlphaAPI.graph.name,
  userUid: getCurrentUserUid(), // roamjs-components
}
```

sid597:
@mdroidian  ready for review

mdroidian:
image

@sid597 `shutdown()` doesn't exist in the `posthog` definition.  Is it required for frontend React?  Or just [Nodejs](https://posthog.com/docs/libraries/node#shutdown)?

I have removed it, but if it is required, let me know.

mdroidian (commented):
🔥

sid597:
good catch my bad, its not required in React only nodejs

PR #62, opened 2025-01-05
disable autocapture and pageview

### LINKED ISSUE #61: PostHog Autocapture
We need to turn off `autocapture` and `capture_pageview`.  It is on by default.

https://posthog.com/docs/product-analytics/autocapture

We don't need (or want) to track every page load, button click, text input enter/leave across the users entire time in Roam.  This is likely a privacy concern but it will also flood our account with tracked events very quickly.

sid597:
closes #61


mdroidian (approved):
👍

PR #81, opened 2025-01-16
Get user's permission for Posthog recording

sid597:


sid597:
@mdroidian Need help, currently I am stuck on how to reactively listen to a value in this case how do I listen to the settings value? 

The idea is (might not be what we want to do from product pov) on plugin install, the value for the setting is `indeterminate` so we show a pop-up to the user asking for their permission, they can say yes or no. We also have an eye icon on the topbar which reflects the user of their choice and they can use it to quickly turn on or off the analytics after their initial decision. 

When the user loads the graph next time they will see the eye icon (eye-off or eye named icons) reflecting their choice. 

What I want to do is read the value of setting reactively, so when the permission popup is open and there is an eye icon in the topbar (reflecting yes by default), and the user chooses in the popup the eye icon should reflect that but with my current code I don't know how to pass this info, talked to all the llms, cursor maybe I am asking them wrong questions.

In clojure I would have a global atom (an observable whose value I can set and then subscribe to in other places, if we change the value from anywhere the new state is shared to all the other places) which makes this straightforward to implement but there is no equivalent  in react and I am not able to find a way to do it using roam-js or even react itself.

mdroidian:
> The idea is (might not be what we want to do from product pov) on plugin install, the value for the setting is `indeterminate` so we show a pop-up to the user asking for their permission, they can say yes or no.

It might be worthwhile to delay this or make it less intrusive/blocking.  Consider this UX: most users will likely be logging in to Roam for the very first time.  It is quite overwhelming in and of itself, so adding additional intrusive popups would contribute negatively to the UX.

Some alternative options:

- include in onboarding
- instead of popup, have a floating action button that is there until they dismiss it 
(example [floating action button](https://mui.com/material-ui/react-floating-action-button/) but we would just use a blueprint icon, on click load dialog/drawer/popup, on explicit dismissal remove floating action button)
- delay the popup `x` time/log in/pageview/etc

While it would likely lower adoption to users who opt-in, it is still worth considering.

> We also have an eye icon on the topbar which reflects the user of their choice and they can use it to quickly turn on or off the analytics after their initial decision.

I like the idea, but maybe we wait for it to be requested?  Not sure it is necessary for the initial release or for the the pilot graphs at all.

> When the user loads the graph next time they will see the eye icon (eye-off or eye named icons) reflecting their choice.

I'd be happy with this just being a regular setting in the main page of our settings/config panel.

> What I want to do is read the value of setting reactively, so when the permission popup is open and there is an eye icon in the topbar (reflecting yes by default), and the user chooses in the popup the eye icon should reflect that but with my current code I don't know how to pass this info, talked to all the llms, cursor maybe I am asking them wrong questions.

I don't think this needs to be reactive.  For our use case it can be simple:

on plugin load
- read roam setting before posthog init
- set posthog init accordingly

on setting change
- update setting
- call `posthog.startSessionRecording()` / `posthog.stopSessionRecording()`

sid597:
@mdroidian  I think we should get the access as soon as possible, many steps would already be done by the onboarding time and I think it would be less overwhelming at the start of roam instead of while onboarding e.g load roam for the first time, popup, yes/no vs load roam, read onboarding, open settings, click dg, find the setting, yes/no

I also don't like the popup approach for ux in general but for this case I think this might not be a bad ux if we are helping with the initial onboarding.

Will remove the eye button, but how can we theoretically add reactivity for such cases in react?

mdroidian:
In React you could use [useContext](https://react.dev/reference/react/useContext).  In Javascript you could use a [Custom Event](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events).

sid597:
I see, reading up the docs and then llm cleared up my understanding, in React we can only update components that are part of a tree, I was thinking we could update components not part of the same tree. In clojure reactivity is much simpler e.g 

```clojure
;; In reagent clojure
(def count-atom (r/atom 0))                    ; Create global state
(defn display [] [:div @count-atom])           ; Any component can read
(defn button [] [:button {:on-click #(swap! count-atom inc)} "+"]) ; Any component can write
```

```typescript
// In react
const CountContext = createContext(null);                    // Create context
const CountProvider = ({children}) => {                      // Wrap tree with provider
  const [count, setCount] = useState(0);
  return {children};
};
const Display = () => useContext(CountContext).count;        // Must be under provider to read
const Button = () => <button onClick={() => useContext(CountContext).setCount(c => c + 1)}>+;
```

👍🏽

sid597:
how is it turning off in a week?
what happens when a user toggles off?
does it stop?
looks like init fires again
what happens if they toggle on again?
looks like init fires again
why intermediate if true?
no need for intermediate at all
I didn't get the popup because initial value isn't "indeterminate"
popup is ambiguous as to is doing the recording

sid597:
buttons to primary 
for a normal user? no, then how do we determine? whitelist by graph name because we would know the lab's graph name but its manual,

mdroidian:
Closing as it looks like we are no longer going down this route.

PR #85, opened 2025-02-11
Add team members to web

trangdoan982:
## Summary by CodeRabbit

- **New Features**
	- Added a “Team” navigation link and a dedicated section on the homepage that displays team member profiles with images, titles, and social links.
- **Style**
	- Updated modal styling with a red background for enhanced visual distinction.


trangdoan982:
@coderabbitai full review

apps/obsidian/.gitignore
```diff
@@ -0,0 +1,2 @@
+plugin
+main.js
```
mdroidian: I'm guessing this was from the first `.env` folder setting?  Let's remove it if it is no longer necessary.

apps/obsidian/styles.css
```diff
@@ -0,0 +1,3 @@
+.modal-content {
+  background-color: red;
+}
```
mdroidian: I'm guessing this was from the first `.env` folder setting?  Let's remove it if it is no longer necessary.

apps/website/app/components/TeamPerson.tsx
```diff
@@ -0,0 +1,81 @@
...
+          className="object-cover"
+        />
+      </div>
+      <div>
+        <h3 className="text-xl font-semibold text-neutral-dark">
+          {member.name}
+        </h3>
+        <p className="text-neutral-dark/80">{member.title}</p>
+      </div>
+      {member.links && (
```
mdroidian, line 33: Links is a nice touch in case we want to add it later.  I'm fine with leaving it, but the links to the images are broken.
We should either fix the links and add the images, leave a comment noting it is for future use, or remove it altogether.

We'll bring it up in the next all hands to see what the team thinks of adding them:
https://roamresearch.com/#/app/discourse-graphs/page/ENJzRTJ0m

mdroidian (changes requested):
Nice work!

A couple minor changes, then we should be good to go.

apps/website/app/data/constants.ts
```diff
@@ -1,5 +1,45 @@
...
+    title: "Developer",
+    image: "/team/trang.jpg",
+  },
+  {
+    name: "David Vargas",
+    title: "Advisor",
+    image: "/team/david.png",
+  },
+  {
+    name: "John Morabito",
```
mdroidian: let's move John between "Michael Gartner" and "Siddharth Yadav"

apps/obsidian/.gitignore
trangdoan982: oh these files come from turbo dev. maybe we can just commit them once?

mdroidian: That is interesting.  I am not replicating this on my end.  Could you delete them and try running `turbo dev` again, this time with the updated Obsidian `.env` file?  If they are still getting created we can look into the issue further at tomorrow's standup.

Regardless, let's remove them from this PR.

apps/website/app/components/TeamPerson.tsx
trangdoan982: i fixed the social image links

apps/obsidian/styles.css
trangdoan982: same as above. it's auto-generated from turbo dev. i can uncommit it but this will probably pop up everytime we run turbo dev, so i think we might as well just commit once and not have to deal with it locally anymore

mdroidian: Although this may be happening in your local environment, `styles.css` in the Obsidian root, `main.js`, and an additional `/plugin` folder are not our desired state now or going forward.

Instead of adding these files to repo to bypass the problem, I suggest we find and address the root cause of the problem. 👍  We can do that together at tomorrow's standup.

mdroidian (approved):
🔥

PR #93, opened 2025-02-25
Enable users to send feedback

sid597:
image


apps/roam/package.json
```diff
@@ -25,6 +25,10 @@
     "widgets"
   ],
   "dependencies": {
+    "@emotion/react": "^11.14.0",
+    "@emotion/styled": "^11.14.0",
+    "@mui/icons-material": "^6.4.5",
+    "@mui/material": "^6.4.5",
```
mdroidian: @sid597 Sorry, I should have been clear.  The link to the FAB was just as an example of the idea of what we wanted.  Let's stick to BlueprintJS for Roam and not import any new UI libraries.

mdroidian (changes requested):
🔥
as per our conversation, lets
- use type instead of interface
- nix the comments
- rename the component/function to be more descriptive of what it does instead of the provider
- fix the shadow css issue when moving the component

Additions:
- add reference to Discourse Graphs so the user knows who the form goes to
- add dropdown with options (if we can customize the text)

sid597:
@mdroidian re: add reference to Discourse Graphs so the user knows who the form goes to

This is not something I am able to do, the docs say it has to be set from the dashboard and when I try to go there I get the permissions error. 

> 
> There are also two input fields that are configured from the server side:
> 
> - The collection to which the session will be uploaded
> -  Labels

> These inputs are configured in the [SDK settings of the web app](https://app.birdeatsbug.com/workspaces/settings/sdk): Inputs to choose a selection or labels will be displayed automatically in the SDK, if at least two collections or labels are marked as selectable in the web app. Both inputs are optional. The default collection for SDK sessions is also configured in the web app.

https://docs.birdeatsbug.com/latest/sdk/customization.html#customize-the-form

mdroidian (approved):
Good stuff.  I added a couple of changes/additions.
Also, remember to format with prettier 👍

PR #99, opened 2025-03-05
[ENG-42] Create setting to define a relationship

trangdoan982:
https://linear.app/discourse-graphs/issue/ENG-42/create-setting-to-define-a-relationship

## Node types
Screenshot 2025-03-06 at 17 05 01
- valid format: no name or format duplications, has to have {content} in the format, no [[ or ]] in format
- can't delete if there's a Discourse Relation using this Node Type

## Relation Type
image
- check for duplication
- can't delete if there's a Discourse Relation using this Relation Type

## Discourse relation definition
Screenshot 2025-03-06 at 17 05 44
- because we use id, it's always synced with Node types and Relation Types
- check for duplicate

Confirm before delete:
image
Screenshot 2025-03-17 at 13 38 41
Screenshot 2025-03-17 at 13 38 52


trangdoan982:
@coderabbitai review

apps/obsidian/src/components/PluginContext.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { createContext, useContext, ReactNode } from "react";
```
mdroidian: . React, unused

apps/obsidian/src/components/RelationshipSettings.tsx
```diff
@@ -0,0 +1,296 @@
+import { useState } from "react";
+import {
+  DiscourseRelation,
+  DiscourseNode,
+  DiscourseRelationType,
+} from "../types";
```
mdroidian: Let's use the `~` convention instead of `..` for imports to reference the src directory consistently.

apps/obsidian/src/components/RelationshipSettings.tsx
```diff
@@ -0,0 +1,296 @@
...
+  };
+
+  return (
+    <div className="discourse-relations">
+      <h3>Node Type Relations</h3>
+
+      {plugin.settings.nodeTypes.length === 0 && (
+        <div className="setting-item">
+          <div className="setting-item-info">
+            <div className="setting-item-description">
```
mdroidian: is there a need for all these `div`s?

apps/obsidian/src/components/RelationshipSettings.tsx
```diff
@@ -0,0 +1,296 @@
...
+              You need to create some node types first.
+            </div>
+          </div>
+        </div>
+      )}
+
+      {plugin.settings.relationTypes.length === 0 && (
+        <div className="setting-item">
+          <div className="setting-item-info">
+            <div className="setting-item-description">
```
mdroidian: is there a need for all these `div`s?

apps/obsidian/src/components/RelationshipSettings.tsx
```diff
@@ -0,0 +1,296 @@
...
+    await plugin.saveSettings();
+    new Notice("Relations saved");
+    setHasUnsavedChanges(false);
+  };
+
+  return (
+    <div className="discourse-relations">
+      <h3>Node Type Relations</h3>
+
+      {plugin.settings.nodeTypes.length === 0 && (
```
mdroidian: Let's combine these with a ternary so we don't get both.
image.png

apps/obsidian/src/components/RelationshipTypeSettings.tsx
```diff
@@ -0,0 +1,147 @@
+import { useState } from "react";
+import { DiscourseRelationType } from "../types";
```
mdroidian: `~`

apps/obsidian/src/components/RelationshipTypeSettings.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { Notice } from "obsidian";
+import { usePlugin } from "./PluginContext";
+import generateUid from "../utils/generateUid";
```
mdroidian: `~`

apps/obsidian/src/components/Settings.tsx
```diff
@@ -1,181 +1,92 @@
...
-    if (hasErrors) {
-      return;
-    }
-
-    plugin.settings.nodeTypes = nodeTypes;
-    await plugin.saveSettings();
-    setHasUnsavedChanges(false);
-  };
+const Settings = () => {
+  const plugin = usePlugin();
```
mdroidian: . not used

apps/obsidian/src/utils/generateUid.ts
```diff
@@ -0,0 +1,5 @@
+const generateUid = (prefix = "dg") => {
+  return `${prefix}_${Date.now()}_${Math.random().toString(36).substring(2, 9)}`;
```
mdroidian: lets import and use nanoid for this

apps/obsidian/src/utils/validateNodeType.ts
```diff
@@ -0,0 +1,100 @@
+import { DiscourseNode } from "~/types";
+
+export function validateNodeFormat(
```
mdroidian, line 3: change to arrow function for consistency

apps/obsidian/src/components/RelationshipSettings.tsx
```diff
@@ -0,0 +1,296 @@
...
+        sourceId: "",
+        destinationId: "",
+        relationshipTypeId: "",
+      },
+    ];
+    setDiscourseRelations(updatedRelations);
+    setHasUnsavedChanges(true);
+  };
+
+  const handleDeleteRelation = async (index: number): Promise<void> => {
```
mdroidian, line 92: nit - we can address this later
since we are using the unsaved changes pattern, we should probably also get confirmation before deleting

apps/obsidian/src/components/Settings.tsx
```diff
@@ -1,181 +1,92 @@
...
+  const [activeTab, setActiveTab] = useState("nodeTypes");
```
mdroidian, line 12: nit - we can address this later
since we are using the unsaved changes pattern, we should probably also warn before tab switch if unsaved changes

mdroidian (changes requested):
🔥 Good stuff, mostly just nits.

apps/obsidian/src/components/RelationshipSettings.tsx
mdroidian: image

PR #104, opened 2025-03-19
moved canvas setting to global setting

sid597:
## Summary by CodeRabbit

- **New Features**
  - Introduced a "Canvas Page Format" configuration option, allowing users to specify the format for canvas pages with clear titles, descriptions, and default values.
  
- **Refactor**
  - Streamlined user settings by removing the previous "Canvas Page Format" controls from personal settings and consolidating the configuration in a dedicated panel.

### DIFF
--- a/apps/roam/src/components/Export.tsx
+++ b/apps/roam/src/components/Export.tsx
@@ -153,7 +153,6 @@ const ExportDialog: ExportDialogComponent = ({
   const [selectedPageUid, setSelectedPageUid] = useState(currentPageUid);
   const isCanvasPage = checkIfCanvasPage({
     title: selectedPageTitle,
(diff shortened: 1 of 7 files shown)


sid597:
@coderabbitai review

mdroidian:
@sid597 The implementation of this setting has not been changed, so the places where the feature is used will not work (they still use the old setting):

https://www.loom.com/share/38e11b68316e44fab63c17063f554054

Also, when you remove a function call or variable, be sure to take a look at where else it might be being used, or where it was declared.  If it is no longer being used, remove the declaration.  In VSCode/Cursor, it should be slightly transparent so it is easier to notice. 

image

 I'll encode this into our typescript config so that it gives lint warnings.

mdroidian (commented):
👍

PR #108, opened 2025-03-23
Roam: Add zoom-path column view option

mdroidian:
image

temp draft to allow users to view/test
`view.value` shape has to change to support multiple options


mdroidian:
https://linear.app/discourse-graphs/issue/ENG-130/add-zoom-path-option-to-query-builder-column-views backlogged, closing this until it is in scope.

PR #115, opened 2025-03-28
ENG-96 Create new relationship between nodes

trangdoan982:
- because there's no datacore API yet, we're using obsidian-dataview
image

Screenshot 2025-03-28 at 13 51 02
Screenshot 2025-03-28 at 13 51 17
image

## Summary by CodeRabbit

- **New Features**
  - Introduced a dedicated relationships section to visually manage connections between nodes.
  - Added an interactive interface for creating and editing relationships, offering guided selections and notifications.
  - Launched a customizable search bar with enhanced fuzzy matching for faster, more accurate results.

- **Chores**
  - Improved the plugin shutdown process to ensure views are cleanly removed, enhancing overall stability.

### DIFF
--- a/apps/obsidian/src/components/DropdownSelect.tsx
+++ b/apps/obsidian/src/components/DropdownSelect.tsx
@@ -1,0 +1,85 @@
+import { App, DropdownComponent } from "obsidian";
+import React, { useEffect, useRef, useState } from "react";
+
+interface DropdownSelectProps<T> {
+  options: T[];
+  onSelect: (item: T | null) => void;
+  placeholder?: string;
+  app: App;
+  getItemText: (item: T) => string;
+  renderItem?: (item: T, el: HTMLElement) => void;
+}
+
+const DropdownSelect = <T,>({
+  options,
+  onSelect,
+  placeholder = "Select...",
+  app,
+  getItemText,
+  renderItem,
+}: DropdownSelectProps<T>) => {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const dropdownRef = useRef<DropdownComponent | null>(null);
+
+  useEffect(() => {
+    if (!containerRef.current) return;
+
+    if (!dropdownRef.current) {
+      dropdownRef.current = new DropdownComponent(containerRef.current);
+    }
+
+    const dropdown = dropdownRef.current;
+    const currentValue = dropdown.getValue();
+
+    dropdown.selectEl.empty();
+
+    dropdown.addOption("", placeholder);
+
+    options.forEach((option) => {
+      const text = getItemText(option);
+      dropdown.addOption(text, text);
+    });
+
+    if (
+      currentValue &&
+      options.some((opt) => getItemText(opt) === currentValue)
+    ) {
+      dropdown.setValue(currentValue);
+    }
+
+    const onChangeHandler = (value: string) => {
+      const selectedOption =
+        options.find((opt) => getItemText(opt) === value) || null;
+      dropdown.setValue(value); // Set the dropdown's displayed value
+      onSelect(selectedOption);
+    };
+
+    dropdown.onChange(onChangeHandler);
+
+    return () => {
+      dropdown.onChange(() => {});
+    };
+  }, [options, onSelect, getItemText, placeholder]);
+
+  useEffect(() => {
+    return () => {
+      if (dropdownRef.current) {
+        dropdownRef.current.selectEl.empty();
+        dropdownRef.current = null;
+      }
+    };
+  }, []);
+
+  return (
+    <div
+      ref={containerRef}
+      className="dropdown-select"
+      style={{
+        width: "100%",
+        position: "relative",
+      }}
+    />
+  );
+};
+
+export default DropdownSelect;
--- a/apps/obsidian/src/components/DropdownSelect.tsx
+++ b/apps/obsidian/src/components/DropdownSelect.tsx
@@ -1,22 +1,18 @@
-import { App, DropdownComponent } from "obsidian";
+import { DropdownComponent } from "obsidian";
 import { useEffect, useRef } from "react";
 
 type DropdownSelectProps<T> = {
   options: T[];
   onSelect: (item: T | null) => void;
   placeholder?: string;
-  app: App;
   getItemText: (item: T) => string;
-  renderItem?: (item: T, el: HTMLElement) => void;
 };
 
 const DropdownSelect = <T,>({
   options,
   onSelect,
   placeholder = "Select...",
-  app,
   getItemText,
-  renderItem,
 }: DropdownSelectProps<T>) => {
   const containerRef = useRef<HTMLDivElement>(null);
   const dropdownRef = useRef<DropdownComponent | null>(null);
--- a/apps/obsidian/src/components/DropdownSelect.tsx
+++ b/apps/obsidian/src/components/DropdownSelect.tsx
@@ -1,14 +1,14 @@
 import { App, DropdownComponent } from "obsidian";
-import React, { useEffect, useRef, useState } from "react";
+import { useEffect, useRef } from "react";
 
-interface DropdownSelectProps<T> {
+type DropdownSelectProps<T> = {
   options: T[];
   onSelect: (item: T | null) => void;
   placeholder?: string;
   app: App;
   getItemText: (item: T) => string;
   renderItem?: (item: T, el: HTMLElement) => void;
-}
+};
 
 const DropdownSelect = <T,>({
   options,
@@ -50,11 +50,15 @@ const DropdownSelect = <T,>({
     const onChangeHandler = (value: string) => {
       const selectedOption =
         options.find((opt) => getItemText(opt) === value) || null;
-      dropdown.setValue(value); // Set the dropdown's displayed value
+      dropdown.setValue(value);
       onSelect(selectedOption);
     };
 
     dropdown.onChange(onChangeHandler);
+
+    if (options && options.length === 1 && !currentValue) {
+      dropdown.setValue(getItemText(options[0] as T));
+    }
 
     return () => {
       dropdown.onChange(() => {});
--- a/apps/obsidian/src/components/RelationshipSection.tsx
+++ b/apps/obsidian/src/components/RelationshipSection.tsx
@@ -1,10 +1,10 @@
 import { TFile, Notice } from "obsidian";
-import { useState, useRef, useEffect } from "react";
-import DiscourseGraphPlugin from "~/index";
+import { useState, useRef, useEffect, useCallback, useMemo } from "react";
 import { QueryEngine } from "~/services/QueryEngine";
 import SearchBar from "./SearchBar";
 import { DiscourseNode } from "~/types";
 import DropdownSelect from "./DropdownSelect";
+import { usePlugin } from "./PluginContext";
 
 type RelationTypeOption = {
   id: string;
@@ -13,11 +13,12 @@ type RelationTypeOption = {
 };
 
 type RelationshipSectionProps = {
-  plugin: DiscourseGraphPlugin;
   activeFile: TFile;
 };
 
-const AddRelationship = ({ plugin, activeFile }: RelationshipSectionProps) => {
+const AddRelationship = ({ activeFile }: RelationshipSectionProps) => {
+  const plugin = usePlugin();
+
   const [selectedRelationType, setSelectedRelationType] = useState<string>("");
   const [selectedNode, setSelectedNode] = useState<TFile | null>(null);
   const [isAddingRelation, setIsAddingRelation] = useState(false);
@@ -70,7 +71,7 @@ const AddRelationship = ({ plugin, activeFile }: RelationshipSectionProps) => {
     setCompatibleNodeTypes(compatibleNodeTypes);
   }, [selectedRelationType, activeNodeTypeId, plugin.settings]);
 
-  const getAvailableRelationTypes = () => {
+  const getAvailableRelationTypes = useCallback(() => {
     if (!activeNodeTypeId) return [];
 
     const options: RelationTypeOption[] = [];
@@ -104,9 +105,12 @@ const AddRelationship = ({ plugin, activeFile }: RelationshipSectionProps) => {
     });
 
     return options;
-  };
+  }, [activeNodeTypeId, plugin.settings]);
 
-  const availableRelationTypes = getAvailableRelationTypes();
+  const availableRelationTypes = useMemo(
+    () => getAvailableRelationTypes(),
+    [getAvailableRelationTypes],
+  );
 
   // Auto-select the relation type if there's only one option
   useEffect(() => {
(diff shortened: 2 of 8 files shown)


trangdoan982:
@coderabbitai review

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -0,0 +1,79 @@
...
+export type SearchNodeOptions = {
+  excludeFile?: TFile;
+  minQueryLength?: number;
+};
+
+export class QueryEngine {
+  private dv: any;
+
+  constructor(app: App) {
+    this.dv = getAPI(app);
```
mdroidian, line 13: `this.dv` isn't initializing for me.
image

My suggestions would be to change this to a utility function (`utils/queryEngine.ts` or individual `utils/searchNodeByTitle.ts`).  Unless you have a reason that a class would be better.

mdroidian, line 13: `this.dv` isn't initializing for me.

image

This could be because we are trying to initialize it on every render of `RelationshipSection.tsx`, and then again on every render of `SearchBar`.

My suggestions would be to change this to a utility function (`utils/queryEngine.ts` or individual `utils/searchNodeByTitle.ts`). Unless you have a reason that a class would be better.

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,231 @@
...
+        });
+      }
+    });
+
+    return options;
+  };
+
+  const availableRelationTypes = getAvailableRelationTypes();
+
+  const searchNodes = async (query: string): Promise<TFile[]> => {
```
mdroidian, line 70: This is being created here just to be passed to the `SearchBar`.  Let's move this to `SearchBar`.
Either pass the `activeFile` or just grab it with the `app` object.

mdroidian: This is being created here just to be passed to the `SearchBar`. Let's move this to `SearchBar`.
Either pass the `activeFile` or (preferably) grab it with the `app` object.

apps/obsidian/src/services/QueryEngine.ts
mdroidian: this.dv isn't initializing for me.

image


This could be because we are trying to initialize it on every render of `RelationshipSection.tsx`, and then again on every render of `SearchBar`.

My suggestions would be to change this to a utility function (`utils/queryEngine.ts` or individual `utils/searchNodeByTitle.ts`). Unless you have a reason that a class would be better.

apps/obsidian/src/components/SearchBar.tsx
```diff
@@ -0,0 +1,224 @@
+import { App, debounce } from "obsidian";
+import React, { useEffect, useState, useMemo, ReactNode } from "react";
+import { QueryEngine } from "~/services/QueryEngine";
+
+type SearchBarBaseProps<T> = {
+  app: App;
+  placeholder?: string;
+  onNodeSelect: (item: T | null) => void;
+  debounceMs?: number;
+  minQueryLength?: number;
```
mdroidian: These aren't being customized by any implementations so let's just set them in the function and remove them from props.

Let's add additional props only when required.

apps/obsidian/src/components/SearchBar.tsx
```diff
@@ -0,0 +1,224 @@
...
+type FunctionSearchBarProps<T> = SearchBarBaseProps<T> & {
+  options?: never;
+  searchFunction: (query: string) => Promise<T[]>;
+};
+
+export type SearchBarProps<T> =
+  | OptionsSearchBarProps<T>
+  | FunctionSearchBarProps<T>;
+
+export function SearchBar<T extends {}>({
```
mdroidian: prefer arrow function for codebase consistency

apps/obsidian/src/components/SearchBar.tsx
```diff
@@ -0,0 +1,224 @@
...
+  app,
```
mdroidian: Remove.  Grab from `appContext`.

apps/obsidian/src/components/SearchBar.tsx
```diff
@@ -0,0 +1,224 @@
...
+  onNodeSelect: (item: T | null) => void;
+  debounceMs?: number;
+  minQueryLength?: number;
+  getItemText: (item: T) => string;
+  getItemKey: (item: T) => string;
+  renderItemContent?: (item: T, isSelected?: boolean) => ReactNode;
+  itemStyle?: React.CSSProperties;
+  containerStyle?: React.CSSProperties;
+  inputStyle?: React.CSSProperties;
+  resultsContainerStyle?: React.CSSProperties;
```
mdroidian: Remove.  Same as above.  It could be the case these are never used, for example.

apps/obsidian/src/components/SearchBar.tsx
mdroidian: Are there any Obsidian native classes we can take advantage of here instead of creating our own?  Maybe something like `PopoverSuggest` or `AbstractInputSuggest` or similar?

mdroidian (changes requested):
I couldn't get it to work on my end (noted below) so didn't do a complete review.  I think we should move `QueryEngine` to a util function and look at it again after.  Also, we should be looking for Obsidian native function/class/components when possible (for `SearchBar` in this case).  The latter we can discuss in our next meeting.

apps/obsidian/src/services/QueryEngine.ts
trangdoan982: oh you do need to install dataview for this to work. that's the limitation right now because we don't have the independent datacore npm yet. this is the only way we can build on top of dataview/datacore right now.

my argument for using class is that in the future we will build many search functions. they all use the same dv/ dc instance instead of having to import from different utils files. but let me know if you think we'll still be better off using utils file

apps/obsidian/src/components/RelationshipSection.tsx
trangdoan982: these availableRelationTypes options are specific to the Relationship Type search (we only give them relations that are defined from plugin.settings.discourseRelations) and not all relations

apps/obsidian/src/components/SearchBar.tsx
trangdoan982: yes, updated the SearchBar using AbastractInputSuggest

trangdoan982:
@mdroidian fix a few issues:
- the search now only shows nodes compatible types (defined from the discourseRelations)
- i also added a small UI to show users what node types are compatible with the active file's node type
- fixed a few display issues

trangdoan982:
@mdroidian i fixed a few issues since you last review:
- search allows only compatible node types, defined in the discourseRelations now
- i added a little UI signal to show them what node types are compatible
- fix a few display issues
Screenshot 2025-03-31 at 15 17 38

apps/obsidian/src/services/QueryEngine.ts
mdroidian: It is likely the current function/methods will be temporary for v0.  We should be able to support arbitrary queries in the near future, so we'll moving towards a query compiler.  Then when we get to inter-graph functionality we'll need a query translator.  For v0, util functions would have less complexity than classes, which is why I would recommend them.  Plus I personally prefer util functions over classes 😁,

That being said, it is 100% your call either way.  Feel free to continue using `Class`.

I'll explain in more detail and give examples from the roam codebase on our next sync.

mdroidian: > this is the only way we can build on top of dataview/datacore right now.

datacore has a `window.datacore` object, so this might be an option as well.  but the user would require installing datacore via BRAT.  pros and cons, we can discuss this as well at our next sync.

apps/obsidian/src/components/RelationshipSection.tsx
mdroidian: Sorry, I should have been more specific.  Just referring to ln 70, `searchNodes`

trangdoan982: yeah same thing. i can probably name this better. but this searchNodes function is specific to the SearchBar for the nodes to link with the activeFile. but we also use SearchBar for Relationship Type. not sure if this is what you mean? We can elaborate more in the meeting

mdroidian: I see.  One is use `asyncSearch` and the other is using `options`.  This looks like it is two separate components.  A Select/Dropdown (with optional autocomplete if you like) for the Relations and a Search/Autocomplete Input.

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -0,0 +1,119 @@
...
+  private app: App;
+  private dc: any;
+
+  constructor(app: App) {
+    const appWithPlugins = app as AppWithPlugins;
+    this.dc = appWithPlugins.plugins?.plugins?.["datacore"]?.api;
+    this.app = app;
+  }
+
+  async searchCompatibleNodeByTitle(
```
mdroidian, line 26: This method is only used in one place, so having `minQueryLength` as a param is unnecessary.  Same with having `excludeFile` as optional.

Unless you already know where this will be used again where we will not exclude the active file (or other file(s)) and where the minQueryLength won't be 2.  If so, ignore this.

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -0,0 +1,119 @@
...
+    query: string,
+    minQueryLength: number = 2,
+    compatibleNodeTypeIds: string[],
+    excludeFile?: TFile,
```
mdroidian: If more than 2-3 params, use named params.  They are more flexible, scalable, and self documenting.

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -0,0 +1,119 @@
+import { TFile, App } from "obsidian";
+import { getAPI } from "obsidian-dataview";
+
+interface AppWithPlugins extends App {
```
mdroidian: type

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -0,0 +1,119 @@
...
+  plugins: {
+    plugins: {
+      [key: string]: {
+        api: any;
+      };
+    };
+  };
+}
```
mdroidian: Is this temporary? Stated differently, will this in the official docs/types?
Or could this be removed?

This would be a good place to make a comment explaining why we are adding type definitions to the Obsidian API.

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -0,0 +1,119 @@
...
+    minQueryLength: number = 2,
+    compatibleNodeTypeIds: string[],
+    excludeFile?: TFile,
+  ): Promise<TFile[]> {
+    if (!query || query.length < minQueryLength) {
+      return [];
+    }
+    if (!this.dc) {
+      console.warn(
+        "Datacore API not available. Search functionality is limited.",
```
mdroidian: nit: technically there will be no searching in this situation.

apps/obsidian/src/components/DropdownSelect.tsx
```diff
@@ -0,0 +1,85 @@
+import { App, DropdownComponent } from "obsidian";
+import React, { useEffect, useRef, useState } from "react";
+
+interface DropdownSelectProps<T> {
```
mdroidian: type

apps/obsidian/src/components/DropdownSelect.tsx
```diff
@@ -0,0 +1,85 @@
...
+  options: T[];
+  onSelect: (item: T | null) => void;
+  placeholder?: string;
+  app: App;
```
mdroidian: unused

apps/obsidian/src/components/DropdownSelect.tsx
```diff
@@ -0,0 +1,85 @@
...
+  getItemText: (item: T) => string;
+  renderItem?: (item: T, el: HTMLElement) => void;
```
mdroidian: unused

apps/obsidian/src/components/DropdownSelect.tsx
```diff
@@ -0,0 +1,85 @@
...
+      currentValue &&
+      options.some((opt) => getItemText(opt) === currentValue)
+    ) {
+      dropdown.setValue(currentValue);
+    }
+
+    const onChangeHandler = (value: string) => {
+      const selectedOption =
+        options.find((opt) => getItemText(opt) === value) || null;
+      dropdown.setValue(value); // Set the dropdown's displayed value
```
mdroidian: .

apps/obsidian/src/components/DropdownSelect.tsx
```diff
@@ -0,0 +1,85 @@
+import { App, DropdownComponent } from "obsidian";
+import React, { useEffect, useRef, useState } from "react";
```
mdroidian: remove unused imports

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,396 @@
...
+        className="relationship-type-selector"
+        style={{ marginBottom: "1rem" }}
+      >
+        <label style={{ display: "block", marginBottom: "0.5rem" }}>
+          Relationship Type:
+        </label>
+        <DropdownSelect<RelationTypeOption>
+          options={availableRelationTypes}
+          onSelect={(option) => option && setSelectedRelationType(option.id)}
+          placeholder="Search available relationship types..."
```
mdroidian: nit: it is no longer a search, and we could keep it simple "Select relation" or "Click to Select" or "Choose" etc

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,396 @@
...
+    try {
+      if (!activeNodeTypeId) {
+        setSearchError("Active file does not have a node type");
+        return [];
+      }
+
+      const relationTypeIds = availableRelationTypes.map((option) => option.id);
+      const allCompatibleNodeTypeIds: string[] = [];
+
+      const filteredRelations = plugin.settings.discourseRelations.filter(
```
mdroidian: We should also filter existing relations.  This can be considered out of scope and a separate task.
https://linear.app/discourse-graphs/issue/ENG-198/filter-existing-relations
image

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,396 @@
...
+          </div>
+        </div>
+      )}
+
+      <div style={{ marginBottom: "1rem" }}>
+        <label style={{ display: "block", marginBottom: "0.5rem" }}>
+          Node to link with:
+        </label>
+        <SearchBar<TFile>
+          app={plugin.app}
```
mdroidian: nit: instead of passing `plugin` and `app` all the way down, we could take advantage of `pluginContext` or `appContext`

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,396 @@
...
+          asyncSearch={searchNodes}
+          onSelect={setSelectedNode}
+          placeholder="Search nodes (type at least 2 characters)..."
+          getItemText={(node) => node.basename}
+          renderItem={renderNodeItem}
+          minQueryLength={2}
```
mdroidian: `minQueryLength` is set in 3 different places (`searchNodes`, `placeholder`, and here)
If one were to come back to this, they'd have to remember to change it in 3 places.

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,396 @@
...
+          (type) => type.id === id,
+        );
+        return nodeType;
+      })
+      .filter(Boolean) as DiscourseNode[];
+
+    setCompatibleNodeTypes(compatibleNodeTypes);
+  }, [selectedRelationType, activeNodeTypeId, plugin.settings]);
+
+  const getAvailableRelationTypes = () => {
```
mdroidian: useMemo

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,396 @@
...
+
+    suggestionEl.createEl("div", {
+      text: "📄",
+      attr: { style: "margin-right: 8px;" },
+    });
+
+    suggestionEl.createEl("div", { text: file.basename });
+  };
+
+  const addRelationship = async () => {
```
mdroidian: useCallback

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,396 @@
...
+      }
+    });
+
+    return options;
+  };
+
+  const availableRelationTypes = getAvailableRelationTypes();
+
+  const searchNodes = async (query: string): Promise<TFile[]> => {
+    if (!queryEngineRef.current) {
```
mdroidian: useCallback

apps/obsidian/src/components/SearchBar.tsx
```diff
@@ -0,0 +1,159 @@
...
+          renderItem,
+          asyncSearch,
+          minQueryLength,
+        },
+      );
+      return () => suggest.close();
+    }
+  }, [onSelect, app, getItemText, renderItem, asyncSearch, minQueryLength]);
+
+  const clearSelection = () => {
```
mdroidian: useCallback

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -0,0 +1,119 @@
...
+      return [];
+    }
+  }
+
+  /**
+   * Enhanced fuzzy search implementation
+   * Returns true if the search term is found within the target string
+   * with tolerance for typos and partial matches
+   */
+  fuzzySearch(target: string, search: string): boolean {
```
mdroidian, line 74: consider using Obsidian's `prepareFuzzySearch`.  This can be marked out of scope.

mdroidian (changes requested):
Good stuff!  A few changes requested.

apps/obsidian/src/components/DropdownSelect.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  useEffect(() => {
+    return () => {
+      if (dropdownRef.current) {
+        dropdownRef.current.selectEl.empty();
+        dropdownRef.current = null;
+      }
+    };
+  }, []);
+
```
mdroidian, line 72: Out of scope: consider auto selecting if there is only 1 option (this will be the case very often)

apps/obsidian/src/services/QueryEngine.ts
trangdoan982: yes this would be temporary. once we use npm we can remove this. 
for now i see that the hacky way on discord is people accessing datacore with: `appWithPlugins.plugins?.plugins?.["datacore"]?.api;`
but typescript won't recognize this, hence the type. I will add a comment.

apps/obsidian/src/components/RelationshipSection.tsx
trangdoan982: oh good point. yeah i think there's better way to structure available nodeTypes, we can do that in follow-up task

apps/obsidian/src/components/DropdownSelect.tsx
trangdoan982: added

apps/obsidian/src/components/DropdownSelect.tsx
```diff
@@ -0,0 +1,85 @@
...
+      const text = getItemText(option);
+      dropdown.addOption(text, text);
+    });
+
+    if (
+      currentValue &&
+      options.some((opt) => getItemText(opt) === currentValue)
+    ) {
+      dropdown.setValue(currentValue);
+    }
```
mdroidian, line 44: nit: `getItemText` is running on all `options` three times in this component.  This could be optimized.
Not a big problem for now as we are just doing `option.label`.

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,399 @@
...
+
+        // Use the already calculated compatibleNodeTypes state
+        if (compatibleNodeTypes.length === 0) {
+          setSearchError(
+            "No compatible node types available for the selected relation type",
+          );
+          return [];
+        }
+
+        // Extract the IDs from the compatibleNodeTypes
```
mdroidian: .

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -0,0 +1,399 @@
...
+          setSearchError("Active file does not have a node type");
+          return [];
+        }
+
+        if (!selectedRelationType) {
+          setSearchError("Please select a relationship type first");
+          return [];
+        }
+
+        // Use the already calculated compatibleNodeTypes state
```
mdroidian: .

apps/obsidian/src/components/SearchBar.tsx
```diff
@@ -0,0 +1,165 @@
...
+  }
+}
+
+const SearchBar = <T,>({
+  onSelect,
+  placeholder,
+  getItemText,
+  renderItem,
+  asyncSearch,
+  minQueryLength = 0,
```
mdroidian: .

package.json
```diff
@@ -29,6 +29,7 @@
     "tailwindConfig": "./packages/tailwind-config/tailwind.config.ts"
   },
   "dependencies": {
+    "obsidian-dataview": "^0.5.68",
```
mdroidian: .

mdroidian (approved):
🔥

PR #116, opened 2025-03-31
[ENG-44] Display relations

trangdoan982:
image

### DIFF
--- a/apps/obsidian/src/components/RelationshipSection.tsx
+++ b/apps/obsidian/src/components/RelationshipSection.tsx
@@ -474,8 +474,11 @@ const CurrentRelationships = ({
 
         if (!linkedFile) continue;
 
-        const group = tempRelationships.get(relationKey)!;
-        if (!group.linkedFiles.some((file) => file.path === linkedFile.path)) {
+        const group = tempRelationships.get(relationKey);
+        if (
+          group &&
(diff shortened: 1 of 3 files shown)


trangdoan982:
@mdroidian this is the follow up PR to display relations. you can review the create separately or just review this Eng-44 instead

apps/obsidian/src/components/RelationshipSection.tsx
```diff
@@ -384,12 +384,191 @@ const AddRelationship = ({ plugin, activeFile }: RelationshipSectionProps) => {
...
+
+        const linkedFileName = match[1];
+        const linkedFile = plugin.app.metadataCache.getFirstLinkpathDest(
+          linkedFileName,
+          activeFile.path,
+        );
+
+        if (!linkedFile) continue;
+
+        const group = tempRelationships.get(relationKey)!;
```
mdroidian: nit: let's do a `if (!group)`  or `group && ` instead to future proof.

mdroidian (approved):
👍

mdroidian:
Ok, because this was based off #115 we have some cleanup to do.
After #115 is ready to merge to `main`, we'll rebase #116  onto #115.
Then we'll merge #115 to `main`.
Then rebase #116 onto `main`.
Then merge #116  onto `main`.

Ideally, in the future, we'll try to base feature branches off main and have them be independent of each other.

PR #124, opened 2025-04-11
[ENG-207] Move Github sync setting to individual nodes

trangdoan982:
https://www.loom.com/share/19f37846ff0047fb82c50e51e6bba4a1?sid=f9cf3658-e53f-4fb3-8a07-fd0b5add284b

## Summary by CodeRabbit

- **New Features**
	- Added a dedicated GitHub integration tab in settings, allowing users to enable syncing with GitHub Issues and configure comment display.
	- Introduced user notifications that alert when configuration changes are needed.
  
- **Refactor**
	- Streamlined the synchronization process by dynamically linking settings to the current page, simplifying setup and removing unnecessary configuration options.

### DIFF
--- a/apps/roam/src/components/GitHubSync.tsx
+++ b/apps/roam/src/components/GitHubSync.tsx
@@ -48,8 +48,7 @@ import {
 } from "roamjs-components/components/ConfigPanels/types";
 import CustomPanel from "roamjs-components/components/ConfigPanels/CustomPanel";
 import getShallowTreeByParentUid from "roamjs-components/queries/getShallowTreeByParentUid";
-import CommentsQuery from "./GitHubSyncCommentsQuery";
-import getSubTree from "roamjs-components/util/getSubTree";
+
 import isFlagEnabled from "~/utils/isFlagEnabled";
 
 const CommentUidCache = new Set<string>();
@@ -110,22 +109,40 @@ const getRoamCommentsContainerUid = async ({
   extensionAPI: OnloadArgs["extensionAPI"];
 }) => {
   const pageTitle = getPageTitleByPageUid(pageUid);
-  const configUid = getPageUidByPageTitle(CONFIG_PAGE);
-  const configTree = getBasicTreeByParentUid(configUid);
-  const queryNode = getSubTree({
-    tree: configTree,
-    key: "Comments Block",
-  });
-  if (!queryNode) {
+
+  // Find the node type that matches this page
+  const discourseNodes = getDiscourseNodes();
+  let matchingNode;
+
+  for (const node of discourseNodes) {
+    if (node.githubSync) {
+      const isMatch = matchDiscourseNode({
+        format: node.format || "",
+        specification: node.specification || [],
+        text: node.text || "",
+        title: pageTitle,
+      });
+
+      if (isMatch) {
+        matchingNode = node;
+        break;
+      }
+    }
+  }
+
+  if (!matchingNode || !matchingNode.githubCommentsQueryUid) {
     renderToast({
       id: "github-issue-comments",
-      content: `Comments Block query not set. Set it in ${CONFIG_PAGE}`,
+      content:
+        "Comments Block query not set. Configure it in the Discourse Graph settings.",
     });
     return;
   }
+
+  // Use the node-specific query
   const results = await runQuery({
     extensionAPI,
-    parentUid: queryNode.uid,
+    parentUid: matchingNode.githubCommentsQueryUid,
     inputs: { NODETEXT: pageTitle, NODEUID: pageUid },
   });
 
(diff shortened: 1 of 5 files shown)


trangdoan982:
@coderabbitai review

apps/roam/src/components/GitHubSync.tsx
```diff
@@ -48,8 +48,7 @@ import {
 } from "roamjs-components/components/ConfigPanels/types";
 import CustomPanel from "roamjs-components/components/ConfigPanels/CustomPanel";
 import getShallowTreeByParentUid from "roamjs-components/queries/getShallowTreeByParentUid";
-import CommentsQuery from "./GitHubSyncCommentsQuery";
-import getSubTree from "roamjs-components/util/getSubTree";
+
```
mdroidian: .

apps/roam/src/components/GitHubSync.tsx
```diff
@@ -110,22 +109,33 @@ const getRoamCommentsContainerUid = async ({
...
+        text: node.text || "",
+        title: pageTitle,
+      });
+
+      if (isMatch) {
+        matchingNode = node;
+        break;
+      }
+    }
+  }
```
mdroidian: We are running this in the page observer already for every page (`initializeObserversAndListeners.ts`).  Let's minimize how many times we have to run this and pass the `node` down from `initializeObserversAndListeners.ts`.

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -31,26 +33,33 @@ const NodeConfig = ({
   const templateUid = getUid("Template");
   const overlayUid = getUid("Overlay");
   const canvasUid = getUid("Canvas");
-  const graphOverviewUid = getUid("Graph Overview");
+  const graphOverviewUid = getBlockUidIfExists(node.type, "Graph Overview");
+  const githubSyncUid = getBlockUidIfExists(node.type, "GitHub Sync");
```
mdroidian: Generally a good solution, but a few things:
- roamjs-components has `getSubTree`, so we generally use that to grab the node and then just `.uid` from there.
- RE: bug — I think you're right about the Graph Overview because the behavior for single flags was just the existence of the node. Nice find. So, let's use `getSubTree` there.
- I think we should stick with the pattern above of `getUid` and keep everything under a GitHub Sync block. There will likely be more settings to come.
- Then we can just change the flag to "Enabled."
```
- GitHub Sync
  - Enabled
  - Comments Block
 ```

Comment (mdroidian) [reply apps/roam/src/components/settings/NodeConfig.tsx]: This will mean we should change the `getDiscourseNodes` config and likely remove the flag panel.  More there.

File: apps/roam/src/utils/getDiscourseNodes.ts [hunk 4]
```diff
@@ -95,6 +97,11 @@ const getDiscourseNodes = (relations = getDiscourseRelations()) => {
         ),
         graphOverview:
           children.filter((c) => c.text === "Graph Overview").length > 0,
+        githubSync: children.filter((c) => c.text === "GitHub Sync").length > 0,
+        githubCommentsQueryUid: getSubTree({
```
mdroidian: Let's nest the comments uid under `githubSync`, just like we do for `canvasSettings`

apps/roam/src/components/GitHubSync.tsx
```diff
@@ -944,36 +945,6 @@ const initializeGitHubSync = async (onloadArgs: OnloadArgs) => {
                     },
                   },
                 } as Field<CustomField>,
```
mdroidian: I think you should be able to remove the entire config page observer, no?  I don't believe we are using it anymore

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -160,6 +169,53 @@ const NodeConfig = ({
...
+                parentUid={node.type}
+                uid={githubSyncUid}
+                value={node.githubSync}
+                options={{
+                  onChange: (checked) => setIsGithubSyncEnabled(checked),
+                }}
+              />
+
+              {isGithubSyncEnabled && (
+                <div className="mt-4 border-t pt-4">
```
mdroidian: Let's remove this whitespace space
image

mdroidian (changes requested):
Good stuff, a few changes requested.

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -147,16 +169,114 @@ const NodeConfig = ({
...
               <DiscourseNodeCanvasSettings uid={canvasUid} />
-              <FlagPanel
-                title="Graph Overview"
-                description="Whether to color the node in the graph overview based on canvas color.  This is based on the node's plain title as described by a \`has title\` condition in its specification."
-                order={0}
-                parentUid={node.type}
-                uid={graphOverviewUid}
-                value={node.graphOverview}
-              />
+              <div className="mb-4">
```
mdroidian: If we change the from using the node as the setting to `true`/`false` child, we would also need to make it backwards compatible.

As this is out of scope for the PR and the proper fix for the bug you identified would actually be a larger refactor, I'm just going to revert these changes and open a new ticket for the bug. 👍

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -32,25 +41,38 @@ const NodeConfig = ({
   const overlayUid = getUid("Overlay");
   const canvasUid = getUid("Canvas");
   const graphOverviewUid = getUid("Graph Overview");
+  const githubSyncUid = getUid("GitHub Sync");
+  const githubCommentsFormatUid = getSubTree({
```
mdroidian: This is probably introducing a race condition on first load because `githubSyncUid` is actually creating the node via an `async` function which `githubCommentsFormatUid` is trying to query.

I think the problem here is a bad pattern for setting up the initial configs of this setting page in general, which is out of scope for this PR.  So for now we'll leave it.

apps/roam/src/utils/getDiscourseNodes.ts
mdroidian: I meant both the blocks and return should be nested like `canvasSettings`

eg:
```
  canvasSettings: {
    [k: string]: string;
  };
```

So

```
githubSyncConfig: {
  enabled: boolean;
  commentsQueryUid: string;
};
```

trangdoan982: ah i see. that makes sense.

mdroidian (changes requested):
It looks like https://github.com/DiscourseGraphs/discourse-graph/pull/124/files#r2053103675 still need to be addressed.  Do you have any questions regarding this?

EDIT: I'm going to go ahead and finish this up as it is blocking a few other tasks.

PR #132, opened 2025-04-16
Roam: Bug - When user creates discourse node using dg node menu, the reference to dg node is not added - ENG-9

sid597:
https://www.loom.com/share/d9a921f0833a4b7389afe74ed7cd2083?sid=cbeebf05-1936-43bd-b1cb-4002241aa6c7


mdroidian:
Converting to draft as we are now looking for minimal reproducibility instead of a fix.

mdroidian (changes requested):
It looks like we are still using `DiscourseNodeMenu` and all of the code that is involved there.

To make sure we remove the most amount of variables, this also needs to be replaced with a minimal version from BlueprintJS.

mdroidian:
Fixed with #154

PR #133, opened 2025-04-16
[ENG-204] Move from localStorage to extensionAPI.settings

trangdoan982:
## Summary by CodeRabbit

- **New Features**
  - Added seamless migration of user settings from browser storage to a centralized extension settings system for improved reliability and consistency.
- **Refactor**
  - Updated settings and token management throughout the app to use new utility functions, ensuring smoother storage handling and better compatibility.
  - Enhanced GitHub OAuth flow with improved state persistence, error handling, and detailed logging.
- **Chores**
  - Introduced utility modules to centralize and standardize settings management.


trangdoan982:
@coderabbitai review

apps/roam/src/utils/settings.ts
```diff
@@ -0,0 +1,34 @@
+import getExtensionAPI from "roamjs-components/util/extensionApiContext";
```
mdroidian: nit: can we rename this file to something that is less generic?  `settings` could refer to a number of things.
some ideas: 
- extensionSettings
- roamExtensionConfig
- settingsManager

apps/roam/src/utils/settings.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (localValue !== null && localValue !== undefined) {
+    extensionAPI.settings.set(key, localValue);
+    localStorage.removeItem(key);
+    localStorage.removeItem(roamjsKey);
+    localStorage.removeItem(roamjsKey2);
+    return localValue as T;
+  }
+
+  return defaultValue as T;
+}
```
mdroidian: Nice idea, but might be overkill.  GitHub Sync shouldn't need it, because it is a new feature.  

Do you already envision where/when we might use this?

mdroidian (approved):
Nice, just a couple of thoughts

apps/roam/src/utils/settings.ts
trangdoan982: oh nice. yeah i forgot that's a new feature. i was just looking at the application storage to figure out which one needs migration. maybe we can just remove it? it doesn't seem like we use localstorage for any feature outside of github sync for now

mdroidian: Yeah, let's remove it for now and bring it back if/when needed. 👍

mdroidian:
Quick tip: remember to delete the branch on GitHub when you are done.  If we don't they the branch menus just get noisy.

PR #137, opened 2025-04-26
ENG-248: DGAI:  Testing local vector databases

sid597:


mdroidian:
It has been decided that we will not use local databases at the moment.

PR #138, opened 2025-04-26
[ENG-259] Create Observer and UI pattern

trangdoan982:
https://www.loom.com/share/16c24a6cc7ad4ca6aef6624ffc097bae

## Summary by CodeRabbit

- **New Features**
  - Introduced a searchable popover menu for inserting discourse nodes by typing "@" in a block, with keyboard and mouse navigation.
  - Selecting a node automatically inserts a page reference into the block.
- **Chores**
  - Added new keyboard event listeners to support the discourse node search functionality.

### DIFF
--- a/apps/roam/src/components/DiscourseNodeSearchMenu.tsx
+++ b/apps/roam/src/components/DiscourseNodeSearchMenu.tsx
@@ -5,7 +5,15 @@ import React, {
   useRef,
   useState,
 } from "react";
-import { Menu, MenuItem, Popover, Position, Checkbox } from "@blueprintjs/core";
+import {
+  Menu,
+  MenuItem,
+  Popover,
+  Position,
+  Checkbox,
+  Button,
+  Icon,
+} from "@blueprintjs/core";
 import ReactDOM from "react-dom";
 import getUids from "roamjs-components/dom/getUids";
 import getTextByBlockUid from "roamjs-components/queries/getTextByBlockUid";
@@ -81,6 +89,7 @@ const NodeSearchMenu = ({
   const [checkedTypes, setCheckedTypes] = useState<Record<string, boolean>>(
     DISCOURSE_TYPES.reduce((acc, type) => ({ ...acc, [type.type]: true }), {}),
   );
+  const [isFilterMenuOpen, setIsFilterMenuOpen] = useState(false);
   const menuRef = useRef<HTMLUListElement>(null);
   const { ["block-uid"]: blockUid, ["window-id"]: windowId } = useMemo(
     () =>
@@ -167,7 +176,6 @@ const NodeSearchMenu = ({
             }
           });
 
-          // Analytics
           posthog.capture("Discourse Node: Selected from Search Menu", {
             id: item.id,
             text: item.text,
@@ -220,15 +228,7 @@ const NodeSearchMenu = ({
         e.stopPropagation();
       }
     },
-    [
-      allItems,
-      setActiveIndex,
-      onSelect,
-      onClose,
-      textarea,
-      setSearchTerm,
-      menuRef,
-    ],
+    [allItems, setActiveIndex, onSelect, onClose],
   );
 
   useEffect(() => {
@@ -236,13 +236,6 @@ const NodeSearchMenu = ({
       setActiveIndex(0);
     }
   }, [allItems, activeIndex]);
-
-  useEffect(() => {
-    document.addEventListener("keydown", keydownListener);
-    return () => {
-      document.removeEventListener("keydown", keydownListener);
-    };
-  }, [keydownListener]);
 
   useEffect(() => {
     textarea.addEventListener("input", handleTextAreaInput);
@@ -293,6 +286,24 @@ const NodeSearchMenu = ({
     e.stopPropagation();
   }, []);
 
+  const toggleFilterMenu = useCallback(
+    (e: React.MouseEvent) => {
+      e.preventDefault();
+      e.stopPropagation();
+
+      setIsFilterMenuOpen((prev) => !prev);
+
+      setTimeout(() => {
+        if (textarea) {
+          textarea.focus();
+          const cursorPos = textarea.selectionStart;
+          textarea.setSelectionRange(cursorPos, cursorPos);
+        }
+      }, 50);
+    },
+    [textarea],
+  );
+
   return (
     <Popover
       onClose={onClose}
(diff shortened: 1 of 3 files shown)


trangdoan982:
@coderabbitai review

mdroidian (changes requested):
Generally good stuff!

A couple use cases we should handle before merging
- insert node and keep typing (don't create a new block on enter)
- dealing with text after `@`
- remove node menu when `@` is gone

https://www.loom.com/share/8210d0012149414aae284bd0f31b1d64

mdroidian:
How goes the progress on this?

trangdoan982:
I'm still working on this. there's some cursor position bug I still can't figure out yet.

trangdoan982:
@mdroidian https://www.loom.com/share/1582b3a10eb44abdbd955c16536caa98 new checkbox for filtering node types during search and a few test cases

mdroidian:
Good stuff.  But what happens to the UI when a graph has 9 or 10 nodes?  Could you test this on `plugin-testing-megacoglab`?

trangdoan982:
@mdroidian i set the popover menu to fixed height will scrollable y so it will look the same and users just need to scroll down to see the results:

https://github.com/user-attachments/assets/8f946b93-ff42-4960-be7e-7ec924b0f1a8

mdroidian:
Having to scroll each time is probably something we want to avoid.  Likely we'll want to widen the entire popover and have the check boxes limited to one or two rows, then scroll horizontal.  Or remove the grouping and hide the filters until the user wants them.  We can consider this out of scope for the current PR and ask Johnny his opinion.

trangdoan982:
i did start with responsive sizing for the menu, but it somehow created a bug that mess with the popover's position. So i'd go with overflow-x-auto for the filters.
But I think this filter might need more thoughts on UI (when it appears, how it appears), or saving the selected options so it persists somewhere instead of renewing the state everytime it opens.
We already have this ticket to confirm UI with John: https://linear.app/discourse-graphs/issue/ENG-277/confirm-on-ui-design-of-dg-summoning-menu. I will follow up with him here and we might scope a separate PR for the filters

jsmorabito:
this is fantastic! I see the utility of having filters given a user might want to only see a particular set of nodes and exclude others from the search. My intuition though is that the filtering action will a less frequent operation and could be relegated to the dg settings menu so that users have the option to customize node search to their needs, but the menu itself will not be cluttered with the filter options. 

I'd be content with that solution and then if users request to have filter settings more easily accessible we can explore designs such as these below:
image

For now I'm content with having filter options in the settings menu and not in the node search menu.

apps/roam/src/components/DiscourseNodeSearchMenu.tsx
```diff
@@ -0,0 +1,390 @@
...
+                      newCursorPosition,
+                    );
+                    break;
+                  }
+                }
+              }, 50);
+            }
+          });
+
+          // Analytics
```
mdroidian: .

apps/roam/src/components/DiscourseNodeSearchMenu.tsx
```diff
@@ -0,0 +1,390 @@
...
+      }
+    },
+    [
+      allItems,
+      setActiveIndex,
+      onSelect,
+      onClose,
+      textarea,
+      setSearchTerm,
+      menuRef,
```
mdroidian: remove unused

trangdoan982:
@jsmorabito my instinct is that user might want to change up the filter type when they query quite often. I'm imagining a flow where users would switch the type up as they type down their paper outline.
I would love to implement a "node type X only" feature, like in Google Flights they have options for you to choose 1 stop only or Non-stop only. With graphs that has lots of lists this would be really helpful. Can you develop a mockup for that as well?

apps/roam/src/components/DiscourseNodeSearchMenu.tsx
```diff
@@ -0,0 +1,390 @@
...
+      setActiveIndex(0);
+    }
+  }, [allItems, activeIndex]);
+
+  useEffect(() => {
+    document.addEventListener("keydown", keydownListener);
+    return () => {
+      document.removeEventListener("keydown", keydownListener);
+    };
+  }, [keydownListener]);
```
mdroidian: We are adding the `keydownListener` twice.  Is that on purpose?

mdroidian (changes requested):
🔥 A few minor changes requested.

Also, I don't believe we should be merging this to main as is.  If we want an anchor I'll create a `eng-235-create-dg-summoning-menu` to merge to.

jsmorabito:
> @jsmorabito my instinct is that user might want to change up the filter type when they query quite often. I'm imagining a flow where users would switch the type up as they type down their paper outline. I would love to implement a "node type X only" feature, like in Google Flights they have options for you to choose 1 stop only or Non-stop only. With graphs that has lots of lists this would be really helpful. Can you develop a mockup for that as well?

would this be something similar to a search with advanced options like we see in obsidians search widget? 
image

I can definitely see a use for having some advanced search that let's you have a more specific query which can persist in the sidebar while your outline is in the main window

trangdoan982:
@jsmorabito i was referring to this:
image


if there's an UI pattern that can be clean and allow people to chose only 1 node type (i imagine it useful if there are many node types). But that's more a nice-to-have. I don't think we need to overkill this with an advanced filter option for now.

jsmorabito:
Gotcha! I made a project in Linear where we can continue to expand on this idea: https://linear.app/discourse-graphs/project/advanced-search-4b947899da16/overview

Feel free to add more notes there. Also added a ticket for mockup up this functionality: https://linear.app/discourse-graphs/issue/DES-49/mock-up-advanced-search

jsmorabito:
But yea, for now this node search menu is good without the filters (in the menu itself at least)

trangdoan982:
@mdroidian addressed your PR comments. I also implemented the filter toggle that Johnny mocked. I made sure the focus remains even when you interact with the filter menu.

https://github.com/user-attachments/assets/41c5e018-5a6e-4800-b8df-de10fd0ce2e5

apps/roam/src/components/DiscourseNodeSearchMenu.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+      setIsFilterMenuOpen((prev) => !prev);
+
+      setTimeout(() => {
+        if (textarea) {
+          textarea.focus();
+          const cursorPos = textarea.selectionStart;
+          textarea.setSelectionRange(cursorPos, cursorPos);
+        }
+      }, 50);
```
mdroidian: Could you show me what use case this is solving?  I assumed it was to keep focus on the block, but I removed it and nothing seemed to change for me.

apps/roam/src/components/DiscourseNodeSearchMenu.tsx
```diff
@@ -0,0 +1,420 @@
...
+        setActiveIndex((prev) => (prev + 1) % allItems.length);
+        e.preventDefault();
+        e.stopPropagation();
+      } else if (e.key === "ArrowUp") {
+        setActiveIndex(
+          (prev) => (prev - 1 + allItems.length) % allItems.length,
+        );
+        e.preventDefault();
+        e.stopPropagation();
+      } else if (e.key == "ArrowLeft" || e.key === "ArrowRight") {
```
mdroidian: If this isn't serving a purpose it should be removed so the user can move the cursor to edit text while searching for nodes

mdroidian (approved):
Good stuff!

One thing we should add in the future is capturing clicks and preventing the default so they don't remove focus. (use case: user wants to click the filter button or a checkbox but misses)

https://www.loom.com/share/6a852a8007464f709700f41571c0646c

None of this is blocking.

apps/roam/src/components/DiscourseNodeSearchMenu.tsx
trangdoan982: oh yes, it was to remain in focus, but it the preventDefault()  and stopPropagation() already did the job

PR #147, opened 2025-05-06
DGAI: Create HyDe utility function - ENG-292

sid597:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new capability for generating and ranking related discourse suggestions based on semantic relationships.
	- Enhanced relevance of suggested nodes through improved semantic matching and ranking.
	- Improved error handling and robustness during the suggestion process.
	- Added flexible embedding generation with customizable model, dimensions, and encoding options.
	- Enhanced API stability by handling missing API keys gracefully with clear error responses.

### DIFF
--- a/apps/roam/src/utils/hyde.ts
+++ b/apps/roam/src/utils/hyde.ts
@@ -1,61 +1,120 @@
-export type EmbeddingVector = number[];
-
 import { Result } from "./types";
+import { getNodeEnv } from "roamjs-components/util/env";
+
+type ApiEmbeddingResponse = {
+  data: Array<{
+    embedding: number[];
+  }>;
+};
+
+type ApiSupabaseResultItem = {
+  roam_uid: string;
+  text_content: string;
+  similarity: number;
+};
+
+export type EmbeddingVectorType = number[];
 
 export type CandidateNodeWithEmbedding = Result & {
   type: string;
-  embedding: EmbeddingVector;
+  embedding: EmbeddingVectorType;
 };
 
 export type SuggestedNode = Result & {
   type: string;
 };
 
-export type RelationTriplet = [string, string, string];
+export type RelationDetails = {
+  relationLabel: string;
+  relatedNodeText: string;
+  relatedNodeFormat: string;
+};
 
-export type HypotheticalNodeGenerator = (params: {
-  node: string;
-  relationType: RelationTriplet;
-}) => Promise<string>;
-
-export type EmbeddingFunc = (text: string) => Promise<EmbeddingVector>;
-
-export type SearchResultItem = {
-  object: SuggestedNode;
+export type NodeSearchResult = {
+  object: { uid: string; text: string };
   score: number;
 };
-export type SearchFunc = (params: {
-  queryEmbedding: EmbeddingVector;
+
+type HypotheticalNodeGenerator = (params: {
+  node: string;
+  relationType: RelationDetails;
+}) => Promise<string>;
+
+type EmbeddingFunc = (text: string) => Promise<EmbeddingVectorType>;
+
+type SearchFunc = (params: {
+  queryEmbedding: EmbeddingVectorType;
   indexData: CandidateNodeWithEmbedding[];
-  options: { topK: number };
-}) => Promise<SearchResultItem[]>;
+}) => Promise<NodeSearchResult[]>;
 
-export const ANTHROPIC_API_URL =
-  "https://discoursegraphs.com/api/llm/anthropic/chat";
-export const ANTHROPIC_MODEL = "claude-3-sonnet-20240229";
-export const ANTHROPIC_REQUEST_TIMEOUT_MS = 30_000;
+const API_CONFIG = {
+  LLM: {
+    URL: "https://discoursegraphs.com/api/llm/openai/chat",
+    MODEL: "gpt-4.1",
+    TIMEOUT_MS: 30_000,
+    MAX_TOKENS: 104,
+    TEMPERATURE: 0.9,
+  },
+  BASE_URL: {
+    DEV: "http://localhost:3000",
+    PROD: "https://discoursegraphs.com",
+  },
+  EMBEDDINGS: {
+    PATH: "/api/embeddings/openai/small",
+  },
+  SUPABASE: {
+    MATCH_EMBEDDINGS_PATH:
+      "/api/supabase/rpc/match-embeddings-for-subset-nodes",
+  },
+} as const;
 
-export const generateHypotheticalNode: HypotheticalNodeGenerator = async ({
+const getBaseUrl = (): string => {
+  const isDevelopment = getNodeEnv() === "development";
+  return isDevelopment ? API_CONFIG.BASE_URL.DEV : API_CONFIG.BASE_URL.PROD;
+};
+
+const handleApiError = async (
+  response: Response,
+  context: string,
+): Promise<never> => {
+  const errorText = await response.text();
+  let errorData;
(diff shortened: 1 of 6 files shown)


sid597:
@coderabbitai full review

sid597:
@coderabbitai full review

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
...
+      return `Error: Failed to generate hypothetical node. Request timed out.`;
+    }
+    console.error("Error during fetch for Claude API:", error);
+    return `Error: Failed to generate hypothetical node. ${
+      error instanceof Error ? error.message : String(error)
+    }`;
+  }
+};
+
+async function searchAgainstCandidates(
```
mdroidian: Let's keep consistent (with this file and the whole repo) and use arrow functions

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
...
+          error,
+        );
+        return [];
+      }
+    }),
+  );
+  return allSearchResults;
+}
+
+function combineScores(
```
mdroidian: Let's keep consistent (with this file and the whole repo) and use arrow functions

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
+export type EmbeddingVector = number[];
```
mdroidian: Are all of these types required to be exported?

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
...
+export const ANTHROPIC_MODEL = "claude-3-sonnet-20240229";
+export const ANTHROPIC_REQUEST_TIMEOUT_MS = 30_000;
+
+export const generateHypotheticalNode: HypotheticalNodeGenerator = async (
+  node: string,
+  relationType: RelationTriplet,
+): Promise<string> => {
+  const [relationLabel, relatedNodeText, relatedNodeFormat] = relationType;
+
+  const userPromptContent = `Given the source discourse node \\\`\\\`\\\`${node}\\\`\\\`\\\`, \nand considering the relation \\\`\\\`\\\`${relationLabel}\\\`\\\`\\\` \nwhich typically connects to a node of type \\\`\\\`\\\`${relatedNodeText}\\\`\\\`\\\` \n(formatted like \\\`\\\`\\\`${relatedNodeFormat}\\\`\\\`\\\`), \ngenerate a hypothetical related discourse node text that would plausibly fit this relationship. \nOnly return the text of the hypothetical node.`;
```
mdroidian: nit: What is the purpose of \\\`\\\`\\\`?  Can this purpose be achieved with less characters?

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
+export type EmbeddingVector = number[];
+
+export type CandidateNodeWithEmbedding = {
```
mdroidian: Let's build off the existing `Result` type https://github.com/DiscourseGraphs/discourse-graph/blob/eng-261-menu-query/apps/roam/src/utils/types.ts#L42

eg: ```type MyType = Result & { type: string };```

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
...
+  text: string;
+  uid: string;
+  type: string;
+  embedding: EmbeddingVector;
+};
+
+export type SuggestedNode = {
```
mdroidian: Let's build off the existing `Result` type https://github.com/DiscourseGraphs/discourse-graph/blob/eng-261-menu-query/apps/roam/src/utils/types.ts#L42

eg: ```type MyType = Result & { type: string };```

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
...
+  text: string;
+  uid: string;
+  type: string;
+};
+
+export type RelationTriplet = [string, string, string];
```
mdroidian: I image we have access to the DiscourseRelation during this process.
Could we reuse https://github.com/DiscourseGraphs/discourse-graph/blob/hyde-utility/apps/roam/src/utils/getDiscourseRelations.ts#L12 instead creating a new type?

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+    console.error("Error during fetch for Claude API:", error);
+    return `Error: Failed to generate hypothetical node. ${
+      error instanceof Error ? error.message : String(error)
+    }`;
+  }
+};
+
+async function searchAgainstCandidates(
+  hypotheticalTexts: string[],
```
mdroidian: Also, when 2+ params let's used named params instead.

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
...
+  combinedResults.sort((a, b) => b.score - a.score);
+  return combinedResults.map((item) => ({
+    text: item.node.text,
+    uid: item.node.uid,
+    type: item.node.type,
+  }));
+}
+
+export const findSimilarNodesUsingHyde = async (
+  candidateNodes: CandidateNodeWithEmbedding[],
```
mdroidian: named params

apps/roam/src/utils/hyde.ts
```diff
@@ -0,0 +1,217 @@
...
+  currentNodeText: string,
+  relationTriplets: RelationTriplet[],
+  options: {
```
mdroidian: why are these functions being passed as options?  what is the use case here?

mdroidian (changes requested):
A few questions/change requests

apps/roam/src/utils/hyde.ts
sid597: For testing purposes, like if we want to test a local embedding setup or a dummy setup to test the functionality of this function. I am not sure if that's the goal, it can definitely be made not modular if that is not a usecase

apps/roam/src/utils/hyde.ts
sid597: I was addressing https://github.com/DiscourseGraphs/discourse-graph/pull/147#discussion_r2075675383 

🛠️ Refactor suggestion

Prompt could leak user input or break format – escape or fence variables

The prompt directly interpolates user-supplied text (node, relationLabel, …).
If any of those strings contains quotes or Markdown delimiters, the prompt structure might break or allow prompt-injection. Consider surrounding dynamic inserts with triple back-tick fences or another delimiter recognised by the LLM.

apps/roam/src/utils/hyde.ts
sid597: This triplet is of type  [relation-label, related-node-text, related-node-format] e.g [Informs, Claim, [[CLM]] - {}]

apps/roam/src/utils/hyde.ts
mdroidian: That is fine.  But we should set defaults then.  They shouldn't need to be imported here just to be passed back.

apps/roam/src/utils/hyde.ts
mdroidian: The meaning of each element in this triplet isn’t immediately clear from the type definition. Would it make sense to represent this as an object with named keys instead? That might make the structure more self-documenting.

sid597:
@coderabbitai full review

PR #148, opened 2025-05-06
DGAI: Integrate HyDe utility with discourse node suggestions component - ENG-293

sid597:
## Summary by CodeRabbit

- **New Features**
	- Introduced a ranked suggestion system for relationships using HyDE-based similarity search, providing more relevant discourse node recommendations.
	- Added a loading indicator while HyDE search results are being retrieved.
	- Updated the "Suggested Relationships" section to display results ranked by HyDE after search completion.

- **Bug Fixes**
	- Improved filtering to exclude nodes of invalid types and those already present in the current results.

### DIFF
--- a/apps/roam/src/components/DiscourseContextOverlay.tsx
+++ b/apps/roam/src/components/DiscourseContextOverlay.tsx
@@ -124,18 +124,18 @@ const getAllReferencesOnPage = (pageTitle: string) => {
   })) as Result[];
 };
 
-interface Relation {
+type Relation = {
   label: string;
   source: string;
   destination: string;
-}
+};
 
-interface DiscourseNodeInfo {
+type DiscourseNodeInfo = {
   format: string;
   text: string;
   type: string;
   [key: string]: any;
-}
+};
 
 const getUniqueLabelTypeTriplets = (
   relations: Relation[],
@@ -185,7 +185,6 @@ const getUniqueLabelTypeTriplets = (
     }
   });
 
-  console.log("uniqueTriplets", uniqueTriplets);
   return uniqueTriplets;
 };
 
@@ -308,8 +307,6 @@ const DiscourseContextOverlay = ({
           type: node.type,
         };
       })
-      .filter((node) => node !== null)
-      .filter((node) => validTypes.includes(node.type))
       .filter((node): node is SuggestedNode => node !== null)
       .filter(
         (node) =>
--- a/apps/roam/src/components/DiscourseContextOverlay.tsx
+++ b/apps/roam/src/components/DiscourseContextOverlay.tsx
@@ -5,6 +5,7 @@ import {
   Position,
   Tooltip,
   ControlGroup,
+  Spinner,
 } from "@blueprintjs/core";
 import React, { useCallback, useEffect, useMemo, useState } from "react";
 import ReactDOM from "react-dom";
@@ -27,6 +28,42 @@ import getAllPageNames from "roamjs-components/queries/getAllPageNames";
 import { Result } from "roamjs-components/types/query-builder";
 import createBlock from "roamjs-components/writes/createBlock";
 import { getBlockUidFromTarget } from "roamjs-components/dom";
+import {
+  findSimilarNodesUsingHyde,
+  SuggestedNode,
+  generateHypotheticalNode,
+  CandidateNodeWithEmbedding,
+  SearchResultItem,
+  EmbeddingVector,
+} from "~/utils/hyde";
+
+// Placeholder for future implementation
+const mockCreateEmbedding = async (text: string): Promise<number[]> => {
+  console.warn(
+    `mockCreateEmbedding called for text: "${text}". Using dummy random embedding.`,
+  );
+  // Return a dummy embedding, matching the dimension (5) used elsewhere for candidate embeddings.
+  return Array.from({ length: 5 }, () => Math.random());
+};
+
+// Placeholder for future implementation
+const mockVectorSearch = async (
+  _queryEmbedding: EmbeddingVector,
+  candidates: CandidateNodeWithEmbedding[],
+  options: { topK: number } = { topK: 3 },
+): Promise<SearchResultItem[]> => {
+  const { topK } = options;
+  const numToReturn = Math.min(topK, candidates.length);
+  console.warn(
+    `mockVectorSearch called with ${candidates.length} candidates. Returning top ${numToReturn} as dummy results.`,
+  );
+  // For a simple mock, return the first few candidates, wrapping them as SearchResultItem.
+  const resultCandidates = candidates.slice(0, numToReturn);
+  return resultCandidates.map((node) => ({
+    object: { uid: node.uid, text: node.text, type: node.type },
+    score: Math.random(), // Assign a dummy score
+  }));
+};
 
 type DiscourseData = {
   results: Awaited<ReturnType<typeof getDiscourseContextResults>>;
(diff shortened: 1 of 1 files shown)


sid597:
@coderabbitai full review

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -87,6 +124,70 @@ const getAllReferencesOnPage = (pageTitle: string) => {
   })) as Result[];
 };
 
+type Relation = {
```
mdroidian, line 127: Relations are already of type DiscourseRelation, so lets reuse that

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -139,11 +245,17 @@ const DiscourseContextOverlay = ({
   }, [refresh, getInfo]);
 
   // Suggestive Mode
-  const validTypes = useMemo(() => {
-    if (!discourseNode) return [];
+  const memoizedData = useMemo(() => {
```
mdroidian, line 248: this seems pretty overloaded, as noted by the ambiguous naming.  let's discuss this at our next meeting.

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -87,6 +124,70 @@ const getAllReferencesOnPage = (pageTitle: string) => {
...
+};
+
+type DiscourseNodeInfo = {
+  format: string;
+  text: string;
+  type: string;
+  [key: string]: any;
+};
+
+const getUniqueLabelTypeTriplets = (
```
mdroidian, line 140: I'm interested in exploring if most of the data required could be provided by `validTypes` and/or `validRelations`.  It seems like we are redoing a lot of work here.

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -102,6 +203,11 @@ const DiscourseContextOverlay = ({
   const [results, setResults] = useState<DiscourseData["results"]>([]);
   const [refs, setRefs] = useState(0);
   const [score, setScore] = useState<number | string>(0);
+  const [isSearchingHyde, setIsSearchingHyde] = useState(false);
+  const [suggestedNodes, setSuggestedNodes] = useState<SuggestedNode[]>([]);
```
mdroidian, line 207: It looks like `suggestedNodes` has been replaced with `hydeFilteredNodes` but this state still exists.

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -171,32 +292,87 @@ const DiscourseContextOverlay = ({
...
+            Object.values(r.results).some((result) => result.uid === node.uid),
+          ),
+      );
 
     setSuggestedNodes(nodes);
-  }, [selectedPage, discourseNode, relations]);
 
-  const handleCreateBlock = async (node: { uid: string; text: string }) => {
+    if (nodes.length > 0 && uniqueRelationTypeTriplets.length > 0) {
+      runHydeSearch(nodes, tag, uniqueRelationTypeTriplets);
```
mdroidian, line 321: This would be easier to debug if this was a pure function with no side effects.  It should return the nodes and the state changes should be done outside of function.

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -171,32 +292,87 @@ const DiscourseContextOverlay = ({
...
     await createBlock({
       parentUid: blockUid,
-      node: { text: `[[${node.text}]]` },
+      node: { text: `[[${nodeText}]]` },
     });
-    setSuggestedNodes(suggestedNodes.filter((n) => n.uid !== node.uid));
+  };
+
+  const runHydeSearch = async (
+    currentSuggestions: SuggestedNode[],
```
mdroidian, line 333: named params

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -171,32 +292,87 @@ const DiscourseContextOverlay = ({
...
+    }
+  }, [selectedPage, results, validTypes, tag, uniqueRelationTypeTriplets]);
+
+  const handleCreateBlock = async (nodeText: string) => {
     await createBlock({
       parentUid: blockUid,
-      node: { text: `[[${node.text}]]` },
+      node: { text: `[[${nodeText}]]` },
     });
-    setSuggestedNodes(suggestedNodes.filter((n) => n.uid !== node.uid));
```
mdroidian, line 199: Why was this removed?  It's function was removing the suggested node from the list when the user added it to the page.

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -27,6 +28,42 @@ import getAllPageNames from "roamjs-components/queries/getAllPageNames";
 import { Result } from "roamjs-components/types/query-builder";
 import createBlock from "roamjs-components/writes/createBlock";
 import { getBlockUidFromTarget } from "roamjs-components/dom";
+import {
```
mdroidian, line 31: It is unclear to me what the use case is of importing these functions just to pass them back to the same util file.

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -87,6 +124,70 @@ const getAllReferencesOnPage = (pageTitle: string) => {
   })) as Result[];
 };
 
+type Relation = {
+  label: string;
+  source: string;
+  destination: string;
+};
+
+type DiscourseNodeInfo = {
```
mdroidian, line 133: Looks like we could reuse the DiscourseNode type here

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -244,23 +420,27 @@ const DiscourseContextOverlay = ({
             {selectedPage && (
               <div className="mt-6">
                 <h3 className="mb-2 text-base font-semibold">
-                  Suggested Relationships
+                  Suggested Relationships (Ranked by HyDE)
```
mdroidian, line 423: This will likely not mean anything to majority of end users. (and could just confuse)

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -244,23 +420,27 @@ const DiscourseContextOverlay = ({
...
+                            minimal
+                            icon="add"
+                            onClick={() => handleCreateBlock(node.text)}
+                            className="ml-2"
+                          />
+                        </li>
+                      ))
+                    : null}
+                  {!isSearchingHyde && hydeFilteredNodes.length === 0 && (
+                    <li>No relevant relations found using HyDE.</li>
```
mdroidian, line 443: HyDe should just be used for sorting/ranking, so this is misleading.  Also "HyDe" will likely not mean anything to most end users.

mdroidian (changes requested):
Lots going on here, nice work!  I've made a few comments.

In our next meeting let's talk about `memoizedData` and `getUniqueLabelTypeTriplets`, but there are still many changes that can be made independent of those.

apps/roam/src/components/DiscourseContextOverlay.tsx
sid597: Note: all info present in validRelations and nodeType to get the triplets

sid597: refactor this fat function and memoize the common data needed its easier to follow what's happening why

mdroidian:
Creating a new branch.  This branch / PR will be closed eventually.

PR #157, opened 2025-05-14
[ENG-310] Preload nodes and relations in setting

trangdoan982:
https://www.loom.com/share/91210e4a0c864615ab9364b66723834d?sid=84cc7718-ba22-433a-9a86-ca3b222d9fbb

# Context:
- We're preloading the plugin with node types, relation types, and discourse relations, as defined in the Linear ticket
- This feature is part of the preparation for Obsidian beta launch for test users.

### DIFF
--- a/apps/obsidian/src/constants.ts
+++ b/apps/obsidian/src/constants.ts
@@ -1,7 +1,7 @@
-import { DiscourseNode, DiscourseRelationType, Settings } from "./types";
-import generateUid from "./utils/generateUid";
+import { DiscourseNode, DiscourseRelationType, Settings } from "~/types";
+import generateUid from "~/utils/generateUid";
 
-export const defaultNodeTypes: Record<string, DiscourseNode> = {
+export const DEFAULT_NODE_TYPES: Record<string, DiscourseNode> = {
   Question: {
     id: generateUid("node"),
     name: "Question",
@@ -18,7 +18,7 @@ export const defaultNodeTypes: Record<string, DiscourseNode> = {
     format: "EVD - {content}",
   },
 };
-export const defaultRelationTypes: Record<string, DiscourseRelationType> = {
+export const DEFAULT_RELATION_TYPES: Record<string, DiscourseRelationType> = {
   supports: {
     id: generateUid("relation"),
     label: "supports",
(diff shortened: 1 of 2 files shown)


apps/obsidian/src/constants.ts
```diff
@@ -0,0 +1,59 @@
+import { DiscourseNode, DiscourseRelationType, Settings } from "./types";
+import generateUid from "./utils/generateUid";
```
mdroidian: nit `~`

apps/obsidian/src/constants.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const defaultNodeTypes: Record<string, DiscourseNode> = {
```
mdroidian: nit
- `DEFAULT_NODE_TYPES` to follow convention
- could be an array to keep it simple

apps/obsidian/src/constants.ts
```diff
@@ -0,0 +1,59 @@
...
+    name: "Claim",
+    format: "CLM - {content}",
+  },
+  Evidence: {
+    id: generateUid("node"),
+    name: "Evidence",
+    format: "EVD - {content}",
+  },
+};
+export const defaultRelationTypes: Record<string, DiscourseRelationType> = {
```
mdroidian: nit
- `DEFAULT_RELATION_TYPES` to follow convention
- could be an array to keep it simple

apps/obsidian/src/index.ts
```diff
@@ -1,16 +1,11 @@
...
-
-const DEFAULT_SETTINGS: Settings = {
-  nodeTypes: [],
-  discourseRelations: [],
-  relationTypes: [],
-};
+import { DEFAULT_SETTINGS } from "./constants";
 
 export default class DiscourseGraphPlugin extends Plugin {
   settings: Settings = { ...DEFAULT_SETTINGS };
```
mdroidian, line 11: It looks like this doesn't create a `data.json` so on every graph load we are generating a new `uid`, which renders previously created nodes unrecognizable.

https://www.loom.com/share/a6e0302ecb2b428a9b30c8d20ccf4dab

We could either save the settings or use a static default `uid`

mdroidian (changes requested):
Just one blocking comment where new `uid`s will be created until a user edits the setting which will save the `data.json`

apps/obsidian/src/index.ts
trangdoan982: i created data.json by this.saveSettings() if it doesn't exist yet.

mdroidian (approved):
👍

PR #162, opened 2025-05-16
[ENG-299] Trigger setting for DG summoning menu

trangdoan982:
## Context:
- The DG summoning menu. This PR is the final piece for our v0!!
- Creates the custom trigger setting to open the node search menu.
## Testing:
https://www.loom.com/share/9f23f6b7dab8401c97918f3fd2f5bd8b 

## Summary by CodeRabbit

- **New Features**
  - Added the ability for users to customize the trigger string that opens the Node Search Menu, allowing for multi-character triggers.
  - Introduced a new settings section in personal settings where users can configure the Node Search Menu trigger.

- **Refactor**
  - Improved detection logic for the Node Search Menu trigger, supporting dynamic and multi-character triggers instead of a hardcoded "@".

- **Chores**
  - Updated event handling to listen for input events instead of keydown events for better trigger detection.

### DIFF
--- a/apps/roam/src/components/DiscourseNodeSearchMenu.tsx
+++ b/apps/roam/src/components/DiscourseNodeSearchMenu.tsx
@@ -256,8 +256,6 @@ const NodeSearchMenu = ({
     }
   }, [textarea, onClose, debouncedSearchTerm, triggerPosition, triggerText]);
 
-  console.log(activeIndex);
-
   const keydownListener = useCallback(
     (e: KeyboardEvent) => {
       if (e.key === "ArrowDown" && allItems.length) {
@@ -282,7 +280,7 @@ const NodeSearchMenu = ({
         e.stopPropagation();
       }
     },
-    [allItems, setActiveIndex, onSelect, onClose],
+    [allItems, activeIndex, setActiveIndex, onSelect, onClose],
   );
 
   useEffect(() => {
--- a/apps/roam/src/components/DiscourseNodeSearchMenu.tsx
+++ b/apps/roam/src/components/DiscourseNodeSearchMenu.tsx
@@ -30,6 +30,7 @@ type Props = {
   textarea: HTMLTextAreaElement;
   triggerPosition: number;
   onClose: () => void;
+  triggerText: string;
 };
 
 const waitForBlock = (
@@ -53,6 +54,7 @@ const NodeSearchMenu = ({
   onClose,
   textarea,
   triggerPosition,
+  triggerText,
 }: { onClose: () => void } & Props) => {
   const [activeIndex, setActiveIndex] = useState(0);
   const [searchTerm, setSearchTerm] = useState("");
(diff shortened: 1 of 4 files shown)


trangdoan982:
@coderabbitai review

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -189,27 +189,48 @@ export const initObservers = async ({
...
-        if (isBeginningOrAfterSpace) {
+    if (
+      target.tagName === "TEXTAREA" &&
+      target.classList.contains("rm-block-input")
+    ) {
+      const textarea = target as HTMLTextAreaElement;
+      const location = window.roamAlphaAPI.ui.getFocusedBlock();
+      if (!location) return;
+
+      const customTrigger = onloadArgs.extensionAPI.settings.get(
```
mdroidian: Let's move this outside of the listener.  Right now this is being called on every keystroke.

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -189,27 +189,48 @@ export const initObservers = async ({
...
-          cursorPos === 0 ||
-          textarea.value.charAt(cursorPos - 1) === " " ||
-          textarea.value.charAt(cursorPos - 1) === "\n";
-        if (isBeginningOrAfterSpace) {
+    if (
+      target.tagName === "TEXTAREA" &&
+      target.classList.contains("rm-block-input")
+    ) {
+      const textarea = target as HTMLTextAreaElement;
+      const location = window.roamAlphaAPI.ui.getFocusedBlock();
```
mdroidian: Ensure this is only called when `customTrigger` the `evt.key`.  Right now this is being called on every keystroke.

mdroidian (changes requested):
Good stuff.  A couple of quick fixes.

apps/roam/src/utils/initializeObserversAndListeners.ts
trangdoan982: i think it needs to check every keystroke because we're trying to match a pattern (the key could be multiple chars, similar to smartblock) instead of listening to just 1 key.

mdroidian: Can you walk me through the reason we need to call  `window.roamAlphaAPI.ui.getFocusedBlock()` on every keystroke?

trangdoan982: ahh i see what you mean. yes good point. I moved getFocusedBlock() check to the last step right before rendering the search menu to optimize

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -186,30 +186,52 @@ export const initObservers = async ({
...
+        const isValidTriggerPosition =
+          lastTriggerPos === 0 ||
+          charBeforeTrigger === " " ||
+          charBeforeTrigger === "\n";
+
+        const isCursorAfterTrigger =
+          cursorPos === lastTriggerPos + customTrigger.length;
+
+        if (isValidTriggerPosition && isCursorAfterTrigger) {
+          const location = window.roamAlphaAPI.ui.getFocusedBlock();
```
mdroidian: Upon further inspection, I'm unsure we actually need this at all.

If it is the case that we are using this as an additional check to be sure the user is actually editing a block (for some edge case that `target.tagName === "TEXTAREA" &&` and `target.classList.contains("rm-block-input")` do not cover, for example) then we should cooerce the value into a boolean (`!!`) and name it as such (eg: `isEditingBlock`).  Even add a comment about the specific use case.  This will help convey the use to other devs.

If it is not needed, we should probably remove it.

I'll approve the PR and leave the decision up to you.

apps/roam/src/components/DiscourseNodeSearchMenu.tsx
```diff
@@ -520,4 +526,38 @@ export const renderDiscourseNodeSearchMenu = (props: Props) => {
...
+      .trim();
+
+    setNodeSearchTrigger(trigger);
+    extensionAPI.settings.set("node-search-trigger", trigger);
+  };
+  return (
+    <InputGroup
+      value={nodeSearchTrigger}
+      onChange={handleNodeSearchTriggerChange}
+      placeholder="@"
```
mdroidian: As we are not currently setting a default, I think this would be confusing to the user.  Let's change it to "Click to set trigger" to match
image

mdroidian (approved):
Couple nits, feel free to edit them now or defer.

PR #163, opened 2025-05-17
DGAI: Integrate HyDe utility with discourse node suggestions component - ENG-293

sid597:
## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced suggestion functionality for discourse relationships, providing more relevant node suggestions using similarity-based search.
  - Added visual feedback with a loading spinner during suggestion searches and clearer messaging when no relevant relations are found.

- **Bug Fixes**
  - Improved error handling to ensure the suggestion list resets gracefully if a search fails.

### DIFF
--- a/apps/roam/src/components/DiscourseContextOverlay.tsx
+++ b/apps/roam/src/components/DiscourseContextOverlay.tsx
@@ -19,10 +19,8 @@ import nanoid from "nanoid";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
 import getDiscourseContextResults from "~/utils/getDiscourseContextResults";
 import findDiscourseNode from "~/utils/findDiscourseNode";
-import getDiscourseNodes, { DiscourseNode } from "~/utils/getDiscourseNodes";
-import getDiscourseRelations, {
-  DiscourseRelation,
-} from "~/utils/getDiscourseRelations";
+import getDiscourseNodes from "~/utils/getDiscourseNodes";
+import getDiscourseRelations from "~/utils/getDiscourseRelations";
 import ExtensionApiContextProvider from "roamjs-components/components/ExtensionApiContext";
 import { OnloadArgs } from "roamjs-components/types/native";
 import AutocompleteInput from "roamjs-components/components/AutocompleteInput";
@@ -34,7 +32,6 @@ import {
   findSimilarNodesUsingHyde,
   SuggestedNode,
   RelationDetails,
-  CandidateNodeWithEmbedding,
 } from "~/utils/hyde";
 
 type DiscourseData = {
@@ -152,87 +149,72 @@ const DiscourseContextOverlay = ({
     getInfo();
   }, [refresh, getInfo]);
 
-  const suggestionContextData = useMemo(() => {
-    const selfNode = discourseNode;
+  const validRelations = useMemo(() => {
+    if (!discourseNode) return [];
+    const selfType = discourseNode.type;
 
-    if (!selfNode || !selfNode.text || !selfNode.format) {
-      return {
-        validTypes: [],
-        uniqueRelationTypeTriplets: [],
-      };
-    }
-    const selfType = selfNode.type;
-
-    const relationsConnectingToSelf = relations.filter(
+    return relations.filter(
       (relation) =>
         relation.source === selfType || relation.destination === selfType,
     );
+  }, [relations, discourseNode]);
 
-    const uniqueTriplets = useMemo(() => {
-      const relatedNodeType = selfNode.type;
+  const uniqueRelationTypeTriplets = useMemo(() => {
+    if (!discourseNode) return [];
+    const relatedNodeType = discourseNode.type;
 
-      return relationsConnectingToSelf.flatMap((relation) => {
-        const isSelfSource = relation.source === relatedNodeType;
-        const isSelfDestination = relation.destination === relatedNodeType;
+    return validRelations.flatMap((relation) => {
+      const isSelfSource = relation.source === relatedNodeType;
+      const isSelfDestination = relation.destination === relatedNodeType;
 
-        let targetNodeType: string;
-        let currentRelationLabel: string;
+      let targetNodeType: string;
+      let currentRelationLabel: string;
 
-        if (isSelfSource) {
-          targetNodeType = relation.destination;
-          currentRelationLabel = relation.label;
-        } else if (isSelfDestination) {
-          targetNodeType = relation.source;
-          currentRelationLabel = relation.complement;
-        } else {
-          return [];
-        }
+      if (isSelfSource) {
+        targetNodeType = relation.destination;
+        currentRelationLabel = relation.label;
+      } else if (isSelfDestination) {
+        targetNodeType = relation.source;
+        currentRelationLabel = relation.complement;
+      } else {
+        return [];
+      }
 
-        const identifiedTargetNode = allNodes.find(
-          (node) => node.type === targetNodeType,
-        );
+      const identifiedTargetNode = allNodes.find(
+        (node) => node.type === targetNodeType,
+      );
 
-        if (!identifiedTargetNode) {
-          return [];
-        }
+      if (!identifiedTargetNode) {
+        return [];
+      }
 
-        const mappedItem: RelationDetails = {
-          relationLabel: currentRelationLabel,
-          relatedNodeText: identifiedTargetNode.text,
-          relatedNodeFormat: identifiedTargetNode.format,
-        };
-        return [mappedItem];
-      });
-    }, [relationsConnectingToSelf, selfNode.type, allNodes]);
+      const mappedItem: RelationDetails = {
+        relationLabel: currentRelationLabel,
+        relatedNodeText: identifiedTargetNode.text,
+        relatedNodeFormat: identifiedTargetNode.format,
+      };
+      return [mappedItem];
+    });
+  }, [validRelations, discourseNode, allNodes]);
 
-    const relationsInvolvingSelfBroadly = relations.filter((relation) =>
-      [relation.source, relation.destination, relation.label].includes(
-        selfType,
-      ),
-    );
-    const hasSelfRelation = relationsInvolvingSelfBroadly.some(
+  const validTypes = useMemo(() => {
+    if (!discourseNode) return [];
+    const selfType = discourseNode.type;
+
+    const hasSelfRelation = validRelations.some(
       (relation) =>
         relation.source === selfType && relation.destination === selfType,
     );
     const types = Array.from(
       new Set(
-        relationsInvolvingSelfBroadly.flatMap((relation) => [
+        validRelations.flatMap((relation) => [
           relation.source,
           relation.destination,
         ]),
       ),
     );
-    const filteredTypes = hasSelfRelation
-      ? types
-      : types.filter((type) => type !== selfType);
-
-    return {
-      validTypes: filteredTypes,
-      uniqueRelationTypeTriplets: uniqueTriplets,
-    };
-  }, [discourseNode, relations, allNodes]);
-
-  const { validTypes, uniqueRelationTypeTriplets } = suggestionContextData;
+    return hasSelfRelation ? types : types.filter((type) => type !== selfType);
+  }, [discourseNode, validRelations]);
 
   const [currentPageInput, setCurrentPageInput] = useState("");
   const [selectedPage, setSelectedPage] = useState<string | null>(null);
@@ -271,11 +253,27 @@ const DiscourseContextOverlay = ({
         setIsSearchingHyde(true);
         setHydeFilteredNodes([]);
         try {
-          const foundNodes = await runHydeSearch({
-            currentSuggestions: nodes,
-            currentNodeText: tag,
-            relationDetails: uniqueRelationTypeTriplets,
-          });
+          const candidateNodesForHyde = nodes.map((node) => ({
+            uid: node.uid,
+            text: node.text,
+            type: node.type,
+          }));
+
+          // TODO: Remove this once the HyDE search is working
+          const foundNodes: SuggestedNode[] =
+            await tempFindSimilarNodesUsingHyde({
+              candidateNodes: candidateNodesForHyde,
+              currentNodeText: tag,
+              relationDetails: uniqueRelationTypeTriplets,
+            });
+
+          // TODO: Uncomment this once the HyDE search is working
+          // const foundNodes: SuggestedNode[] = await findSimilarNodesUsingHyde({
+          //   candidateNodes: candidateNodesForHyde,
+          //   currentNodeText: tag,
+          //   relationDetails: uniqueRelationTypeTriplets,
+          // });
+
           setHydeFilteredNodes(foundNodes);
         } catch (error) {
           console.error(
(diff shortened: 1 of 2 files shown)


sid597:
@coderabbitai full review

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -87,6 +96,67 @@ const getAllReferencesOnPage = (pageTitle: string) => {
   })) as Result[];
 };
 
+type SelfNodeType = {
```
mdroidian: Is this type necessary?  Or can we use the existing `DiscourseNode` type which selfNode already is?

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -87,6 +96,67 @@ const getAllReferencesOnPage = (pageTitle: string) => {
...
+  type: string;
+  text: string;
+  format: string;
+};
+
+type AllNodesType = ReadonlyArray<SelfNodeType>;
```
mdroidian: Is this type necessary?  Or can we use the existing `DiscourseNode[]` type which allNodes already is?

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -87,6 +96,67 @@ const getAllReferencesOnPage = (pageTitle: string) => {
...
+    let targetNodeDetails: { text: string; format: string } | undefined;
+    let finalRelationLabel: string;
+
+    if (relation.source === relatedNodeType) {
+      targetNodeIdentifier = relation.destination;
+      finalRelationLabel = relation.label;
+      const destinationNode = allNodes.find(
+        (n) => n.type === targetNodeIdentifier,
+      );
+      if (destinationNode && destinationNode.text && destinationNode.format) {
```
mdroidian: If we use the existing type (`DiscourseNode`), I believe this check would be unnecessary. (or at least the additional `.text` and `.format`)

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -87,6 +96,67 @@ const getAllReferencesOnPage = (pageTitle: string) => {
...
+        targetNodeDetails = {
```
mdroidian: If we use the existing type (`DiscourseNode`), we could probably just pass the whole `DiscourseNode`

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -87,6 +96,67 @@ const getAllReferencesOnPage = (pageTitle: string) => {
...
+          text: destinationNode.text,
+          format: destinationNode.format,
+        };
+      }
+    } else if (relation.destination === relatedNodeType) {
+      targetNodeIdentifier = relation.source;
+      finalRelationLabel = relation.complement;
+      const sourceNode = allNodes.find((n) => n.type === targetNodeIdentifier);
+      if (sourceNode && sourceNode.text && sourceNode.format) {
```
mdroidian: If we use the existing type (`DiscourseNode`), I believe this check would be unnecessary. (or at least the additional `.text` and `.format`)

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -138,29 +213,56 @@ const DiscourseContextOverlay = ({
...
-    if (!discourseNode) return [];
-    const selfType = discourseNode.type;
-    const validRelations = relations.filter((relation) =>
-      [relation.source, relation.destination].includes(selfType),
+  const suggestionContextData = useMemo(() => {
+    const selfNode = discourseNode;
+
+    if (!selfNode || !selfNode.text || !selfNode.format) {
+      return {
+        validTypes: [] as string[],
```
mdroidian: Are these type assertions required?  If so, why?

### Some background on Type assertions
Type assertions should only be used when absolutely necessary and only when we're very confident about the value's type.

Using a type assertion is like telling TypeScript: "Trust me — I know what I'm doing, and you're wrong about this type."

It's a way to override TypeScript's type checking, so we should use it cautiously and only when we can't express the correct type in another way.

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -87,6 +96,67 @@ const getAllReferencesOnPage = (pageTitle: string) => {
...
 
+type SelfNodeType = {
+  type: string;
+  text: string;
+  format: string;
+};
+
+type AllNodesType = ReadonlyArray<SelfNodeType>;
+
+const getUniqueLabelTypeTriplets = (
```
mdroidian: remember: named params if more than 2

Also, if we don't need to export this, it could be created inside the function, memoized, and then the params are unnecessary.

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -170,33 +272,106 @@ const DiscourseContextOverlay = ({
...
+
+    try {
+      const candidateNodesForHyde = currentSuggestions.map((node) => {
+        const nodeWithEmbedding = {
+          uid: node.uid,
+          text: node.text,
+          type: node.type,
+          embedding: [] as number[],
+        };
+        return nodeWithEmbedding as CandidateNodeWithEmbedding;
```
mdroidian: I know I have previously suggested we use `Result` and work from there.  Generally that is the case.

It looks like in this case, however, that we cannot do that because `embedding` is an array and `Result` does not accept that.

So for this case we could change `CandidateNodeWithEmbedding` to be it's own type which would allow us to define the variable as this type instead of type asserting.

```typescript
export type CandidateNodeWithEmbedding = {
  uid: string;
  text: string;
  type: string;
  embedding: number[];
};
```

If we continue to use `CandidateNodeWithEmbedding`, we should do this.  (which depends on [this](https://github.com/DiscourseGraphs/discourse-graph/pull/163#discussion_r2096524403) comment)

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -170,33 +272,106 @@ const DiscourseContextOverlay = ({
...
+      return [];
+    }
+
+    try {
+      const candidateNodesForHyde = currentSuggestions.map((node) => {
+        const nodeWithEmbedding = {
+          uid: node.uid,
+          text: node.text,
+          type: node.type,
+          embedding: [] as number[],
```
mdroidian: Can you show me where this value (`embedding`) gets updated/used?

I spent a good deal of time searching but could not find it.

If it not used, we can simplify the types.

mdroidian (changes requested):
Multiple clarifying questions/context needed which will likely trigger changes and another review.

mdroidian:
Reminder: use the linear `copy git branch name` when creating the branch. (as specified in [DevOps](https://linear.app/discourse-graphs/document/project-handbook-dd139f6e26fc#heading-devops-aec020c3))

image

mdroidian:
Reminder: [add a loom video to the PR](https://linear.app/discourse-graphs/document/project-handbook-dd139f6e26fc#heading-pr-workflow-guidelines-38d89c91) to help explain the context.  Especially for anything user-facing.

apps/roam/src/components/DiscourseContextOverlay.tsx
sid597: Yeah we are not using it anywhere

sid597:
> Reminder: [add a loom video to the PR](https://linear.app/discourse-graphs/document/project-handbook-dd139f6e26fc#heading-pr-workflow-guidelines-38d89c91) to help explain the context. Especially for anything user-facing.

I want to defer this to a later pr maybe the api routes one, the functionality would not work on this one, would require some merging to another branch resolving conflits then demoing it, other way is using dummy functions but then would have to remove them. Both cases require work that would only be for demo and would also be done again in a later branch anyway.

mdroidian:
> > Reminder: [add a loom video to the PR](https://linear.app/discourse-graphs/document/project-handbook-dd139f6e26fc#heading-pr-workflow-guidelines-38d89c91) to help explain the context. Especially for anything user-facing.
> 
> I want to defer this to a later pr maybe the api routes one, the functionality would not work on this one, would require some merging to another branch resolving conflits then demoing it, other way is using dummy functions but then would have to remove them. Both cases require work that would only be for demo and would also be done again in a later branch anyway.

@sid597 This is all context that should be added to the PR so that the reviewer (or anyone viewing it for that matter) would understand the intentions and choices made.

mdroidian:
@coderabbit full review

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -18,15 +19,23 @@ import nanoid from "nanoid";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
 import getDiscourseContextResults from "~/utils/getDiscourseContextResults";
 import findDiscourseNode from "~/utils/findDiscourseNode";
-import getDiscourseNodes from "~/utils/getDiscourseNodes";
-import getDiscourseRelations from "~/utils/getDiscourseRelations";
+import getDiscourseNodes, { DiscourseNode } from "~/utils/getDiscourseNodes";
```
mdroidian: Why are we importing the `DiscourseNode` type if we are not going to use it?

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -18,15 +19,23 @@ import nanoid from "nanoid";
...
 import AutocompleteInput from "roamjs-components/components/AutocompleteInput";
 import getAllPageNames from "roamjs-components/queries/getAllPageNames";
 import { Result } from "roamjs-components/types/query-builder";
 import createBlock from "roamjs-components/writes/createBlock";
 import { getBlockUidFromTarget } from "roamjs-components/dom";
+import {
+  findSimilarNodesUsingHyde,
+  SuggestedNode,
+  RelationDetails,
+  CandidateNodeWithEmbedding,
```
mdroidian: Why are we importing `CandidateNodeWithEmbedding` if we are not using it?

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -138,29 +152,88 @@ const DiscourseContextOverlay = ({
...
+      };
+    }
+    const selfType = selfNode.type;
+
+    const relationsConnectingToSelf = relations.filter(
+      (relation) =>
+        relation.source === selfType || relation.destination === selfType,
+    );
+
+    const uniqueTriplets = useMemo(() => {
```
mdroidian: `useMemo` inside another `useMemo` doesn’t give us any performance benefit — the inner memoization will rerun every time the outer one does. In most cases, this just adds complexity without gain.

Let’s refactor this section with clarity and separation of concerns in mind.
- Start with `const validTypes` as it was before this PR.
- Identify what problem you are trying to solve, write it down.
  - What inputs do you need? What steps are involved?
- Identify how you are going to solve that problem, write it down.
- Scan for overlap: Is this logic already partly being done somewhere else, like inside the calculation of `validTypes`?
  - If the answer is `yes` and that value lives inside of `validTypes`, pull it out of `validTypes` into it's own `const` so you can use it.
- Create a new `const` (ouside of `validTypes`) that returns the value you need and place it below `validTypes`

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -138,29 +152,88 @@ const DiscourseContextOverlay = ({
...
-  // Suggestive Mode
-  const validTypes = useMemo(() => {
-    if (!discourseNode) return [];
-    const selfType = discourseNode.type;
-    const validRelations = relations.filter((relation) =>
-      [relation.source, relation.destination].includes(selfType),
+  const suggestionContextData = useMemo(() => {
+    const selfNode = discourseNode;
+
+    if (!selfNode || !selfNode.text || !selfNode.format) {
```
mdroidian: same as [this](https://github.com/DiscourseGraphs/discourse-graph/pull/163#discussion_r2096426436)

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -138,29 +152,88 @@ const DiscourseContextOverlay = ({
...
+      return {
+        validTypes: [],
+        uniqueRelationTypeTriplets: [],
+      };
+    }
+    const selfType = selfNode.type;
+
+    const relationsConnectingToSelf = relations.filter(
```
mdroidian: If you rename/refactor a variable/function, but don't change it's purpose, make sure to leave a comment as to why this change was necessary (just in GitHub in the PR, you can click on the `+` to leave a comment on this specific line).

This helps the review understand functionally irrelevant changes much quicker.

apps/roam/src/components/DiscourseContextOverlay.tsx
```diff
@@ -170,33 +243,102 @@ const DiscourseContextOverlay = ({
...
+    setHydeFilteredNodes(hydeFilteredNodes.filter((n) => n.uid !== node.uid));
+  };
+
+  type RunHydeSearchArgs = {
+    currentSuggestions: SuggestedNode[];
+    currentNodeText: string;
+    relationDetails: RelationDetails[];
+  };
+
+  const runHydeSearch = async ({
```
mdroidian: What is the purpose of this function?  To transform `currentSuggestions` then run `findSimilarNodesUsingHyde`?  Seems like this could be done inside the already existing `performSearch`. Then `runHydeSearch` could be removed and we could reduce complexity.

apps/roam/src/utils/hyde.ts
```diff
@@ -308,8 +307,7 @@ const rankNodes = ({
   maxScores.forEach((score, uid) => {
     const fullNode = nodeMap.get(uid);
     if (fullNode) {
-      const { embedding, ...restOfNode } = fullNode;
-      const suggestedNodeObject: SuggestedNode = restOfNode;
+      const suggestedNodeObject: SuggestedNode = fullNode;
```
mdroidian: The `suggestedNodeObject` variable seems redundant here — you're not modifying or reusing it.

mdroidian (changes requested):
https://www.loom.com/share/af0f0551346a4b07b0d122eef86d3121

I get this error on trying to load just the overlay icon.
This means that the rest of the UI cannot be reviewed because it doesn't load.
We also can't confirm that the code is functional in any meaningful sense.

If this is intentional, please comment in the code (or in the GitHub PR UI) where/why this is happening, and how this works into the overall flow.

If a simple stub function can fix this, please add it.  Eg:

```
PR Body / Loom Video:

This PR adds the UI and front end scaffolding for the upcoming backend HyDe search API's.  There is a stub function that just returns the suggested nodes without sorting them.
```

```typescript
  // TODO: Remove this once the HyDE search is working
  const tempFindSimilarNodesUsingHyde = async ({
    candidateNodes,
    currentNodeText,
    relationDetails,
  }: {
    candidateNodes: SuggestedNode[];
    currentNodeText: string;
    relationDetails: RelationDetails[];
  }): Promise<SuggestedNode[]> => {
    return candidateNodes;
  };

  // TODO: Remove this once the HyDE search is working
  const foundNodes: SuggestedNode[] = await tempFindSimilarNodesUsingHyde({
    candidateNodes: candidateNodesForHyde,
    currentNodeText: currentNodeText,
    relationDetails: relationDetails,
  });
  // TODO: Uncomment this once the HyDE search is working
  // const foundNodes: SuggestedNode[] = await findSimilarNodesUsingHyde({
  //   candidateNodes: candidateNodesForHyde,
  //   currentNodeText: currentNodeText,
  //   relationDetails: relationDetails,
  // });
```

sid597:
https://www.loom.com/share/aa5b0523f6ed4d8c829219e70a22ec33?sid=6f258300-ecdd-4492-a959-438d846e8c3a

@mdroidian addressed all the review points, previously I was using cursor for all the work, I reviewed the code, parts of it seemed correct, things which did not I asked it to correct, it corrected the part i wanted it to fix but left the unused code from previous run. In my mind it was reviewed and only some new function was added which lead to these situation. 

For this run I actually reviewed using my brains instead of outsourcing to llm, I think its correct this time and covers the practices we want to follow.

sid597:
@coderabbitai full review

PR #168, opened 2025-05-20
[ENG-300] Update plugin ID to fix install on BRAT

trangdoan982:
- Update the id in `manifest.json` before publishing to the target repo (discourse-graphs-obsidian), so that BRAT can install the plugin correctly


mdroidian (approved):
👍

PR #179, opened 2025-05-26
[ENG-350] Obsidian - Add template to node type setting

trangdoan982:
https://www.loom.com/share/9a5c3b09a6a4458d8592da767af17262

## Summary by CodeRabbit

- **New Features**
  - Users can now associate a template with each node type in the Discourse Graph plugin, leveraging templates from Obsidian's core Templates plugin.
  - When creating a new node, the selected template is automatically applied to the node file if specified.

- **Documentation**
  - Expanded instructions on configuring node types, including how to enable and use templates, with new images and step-by-step guidance.


trangdoan982:
@coderabbitai review

apps/obsidian/src/utils/templates.ts
```diff
@@ -0,0 +1,100 @@
...
+    console.error("Error accessing Templates plugin:", error);
+    return { isEnabled: false, folderPath: "" };
+  }
+};
+
+export const getTemplateFiles = async (app: App): Promise<string[]> => {
+  try {
+    const { isEnabled, folderPath } = getTemplatePluginInfo(app);
+
+    if (!isEnabled) {
```
mdroidian: nit: if you did `if (!isEnabled) return []` you could reduce this to 2 lines instead of 7.  Or combine them with `||` into one line.
I find it easier to read, but that is just my preference.

apps/obsidian/src/utils/templates.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const templateFolder = app.vault.getAbstractFileByPath(folderPath);
+
+    if (!templateFolder || !(templateFolder instanceof TFolder)) {
+      return [];
+    }
+
+    const templateFiles = templateFolder.children
+      .filter(
+        (file: any): file is TFile =>
```
mdroidian: `any` should be avoided wherever possible.  In this case we know that `file` is of type `TAbstractFile`. 

```typescript
    const templateFiles = templateFolder.children
      .filter(
        (file: TAbstractFile): file is TFile =>
          file instanceof TFile && file.extension === "md",
      )
      .map((file) => file.basename)
      .sort();
```

apps/obsidian/src/utils/templates.ts
```diff
@@ -0,0 +1,100 @@
...
+      isEnabled: true,
+      folderPath,
+    };
+  } catch (error) {
+    console.error("Error accessing Templates plugin:", error);
+    return { isEnabled: false, folderPath: "" };
+  }
+};
+
+export const getTemplateFiles = async (app: App): Promise<string[]> => {
```
mdroidian: What part of this function is `async`?

apps/obsidian/src/components/NodeTypeSettings.tsx
```diff
@@ -17,6 +18,15 @@ const NodeTypeSettings = () => {
   );
   const [formatErrors, setFormatErrors] = useState<Record<number, string>>({});
   const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false);
+  const [templateFiles, setTemplateFiles] = useState<string[]>([]);
+
+  useEffect(() => {
+    const loadTemplateFiles = async () => {
```
mdroidian: Does this need to be `async`?

```typescript
    const files = getTemplateFiles(plugin.app);
    setTemplateFiles(files);
```

apps/obsidian/src/utils/templates.ts
```diff
@@ -0,0 +1,100 @@
...
+      .sort();
+
+    return templateFiles;
+  } catch (error) {
+    console.error("Error getting template files:", error);
+    return [];
+  }
+};
+
+export const applyTemplate = async (
```
mdroidian: named params

apps/obsidian/src/utils/templates.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (!templateFile || !(templateFile instanceof TFile)) {
+      console.warn(`Template file not found: ${templateFilePath}`);
+      return false;
+    }
+
+    const templateContent = await app.vault.read(templateFile);
+
+    const currentContent = await app.vault.read(targetFile);
+
+    const newContent = currentContent + templateContent;
```
mdroidian, line 91: Let's bring this up with the UX team to see if this is the expected behavior.

apps/obsidian/src/constants.ts
```diff
@@ -6,16 +6,19 @@ export const DEFAULT_NODE_TYPES: Record<string, DiscourseNode> = {
     id: generateUid("node"),
     name: "Question",
     format: "QUE - {content}",
+    template: "",
```
mdroidian: nit: If this is optional, what is the purpose of putting an empty string here?

mdroidian (approved):
Nice.

Only blocking comments are the `async` function and named params.
Going to approve this so you can merge when complete.

Also, might be worth creating a task to enhance the UX when a user is trying to set a template but doesn't have any templates defined / template plugin is disabled.  I know it's in the readme, but right in the UI somewhere would be ideal.

We could disable the dropdown and have a message on hover, for example.

image

PR #184, opened 2025-05-30
repairs to make routes work; some local testing in ipynb

maparent:
Note: PR churn because of an error naming the branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ yarn-error.log*
 # Misc
 .DS_Store
 *.pem
+*.ipynb
--- a/apps/website/app/utils/supabase/apiUtils.ts
+++ b/apps/website/app/utils/supabase/apiUtils.ts
@@ -30,7 +30,7 @@ export const createApiResponse = <T>(
       { status },
     );
   } else if (data !== undefined && data !== null) {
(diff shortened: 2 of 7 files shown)


maparent:
Coderabbit review done before branch rename. It wanted my ipynb to be more like a full test. We should discuss where and how to make a test suite.

mdroidian:
> We should discuss where and how to make a test suite.

We don't have a standard for testing yet.  Yes, we should discuss this and decide on one.  For the APIs, just manually test until we have this in place.

packages/database/tests.ipynb
```diff
@@ -0,0 +1,204 @@
+{
```
mdroidian: It is unlikely we will using `ipynb`'s as a standard part of our workflow, so let's not commit this to the repo.

If you want to use this locally for yourself to confirm routes are working and such, that is fine.  In that case, let's add `tests.ipynb` to the `gitignore`.

mdroidian (approved):
Just 1 blocking comment.  I've approved the changes so that you can merge once the `tests.ipynb` is removed.

packages/database/tests.ipynb
maparent: done.

PR #194, opened 2025-06-06
[ENG-410] Node creation dialog

trangdoan982:
https://www.loom.com/share/912e32efc0ba4fceacfe322cbac9bce5

## Summary by CodeRabbit

- **New Features**
  - Introduced a modal for creating nodes without requiring a text selection.
  - Added a new command to create nodes directly from the command palette.

- **Improvements**
  - Enhanced the node creation workflow, allowing users to select node type and enter a title in a user-friendly modal.
  - Streamlined command behavior to support node creation both with and without selected text.

- **Bug Fixes**
  - Improved input validation and error handling during node creation.


trangdoan982:
@coderabbitai full review

apps/obsidian/src/components/CreateNodeModal.tsx
```diff
@@ -0,0 +1,143 @@
+import { App, Modal, Notice } from "obsidian";
+import { DiscourseNode } from "~/types";
+import type DiscourseGraphPlugin from "~/index";
+
+interface CreateNodeModalProps {
```
mdroidian: Let's use `type` to be consistent with the repo

apps/obsidian/src/components/CreateNodeModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+    this.nodeTypes = props.nodeTypes;
+    this.plugin = props.plugin;
+    this.onNodeCreate = props.onNodeCreate;
+  }
+
+  onOpen() {
+    const { contentEl } = this;
+    contentEl.empty();
+
+    contentEl.createEl("h2", { text: "Create node" });
```
mdroidian: "Create Discourse Node"

apps/obsidian/src/components/CreateNodeModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+  private confirmButton: HTMLButtonElement | null = null;
+
+  constructor(app: App, props: CreateNodeModalProps) {
+    super(app);
+    this.nodeTypes = props.nodeTypes;
+    this.plugin = props.plugin;
+    this.onNodeCreate = props.onNodeCreate;
+  }
+
+  onOpen() {
```
mdroidian, line 172: One thing to considering here would be returning a react component.
This would allow us to start creating reusable components for common things (`Input`, `Button`, etc), use tailwind, and it would also be easier maintain (less verbose).

Can be deferred.

apps/obsidian/src/components/CreateNodeModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    contentEl.createEl("h2", { text: "Create node" });
+
+    const titleContainer = contentEl.createDiv({ cls: "setting-item" });
+    titleContainer.createEl("div", { text: "Title", cls: "setting-item-name" });
+    const titleInputContainer = titleContainer.createDiv({
+      cls: "setting-item-control",
+    });
+
+    this.titleInput = titleInputContainer.createEl("input", {
```
mdroidian: This felt quite small, especially considering how long some of the nodes in `megacoglab` and `akamatsulab` are.  Likely we'll more room here so the user can see as they type out node names.

Can be deferred.

apps/obsidian/src/utils/createNode.ts
```diff
@@ -130,9 +132,29 @@ export const processTextToDiscourseNode = async ({
...
-  if (newFile) {
+
+  if (newFile && editor) {
     editor.replaceSelection(`[[${formattedNodeName}]]`);
   }
 
   return newFile;
 };
+
+export const processTextToDiscourseNode = async ({
```
mdroidian: We don't need an extra function for this.  We could just use `createDiscourseNode`, make `text` optional and grab the selection if it's null.  `editor` can be required.

```typescript
const nodeText = text ?? editor.getSelection().trim();
```

apps/obsidian/src/utils/registerCommands.ts
```diff
@@ -1,19 +1,52 @@
...
 export const registerCommands = (plugin: DiscourseGraphPlugin) => {
   plugin.addCommand({
     id: "open-node-type-menu",
     name: "Open Node Type Menu",
     hotkeys: [{ modifiers: ["Mod"], key: "\\" }],
     editorCallback: (editor: Editor) => {
-      if (!editor.getSelection()) {
-        new Notice("Please select some text to create a discourse node", 3000);
-        return;
+      const hasSelection = editor.getSelection();
```
mdroidian: Could add `!!` if only using as boolean

```typescript
const hasSelection = !!editor.getSelection();
```

mdroidian (approved):
Nice. 🚀

Couple of small suggestions, then feel free to merge.

PR #196, opened 2025-06-09
ENG-430 Add access-token table

mdroidian:
To be used for GitHub Apps initially.

```
node -e "if (process.env.HOME !== '/vercel') { require('child_process').execSync('supabase start && supabase migrations up && (supabase gen types typescript --local --schema public > types.gen.ts)', { stdio: 'inherit' }) }"`
```
image

## Summary by CodeRabbit

- **New Features**
	- Introduced a new "access-token" table to the database, enabling storage and management of access tokens with associated codes, states, and timestamps.

- **Bug Fixes**
	- Not applicable.

- **Documentation**
	- Not applicable.

- **Chores**
	- Updated database configuration and type definitions to reflect the new "access-token" table.

### DIFF
--- a/packages/database/supabase/config.toml
+++ b/packages/database/supabase/config.toml
@@ -59,6 +59,7 @@ schema_paths = [
     './schemas/contributor.sql',
     './schemas/sync.sql',
     './schemas/upload_temp.sql',
+    './schemas/access_token.sql',
 ]
 
 [db.seed]
--- a/packages/database/supabase/migrations/20250609051037_create_access_table.sql
+++ b/packages/database/supabase/migrations/20250609051037_create_access_table.sql
@@ -1,68 +1,58 @@
-create table "public"."access-token" (
+create table "public"."access_token" (
     "id" bigint generated always as identity not null,
-    "access-token" text not null,
+    "access_token" text not null,
     "code" text,
     "state" text,
     "created_date" timestamp with time zone not null default timezone('utc'::text, now())
 );
 
 
-CREATE UNIQUE INDEX "access-token_pkey" ON public."access-token" USING btree (id);
+CREATE UNIQUE INDEX access_token_access_token_idx ON public.access_token USING btree (access_token);
 
-CREATE UNIQUE INDEX access_token_access_token_idx ON public."access-token" USING btree ("access-token");
+CREATE INDEX access_token_code_idx ON public.access_token USING btree (code);
 
-CREATE INDEX access_token_code_idx ON public."access-token" USING btree (code);
+CREATE INDEX access_token_created_date_idx ON public.access_token USING btree (created_date DESC);
 
-CREATE INDEX access_token_created_date_idx ON public."access-token" USING btree (created_date DESC);
+CREATE UNIQUE INDEX access_token_pkey ON public.access_token USING btree (id);
 
-CREATE INDEX access_token_state_idx ON public."access-token" USING btree (state);
+CREATE INDEX access_token_state_idx ON public.access_token USING btree (state);
 
-alter table "public"."access-token" add constraint "access-token_pkey" PRIMARY KEY using index "access-token_pkey";
+alter table "public"."access_token" add constraint "access_token_pkey" PRIMARY KEY using index "access_token_pkey";
 
-alter table "public"."access-token" add constraint "access_token_code_state_check" CHECK (((code IS NOT NULL) OR (state IS NOT NULL))) not valid;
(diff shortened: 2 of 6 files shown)


mdroidian:
@coderabbit full review

maparent:
Ok, a few questions: Are access tokens unique for spaces? In which case should the Id not be the foreign key to the state instead of an independent identity? 
If they are indeed not unique, would they have an owner? (reference to PlatformAccount)
And would they still not be reference a space in any case?
Either way the space should be indexed, as it is the  most likely point of access.
What is the semantic of code and state? Should those two have an index? (I.e. are they likely to be the identifiers in a query, vs a filter? There is a performance cost to unused indices.)
I think I need to understand more about the intent of the table.
On another note: Right now I have a SpaceAccess table, which has a "editor" boolean. I'm thinking we'll need three states: Reader, Editor, Admin. The latter would gate access to the token table.
I have another thought: Should we instead have a "SpaceAdminSecrets" table with a json or columns? Is there any chance we'll go beyond tokens someday? I'm thinking of state-specific access token for LLM providers, in particular. (which is open-ended, and why I'm leaning on json, but not strongly.)

maparent:
Another issue, on me: I'm maintaining the linkml/svg with tooling that I did not include in the build, because I did not want to impose it to everyone to install. As long as I was mostly manipulating the tables, not an issue. Do you have an opinion on whether I should include it or keep maintaining it in the background?

mdroidian:
> I think I need to understand more about the intent of the table.

For this use case it is authorizing a GitHub App to act on users behalf to sync a Discourse Node to a GitHub Issue.
Flow: 
- user clicks "Install App", generate `state`, get popup
- user installs app on repo
- popup window redirects to our callback page with `code`
- callback page uses `code` to get `accessToken`
- on success
  - save to db
  - post `accessToken` back to browser, store in app
  - when app is used instead of browser, we cannot post back, so client periodically calls table for `state`

> Ok, a few questions: Are access tokens unique for spaces? 

Not necessarily, they could possibly have a PlatformAccount in the future.  Other uses for this table might be tied to spaces.  In either case, we'll address that when if/when it shows up.

> What is the semantic of code and state? Should those two have an index?

Both will be used to query the table and return `accessToken`.

> On another note: Right now I have a SpaceAccess table, which has a "editor" boolean. I'm thinking we'll need three states: Reader, Editor, Admin. The latter would gate access to the token table. 

I don't see the need currently.

>I have another thought: Should we instead have a "SpaceAdminSecrets" table with a json or columns? 

I don't think so.

> Is there any chance we'll go beyond tokens someday? I'm thinking of state-specific access token for LLM providers, in particular. 

Maybe.  We'll cross that bridge when we come to it.

mdroidian:
> Another issue, on me: I'm maintaining the linkml/svg

Any feature in the repo should not be dependent on a single dev.  If we keep it, instructions need to be included.  Ideally this would be automated by an `npm` package and not require additional local tooling.

icejubjub:
i need use access-token to sign in appMonopolyGo can you help me ?

mdroidian:
@maparent added `platform_account_id`, awaiting review.

maparent:
Still trying to understand some of the details of part of the flow:

- user clicks "Install App", generate `state`, get popup _Click presumably in frontend, is popup also under control of frontend?_
- user installs app on repo _That's on GitHub, right?_
- popup window redirects to our callback page with `code` _Who generates `code`? Github or popup?_
- callback page uses code to get `accessToken` _assuming that's a vercel function calling GitHub... but through a browser page. `accessToken from github`_
- on success
   - save to db _still Vercel function_
   - post accessToken back to browser, store in app _inter-page channel?_
   - when app is used instead of browser, we cannot post back, so client periodically calls table for `state`. _I think you mean using state as a key.?_

Anyhow the reason I'm asking for details, besides clarity, is I'm trying to see if there's a primary key in there besides the numeric identity. I suspect `state` is the primary key, and what is used to make queries, or is it `code`? Would it make sense to use `state` directly as a primary key and not use a numeric identity at all?

maparent:
> > Another issue, on me: I'm maintaining the linkml/svg
> 
> Any feature in the repo should not be dependent on a single dev. If we keep it, instructions need to be included. Ideally this would be automated by an `npm` package and not require additional local tooling.

That's why I have not done it yet, LinkML is a python package, no NPM version. I could probably make it run using minimal tooling installs, but I want an ok for installing non-npm tools. Same with plantuml, there's a npm version but it's quite outdated. Npm is a bottleneck.

maparent:
Other than the primary key issue, the rest looks good.

mdroidian:
Reviewed live by @maparent

PR #201, opened 2025-06-09
ENG-417 make sure the temporary upsert functions live in the schema

maparent:
as well as in the migrations. 
A `supabase db diff` should return nothing after merge in general.


maparent:
This is just duplicating code over from one place to another, no need for a review (beyond of the principle of this being best practice)

mdroidian (approved):

PR #204, opened 2025-06-10
ENG-417 forgot type generation stage

maparent:


mdroidian:
Taken care of with #196

PR #206, opened 2025-06-12
document sync functions

maparent:


maparent:
will adjust with function changes.

PR #211, opened 2025-06-13
A newer turbo crept in the package-lock, revert.

maparent:
I was experimenting with turbo versions in
bb2b6d9c72d7ce0d69be0b2916e172f598790411
I reverted the package.json and npm installed,
but the newer version stuck in the lock.

The newer turbo version seems to work fine on
vercel, but locally I have issues with the
combineStyles step of the Obsidian compile.
Commenting out compile.ts:120 makes the error
go away but is clearly not a solution.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19576,9 +19576,9 @@
       "license": "0BSD"
     },
     "node_modules/tsx": {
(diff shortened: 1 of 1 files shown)


mdroidian (approved):
👍

PR #218, opened 2025-06-17
[ENG-412] Node creation popover menu

trangdoan982:
https://www.loom.com/share/ea08a74d1e584f8ab58702b70cf7bf27

## Summary by CodeRabbit

- **New Features**
  - Introduced a popup menu that appears when text is selected within a block, providing quick access to relevant actions.
  - Added keyboard navigation enhancements for the popup menu, supporting up/down arrow navigation.
  - Popup menu now restores and focuses the original text selection when triggered.
- **Style**
  - Added new styles for the text selection popup, improving its appearance and visibility in the interface.


trangdoan982:
@coderabbit full review

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -111,42 +117,69 @@ const NodeMenu = ({
           menuRef.current?.getAttribute("data-active-index"),
         );
         onSelect(index);
-        // Remove focus from the block to ensure updateBlock works properly
```
mdroidian, line 114: Let's keep this comment, as it explains why we are doing `document.body.click()`

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -111,42 +117,69 @@ const NodeMenu = ({
...
+        document.body.click();
+      } else if (e.key === "Escape") {
+        onClose();
         document.body.click();
       } else if (shortcuts.has(e.key.toUpperCase())) {
         onSelect(indexBySC[e.key.toUpperCase()]);
-        // Remove focus from the block to ensure updateBlock works properly
```
mdroidian, line 118: Let's keep this comment, as it explains why we are doing `document.body.click()`

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -90,17 +95,18 @@ const NodeMenu = ({
...
       onClose();
     },
-    [menuRef, blockUid, onClose],
+    [menuRef, blockUid, onClose, textarea, extensionAPI],
   );
+
   const keydownListener = useCallback(
     (e: KeyboardEvent) => {
-      if (e.key === "ArrowRight" || e.key === "ArrowDown") {
+      if (e.key === "ArrowDown") {
```
mdroidian, line 103: Any specific reason ArrowRight/ArrowLeft was removed?

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -111,42 +117,69 @@ const NodeMenu = ({
...
     },
-    [menuRef, setActiveIndex],
+    [onSelect, onClose, indexBySC],
   );
   useEffect(() => {
-    textarea.addEventListener("keydown", keydownListener);
-    textarea.addEventListener("input", onClose);
+    const eventTarget = trigger ? document : textarea;
+    const keydownHandler = (e: Event) => {
+      if (trigger && e instanceof KeyboardEvent) {
```
mdroidian: When would `e` not be a `KeyboardEvent`?

Regardless, the logic of the `keydownHandler` will always call `keydownListener(e)`

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -62,6 +62,25 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
...
+        onChange={(e) => {
+          const target = e.target as HTMLInputElement;
+          extensionAPI.settings.set("text-selection-popup", target.checked);
+        }}
+        labelElement={
+          <>
+            Text Selection Popup
+            <Description
+              description={
+                "Whether or not to show the discourse node creation popup when selecting text."
```
mdroidian: always capitalize Discourse Node

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -240,6 +247,57 @@ export const initObservers = async ({
     }
   };
 
+  const selectionChangeListener = () => {
+    // Check if text selection popup is enabled (default to true for backward compatibility)
```
mdroidian: Backward compatibility to what?  This feature is being introduced now.

apps/roam/src/utils/renderTextSelectionPopup.tsx
```diff
@@ -0,0 +1,81 @@
...
+      break;
+    }
+    currentElement = currentElement.parentElement;
+  }
+
+  return blockElement;
+};
+
+export const renderTextSelectionPopup = (
+  extensionAPI: OnloadArgs["extensionAPI"],
```
mdroidian: named params

apps/roam/src/utils/renderTextSelectionPopup.tsx
```diff
@@ -0,0 +1,81 @@
...
+  blockElement?: Element | null,
```
mdroidian: When would blockElement be `null`?

mdroidian: And if `blockElement` is never `null`, then we can remove some redudant code below, including `targetblockElement.querySelector("textarea")` as this was just ran before calling `renderTextSelectionPopup`

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -200,6 +233,51 @@ export const render = (props: Props) => {
...
+            height="19"
+            viewBox="0 0 18 19"
+            fill="none"
+            xmlns="http://www.w3.org/2000/svg"
+          >
+            <path
+              fillRule="evenodd"
+              clipRule="evenodd"
+              d="M11.0389 17.7527C9.91285 18.8786 8.08718 18.8786 6.96115 17.7527L0.844529 11.6361C-0.281509 10.51 -0.28151 8.68435 0.844528 7.55832L3.90284 4.50001C4.46586 3.93698 5.37868 3.93698 5.94171 4.50001C6.50473 5.06302 6.50473 5.97585 5.94171 6.53888L4.92228 7.55832C3.79623 8.68435 3.79623 10.51 4.92228 11.6361L7.98057 14.6944C8.5436 15.2574 9.45643 15.2574 10.0194 14.6944C10.5825 14.1313 10.5825 13.2185 10.0194 12.6555L9.00001 11.6361C7.87398 10.51 7.87398 8.68435 9.00001 7.55832C10.1261 6.43227 10.1261 4.6066 9.00001 3.48057L7.98057 2.46114C7.41756 1.89812 7.41756 0.985283 7.98057 0.422263C8.5436 -0.140755 9.45643 -0.140754 10.0194 0.422265L17.1555 7.55832C18.2815 8.68435 18.2815 10.51 17.1555 11.6361L11.0389 17.7527ZM14.0972 8.57776C13.5342 8.01473 12.6213 8.01473 12.0583 8.57776C11.4953 9.14077 11.4953 10.0536 12.0583 10.6166C12.6213 11.1796 13.5342 11.1796 14.0972 10.6166C14.6602 10.0536 14.6602 9.14077 14.0972 8.57776Z"
```
mdroidian: We should try and reuse the path, if possible: https://github.com/DiscourseGraphs/discourse-graph/blob/eng-412-node-popover-create/apps/roam/src/components/BirdEatsBugs.tsx#L174

apps/roam/src/styles/discourseGraphStyles.css
```diff
@@ -177,3 +177,24 @@ div.roamjs-discourse-drawer div.bp3-drawer {
...
+
+/* Text Selection Popup Styles */
+.discourse-text-selection-popup {
+  position: fixed;
+  z-index: 9999;
+  max-width: none;
+  font-family: inherit;
+}
+
+.discourse-text-selection-popup .bp3-button {
```
mdroidian: can we inline these as tailwind on the `<Button>`?

apps/roam/src/styles/discourseGraphStyles.css
```diff
@@ -177,3 +177,24 @@ div.roamjs-discourse-drawer div.bp3-drawer {
...
+  background: white;
+  border: 1px solid #d3d8de;
+  border-radius: 4px;
+  box-shadow: 0 2px 8px rgba(0, 0, 0, 0.15);
+  padding: 4px 8px;
+}
+
+.discourse-text-selection-popup .bp3-button:hover {
```
mdroidian: can we inline these as tailwind on the `<Button>`?

apps/roam/src/styles/discourseGraphStyles.css
```diff
@@ -177,3 +177,24 @@ div.roamjs-discourse-drawer div.bp3-drawer {
 .bp3-popover-wrapper + .referenced-node-autocomplete {
   margin-top: 1rem;
 }
+
+/* Text Selection Popup Styles */
+.discourse-text-selection-popup {
```
mdroidian: Let's co locate these with the element.  Looks like there are some crossovers already:

image

You can use tailwind classes or inline them with styles.

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -240,6 +247,57 @@ export const initObservers = async ({
...
+
+    if (blockElement) {
+      const textarea = blockElement.querySelector("textarea");
+      renderTextSelectionPopup(onloadArgs.extensionAPI, blockElement, textarea);
+    } else {
+      removeTextSelectionPopup();
+    }
+  };
+
+  const documentClickListener = (e: Event) => {
```
mdroidian: Is this listener required?  Or is this taken care of by the `selectionChangeListener`?  When a user clicks away, isn't the selection removed, which calls `removeTextSelectionPopup()`?

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -240,6 +247,57 @@ export const initObservers = async ({
     }
   };
 
+  const selectionChangeListener = () => {
```
mdroidian: Let's be more specific about what this listener's function is.

Take a look at some of the other listener names:
`nodeMenuTriggerListener`
`discourseNodeSearchTriggerListener`

mdroidian (changes requested):
A few requested changes and comments, but great stuff!

I'm wondering if we couldn't directly call the `NodeMenu` from the `TextSelectionNodeMenu` in a way that wouldn't require changes to `isOpen`, `target`, and the addition of `onInteraction`.  But this could be a later optimization.

We should also remove the `TextSelectionNodeMenu` if a user selects text and hits the keyboard hotkey.  I think this might be solved with a more direct call as well.

image

apps/roam/src/components/DiscourseNodeMenu.tsx
trangdoan982: yeah i often select with Shift+ arrow left/right, to select up to specific characters using keyboard-only. adding key listener for ArrowLeft and ArrowRight prevents this flow.

mdroidian (approved):
🔥

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -113,6 +119,9 @@ const NodeMenu = ({
         onSelect(index);
         // Remove focus from the block to ensure updateBlock works properly
         document.body.click();
+      } else if (e.key === "Escape") {
+        onClose();
+        document.body.click();
```
mdroidian, line 124: @trangdoan982 do you remember why `document.body.click()` was added here?  I imagine a user would like to keep editing the block so I want to remove it, but I don't want to re-introduce a bug that this possibly fixing.

mdroidian: I removed it here: https://github.com/DiscourseGraphs/discourse-graph/pull/477

Let me know if that is in error.

trangdoan982: i don't think it was specifically to address any bug but just as general behavior to de-focus. I think it's okay that you remove it

PR #219, opened 2025-06-17
[ENG-425] Fix template frontmatter bug

trangdoan982:
https://www.loom.com/share/bf5f5e0a48cc425da3104872ff58e74f

## Summary by CodeRabbit

- **New Features**
  - Improved template application now merges frontmatter metadata from templates and target files, prioritizing existing file values.
  - Template body content is appended only if non-empty, preserving existing file content outside frontmatter.

- **Bug Fixes**
  - Prevents overwriting of entire file content when applying templates, ensuring existing content is retained.

### DIFF
--- a/apps/obsidian/src/utils/templates.ts
+++ b/apps/obsidian/src/utils/templates.ts
@@ -100,12 +100,21 @@ export const applyTemplate = async ({
       Object.assign(fm, mergedFrontmatter);
     });
 
-    const templateBodyMatch = templateContent.match(
-      /^---\s*\n([\s\S]*?)\n---\s*\n?([\s\S]*)$/,
+    const normalizedContent = templateContent
+      .replace(/\r\n/g, "\n")
+      .replace(/\r/g, "\n");
+
+    const templateBodyMatch = normalizedContent.match(
+      /^---\s*\n([\s\S]*?)\n---\s*(?:\n|$)([\s\S]*)$/,
     );
-    const templateBody = templateBodyMatch
-      ? templateBodyMatch[2] || ""
-      : templateContent;
+
+    let templateBody = "";
+
+    if (templateBodyMatch) {
+      templateBody = templateBodyMatch[2] || "";
+    } else {
+      templateBody = normalizedContent;
+    }
 
     if (templateBody.trim()) {
       await app.vault.append(targetFile, templateBody);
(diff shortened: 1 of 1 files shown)


trangdoan982:
@coderabbitai review

apps/obsidian/src/utils/templates.ts
```diff
@@ -86,11 +86,39 @@ export const applyTemplate = async ({
...
+
+    await app.fileManager.processFrontMatter(targetFile, (fm) => {
+      Object.assign(fm, mergedFrontmatter);
+    });
+
+    const normalizedContent = templateContent
+      .replace(/\r\n/g, "\n")
+      .replace(/\r/g, "\n");
+
+    const templateBodyMatch = normalizedContent.match(
```
mdroidian: Let's use obsidian's [getFrontMatterInfo](https://docs.obsidian.md/Reference/TypeScript+API/getFrontMatterInfo) / [FrontMatterInfo](https://docs.obsidian.md/Reference/TypeScript+API/FrontMatterInfo) here to future proof for any changes, instead of normalizing and matching ourselves

something like 
```typescript
    const frontmatterInfo = getFrontMatterInfo(templateContent);
    let templateBody = "";

    if (frontmatterInfo.exists) {
      templateBody = templateContent.slice(frontmatterInfo.contentStart);
    } else {
      templateBody = templateContent;
    }
```

apps/obsidian/src/utils/templates.ts
```diff
@@ -86,11 +86,39 @@ export const applyTemplate = async ({
 
     const templateContent = await app.vault.read(templateFile);
 
-    const currentContent = await app.vault.read(targetFile);
+    const templateFrontmatter =
+      app.metadataCache.getFileCache(templateFile)?.frontmatter || {};
+    const currentFrontmatter =
+      app.metadataCache.getFileCache(targetFile)?.frontmatter || {};
+
+    const mergedFrontmatter = {
```
mdroidian: This approach will overwrite, rather than combine, arrays/lists.

eg
```
# template.md
tags: [note, template]

# current.md
tags: [important]
```

I'd be curious if Obsidian / the Obsidian community has already solved this in a way we can reuse.

mdroidian (approved):
Not necessarily blocking, but both issues should be looked into and any changes should be tested.

PR #220, opened 2025-06-17
ENG-332 content to concept conversion functions

maparent:
These are the proposed functions to materialize nodes in the concept database.

## Summary by CodeRabbit

- **New Features**
  - Added new functionality to convert Roam Research discourse nodes and relations into a local concept format for database integration.
  - Enhanced extraction and aggregation of related concept information, including metadata such as author, creation time, and references.

Also wrote functions to order concepts in order of dependency. Not strictly needed in the current scenario, but could be needed down the road, and adds to the understanding of constraints of the sync function.

### DIFF
--- a/apps/roam/src/utils/conceptConversion.ts
+++ b/apps/roam/src/utils/conceptConversion.ts
@@ -225,7 +225,7 @@ export const orderConceptsByDependency = (
   const ordered: LocalConceptDataInput[] = [];
   let missing: Set<string> = new Set();
   while (Object.keys(conceptById).length > 0) {
-    const first = conceptById[concepts[0].represented_by_local_id!];
+    const first = Object.values(conceptById)[0];
     missing = missing.union(orderConceptsRec(ordered, first, conceptById));
   }
   return { ordered, missing: [...missing] };
--- a/apps/roam/src/utils/conceptConversion.ts
+++ b/apps/roam/src/utils/conceptConversion.ts
@@ -60,9 +60,15 @@ export function discourseNodeSchemaToLocalConcept(
 
 export function discourseNodeBlockToLocalConcept(
   context: SupabaseContext,
-  node_uid: string,
-  schema_uid: string,
-  text: string
+  {
+    node_uid,
+    schema_uid,
+    text
+  }: {
+    node_uid: string,
+    schema_uid: string,
+    text: string
+  }
 ): LocalConceptDataInput {
   const extra_data = getNodeExtraData(node_uid);
   return {
--- a/apps/roam/src/utils/conceptConversion.ts
+++ b/apps/roam/src/utils/conceptConversion.ts
@@ -181,18 +181,24 @@ const orderConceptsRec = (
   ordered: LocalConceptDataInput[],
   concept: LocalConceptDataInput,
   remainder: { [key: string]: LocalConceptDataInput },
-): undefined => {
+): Set<string> => {
   const relatedConceptIds = relatedConcepts(concept);
+  let missing: Set<string> = new Set();
   while (relatedConceptIds.length > 0) {
-    const relatedConceptId = relatedConceptIds.unshift();
+    const relatedConceptId = relatedConceptIds.pop()!;
     const relatedConcept = remainder[relatedConceptId];
-    if (relatedConcept !== undefined) {
-      orderConceptsRec(ordered, relatedConcept, remainder);
+    if (relatedConcept === undefined) {
+      missing.add(relatedConceptId);
+    } else {
+      missing = missing.union(
+        orderConceptsRec(ordered, relatedConcept, remainder),
+      );
       delete remainder[relatedConceptId];
     }
   }
   ordered.push(concept);
   delete remainder[concept.represented_by_local_id!];
+  return missing;
 };
 
 // This can be used by a sync function to order upserts.
@@ -200,14 +206,14 @@ const orderConceptsRec = (
 // and that nodes that are not in the upsert set are already in the database.
 export const orderConceptsByDependency = (
   concepts: LocalConceptDataInput[],
-): LocalConceptDataInput[] => {
+): { ordered: LocalConceptDataInput[]; missing: string[] } => {
   if (concepts.length === 0) return concepts;
   const conceptById: { [key: string]: LocalConceptDataInput } =
     Object.fromEntries(concepts.map((c) => [c, c.represented_by_local_id]));
   const ordered: LocalConceptDataInput[] = [];
   const first = conceptById[concepts[0].represented_by_local_id!];
-  orderConceptsRec(ordered, first, conceptById);
-  return ordered;
+  const missing = orderConceptsRec(ordered, first, conceptById);
+  return { ordered, missing: [...missing] };
 };
 
 // the input to the upsert method would look like this:
--- a/apps/roam/src/utils/conceptConversion.ts
+++ b/apps/roam/src/utils/conceptConversion.ts
@@ -4,12 +4,18 @@ import type { DiscourseRelation } from "./getDiscourseRelations";
 import { Database } from "@repo/database/types.gen";
 import type { SupabaseContext } from "~/utils/supabaseContext";
 
-// When it's merged: import type { LocalConceptDataInput } from '@repo/database/input_types';
-type LocalConceptDataInput = Partial<Database["public"]["CompositeTypes"]["concept_local_input"]>;
+import type { LocalConceptDataInput } from "@repo/database/input_types";
 
-// Get data from roam about the node
-const getNodeExtraData = (node_uid: string): { author_uid: string, created: string, last_modified: string, page_uid: string } => {
-  const result = window.roamAlphaAPI.q(`[
+const getNodeExtraData = (
+  node_uid: string,
+): {
+  author_uid: string;
+  created: string;
+  last_modified: string;
+  page_uid: string;
+} => {
+  const result = window.roamAlphaAPI.q(
+    `[
     :find
       ?author_uid
       ?page_uid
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,170 @@
...
+  const title_parts = node.text.split('/');
+  return {
+    space_id: context.spaceId,
+    author_local_id: extra_data.author_uid,
+    created: extra_data.created,
+    last_modified: extra_data.last_modified,
+    name: title_parts[title_parts.length - 1],
+    represented_by_local_id: node.type,
+    is_schema: true,
+  };
```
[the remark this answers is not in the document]
maparent: Wrong, there is no node.uid field in the DiscourseNode type. node.type plays that role, strange as it is.

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,170 @@
...
+    local_reference_content: Object.fromEntries(
+      Object.entries(relation)
+        .filter(([key, v]) => STANDARD_ROLES.includes(key))
+      ) as { [key: string]: string },
+    literal_content: {
+      roles: STANDARD_ROLES,
+      label: relation.label,
+      complement: relation.complement,
+      representation: relation.triples.map(t=>t[0]),
+    },
```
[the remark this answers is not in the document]
maparent: What makes you think of string[] being among the types? All DiscourseRelation types are string or Triple[], but that latter case  is excluded by the STANDARD_ROLES filter.

mdroidian:
@maparent  I'm curious why this is merging into [ENG-446-bulk-db-concept-upsert](https://github.com/DiscourseGraphs/discourse-graph/tree/ENG-446-bulk-db-concept-upsert) instead of `main`.

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,180 @@
+import { DiscourseNode } from "./getDiscourseNodes";
+import getDiscourseRelations from "./getDiscourseRelations";
+import type { DiscourseRelation } from "./getDiscourseRelations";
+import { Database } from "@repo/database/types.gen";
+import type { SupabaseContext } from "~/utils/supabaseContext";
+
+type roamAlphaAPI = Window["roamAlphaAPI"];
```
mdroidian: This shouldn't be needed.  `roamAlphaAPI` is already declared on the Window object in roamjs-components. See other util examples using `window.roamAlphaAPI`

apps/roam/src/utils/supabaseContext.ts
```diff
@@ -5,8 +5,10 @@ import getRoamUrl from "roamjs-components/dom/getRoamUrl";
 import { Database } from "@repo/database/types.gen";
 
 type Platform = Database["public"]["Enums"]["Platform"];
+type roamAlphaAPI = Window["roamAlphaAPI"];
 
 export type SupabaseContext = {
+  api: roamAlphaAPI;
```
mdroidian: This shouldn't be needed.  `roamAlphaAPI` is always available on the `window` object

apps/roam/src/utils/supabaseContext.ts
```diff
@@ -5,8 +5,10 @@ import getRoamUrl from "roamjs-components/dom/getRoamUrl";
 import { Database } from "@repo/database/types.gen";
 
 type Platform = Database["public"]["Enums"]["Platform"];
+type roamAlphaAPI = Window["roamAlphaAPI"];
```
mdroidian: .

apps/roam/src/utils/supabaseContext.ts
```diff
@@ -105,7 +107,7 @@ export const getSupabaseContext = async (): Promise<SupabaseContext | null> => {
         personName,
         personEmail,
       });
-      CONTEXT_CACHE = { platform: "Roam", spaceId, userId };
+      CONTEXT_CACHE = { api: window.roamAlphaAPI, platform: "Roam", spaceId, userId };
```
mdroidian: This shouldn't be needed.  `roamAlphaAPI` is always available on the `window` object

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,180 @@
...
+    created: extra_data.created,
+    last_modified: extra_data.last_modified,
+    name: title_parts[title_parts.length - 1],
+    represented_by_local_id: node.type,
+    is_schema: true,
+  };
+}
+
+
+export function discourseNodeBlockToLocalConcept(
```
mdroidian: Please review for Style Guide adherence (arrow function, named params, variable names, etc)

mdroidian (changes requested):
Let's do a first pass before reviewing the functionality.

maparent:
> @maparent I'm curious why this is merging into [ENG-446-bulk-db-concept-upsert](https://github.com/DiscourseGraphs/discourse-graph/tree/ENG-446-bulk-db-concept-upsert) instead of `main`.

Because it relies on the on the concept local input types defined in 446.

apps/roam/src/utils/conceptConversion.ts
maparent: Coderabbit suggested injecting the roamAlphaAPI instead of using it directly in the function, to make mocking and debugging easier, and I thought it was a good idea. In the original coderabbit suggestion, it had defined an _ad hoc_ roamAlphaApi type for the injection. I wanted to type the injected api properly, and this where I could find it. I could revert those changes, but I still think coderabbit had a point there.

apps/roam/src/utils/supabaseContext.ts
maparent: see above.

apps/roam/src/utils/supabaseContext.ts
maparent: see above.

apps/roam/src/utils/supabaseContext.ts
maparent: ditto.

apps/roam/src/utils/conceptConversion.ts
maparent: dammit. I really want the linter to tell me this.

apps/roam/src/utils/conceptConversion.ts
maparent: note that a lot of those changes were in the second commit, which I did not yet squash, so you can have a before/after and decide which you prefer. Again, I think mockability is good.

mdroidian: How would this make mocking and debugging easier?

maparent: Because we can send a mock roam api object in a non-browser testing environment

maparent:
actually wait will make minor adjustments

maparent:
added comments as in the other one.

apps/roam/src/utils/conceptConversion.ts
mdroidian: We currently aren't doing this and have no immediate plans to. There is also clear precedent in the codebase: we access the API directly rather than passing it as a parameter (hundreds of examples). It's equally plausible that we solve this differently in the future—or never need to test this function at all. If and when we do adopt this kind of testing, we can add the necessary functionality then. For now, including it here introduces unnecessary and inconsistent code. Let's remove it.

maparent:
Hmm... I did as you asked and resolved the conversations, and they're all marked as outdated, but I still get that there's a review requesting changes. Anyhow, maybe time for another review?

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+      :where
+      [?block :block/uid ?block_uid]
+      [?block :create/user ?author_id]
+      [?author_id :user/uid ?author_uid]
+      [?block :create/time ?created]
+      [?block :edit/time ?last_modified]
+      [(get-else $ ?block :block/page ?block) ?page_id]
+      [?page_id :block/uid ?page_uid]
+  ]`, node_uid);
+  if (result.length != 1 || result[0].length != 4)
```
mdroidian: !==

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+    created,
+    last_modified,
+    page_uid
+  }
+}
+
+
+export function discourseNodeSchemaToLocalConcept(
+  context: SupabaseContext,
+  node: DiscourseNode, // contains type, uid, text
```
mdroidian: Is this comment necessary?

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+  return {
+    author_uid,
+    created,
+    last_modified,
+    page_uid
+  }
+}
+
+
+export function discourseNodeSchemaToLocalConcept(
```
mdroidian: arrow function

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+  context: SupabaseContext,
+  node: DiscourseNode, // contains type, uid, text
+): LocalConceptDataInput {
+  const extra_data = getNodeExtraData(node.type);
```
mdroidian: camel case

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+  const title_parts = node.text.split('/');
```
mdroidian: camel case

apps/roam/src/utils/conceptConversion.ts
mdroidian: Was this updated?  GitHub is still showing `function`

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+      [?block :edit/time ?last_modified]
+      [(get-else $ ?block :block/page ?block) ?page_id]
+      [?page_id :block/uid ?page_uid]
+  ]`, node_uid);
+  if (result.length != 1 || result[0].length != 4)
+    throw new Error("Invalid result from Roam query");
+
+  const [author_uid, page_uid, created_t, last_modified_t] = result[0] as [string, string, number, number];
+  const created = new Date(created_t).toISOString();
+  const last_modified = new Date(last_modified_t).toISOString();
```
mdroidian, line 45: camel case

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+    node_uid,
+    schema_uid,
+    text
+  }: {
+    node_uid: string,
+    schema_uid: string,
+    text: string
+  }
+): LocalConceptDataInput {
+  const extra_data = getNodeExtraData(node_uid);
```
mdroidian: camel case

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+    represented_by_local_id: node_uid,
+    schema_represented_by_local_id: schema_uid,
+    is_schema: false,
+  };
+
+}
+
+const STANDARD_ROLES = ["source", "target"];
+
+export function discourseRelationSchemaToLocalConcept(
```
mdroidian: arrow

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+  context: SupabaseContext,
+  relation: DiscourseRelation, // contains type, uid, text
+): LocalConceptDataInput {
+  const extra_data = getNodeExtraData(relation.id);
```
mdroidian: camel case

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+    is_schema: false,
+  };
+
+}
+
+const STANDARD_ROLES = ["source", "target"];
+
+export function discourseRelationSchemaToLocalConcept(
+  context: SupabaseContext,
+  relation: DiscourseRelation, // contains type, uid, text
```
mdroidian: Is this comment required?  Not sure what it is referring to.

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+): LocalConceptDataInput {
+  const extra_data = getNodeExtraData(relation.id);
+  return {
+    space_id: context.spaceId,
+    represented_by_local_id: relation.id,
+    author_local_id: extra_data.author_uid,
+    created: extra_data.created,
+    last_modified: extra_data.last_modified,
+    name: `${relation.id}-${relation.label}`,  // the label is not unique
```
mdroidian: What purpose is this comment serving?

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+    literal_content: {
+      roles: STANDARD_ROLES,
+      label: relation.label,
+      complement: relation.complement,
+      representation: relation.triples.map(t=>t[0]),
+    },
+  };
+}
+
+export function discourseRelationDataToLocalConcept(
```
mdroidian: arrow

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+  context: SupabaseContext,
```
mdroidian, line 118: params

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+  relationSchemaUid: string,
+  relationNodes: { [role: string]: string },
+): LocalConceptDataInput {
+  const roamRelation = getDiscourseRelations().find(r => r.id === relationSchemaUid);
+  if (roamRelation === undefined) {
+    throw new Error(`Invalid roam relation id ${relationSchemaUid}`)
+  }
+  const relation = discourseRelationSchemaToLocalConcept(context, roamRelation);
+  const lit_content = ((relation.literal_content)? relation.literal_content: {}) as unknown as {[key: string]: any};
```
mdroidian: camel case

mdroidian: Why is this assertion necessary?  Can we not achieve this with types?

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+      [?author_id :user/uid ?author_uid]
+      [?block :create/time ?created]
+      [?block :edit/time ?last_modified]
+      [(get-else $ ?block :block/page ?block) ?page_id]
+      [?page_id :block/uid ?page_uid]
+  ]`, node_uid);
+  if (result.length != 1 || result[0].length != 4)
+    throw new Error("Invalid result from Roam query");
+
+  const [author_uid, page_uid, created_t, last_modified_t] = result[0] as [string, string, number, number];
```
mdroidian: came case

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
+import { DiscourseNode } from "./getDiscourseNodes";
+import getDiscourseRelations from "./getDiscourseRelations";
+import type { DiscourseRelation } from "./getDiscourseRelations";
+import { Database } from "@repo/database/types.gen";
+import type { SupabaseContext } from "~/utils/supabaseContext";
+
+// When it's merged: import type { LocalConceptDataInput } from '@repo/database/input_types';
```
mdroidian: When what is merged?

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+type LocalConceptDataInput = Partial<Database["public"]["CompositeTypes"]["concept_local_input"]>;
+
+// Get data from roam about the node
```
mdroidian: Is this comment necessary?

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+      .map((role) => [role, relationNodes[role]])
+      .filter(([, uid]) => uid !== undefined)
+  );
+  if (Object.keys(casting).length === 0) {
+    throw new Error(`No valid node UIDs supplied for roles ${roles.join(", ")}`);
+  }
+  // TODO: Also get the nodes from the representation, using QueryBuilder. That will likely give me the relation object
+  const nodeData = Object.values(casting).map((v) => getNodeExtraData(v));
+  // roundabout way to do a max from stringified dates
+  const last_modified = new Date(Math.max(...nodeData.map((nd) => new Date(nd.last_modified).getTime()))).toISOString();
```
mdroidian: camel case

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+  // creation is actually creation of the relation node, not the rest of the cast, but this will do as a first approximation.
+  // Still using max, since the relation cannot be created before its cast
+  const created = new Date(Math.max(...nodeData.map((nd) => new Date(nd.created).getTime()))).toISOString();
+  const author_local_id: string = nodeData[0].author_uid;  // take any one; again until I get the relation object
```
mdroidian: camel case

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+  const represented_by_local_id = (casting["target"] || Object.values(casting)[0]); // This one is tricky. Prefer the target for now.
```
mdroidian: camel case

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+    local_reference_content: casting,
+  };
+}
+
+const relatedConcepts = (concept: LocalConceptDataInput): string[] => {
+  const relations = Object.values(concept.local_reference_content || {}).flat();
+  if (concept.schema_represented_by_local_id) {
+    relations.push(concept.schema_represented_by_local_id)
+  }
+  // remove duplicates
```
mdroidian, line 177: Is this comment necessary?

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+
+const contentRelatedToConcept = (concept: LocalConceptDataInput): string[] => {
+  const relations = relatedConcepts(concept);
+  if (concept.represented_by_local_id && !relations.includes(concept.represented_by_local_id)) {
+    relations.push(concept.represented_by_local_id)
+  }
+  return relations;
+}
+
+// From those two, a function could in theory ensure that a concept's dependent concepts are already in the database.
```
mdroidian: For what two?

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,184 @@
...
+const relatedConcepts = (concept: LocalConceptDataInput): string[] => {
+  const relations = Object.values(concept.local_reference_content || {}).flat();
+  if (concept.schema_represented_by_local_id) {
+    relations.push(concept.schema_represented_by_local_id)
+  }
+  // remove duplicates
+  return [...new Set(relations)];
+}
+
+const contentRelatedToConcept = (concept: LocalConceptDataInput): string[] => {
```
mdroidian: This is currently unused.  Will it be used somewhere in the future?  If so, by whom?

mdroidian (changes requested):
Lots of stylistic change requests still.

A question as well:
- What does "Local" mean in this context?  Is it different than the db schema?

mdroidian:
@maparent also be sure to format with Prettier.  It looks like this PR was missed.  eg: [https://github.com/DiscourseGraphs/discourse-graph/pull/220/files#diff-688e149bb8d226571fd59b471992e057cd6e21838e8dbaf92ecf28bf6f8f61a9R11](https://github.com/DiscourseGraphs/discourse-graph/pull/220/files#diff-688e149bb8d226571fd59b471992e057cd6e21838e8dbaf92ecf28bf6f8f61a9R11)

apps/roam/src/utils/conceptConversion.ts
maparent: For someone unfamiliar with the codebase, yes. Will remove.

apps/roam/src/utils/conceptConversion.ts
maparent: It referred to the input_types file. I thought it was obvious. It's in ENG-415, which I guess I should have specified.

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,170 @@
...
+  const [author_uid, page_uid, created_t, last_modified_t] = result[0] as [string, string, number, number];
+  const created = new Date(created_t).toISOString();
+  const last_modified = new Date(last_modified_t).toISOString();
+  return {
+    author_uid,
+    created,
+    last_modified,
+    page_uid
+  }
+}
```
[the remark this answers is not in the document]
maparent: @coderabbit I could not agree more, but it is the pattern in the codebase right now. That change will not be made until it is scheduled to be made globally.

apps/roam/src/utils/conceptConversion.ts
maparent: This one will be used as-is in the output, without equals.

apps/roam/src/utils/conceptConversion.ts
maparent: removed those variables.

apps/roam/src/utils/conceptConversion.ts
maparent: removed

apps/roam/src/utils/conceptConversion.ts
maparent: expanded.

apps/roam/src/utils/conceptConversion.ts
maparent: I thought I was allowed three params?

apps/roam/src/utils/conceptConversion.ts
maparent: The type of `literal_content` is `Json | undefined`. I have to assert typing to tell it about the expected Json structure, otherwise I run in even more trouble in the next line.
Do you think that would warrant a comment? (I'm really confused as to what should be considered obvious or comment-worthy.)

apps/roam/src/utils/conceptConversion.ts
maparent: those are designed to fit the the LocalConceptInput types.

apps/roam/src/utils/conceptConversion.ts
maparent: also fits output type.

apps/roam/src/utils/conceptConversion.ts
maparent: also fits output type.

apps/roam/src/utils/conceptConversion.ts
maparent: output type.

apps/roam/src/utils/conceptConversion.ts
maparent: I don't know what will be unclear to the code reader. I think indicating intent when not obvious from syntax is always worth doing, no?

apps/roam/src/utils/conceptConversion.ts
maparent: Those two functions. I decided one was less likely to be used, and that the usefulness of the other would be in the context of another function which I then provided. I hope that clarifies.

apps/roam/src/utils/conceptConversion.ts
maparent: Used by orderConceptsByDependency, which should be used by sync.

maparent:
Local id means platform-local, same as in the db schema. (Tempted to do a global rename of local to platform, but clearly out of scope.)
Added a function. 
Prettier was a vscode mis-configuration (it wanted .prettierrc to exist), but the style guide... I'll prioritize 369, it's wasting both our time.

maparent:
Also note: many of the non-camel-case I use so I can use the variable name as-is in the output, and I think that saves optimized JS volume. But yeah, it's going to be annoying to tell the linter or the reviewer why this var is not camel-cased. Still probably necessary.

maparent:
@coderabbit review function orderConceptsByDependency

maparent:
@coderabbit thank you, those were serious concerns indeed. I stayed a bit simpler than what you proposed, because I know I can assume acyclicity. I also gave more context. How does it look now?

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,241 @@
...
+    represented_by_local_id: node.type,
+    is_schema: true,
+    ...getNodeExtraData(node.type),
+  };
+};
+
+export const discourseNodeBlockToLocalConcept = (
+  context: SupabaseContext,
+  {
+    node_uid,
```
mdroidian: On second thought, I disagree with using `_` even if they will be `_` in some other function.  This is introducing subjective interpretation on when to use `_` (like `node_uid` which is not the actual output, same with `created_t`, `last_modified_t`, etc).  But I will not block the PR further.  Just noting my thoughts.

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,241 @@
+import { DiscourseNode } from "./getDiscourseNodes";
+import getDiscourseRelations from "./getDiscourseRelations";
+import type { DiscourseRelation } from "./getDiscourseRelations";
+import { Database } from "@repo/database/types.gen";
```
mdroidian: remove, unused.

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -0,0 +1,241 @@
...
+import type { SupabaseContext } from "~/utils/supabaseContext";
+
+import type { LocalConceptDataInput } from "@repo/database/input_types";
```
mdroidian: update to `inputTypes`

mdroidian (approved):
Remove the unused import and update the `inputTypes` import, then feel free to merge.

apps/roam/src/utils/conceptConversion.ts
maparent: Actually I see it. Changing.

PR #221, opened 2025-06-17
Upgrade supabase. Also remove jwt, as it blocks a postgres upgrade.

maparent:


mdroidian:
@maparent have you tested the build after upgrading?

maparent:
I often use the new version locally because it gets installed by homebrew. It's actually why I'm so keen to update, I get mismatch issues.

mdroidian:
In PR's like this, could you run the `turbo build`, screenshot the results, and paste into the PR body.  Or use some other way to convey that this has indeed been tested.

mdroidian:
Example: DiscourseGraphs/discourse-graph#196

mdroidian (approved):
👍

maparent:
@mdroidian And you were right to ask to check. supabase is fine, but the drop extension migration failed. I did try locally, but in the window; made a correction in the window, and iirc in the file, but somehow must have forgotten to save it, so the correction did not make it in the commit. With idempotent migrations, I sometimes reapply the migration after my local tests to be sure, but that one was not idempotent so I did not. Sigh. There will be another minor commit. More important, this means we _really_ need to have either a dev branch or supabase branches, it would have caught this (and have saved us both this and the above conversation.)

mdroidian:
> @mdroidian And you were right to ask to check. supabase is fine, but the drop extension migration failed. I did try locally, but in the window; made a correction in the window, and iirc in the file, but somehow must have forgotten to save it, so the correction did not make it in the commit. With idempotent migrations, I sometimes reapply the migration after my local tests to be sure, but that one was not idempotent so I did not. Sigh. There will be another minor commit. More important, this means we _really_ need to have either a dev branch or supabase branches, it would have caught this (and have saved us both this and the above conversation.)

How would a dev branch be better than running `turbo build` before submitting the PR?

PR #222, opened 2025-06-17
turbo is complaning about ouputs in packages/database.

maparent:
Specifying this file silences the warning.
 ETA: can be tested with `turbo -F @repo/database`.


mdroidian (approved):
👍

PR #223, opened 2025-06-18
correction: drop extension is not namespaced

maparent:
Correction to https://github.com/DiscourseGraphs/discourse-graph/pull/221
Added a message there to describe the situation.


mdroidian (approved):
👍

PR #215, opened 2025-06-16
ENG-446 concept upsert functions

maparent:
## Summary by CodeRabbit

- **New Features**
  - Added support for batch upserting multiple concepts using a new function, allowing efficient import or update of concept data from JSON input.
  - Introduced enhanced handling of local identifiers and references, enabling automatic resolution of related entities during concept creation or update.

- **Documentation**
  - Expanded type definitions to reflect new functions and input types for improved clarity and type safety in database interactions.

PR #224, opened 2025-06-19
two migration errors

maparent:


maparent:
This is about getting the db diff to say the schema is exactly the same as running the migrations. Did the rename too fast.

maparent:
(not urgent)

packages/database/supabase/migrations/20250619021503_correct_migrations.sql
```diff
@@ -0,0 +1,69 @@
+-- complete migration 20250610201801_create_access_table.sql
+
+revoke delete on table "public"."access_token" from "anon";
```
mdroidian, line 3: What error is this fixing?  Could you link the the source of the error?

packages/database/supabase/migrations/20250619021503_correct_migrations.sql
```diff
@@ -0,0 +1,69 @@
...
+
+revoke trigger on table "public"."access_token" from "anon";
+
+revoke truncate on table "public"."access_token" from "anon";
+
+revoke update on table "public"."access_token" from "anon";
+
+-- rename in schema was not ported to migration 20250611180757_concept_upsert_functions.sql
+
+alter function "public".local_concept_to_db_concept(data concept_local_input) rename to _local_concept_to_db_concept;
```
mdroidian, line 15: What error is this fixing?  Could you link the the source of the error?

packages/database/supabase/migrations/20250619021503_correct_migrations.sql
```diff
@@ -0,0 +1,69 @@
...
+
+CREATE OR REPLACE FUNCTION public.upsert_concepts(v_space_id bigint, data jsonb)
```
mdroidian, line 17: What error is this fixing?  Could you link the the source of the error?

packages/database/supabase/migrations/20250619021503_correct_migrations.sql
maparent: Before and after work on a migration, `supabase db diff` (with supabase in stopped state) should say there are no differences between applying the schema directly, or through the sequence of migrations. 
In this case, the content of schema and migration have different permissions; so the `db diff` provided these changes to align the migrations with the schema. It's entirely possible that the migration was correct and the schema was wrong; in which case we should change the schema and not the migration.

packages/database/supabase/migrations/20250619021503_correct_migrations.sql
maparent: This is because I renamed the function to indicate it's private in the schema but forgot to do so in the migration. So in this migration I rename the private function, and I have to re-define the function that uses  it to use the new name. It is the only change to that function. Here I'm clearly migrating to align to the schema as source of truth.

packages/database/supabase/migrations/20250619021503_correct_migrations.sql
mdroidian: Right.  What I'm saying is that we should link to where this happened so that someone inspecting this PR has a breadcrumb and as a way to ensure these changes actually reflect remote schema.

EG: https://github.com/DiscourseGraphs/discourse-graph/pull/215/files#diff-22bac1fb1de78f55daaeedaaa408e1bd563554544a152bf0c82630d8675a69daR139

packages/database/supabase/migrations/20250619021503_correct_migrations.sql
mdroidian: https://github.com/DiscourseGraphs/discourse-graph/pull/215/files#diff-22bac1fb1de78f55daaeedaaa408e1bd563554544a152bf0c82630d8675a69daR139

packages/database/supabase/migrations/20250619021503_correct_migrations.sql
mdroidian: https://github.com/DiscourseGraphs/discourse-graph/pull/196/files#diff-afcd75279c042dc0e5f9dd17b7836f4c3aec96eeb36abab541ab85516798e77dR20

mdroidian (approved):

PR #231, opened 2025-06-22
ENG-449: Calculate the embeddings for updated discourse nodes

sid597:
https://www.loom.com/share/44125280d2eb4aec921b0933e8e27ee3?sid=b17f08dc-372c-45ab-88c0-4209cb50e250

Loom does not have access to console log so adding a screenshot of the console.log 
 
 
image

Tested using stub code that fetches 5 discourse nodes from graph and creates embeddings for them using this function

## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to fetch and attach vector embeddings to content nodes, enhancing data with additional model and vector information.

PR #232, opened 2025-06-22
ENG-448: Query Roam Discourse Nodes created or modified since the last update

sid597:
Here is the result of using this function 

image

## Summary by CodeRabbit

- **New Features**
  - Added the ability to fetch and display discourse nodes that have been edited since a specific date, including details such as author, timestamps, and content.

### DIFF
--- a/apps/roam/src/utils/getAllDiscourseNodesSince.ts
+++ b/apps/roam/src/utils/getAllDiscourseNodesSince.ts
@@ -1,22 +1,22 @@
 import isDiscourseNode from "./isDiscourseNode";
 
-type RoamEntityFromQuery = {
-  uid: string;
-  title?: string;
-  createTime?: number;
-  editTime?: number;
-  userUuid?: string;
-  username?: string;
+export type DiscourseGraphContent = {
+  author_local_id: string;
+  source_local_id: string;
+  scale: string;
+  created: string;
+  last_modified: string;
+  text: string;
 };
 
 export const getAllDiscourseNodesSince = async (
   since: string,
-): Promise<RoamEntityFromQuery[]> => {
+): Promise<DiscourseGraphContent[]> => {
   const roamAlpha = (window as any).roamAlphaAPI;
   const sinceMs = new Date(since).getTime();
 
   const query = `[:find ?uid ?create-time ?edit-time ?user-uuid ?username ?title
-     :keys  uid   createTime    editTime    userUuid   username title
+     :keys  source_local_id created last_modified author_local_id author_name text 
      :in $ ?since 
      :where
       [?e :node/title ?title]
@@ -28,13 +28,13 @@ export const getAllDiscourseNodesSince = async (
       [?e :edit/time ?edit-time]
       [(> ?edit-time ?since)]]`;
 
-  const result = roamAlpha.data.q(query, sinceMs) as RoamEntityFromQuery[];
+  const result = roamAlpha.data.q(query, sinceMs) as DiscourseGraphContent[];
 
   return result.filter(
     (entity) =>
-      entity.uid &&
-      isDiscourseNode(entity.uid) &&
-      entity.title &&
-      entity.title.trim() !== "",
+      entity.source_local_id &&
+      isDiscourseNode(entity.source_local_id) &&
+      entity.text &&
+      entity.text.trim() !== "",
   );
 };
--- a/apps/roam/src/utils/getAllDiscourseNodesSince.ts
+++ b/apps/roam/src/utils/getAllDiscourseNodesSince.ts
@@ -1,18 +1,20 @@
-import isDiscourseNode from "./isDiscourseNode";
+import getDiscourseNodes from "./getDiscourseNodes";
+import findDiscourseNode from "./findDiscourseNode";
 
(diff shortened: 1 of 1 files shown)


sid597:
@coderabbitai full review

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -0,0 +1,40 @@
...
+  created: string;
+  last_modified: string;
+  text: string;
+};
+
+export const getAllDiscourseNodesSince = async (
+  since: string,
+): Promise<DiscourseGraphContent[]> => {
+  const roamAlpha = (window as any).roamAlphaAPI;
+  const sinceMs = new Date(since).getTime();
```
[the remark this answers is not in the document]
sid597: date will be validate before reaching this function

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -0,0 +1,40 @@
...
+      [?e :node/title ?title]
+      [?e :block/uid ?uid] 
+      [?e :create/user ?user-id]
+      [?user-id :user/uid ?user-uuid]
+      [?user-id :user/display-name ?username]
+      [?e :create/time ?create-time]
+      [?e :edit/time ?edit-time]
+      [(> ?edit-time ?since)]]`;
+
+  const result = roamAlpha.data.q(query, sinceMs) as DiscourseGraphContent[];
```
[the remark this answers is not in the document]
sid597: nah all good

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -0,0 +1,40 @@
...
+  author_local_id: string;
+  source_local_id: string;
+  scale: string;
+  created: string;
+  last_modified: string;
+  text: string;
+};
+
+export const getAllDiscourseNodesSince = async (
+  since: string,
```
mdroidian: Let's alias this to help the caller know what to input

```typescript
type ISODateString = string;

export const getAllDiscourseNodesSince = async (
  since: ISODateString,
//...
```

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -0,0 +1,40 @@
...
+): Promise<DiscourseGraphContent[]> => {
+  const roamAlpha = (window as any).roamAlphaAPI;
```
mdroidian: Avoid `any` at all costs.  

Let's cast the results to `as unknown[][] as DiscourseGraphContent[];` to explicitly show that we are doing this on purpose.

apps/roam/src/utils/getAllDiscourseNodesSince.ts
mdroidian: use `data.backend.q`, it returns a promise and is async and non blocking
https://roamresearch.com/#/app/developer-documentation/page/1gcc_o62N

it isn't in the roamjs-components types yet, so you'll need to add `// @ts-ignore` in the line above

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -0,0 +1,40 @@
...
+  const sinceMs = new Date(since).getTime();
+
+  const query = `[:find ?uid ?create-time ?edit-time ?user-uuid ?username ?title
+     :keys  source_local_id created last_modified author_local_id author_name text
```
mdroidian, line 21: the key's don't match the type definition you are casting to. (author_name, scale)

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -0,0 +1,40 @@
+import isDiscourseNode from "./isDiscourseNode";
+
+export type DiscourseGraphContent = {
```
mdroidian: Why is this exported?  Is it to be used somewhere else?  Also, it is named the same as in https://github.com/DiscourseGraphs/discourse-graph/pull/231 with different keys.  Let's rename it to something more specific to this function.

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -0,0 +1,40 @@
...
+      [?e :create/time ?create-time]
+      [?e :edit/time ?edit-time]
+      [(> ?edit-time ?since)]]`;
+
+  const result = roamAlpha.data.q(query, sinceMs) as DiscourseGraphContent[];
+
+  return result.filter(
+    (entity) =>
+      entity.source_local_id &&
+      isDiscourseNode(entity.source_local_id) &&
```
mdroidian: This will call `getDiscourseNodes()` on every node.

Let's optimize this by calling `getDiscourseNodes()` in this file then filtering each node:

```typescript
  const node = findDiscourseNode(uid, nodes);
  if (!node) return false;
  return node.backedBy !== "default";
```

mdroidian (changes requested):
Some changes requested.

mdroidian (approved):

PR #234, opened 2025-06-22
Alpha Suggestive Discourse graph: ENG-433 : Test different type of UIs for suggestive discourse graph

sid597:
Creating a PR so that it gets deployed to vercel and we can use a link to use it, once its there, tested that the link is working as is should then I will merge this to the og suggestive mode pr, and delete this branch


mdroidian:
@sid597
I'm confused.  I thought you were using the `store-in-supabase` branch for testing?

sid597:
I did not want to take down that branch if there was any bug that took time to resolve, now this one is tested so merging and closing

PR #240, opened 2025-06-25
schema linting

maparent:
This is just applying the `sqruff fix` to the sql schemas, because it errors and fails the lint step, making developing the lint (ENG-369) harder.
I'm ok with the lint being imposed; but we could also do the equivalent of only-warn and not fail the lint, either way. But for now just checking this in, to distinguish it from the rest of 369, to which it's not that deeply related.


mdroidian (approved):
👍

PR #244, opened 2025-06-25
ENG-520: break-out infrastructure from ENG-373

maparent:
There are many branches (notably #238, which may be broken in three before merge; and an upcoming branch for ENG-420 which I need to base on both this one and #245)  that assume access to supabase infrastructure in the front-end; I just wanted to isolate the infrastructure work from the ui work in the ENG-373 branch, so those other branches build on it.

An unfortunate consequence of this is that turbo now expects `SUPABASE_URL` and `SUPABASE_ANON_KEY` to be defined. We will all need to update our `.env` accordingly. This should be fine with vercel and github actions.

Note: There are a few non-obvious choices, such as putting the client code in the ui, which I think will benefit obsidian in the longer term.

## Summary by CodeRabbit

- **New Features**
	- Added utility functions for creating and managing Supabase clients in both the website and UI packages.
	- Introduced session management middleware for user authentication with Supabase in the website app.

- **Chores**
	- Updated environment variable handling across build, development, deployment, and CI workflows to support Supabase integration.
	- Added Supabase-related dependencies to project packages.
	- Enhanced TypeScript configurations with new path aliases for simplified imports.

### DIFF
--- a/apps/website/app/utils/supabase/client.ts
+++ b/apps/website/app/utils/supabase/client.ts
@@ -1,0 +1,14 @@
+import { createClient as createSupabaseClient } from "@supabase/supabase-js";
+import { Database } from "@repo/database/types.gen.ts";
+
+// Inspired by https://supabase.com/ui/docs/nextjs/password-based-auth
+
+export const createClient = () => {
+  const url = process.env.NEXT_PUBLIC_SUPABASE_URL;
+  const key = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY;
+
+  if (!url || !key) {
+    throw new Error("Missing required Supabase environment variables");
+  }
+  return createSupabaseClient<Database, "public", Database["public"]>(url, key);
+};
--- a/apps/website/app/utils/supabase/middleware.ts
+++ b/apps/website/app/utils/supabase/middleware.ts
@@ -1,0 +1,70 @@
+import { createServerClient } from "@supabase/ssr";
+import { NextResponse, type NextRequest } from "next/server";
+
+// Inspired by https://supabase.com/ui/docs/nextjs/password-based-auth
+
+export const updateSession = async (request: NextRequest) => {
+  const supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL;
+  const supabaseKey = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY;
+
+  if (!supabaseUrl || !supabaseKey) {
+    throw new Error("Missing required Supabase environment variables");
+  }
+
+  let supabaseResponse = NextResponse.next({ request });
+
+  const supabase = createServerClient(supabaseUrl, supabaseKey, {
+    cookies: {
+      getAll() {
+        return request.cookies.getAll();
+      },
+      setAll(cookiesToSet) {
+        cookiesToSet.forEach(({ name, value }) =>
+          request.cookies.set(name, value),
+        );
+        supabaseResponse = NextResponse.next({
+          request,
+        });
+        cookiesToSet.forEach(({ name, value, options }) =>
+          supabaseResponse.cookies.set(name, value, options),
+        );
+      },
+    },
+  });
+
+  // Do not run code between createServerClient and
+  // supabase.auth.getUser(). A simple mistake could make it very hard to debug
+  // issues with users being randomly logged out.
+
+  // IMPORTANT: DO NOT REMOVE auth.getUser()
+
+  const {
+    data: { user },
+  } = await supabase.auth.getUser();
+
+  if (
+    !user &&
+    !request.nextUrl.pathname.startsWith("/login") &&
+    !request.nextUrl.pathname.startsWith("/auth")
+  ) {
+    // no user, potentially respond by redirecting the user to the login page
+    const url = request.nextUrl.clone();
+    url.pathname = "/auth/login";
+    return NextResponse.redirect(url);
+  }
+
+  // IMPORTANT: You *must* return the supabaseResponse object as it is.
+  // If you're creating a new response object with NextResponse.next() make sure to:
+  // 1. Pass the request in it, like so:
+  //    const myNewResponse = NextResponse.next({ request })
+  // 2. Copy over the cookies, like so:
+  //    myNewResponse.cookies.setAll(supabaseResponse.cookies.getAll())
+  // 3. Change the myNewResponse object to fit your needs, but avoid changing
+  //    the cookies!
+  // 4. Finally:
+  //    return myNewResponse
+  // If this is not done, you may be causing the browser and server to go out
+  // of sync and terminate the user's session prematurely!
+
+  return supabaseResponse;
+};
--- a/apps/website/tsconfig.json
+++ b/apps/website/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "@repo/typescript-config/nextjs.json",
   "compilerOptions": {
+    "baseUrl": ".",
     "paths": {
       "~/*": ["./app/*"]
     },
--- a/packages/ui/src/lib/supabase/client.ts
+++ b/packages/ui/src/lib/supabase/client.ts
@@ -1,0 +1,14 @@
+import { createClient as createSupabaseClient } from "@supabase/supabase-js";
+import { Database } from "@repo/database/types.gen.ts";
+
+declare const SUPABASE_URL: string;
+declare const SUPABASE_ANON_KEY: string;
+
+// Inspired by https://supabase.com/ui/docs/react/password-based-auth
+
+export const createClient = () => {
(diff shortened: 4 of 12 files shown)


maparent:
@coderabbit review

apps/website/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "@repo/typescript-config/nextjs.json",
   "compilerOptions": {
     "paths": {
-      "~/*": ["./app/*"]
+      "~/*": ["./app/*"],
+      "@/*": ["../../packages/ui/src/*"]
```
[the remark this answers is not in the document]
maparent: I think I prefer tree-shaking to compile time in this case, but it could be revisited. For now I'll use the export map.

apps/website/app/utils/supabase/client.ts
```diff
@@ -0,0 +1,9 @@
+import { createClient as createSupabaseClient } from '@supabase/supabase-js'
+import { Database } from "@repo/database/types.gen.ts";
+
+export function createClient() {
```
mdroidian: .

apps/website/tsconfig.json
```diff
@@ -1,8 +1,10 @@
 {
   "extends": "@repo/typescript-config/nextjs.json",
   "compilerOptions": {
+    "baseUrl": ".",
     "paths": {
-      "~/*": ["./app/*"]
+      "~/*": ["./app/*"],
+      "@/*": ["@repo/ui/src/*"]
```
mdroidian: Interesting choice.  Any specific reason why the `ui` library should have `@` path?  Are you using this anywhere?

packages/ui/src/lib/supabase/client.ts
```diff
@@ -0,0 +1,12 @@
+import { createClient as createSupabaseClient } from "@supabase/supabase-js";
+import { Database } from "@repo/database/types.gen.ts";
+
+declare const SUPABASE_URL: string;
+declare const SUPABASE_ANON_KEY: string;
+
+export function createClient() {
```
mdroidian: .

packages/ui/src/lib/supabase/client.ts
```diff
@@ -0,0 +1,12 @@
+import { createClient as createSupabaseClient } from "@supabase/supabase-js";
+import { Database } from "@repo/database/types.gen.ts";
+
+declare const SUPABASE_URL: string;
```
mdroidian, line 4: Will this work in Obsidian right now?

apps/website/app/utils/supabase/middleware.ts
```diff
@@ -0,0 +1,65 @@
+import { createServerClient } from '@supabase/ssr'
+import { NextResponse, type NextRequest } from 'next/server'
+
+export async function updateSession(request: NextRequest) {
```
mdroidian: .

mdroidian (changes requested):
> There are many branches that assume access to supabase infrastructure in the front-end

Could you give an example (link the file or even directly to the code as GitHub will highlight it or embed it)  EG:

highlight from another PR
https://github.com/DiscourseGraphs/discourse-graph/pull/238#diff-f8c2d7f6d3c4f5f77bb660ebf330b6249b39e5faac089109479631cc05fbe994R20-R21

embedded
https://github.com/DiscourseGraphs/discourse-graph/blob/9cf83348366d8f45627f5c7bb3e5ee3c286b352d/.github/workflows/roam-pr.yaml#L20-L21


> ... ui work in the ENG-373 branch, so other branches can benefit from this infrastructure work, notably future ENG-420 work ...

Note: you can use the `#` in GitHub to link to other branches.  This makes it easier for anyone else viewing this PR to navigate to what you are referencing.  EG: #238

apps/website/tsconfig.json
maparent: This is following a pattern set by shadcn: https://ui.shadcn.com/docs/components-json#aliases
Maybe this should have been put in #238, where I introduce shadcn?
Shadcn itself is proposed by https://supabase.com/ui/docs/nextjs/password-based-auth
But I saw a lot of shadcn-related patterns in `@repo/ui`, which made me suspect you may have used it there without making it a dev dependency.

apps/website/app/utils/supabase/client.ts
maparent: Ah, damn. This was copy-pasted from the documentation, so I considered it as almost-library code, hence did not look inside beyond a point. But I agree that if it's checked in it needs to follow standards.
I will also add the provenance in a comment.

maparent:
On PR references: I find using the PR numbers confusing, as I think of the task in terms of linear task numbers. I wish there were a good way to make short linear references in github. But yeah, I also see the value of hyperlinking of course. Just pointing out a cognitive overhead.

apps/website/tsconfig.json
maparent: Yeah, I'll delay it to #238, where it belongs.
Rant: I wish I could make this kind of comment in Json, so I could explain it in-place!

maparent:
sorry, went too fast in re-request, one aspect is missing: Will it work in obsidian? I don't know, and I don't know that part of the codebase enough to test it. But I checked that, as long as the variables as defined (my warning at the top of the PR), obsidian builds correctly. As for calling it, I see no reason why not, but we may need to add environment passthrough here and there. I think we can wait until we actually call it to check that.

mdroidian:
> On PR references: I find using the PR numbers confusing, as I think of the task in terms of linear task numbers. I wish there were a good way to make short linear references in github. But yeah, I also see the value of hyperlinking of course. Just pointing out a cognitive overhead.

Because we are adding the linear id's in the PR titles, `#` will find these as well.  In terms of linking directly to linear, `Ctrl-C` when you are in a linear task will copy a markdown friendly URL to paste anywhere.

apps/website/app/utils/supabase/middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+  } = await supabase.auth.getUser();
+
+  if (
+    !user &&
+    !request.nextUrl.pathname.startsWith("/login") &&
+    !request.nextUrl.pathname.startsWith("/auth")
+  ) {
+    // no user, potentially respond by redirecting the user to the login page
+    const url = request.nextUrl.clone();
+    url.pathname = "/auth/login";
```
mdroidian, line 52: Looks like this will be added in https://github.com/DiscourseGraphs/discourse-graph/pull/238

apps/website/app/utils/supabase/middleware.ts
```diff
@@ -0,0 +1,70 @@
+import { createServerClient } from "@supabase/ssr";
+import { NextResponse, type NextRequest } from "next/server";
+
+// Inspired by https://supabase.com/ui/docs/nextjs/password-based-auth
+
+export const updateSession = async (request: NextRequest) => {
```
mdroidian, line 6: @maparent can you give an example of how / when this is to be used?  I didn't see it in #238

apps/website/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@repo/typescript-config/nextjs.json",
   "compilerOptions": {
+    "baseUrl": ".",
```
mdroidian, line 4: What is the purpose for adding this?  Or problem is this solving?

mdroidian:
> sorry, went too fast in re-request, one aspect is missing: Will it work in obsidian? I don't know, and I don't know that part of the codebase enough to test it. But I checked that, as long as the variables as defined (my warning at the top of the PR), obsidian builds correctly. As for calling it, I see no reason why not, but we may need to add environment passthrough here and there. I think we can wait until we actually call it to check that.

Ah, I was asking because I thought it was added for Obsidian:

> Note: There are a few non-obvious choices, such as putting the client code in the ui, which I think will benefit obsidian in the longer term.

mdroidian (approved):
A few clarifying questions/comments, but not blocking.

apps/website/app/utils/supabase/middleware.ts
maparent: It's not used yet; I was following the documented pattern in the url above.

apps/website/tsconfig.json
maparent: It was necessary for the paths to work well in #238, and also for the eslint work in #246 . Probably not strictly needed here, but I guess I got tired of git surgery.

mdroidian: That is fine, it is just helpful to add the context.

apps/website/app/utils/supabase/middleware.ts
mdroidian: I briefly scanned the URL but didn't explicitly see it's use, only the code.  Do you have a sense of where we will use it in the future?

PR #255, opened 2025-07-01
ENG-488 Change supabase config to agree with new pg17 project.

maparent:
This is a way to review the branch machinery as well as test Postgres 17


maparent:
This change does not trigger a supabase branch. Will wait for an opportune moment for the migration.

maparent:
This is not a test anymore, just a statement of fact. We're actually using 17 now and the local supabase should be made aware.

mdroidian (approved):

PR #268, opened 2025-07-07
ENG-549 Update database README with simplified setup and sync instructions

mdroidian:
Info/discussion: [ENG-549: remove `supabase link` from db steps](https://linear.app/discourse-graphs/issue/ENG-549/remove-supabase-link-from-db-steps)


maparent:
This is duplicating ongoing work in #249. I will use these changes, with modifications.

PR #269, opened 2025-07-07
ENG-544 sql repairs

maparent:
The restore failed because of an error with a missing schema identifier in the compute_arity_local, and also because of function ordering. This corrects the schema identifiers.
While investigating locally, I found many other small errors that were missed earlier, probably because Postgres 17 is better at catching them than 15 was. I also solved a divergence in temporary functions due to formatting.
After this is pushed, we will want to wait for a new backup cycle to happen, and download and test it locally.
(I have a concern that the function ordering issue will not be addressed; they are correctly ordered in schema. But they being named properly in the calling function may impose proper ordering of definitions.)

### DIFF
--- a/packages/database/supabase/migrations/20250707023406_more_corrections.sql
+++ b/packages/database/supabase/migrations/20250707023406_more_corrections.sql
@@ -1,0 +1,558 @@
+set check_function_bodies = off ;
+
+-- this was missing public prefix on functions
+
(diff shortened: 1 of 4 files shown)


maparent:
These are pure corrections, I don't think we need coderabbit to review.

mdroidian (approved):

PR #272, opened 2025-07-08
ENG-567 Refactor getContext so the PlatformAccount is created in a Space.

maparent:
Factor out some of the functions that call next from roam to ui code. 
Also work on a generic way to know the Next endpoint. (as described in #249, in [this file](https://github.com/DiscourseGraphs/discourse-graph/blob/ENG-520-supabase-dev-instructions/packages/database/README.md) )
The reason for doing that it that it is safer, in terms of PlatformAccount access, to always start from a Space login, even though the PlatformAccount is not bound to a single space.

## Summary by CodeRabbit

* **New Features**
  * Added utility to determine the API root URL based on environment, improving environment-based configuration.
  * Introduced functions for streamlined creation and authentication of platform spaces and accounts.
  * Added password validation requiring at least 8 characters during space creation.

* **Refactor**
  * Centralized and simplified Supabase account and space management logic, reducing code duplication and manual error handling.
  * Updated space creation input structure and validation to enhance consistency and security.

* **Bug Fixes**
  * Corrected import paths for utility functions to ensure proper module resolution.

* **Style**
  * Minor formatting improvements for code clarity.

### DIFF
--- a/apps/roam/scripts/compile.ts
+++ b/apps/roam/scripts/compile.ts
@@ -132,6 +132,7 @@ export const compile = ({
       define: {
         "process.env.SUPABASE_URL": `"${process.env.SUPABASE_URL}"`,
         "process.env.SUPABASE_ANON_KEY": `"${process.env.SUPABASE_ANON_KEY}"`,
+        "process.env.NEXT_API_ROOT": `"${process.env.NEXT_API_ROOT || ""}"`,
       },
       sourcemap: process.env.NODE_ENV === "production" ? undefined : "inline",
       minify: process.env.NODE_ENV === "production",
--- a/apps/roam/src/utils/supabaseContext.ts
+++ b/apps/roam/src/utils/supabaseContext.ts
@@ -7,11 +7,7 @@ import { Enums } from "@repo/database/types.gen";
 import { DISCOURSE_CONFIG_PAGE_TITLE } from "~/utils/renderNodeConfigPage";
 import getBlockProps from "~/utils/getBlockProps";
 import setBlockProps from "~/utils/setBlockProps";
-import {
-  createClient,
-  type DGSupabaseClient,
-} from "@repo/ui/lib/supabase/client";
-import { spaceAnonUserEmail } from "@repo/ui/lib/utils";
+import { type DGSupabaseClient } from "@repo/ui/lib/supabase/client";
 import {
   fetchOrCreateSpaceId,
   fetchOrCreatePlatformAccount,
--- a/apps/roam/src/utils/supabaseContext.ts
+++ b/apps/roam/src/utils/supabaseContext.ts
@@ -1,4 +1,3 @@
-import { getNodeEnv } from "roamjs-components/util/env";
 import getCurrentUserEmail from "roamjs-components/queries/getCurrentUserEmail";
 import getCurrentUserDisplayName from "roamjs-components/queries/getCurrentUserDisplayName";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
@@ -11,8 +10,13 @@ import setBlockProps from "~/utils/setBlockProps";
 import {
   createClient,
   type DGSupabaseClient,
-} from "@repo/ui/src/lib/supabase/client";
+} from "@repo/ui/lib/supabase/client";
 import { spaceAnonUserEmail } from "@repo/ui/lib/utils";
+import {
+  fetchOrCreateSpaceId,
+  fetchOrCreatePlatformAccount,
+  createLoggedInClient,
+} from "@repo/ui/lib/supabase/contextFunctions";
 
 declare const crypto: { randomUUID: () => string };
 
@@ -26,12 +30,6 @@ export type SupabaseContext = {
 };
 
 let _contextCache: SupabaseContext | null = null;
-
-// TODO: This should be an util on its own.
-const base_url =
-  getNodeEnv() === "development"
-    ? "http://localhost:3000/api/supabase"
-    : "https://discoursegraphs.com/api/supabase";
 
 const settingsConfigPageUid = getPageUidByPageTitle(
   DISCOURSE_CONFIG_PAGE_TITLE,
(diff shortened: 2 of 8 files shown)


maparent:
@coderabbit review

packages/ui/src/lib/supabase/contextFunctions.ts
```diff
@@ -0,0 +1,111 @@
...
+    {
+      onConflict: "space_id,account_id",
+      ignoreDuplicates: false,
+      count: "estimated",
+    },
+  );
+  if (resultUserSpaceAccess.error)
+    throw Error(resultUserSpaceAccess.error.message); // account created but not connected, try again
+  return account_id;
+};
```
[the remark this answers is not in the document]
maparent: The Supabase client does not offer transaction support.
My second best option is ensure each step is idempotent, with upserts, as done here. Retrying will not lead to orphans.
Another option would be to wrap this in a database function, but this did not feel fully appropriate.

maparent:
@coderabbit: Good catch on the imports, I'm still fighting how vscode and zed handle the configuration differently, and I let that slip.
As for more specific process.env definition: I think that this would add unnecessary coupling.

packages/ui/src/lib/supabase/contextFunctions.ts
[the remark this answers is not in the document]
maparent: Actually... now rethinking the approach and will use a database function, again to coordinate security. So that was a good suggestion.

maparent:
@coderabbit review

packages/ui/src/lib/supabase/contextFunctions.ts
```diff
@@ -0,0 +1,76 @@
+import { Enums } from "@repo/database/types.gen.ts";
+import { nextApiRoot, spaceAnonUserEmail } from "@repo/ui/lib/utils";
+import {
+  createClient,
+  type DGSupabaseClient,
+} from "@repo/ui/lib/supabase/client";
+
+type Platform = Enums<"Platform">;
+
+const base_url = nextApiRoot() + "/supabase";
```
mdroidian: camelCase

packages/ui/src/lib/utils.ts
```diff
@@ -1,9 +1,24 @@
...
+declare const process: { env: Record<string, any> };
+
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs));
 }
 
 export const spaceAnonUserEmail = (platform: string, space_id: number) =>
   `${platform.toLowerCase()}-${space_id}-anon@database.discoursegraphs.com`;
+
+export const IS_DEV = process.env.NODE_ENV !== "production";
```
mdroidian: Let's not export if we aren't explicitly using it somewhere else.

packages/ui/src/lib/supabase/contextFunctions.ts
```diff
@@ -0,0 +1,76 @@
+import { Enums } from "@repo/database/types.gen.ts";
```
mdroidian, line 1: Not a huge fan of this being in `packages/ui`, as it's not a UI component.  I envision `packages/ui` to mostly be made up of `.tsx` React components and their utils/etc.

But, `client.ts` is already there, so I won't block on this.

We'll probably move it in the future.  Maybe put it in `packages/database/utils`?

apps/roam/src/utils/supabaseContext.ts
```diff
@@ -11,8 +10,13 @@ import setBlockProps from "~/utils/setBlockProps";
 import {
   createClient,
   type DGSupabaseClient,
-} from "@repo/ui/src/lib/supabase/client";
+} from "@repo/ui/lib/supabase/client";
 import { spaceAnonUserEmail } from "@repo/ui/lib/utils";
```
mdroidian: unused

apps/roam/src/utils/supabaseContext.ts
```diff
@@ -11,8 +10,13 @@ import setBlockProps from "~/utils/setBlockProps";
 import {
   createClient,
```
mdroidian: unused

packages/ui/src/lib/utils.ts
```diff
@@ -1,9 +1,24 @@
 import { type ClassValue, clsx } from "clsx";
 import { twMerge } from "tailwind-merge";
 
+declare const process: { env: Record<string, any> };
```
mdroidian, line 4: Is this currently assuming that the value will be provided by `apps/roam` 's compile script?  If we go ahead with `NEXT_API_ROOT`, maybe we should add a `.env` file to `packages/ui` and locate `NEXT_API_ROOT` there so it is decoupled.

mdroidian (changes requested):

packages/ui/src/lib/supabase/contextFunctions.ts
maparent: It makes sense to keep `packages/ui` for ui code; but I think at some point we'll want a package for shared code that is non-UI, but maybe not db-specific either. How about having a `packages/utils`, with a database subdirectory of that? Anyway, agree we should leave that refactoring for another PR.

apps/roam/src/utils/supabaseContext.ts
maparent: sorry, remnant of many refactorings.

packages/ui/src/lib/utils.ts
maparent: It is assumed, and as I understand it, `packages/ui` is not compiled individually, but as included by each app using it. So the applicable `.env` would be the app's in all cases.

apps/roam/scripts/compile.ts
```diff
@@ -132,6 +132,7 @@ export const compile = ({
       define: {
         "process.env.SUPABASE_URL": `"${process.env.SUPABASE_URL}"`,
         "process.env.SUPABASE_ANON_KEY": `"${process.env.SUPABASE_ANON_KEY}"`,
+        "process.env.NEXT_API_ROOT": `"${process.env.NEXT_API_ROOT || ""}"`,
```
mdroidian, line 135: I understand you reference [this file](https://github.com/DiscourseGraphs/discourse-graph/blob/ENG-520-supabase-dev-instructions/packages/database/README.md), which says,

> If you are working on frontend code, you may use the environment variable NEXT_API_ROOT (forthcoming) to get the api of the active vercel branch instance (as given in the [vercel interface](https://vercel.com/discourse-graphs/discourse-graph/deployments).)

By "API" do you mean the "API URL"?

Is the intended workflow:
- publish a branch/create a PR
- log in to vercel/use the CLI/navigate to the PR
- find the branch's URL
- add that URL + "/api" to your `.env`/vercel `env` as the value for `NEXT_API_ROOT`?

Is there a plan to update all of the other places in the code that the API URL is referenced?

mdroidian (changes requested):
I was switching back and forth between GitHub's new preview view, looks like this comment got lost somehow.

apps/roam/scripts/compile.ts
maparent: Yes, API URL. Will rephrase.
Correct on workflow. That is in the case where you want to use your local dev code with the vercel branch (talking to the supabase branch.)
And yes, making a global update would be relevant. Presumably in a separate task.

mdroidian: Introducing it now with no mention or reference to a future task to update the rest of the code will likely cause issues and confusion when someone tries to use it (as they would expect the API calls to use `NEXT_API_ROOT`when they don't, which wouldn't be fun to debug).  If you could go ahead and create the future task, assign it to yourself, and link it here, then we can merge this.

mdroidian: > That is in the case where you want to use your local dev code with the vercel branch

It's also worth noting that this case is probably quite rare.  Usually people will use `localhost`.  Maybe useful when you need to redirect a third party to a live URL, but currently something like ngrok is a better dev experience because you don't have to wait for the vercel build every commit.

maparent: Good, created the task https://linear.app/discourse-graphs/issue/ENG-585/generalize-usage-of-nextapiroot

packages/ui/src/lib/supabase/contextFunctions.ts
```diff
@@ -0,0 +1,76 @@
...
+import { nextApiRoot, spaceAnonUserEmail } from "@repo/ui/lib/utils";
+import {
+  createClient,
+  type DGSupabaseClient,
+} from "@repo/ui/lib/supabase/client";
+
+type Platform = Enums<"Platform">;
+
+const baseUrl = nextApiRoot() + "/supabase";
```
mdroidian, line 10: Note, currently only works in if bundled with Roam.

packages/ui/src/lib/utils.ts
```diff
@@ -1,9 +1,24 @@
...
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs));
 }
 
 export const spaceAnonUserEmail = (platform: string, space_id: number) =>
   `${platform.toLowerCase()}-${space_id}-anon@database.discoursegraphs.com`;
+
+const IS_DEV = process.env.NODE_ENV !== "production";
+
+export const nextApiRoot = (): string => {
```
mdroidian, line 15: Note, `NEXT_API_ROOT` only works in `contextFunctions()` at the moment.
Task to add support for the rest of the repo: https://linear.app/discourse-graphs/issue/ENG-585/generalize-usage-of-nextapiroot

mdroidian (approved):

PR #275, opened 2025-07-12
ENG-582 setting an explicit empty search path for each function definition

maparent:
... following supabase recommendation. Note that it cannot be set to empty for the embeddings, because of the <=> operator; and supabase still complains with an explicit non-empty search_path.
Also some linting corrections.


mdroidian:
> ... following supabase recommendation

@maparent could add a link to the Supabase recommendation?

maparent:
Ah, sorry! https://supabase.com/dashboard/project/zytfjzqyijgagqxrzbmz/advisors/security?preset=WARN&id=
It was very visible to me in the local db after I solved the other issues.

mdroidian (approved):

PR #276, opened 2025-07-12
ENG-583-add-target-type-column

maparent:
As described in [eng-583](https://linear.app/discourse-graphs/issue/ENG-583/complete-the-universal-foreign-key-in-the-sync-table):
The sync_info table was designed so the target could be a member of any table, i.e. not an explicit foreign key. This is an implementation of my favoured path, adding an EntityType column so we can ensure proper logic.
Having this allows me to add RLS to that table.


mdroidian (approved):
Let's try to make sure the title/branch name are for the same linear task id.
This is now linking/automating the status' of both task.

image

If you want to reference a related task that is not the primary task, you can add a link to the body comment and it won't trigger the linear app.

PR #257, opened 2025-07-03
ENG-539 ensure all vercel variables are accounted for, mostly globally

maparent:
This is to reduce the number of warnings in vercel deploy, and ensure all relevant variables are available.

PR #277, opened 2025-07-12
ENG-546 reintroduce GITHUB_TOKEN

maparent:
Was removed by mistake in ENG-539
https://github.com/DiscourseGraphs/discourse-graph/pull/257

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -55,7 +55,7 @@
       "cache": false,
       "inputs": ["$TURBO_DEFAULT$", ".env*"],
       "env": ["APP_ID"],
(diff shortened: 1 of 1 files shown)


mdroidian (approved):
@maparent I updated the original PR comment to include a link to the PR that reverts it. When PRs are linked like this, GitHub adds inline references to both, which helps future developers quickly see the relationship between them. If someone is viewing the original PR, they'll know it was reverted. And if they're viewing the revert PR, they’ll easily find the context of the original change. It’s a small habit that really helps with traceability—your future self will thank you! 😄

image

PR #246, opened 2025-06-26
Eng-521 repair and modernize eslint with flat config

maparent:
We have an issue with turbo lint not working well, not linting all subprojects, and not using our eslint settings.
This PR is mostly a modernization of our eslint configuration, with new flat configs.
I have also updated prettier, and made sure all packages were using the same version of typescript.
Then, this did not work as well in vscode, so I had to be more more explicit about subprojects.
It was subsequently aligned with a fresh turbo repo (from `npx create-turbo@latest`) and the differences were minimized, while still maintaining a functional turbo build / dev / lint.

## Summary by CodeRabbit

- **New Features**
  - Introduced new ESLint and TypeScript lint configuration files for improved code linting consistency across multiple projects.
  - Added a multi-root workspace configuration for easier project management.

- **Chores**
  - Migrated ESLint configurations to the new flat config format and updated related settings.
  - Updated and streamlined TypeScript configuration files for better project structure and exclusion of build directories.
  - Revised and upgraded various development dependencies for ESLint, Prettier, and TypeScript.
  - Updated lint scripts and relaxed lint warning limits in some packages.
  - Removed deprecated or redundant configuration files.

PR #293, opened 2025-07-20
ENG-614: harmonize license to apache2 everywhere

maparent:


mdroidian (approved):
Note: this is reverting a change from #246

PR #297, opened 2025-07-21
Roam: [Eng-610] tag assignment in roam dg plugin settings menu

sid597:
https://www.loom.com/share/4630914958e0445981dced88379e536c?sid=f72fb8bc-da8b-48cd-aaa0-b02d2647966e

## Summary by CodeRabbit

* **New Features**
  * Added a "Tag" field to node configuration, allowing users to specify a hashtag for each node type.
  * The "Tag" field is now visible and editable when creating or configuring nodes.

* **Bug Fixes**
  * Introduced validation for the "Tag" and "Format" fields to prevent conflicts, with error messages displayed for invalid input.

* **Style**
  * Updated the layout of the "General" tab to display input panels in a horizontal row for improved usability.


sid597:
@coderabbitai full review

sid597:
@coderabbitai full review

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -40,6 +120,68 @@ const NodeConfig = ({
   });
 
   const [selectedTabId, setSelectedTabId] = useState<TabId>("general");
+  const [tagError, setTagError] = useState("");
+  const [formatError, setFormatError] = useState("");
+  const isValid = !tagError && !formatError;
```
mdroidian: Is what valid?  Let's be more specific.  There are many things this could refer to.

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -1,17 +1,96 @@
...
+  error,
+}: {
+  label: string;
+  description: string;
+  value: string;
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  onBlur: () => void;
+  error: string;
+}) => (
+  <div>
```
mdroidian: Only use `<div>` if it provides a necessary structure.  Use `<>` if the additional DOM element isn't required and we just need to satisfy the JSX constraint.

apps/roam/src/utils/getDiscourseNodes.ts
```diff
@@ -15,6 +15,7 @@ export type DiscourseNode = {
   text: string;
   type: string;
   shortcut: string;
+  tag: string;
```
mdroidian: Let's make this optional, which would more directly reflect it's actual use.
Unfortunately this will have some cascading effects that will requires some code changes, but it will be worth it in the long run.

apps/roam/src/utils/getDiscourseNodes.ts
```diff
@@ -110,6 +114,7 @@ const getDiscourseNodes = (relations = getDiscourseRelations()) => {
           text: r.label,
           type: r.id,
           shortcut: r.label.slice(0, 1),
+          tag: "",
```
mdroidian: can rm if optional

apps/roam/src/utils/getDiscourseNodes.ts
```diff
@@ -32,6 +33,7 @@ const DEFAULT_NODES: DiscourseNode[] = [
     text: "Page",
     type: "page-node",
     shortcut: "p",
+    tag: "",
```
mdroidian: can rm if optional

apps/roam/src/utils/getDiscourseNodes.ts
```diff
@@ -49,6 +51,7 @@ const DEFAULT_NODES: DiscourseNode[] = [
     text: "Block",
     type: "blck-node",
     shortcut: "b",
+    tag: "",
```
mdroidian: can rm if optional

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -40,6 +120,68 @@ const NodeConfig = ({
...
+    const formatUpper = formatWithoutPlaceholders.toUpperCase();
+    const formatParts = formatUpper.split(/[^A-Z0-9]/);
+    const hasConflict = formatParts.includes(cleanTag);
+
+    if (hasConflict) {
+      const formatForMessage = formatWithoutPlaceholders
+        .trim()
+        .replace(/(\s*-)*$/, "");
+      setFormatError(
+        `Format "${formatForMessage}" conflicts with tag: "${tag}". Please use some other format.`,
```
mdroidian: Let's be slightly more specific.  Something like "`${tag}` cannot also be in Format"

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -40,6 +120,68 @@ const NodeConfig = ({
...
+    if (!cleanTag) {
+      setTagError("");
+      setFormatError("");
+      return;
+    }
+
+    const formatWithoutPlaceholders = format.replace(/{[^}]+}/g, "");
+    const formatUpper = formatWithoutPlaceholders.toUpperCase();
+    const formatParts = formatUpper.split(/[^A-Z0-9]/);
+    const hasConflict = formatParts.includes(cleanTag);
```
mdroidian: This is checking if the tag appears anywhere as word in the format.  Which means even this would be invalid:

tag: claim
format: [[CLM]] - my claim is: {content}

What we need to do is check if the tag is **referenced** anywhere in the format.  Eg: if it appears in like `[[claim]]` or `#[[claim]]` or (rare) `#[[claim]]`.

I don't think `roamjs-components` has a regex for this, but there is `extractTag` for going the other way that you can use in the function.

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -1,17 +1,96 @@
...
+  value: string;
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  onBlur: () => void;
+  error: string;
+}) => (
+  <div>
+    <Label>
+      {label}
+      <Description description={description} />
+      <InputGroup value={value} onChange={onChange} onBlur={onBlur} />
```
mdroidian: We could add a `placeholder` that is `#{node.text}`, but fine to consult @jsmorabito first.

mdroidian (changes requested):
Good stuff so far!  The biggest blocker is the validation.  We need to check if the tag is **referenced**, not just if the word shows up. 👍

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -15,6 +15,7 @@ import { OnloadArgs } from "roamjs-components/types";
 import getBasicTreeByParentUid from "roamjs-components/queries/getBasicTreeByParentUid";
 import createBlock from "roamjs-components/writes/createBlock";
 import updateBlock from "roamjs-components/writes/updateBlock";
+import extractTag from "roamjs-components/util/extractTag";
```
mdroidian: unused
image

mdroidian (approved):
👍

PR #298, opened 2025-07-21
Roam: [Eng-621] show node tag options in inline node creation menu

sid597:
https://www.loom.com/share/d3d53d536c1d445c9052e01517eac201?sid=76e85ee4-1412-4a83-b48a-6000774dbe66

## Summary by CodeRabbit

* **New Features**
  * Added the ability to toggle the node menu between showing full discourse node types and a simplified tag view, based on Shift key state and text selection.
  * Menu display dynamically updates according to keyboard interaction and selected text.

* **Improvements**
  * Enhanced keyboard handling for more intuitive menu toggling.
  * Menu now visually distinguishes between node types and tags for easier selection.


sid597:
@coderabbitai full review

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -204,7 +256,9 @@ const NodeMenu = ({
                   />
                 }
                 labelElement={
-                  <span className="font-mono">{item.shortcut}</span>
+                  <span className="font-mono">
+                    {showNodeTypes ? item.shortcut : ""}
```
mdroidian: They should still be able to use the shortcut, regardless if it is the node or tag.

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -32,17 +32,25 @@ type Props = {
...
   extensionAPI,
   trigger,
+  isShift,
+  isTextSelected,
 }: { onClose: () => void } & Props) => {
+  const [showNodeTypes, setShowNodeTypes] = useState(isTextSelected || isShift);
   const discourseNodes = useMemo(
-    () => getDiscourseNodes().filter((n) => n.backedBy === "user"),
-    [],
+    () =>
```
mdroidian: Let's not requery the discourse nodes every time `shift` is selected.
We can query them once in the `useMemo` then just change which ones are shown if shift is clicked.

You can either store two seperate vars (nodes/tags) or something like `nodesToShow` and filter like you are currently doing `&& (showNodeTypes || n.tag)`, or both.

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -56,51 +64,79 @@ const NodeMenu = ({
...
+          });
+
+          createDiscourseNode({
+            text: pageName,
+            configPageUid: nodeUid,
+            extensionAPI,
+          });
         });
+      } else {
+        const menuItem =
```
mdroidian: move this outside the `if` so it is DRY

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -56,51 +64,79 @@ const NodeMenu = ({
...
+            text: pageName,
+          });
+
+          createDiscourseNode({
+            text: pageName,
+            configPageUid: nodeUid,
+            extensionAPI,
+          });
         });
+      } else {
```
mdroidian, line 109: add posthog.capture to the else

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -190,7 +241,8 @@ const NodeMenu = ({
               <MenuItem
                 key={item.text}
                 data-node={item.type}
-                text={item.text}
+                data-tag={item.tag}
+                text={showNodeTypes ? item.text : `#${item.tag || 'untagged'}`}
```
mdroidian: If we are showing items that don't have tags, this should probably just be `""` and the `MenuItem` should be disabled.

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -238,10 +292,12 @@ export const TextSelectionNodeMenu = ({
   textarea,
   extensionAPI,
   onClose,
+  isTextSelected,
 }: {
   textarea: HTMLTextAreaElement;
   extensionAPI: OnloadArgs["extensionAPI"];
   onClose: () => void;
+  isTextSelected?: boolean;
```
mdroidian: In what scenario will this be false?

mdroidian (changes requested):
Good stuff so far!

mdroidian:
@sid597 When no text is selected and you press `Shift` to show nodeTypes, user cannot use up/down/enter or the Keyboard shortcut to select.

https://www.loom.com/share/66c7c188e0cb4fc1a733ddf3ffc3a358

sid597:
> @sid597 When no text is selected and you press Shift to show nodeTypes, user cannot use up/down/enter or the Keyboard shortcut to select.

~Missed this before asking for review will fix~ 
Fixed now

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -42,15 +41,15 @@ const NodeMenu = ({
...
+    () => getDiscourseNodes().filter((n) => n.backedBy === "user"),
+    [],
+  );
   const discourseNodes = useMemo(
-    () =>
-      getDiscourseNodes().filter(
-        (n) => n.backedBy === "user" && (showNodeTypes || n.tag),
-      ),
-    [showNodeTypes],
+    () => userDiscourseNodes.filter((n) => showNodeTypes || n.tag),
```
mdroidian: Because `getDiscourseNodes` is memoized (which is the expensive query), `discourseNodes` doesn't also need to be memoized because it is a simple filter.

mdroidian (approved):
👍

PR #304, opened 2025-07-25
Roam: ENG-622: Node tag to node conversions

sid597:
## Summary by CodeRabbit

* **New Features**
  * Introduced a modal dialog for creating discourse nodes, accessible from tag hover actions.
  * Added a popup button that appears when hovering over certain tags, allowing quick creation of nodes with prefilled information.

* **Style**
  * Updated string formatting for consistency in menu item rendering.


sid597:
@coderabbitai full review

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -289,6 +294,54 @@ export const initObservers = async ({
...
+    if (!target || !target.classList?.contains("rm-page-ref--tag")) return;
+
+    const textContent = target.textContent?.trim() || "";
+    const tagAttr = target.getAttribute("data-tag") || textContent;
+    const tag = tagAttr.replace(/^#/, "").toLowerCase();
+    const discourseNodes = getDiscourseNodes();
+    const discourseTagSet = new Set(
+      discourseNodes
+        .map((n) => n.tag?.toLowerCase())
+        .filter(Boolean) as string[],
```
mdroidian: image

apps/roam/src/utils/renderNodeTagPopup.tsx
```diff
@@ -0,0 +1,44 @@
...
+}: {
+  tagElement: HTMLElement;
+  onClick: () => void;
+  label?: string;
+}) => {
+  removeNodeTagPopup();
+
+  const rect = tagElement.getBoundingClientRect();
+
+  currentPopup = document.createElement("div");
```
mdroidian: Let's use Blueprint's `<Popover>` here.  It's a pattern we've used quite a bit in the repo, and it handles things like positioning based on screen/DOM, click/hover, esc, overflow, etc.

See `NodeMenu` for the `render`/`<Popover>` pattern.

apps/roam/src/utils/renderNodeTagPopup.tsx
```diff
@@ -0,0 +1,44 @@
...
+  currentPopup.id = "discourse-node-tag-popup";
+  currentPopup.className = "absolute z-[9999] max-w-none font-inherit bg-white";
+  currentPopup.style.left = `${rect.left + window.scrollX}px`;
+  currentPopup.style.top = `${rect.bottom + window.scrollY + 4}px`;
```
mdroidian: Let's put this above the text, as per the design.  With `<Popover>` we can set `position:top`

apps/roam/src/utils/renderNodeTagPopup.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  document.body.appendChild(currentPopup);
+
+  currentPopup.addEventListener("mouseleave", removeNodeTagPopup, {
+    once: true,
+  });
+
+  ReactDOM.render(
+    <Button intent="primary" minimal onClick={onClick} text={label} />,
```
mdroidian: Let's give this `outlined` so it is slightly more visible.  Be sure to turn off Roam Studio or any other css plugins when testing.

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+    ).trim();
+
+    let formattedTitle: string;
+    if (!format) {
+      formattedTitle = title.trim();
+    } else if (/{content}/i.test(format)) {
+      formattedTitle = format.replace(/{content}/gi, title.trim());
+    } else {
+      formattedTitle = `${format} ${title.trim()}`.trim();
+    }
```
mdroidian: We already do this in `getNewDiscourseNodeText`.  See what you can take/combine from there.  It also include adding the `{referencedNode}` eg: when an EVD is created in a @Source page.

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+    const newPageUid = await createDiscourseNode({
```
mdroidian, line 62: image

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+    if (blockUid) {
+      const pageRef = `[[${formattedTitle}]]`;
+      await updateBlock({ uid: blockUid, text: pageRef });
+
+      const newCursorPosition = pageRef.length;
+      const windowId =
+        window.roamAlphaAPI.ui.getFocusedBlock?.()?.["window-id"] || "main";
+
+      if (window.roamAlphaAPI.ui.setBlockFocusAndSelection) {
+        window.roamAlphaAPI.ui.setBlockFocusAndSelection({
```
mdroidian: await 
image

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+          location: { "block-uid": blockUid, "window-id": windowId },
+          selection: { start: newCursorPosition },
+        });
+      } else {
+        setTimeout(() => {
+          const textareaElements = document.querySelectorAll("textarea");
+          for (const el of textareaElements) {
+            if (getUids(el as HTMLTextAreaElement).blockUid === blockUid) {
+              (el as HTMLTextAreaElement).focus();
+              (el as HTMLTextAreaElement).setSelectionRange(
```
mdroidian: rm unnecessary assertions

image

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+    if (blockUid) {
+      const pageRef = `[[${formattedTitle}]]`;
+      await updateBlock({ uid: blockUid, text: pageRef });
+
+      const newCursorPosition = pageRef.length;
+      const windowId =
+        window.roamAlphaAPI.ui.getFocusedBlock?.()?.["window-id"] || "main";
+
+      if (window.roamAlphaAPI.ui.setBlockFocusAndSelection) {
```
mdroidian: I don't believe we need this `if` or the entire `else` statement.

apps/roam/src/components/CreateNodeDialog.tsx
mdroidian: But this entire `else` isn't required.

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    const newPageUid = await createDiscourseNode({
+      text: formattedTitle,
+      configPageUid: selectedType.type,
+      extensionAPI,
+    });
+
+    if (blockUid) {
+      const pageRef = `[[${formattedTitle}]]`;
```
mdroidian, line 73: This assumes `format` or `specification` with only `has title`.  EG: if specification was `is in page with title` then actually it would be a block, not a page.  This is fine for now, as we are limiting support to `title`, but let's note this for the future.

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+          }
+        }, 50);
+      }
+    }
+
+    renderToast({
+      id: `discourse-node-created-${Date.now()}`,
+      intent: "success",
+      timeout: 10000,
+      content: `Created node [[${formattedTitle}]]`,
```
mdroidian: Let's make this clickable, something like:

```
 <a
    onClick={(event) => {
      if (event.shiftKey) {
        window.roamAlphaAPI.ui.rightSidebar.addWindow({
          window: {
            "block-uid": newPageUid,
            type: "outline",
          },
        });
      } else {
        window.roamAlphaAPI.ui.mainWindow.openPage({
          page: { uid: newPageUid},
        });
      }
    }}
  >
    [[{formattedTitle}]]
  </a>
```

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+    });
+    setLoading(false);
+    onClose();
+  };
+
+  return (
+    <Dialog
+      isOpen={isOpen}
+      onClose={onClose}
+      title="Create node"
```
mdroidian: "Create Discourse Node"

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+      autoFocus={false}
+    >
+      <div className={Classes.DIALOG_BODY}>
+        <div className="flex flex-col gap-4">
+          <div>
+            <label className="mb-1 block font-bold">Title</label>
```
mdroidian, line 129: Use blueprint `<Label>`

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+            <InputGroup
+              placeholder={`This is a potential ${selectedType.text.toLowerCase()}`}
+              value={title}
+              onChange={(e) => setTitle(e.currentTarget.value)}
+              inputRef={inputRef}
+            />
+          </div>
+
+          <div>
+            <label className="mb-1 block font-bold">Type</label>
```
mdroidian: Use blueprint `<Label>`.  Search for `<Dialog` in the repo to see the pattern in how it's used (`<Label> wraps <MenuItemSelect>`)

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+            <HTMLSelect
```
mdroidian: Let's use `<MenuItemSelect>` from roamjs-components which ties into Blueprints React version of HTMLSelect, and will be more consistent as it it used across the repo.

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { OnloadArgs } from "roamjs-components/types";
+import updateBlock from "roamjs-components/writes/updateBlock";
+import { render as renderToast } from "roamjs-components/components/Toast";
+import getUids from "roamjs-components/dom/getUids";
+import { DiscourseNode } from "~/utils/getDiscourseNodes";
+
+export type CreateNodeDialogProps = {
+  isOpen: boolean;
+  onClose: () => void;
+  nodeTypes: DiscourseNode[];
```
mdroidian: Let's not pass this.  Use this inline:

`const discourseNodes = getDiscourseNodes().filter(excludeDefaultNodes);`

`excludeDefaultNodes` filters `page` and `block`

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -289,6 +294,54 @@ export const initObservers = async ({
...
+
+    const rawBlockText = blockUid ? getTextByBlockUid(blockUid) : "";
+    const cleanedBlockText = rawBlockText.replace(textContent, "").trim();
+
+    renderNodeTagPopup({
+      tagElement: target,
+      label: `+ Create ${matchedNode.text}`,
+      onClick: () => {
+        renderCreateNodeDialog({
+          onClose: removeNodeTagPopup,
```
mdroidian: This should happen before the CreateNodeDialog closes.

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+import renderOverlay from "roamjs-components/util/renderOverlay";
+import createDiscourseNode from "~/utils/createDiscourseNode";
+import { OnloadArgs } from "roamjs-components/types";
+import updateBlock from "roamjs-components/writes/updateBlock";
+import { render as renderToast } from "roamjs-components/components/Toast";
+import getUids from "roamjs-components/dom/getUids";
+import { DiscourseNode } from "~/utils/getDiscourseNodes";
+
+export type CreateNodeDialogProps = {
+  isOpen: boolean;
```
mdroidian: isOpen is probably redundant in this component.  It will always be true.

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+  initialTitle: string;
+};
+
+const CreateNodeDialog = ({
+  isOpen,
+  onClose,
+  nodeTypes,
+  defaultNodeType,
+  extensionAPI,
+  blockUid,
```
mdroidian: nit: `blockUid` is rather ambigious.  Maybe something like `sourceBlockUid` or `originBlockUid`

mdroidian (changes requested):
Good start!  There is lots to consider with something like `CreateNodeDialog` so there's quite a few comment/suggestions.

We generate a type of "CreateNodeDialog" in multiple places (Canvas/DiscourseNodeMenu/DG Popover Menu/command palette (forthcoming)), so there are lots of previous patterns to glean from.

Just FYI, we'll be combing these in the future [ENG-555: Consolidate node creation implementations](https://linear.app/discourse-graphs/issue/ENG-555/consolidate-node-creation-implementations) but your `CreateNodeDialog` looks like a pretty good start 👍

apps/roam/src/components/CreateNodeDialog.tsx
sid597: this does not show up in my ide, are you using a plugin?

mdroidian: I believe this is a part of the linting PR that was recently merged.  Try checking out main, `git pull`, and adding code like this.

apps/roam/src/utils/renderNodeTagPopup.tsx
sid597: tried doing it but not able to get anywhere with it for quite some time, the popover needs a target to bind and creating the target then attaching to the popover leads to popover button rendered but not closing on mouse move from the button. keeping this part the same, need your help. Will give it another try tomorrow.

mdroidian: This should be automatic.  Have you looked over the Popover docs? https://blueprintjs.com/docs/versions/3/#core/components/popover

https://www.loom.com/share/c7b4ec262839441ba81528bb857a4689

mdroidian: You can see this exact pattern in Kanban as well: https://github.com/DiscourseGraphs/discourse-graph/blob/main/apps/roam/src/components/results-view/Kanban.tsx#L651

https://www.loom.com/share/c4b314168eee48108177c368f40b56c3

Attach the to the `span` of the hashtag as the target, as it already exists.

sid597: Yeah the issue is that the span that we get is not a react element whereas the target requires `(property) target?: string | JSX.Element | undefined` so what we have to do is create a dummy element and attach it to dom and then render popover that element

mdroidian: Let's see what you have at our next meeting and walk through it together.

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -0,0 +1,177 @@
...
+    });
+    setLoading(false);
+    onClose();
+  };
+
+  return (
+    <Dialog
+      isOpen={true}
+      onClose={onClose}
+      title="Create Discourse node"
```
mdroidian: "Discourse Node" (capital N)

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -287,6 +292,52 @@ export const initObservers = async ({
     }
   };
 
+  const nodeTagHoverListener = (e: Event) => {
```
mdroidian: As discussed in our last meeting, this should be an html observer instead of a hover listener.  This will make it fire far less.

apps/roam/src/utils/renderNodeTagPopup.tsx
```diff
@@ -0,0 +1,139 @@
...
+}: {
+  onClick: () => void;
+  label: string;
+  onClose: () => void;
+  tagElement: HTMLElement;
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const timeoutRef = useRef<number>();
+
+  useEffect(() => {
```
mdroidian: None of these listeners should be required.  If you rewatch our last meeting you'll see that just the initial observer code + popover will handle this.

apps/roam/src/utils/renderNodeTagPopup.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const rect = tagElement.getBoundingClientRect();
+
+  return (
+    <Popover
+      content={
+        <div
+          onMouseEnter={handlePopoverMouseEnter}
+          onMouseLeave={handlePopoverMouseLeave}
+        >
+          <Button intent="primary" outlined onClick={onClick} text={label} />
```
mdroidian: add `minimal` and remove `intent="primary"`

apps/roam/src/utils/renderNodeTagPopup.tsx
```diff
@@ -0,0 +1,139 @@
...
+        </div>
+      }
+      target={
+        <div
+          style={{
+            position: "fixed",
```
mdroidian: We shouldn't need to set the position like this.  It is not centered.
image

Again, if you see our last meeting, the overlay was perfectly centered automagically.

image

mdroidian (changes requested):
The biggest change we need to do is use an observer instead of a listener.  We talked about this at the Dev meeting and we walked through some example code in our 1:1.

Here is the example code from our meeting:

```typescript
const attributeObserver = createHTMLObserver({
    className: "rm-page-ref--tag",
    tag: "SPAN",
    callback: (s: HTMLSpanElement) => {
      // if (discourseTagSet.has(tag)) {
      renderAttributeButton(s);
      // }
    },
  });
```

I'll also attach the example button and styling code as it had a nice y offset and was centered.

```typescript
export const renderAttributeButton = (parent: HTMLSpanElement) => {
  if (parent.dataset.attributeButtonRendered === "true") return;
  parent.dataset.attributeButtonRendered = "true";
  const wrapper = document.createElement("span");
  wrapper.style.position = "relative";
  wrapper.style.display = "inline-block";
  parent.parentNode?.insertBefore(wrapper, parent);
  wrapper.appendChild(parent);
  const reactRoot = document.createElement("span");
  reactRoot.style.position = "absolute";
  reactRoot.style.top = "0";
  reactRoot.style.left = "0";
  reactRoot.style.width = "100%";
  reactRoot.style.height = "100%";
  reactRoot.style.pointerEvents = "auto";
  reactRoot.style.zIndex = "10";
  wrapper.appendChild(reactRoot);
  ReactDOM.render(<AttributeButtonPopover />, reactRoot);
};
```

Popover modifiers
```
modifiers={{
        offset: {
          offset: "0, 10",
        },
        arrow: {
          enabled: false,
        },
      }}
```

Popover target

```typescript
<span
  style={{
    display: "block",
    width: "100%",
    height: "100%",
 }}
/>
```

apps/roam/src/utils/initializeObserversAndListeners.ts
sid597: yeah my bad, i asked for review before our meeting then we had the review in meeting and i did not make this pr a draft pr after meeting

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -229,7 +229,7 @@ const NodeConfig = ({
                 onChange={handleTagChange}
                 onBlur={handleTagBlur}
                 error={tagError}
-                placeholder={`#${node.text}`}
```
mdroidian, line 232: Actually I liked the `#` 😁.

I'm going to create a task to handle `#`, as we discussed today.  We can re-add it once the use case is handled 😃

mdroidian (approved):
Nice 🚀

apps/roam/src/components/settings/NodeConfig.tsx
mdroidian: [ENG-693: Handle `#` in tags](https://linear.app/discourse-graphs/issue/ENG-693/handle-in-tags)

PR #305, opened 2025-07-26
ENG-642 Minor adjustments to database files

maparent:
https://linear.app/discourse-graphs/issue/ENG-642/minor-adjustments-to-database-files

The new supabase version changes the types.gen file.

correct sqruff configuration file so the faulty rule is ignored.

Apply sqruff cosmetic fixes to schema.

### DIFF
--- a/packages/database/supabase/schemas/access_token.sql
+++ b/packages/database/supabase/schemas/access_token.sql
@@ -27,7 +27,7 @@ revoke trigger on table "public"."access_token" from "anon";
 grant select on table "public"."access_token" to "anon";
 grant insert on table "public"."access_token" to "anon";
 
-ALTER TABLE public.access_token ENABLE ROW LEVEL SECURITY ;
+alter table public.access_token enable row level security ;
 
-DROP POLICY IF EXISTS access_token_policy ON public.access_token ;
-CREATE POLICY access_token_policy ON public.access_token FOR ALL USING (public.my_account(platform_account_id)) ;
+drop policy if exists access_token_policy on public.access_token ;
+create policy access_token_policy on public.access_token for all using (public.my_account(platform_account_id)) ;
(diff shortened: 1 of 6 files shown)


packages/database/supabase/schemas/access_token.sql
```diff
@@ -27,7 +27,7 @@ revoke trigger on table "public"."access_token" from "anon";
 grant select on table "public"."access_token" to "anon";
 grant insert on table "public"."access_token" to "anon";
 
-ALTER TABLE public.access_token ENABLE ROW LEVEL SECURITY ;
+alter table public.access_token enable row level security ;
 
-DROP POLICY IF EXISTS access_token_policy ON public.access_token ;
-CREATE POLICY access_token_policy ON public.access_token FOR ALL USING (public.my_account(platform_account_id)) ;
+drop policy if exists access_token_policy on public.access_token ;
+create policy access_token_policy on public.access_token for all using (public.my_account(platform_account_id)) ;
```
mdroidian: Curious, why is it lowercase in some places but UPPERCASE in others?

mdroidian (approved):

packages/database/supabase/schemas/access_token.sql
maparent: sqruff asks for each file to be consistent, but does not enforce consistency across files. First token sets the tone.
You started access_token as lowercase, and I added the policy in upper case later, so sqruff made my lines follow your local standard.
I could make that file all uppercase easily, and I think that would be preferable; but not a decision I wanted to take without discussion.

maparent: (I will wait for a reply to merge this.)

mdroidian: UPPERCASE sounds good to me 👍

PR #308, opened 2025-07-28
[ENG-651] Fix horizontal scroll bar and excessive spacing in setting tab list

trangdoan982:
Before: 
image

After:
- removed excessive spacing between sections
- remove the horizontal scroll bar
Screenshot 2025-07-28 at 11 13 25

### DIFF
--- a/apps/roam/src/components/settings/Settings.tsx
+++ b/apps/roam/src/components/settings/Settings.tsx
@@ -100,6 +100,7 @@ export const SettingsDialog = ({
         <style>{`
           .dg-settings-tabs .bp3-tab-list {
             overflow-y: auto;
+            overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


apps/roam/src/components/settings/Settings.tsx
```diff
@@ -100,6 +100,7 @@ export const SettingsDialog = ({
         <style>{`
           .dg-settings-tabs .bp3-tab-list {
             overflow-y: auto;
+            overflow-x: hidden;
```
mdroidian, line 103: This works, but seeing as none of the text is long enough to cause the issue, this might be more of a bandaid.  And might bite us later.  I won't block on this, but I wonder what the root cause actually is?

mdroidian (approved):

PR #315, opened 2025-08-03
Roam: Eng-450 remove unused db functions

sid597:
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed several outdated database functions related to content synchronization, embedding matching, and discourse node management.
  * Cleaned up associated configuration and type definitions to reflect these removals.

* **Chores**
  * Updated database schema and migration scripts to ensure deprecated functions are fully dropped. 

These changes streamline the database and remove unused or obsolete features, improving maintainability. No user-facing features or workflows are affected.


sid597:
@coderabbitai full review

packages/database/supabase/schemas/sync.sql
```diff
@@ -258,10 +219,7 @@ GRANT ALL ON FUNCTION public.propose_sync_task(
     "task_interval" interval
 ) TO "service_role";
 
-GRANT ALL ON FUNCTION public.get_nodes_needing_sync(nodes_from_roam jsonb) TO "anon";
-GRANT ALL ON FUNCTION public.get_nodes_needing_sync(nodes_from_roam jsonb) TO "authenticated";
-GRANT ALL ON FUNCTION public.get_nodes_needing_sync(nodes_from_roam jsonb) TO "service_role";
-
+DROP FUNCTION IF EXISTS public.get_nodes_needing_sync(jsonb);
```
maparent: I sometimes add drops before creates in the schema, to make the schema closer to idempotent; 
but in general the drop belongs in migrations, not in schemas.
So I think this line should go.

packages/database/supabase/migrations/20250803100134_drop-get_nodes_needing_sync.sql
```diff
@@ -0,0 +1,3 @@
+drop function if exists "public"."get_nodes_needing_sync"(nodes_from_roam jsonb);
```
maparent: nbd but why two migration files?

packages/database/supabase/schemas/embedding.sql
```diff
@@ -30,40 +30,6 @@ GRANT ALL ON TABLE public."ContentEmbedding_openai_text_embedding_3_small_1536"
 
 set search_path to public, extensions ;
 
-CREATE OR REPLACE FUNCTION public.match_content_embeddings (
```
maparent: I don't know if this function should be removed.
On the one hand you are correct that it is not currently used. OTOH It allows functionality that is not there otherwise, and _could_ be used.
That said... Search architecture will be rethought soon, so it's not like we're going to get to use it in this exact shape.
I guess I'm a -0 on this. Not going to fight it.

packages/database/supabase/migrations/20250803092544_drop-unused-functions.sql
```diff
@@ -0,0 +1,11 @@
+drop function if exists "public"."alpha_delete_by_source_local_ids"(p_space_name text, p_source_local_ids text[]);
```
maparent, line 1: We're standardizing towards all-caps for sql. I wish I could tell sqruff this.
Trick: if your sqruff is up to date, (>= 0.29.2) you could change the first word and then do a sqruff fix.
Btw if you don't have sqruff handy, I'm happy to add that commit to your branch.

packages/database/supabase/migrations/20250803100134_drop-get_nodes_needing_sync.sql
maparent: (Also upper case)

maparent (changes requested):
The drop in schema is the only one I'd block on, but the upper-case should not be hard.
Otherwise I'm good with this.

packages/database/supabase/migrations/20250803092544_drop-unused-functions.sql
maparent: On second thought... this convention is needed for schemas, not migrations.

sid597: Yeah I noticed this convention in schemas, I don't touch the migration files apart from checking if the migration contains only the changes that I made

packages/database/supabase/migrations/20250803100134_drop-get_nodes_needing_sync.sql
sid597: I forgot migrating this file in the first go, and had a seperate migration for it which created this file, merged to a single one now.

packages/database/supabase/schemas/embedding.sql
sid597: I mis-remembered and thought I created both these functions and now the other one is being used, I don't have any opinions on which one to keep. Since you see a use for this in future I will keep this one and remove the other one

maparent: I would not remove the other one! It fits a real and distinct use case.
I originally had a single function which handled both use cases, but @mdroidian thought it was confusing.

sid597: Ok no embedding function is being removed now.

maparent (approved):
LGTM, thank you!

maparent:
The failure is due to merging the migrations. Trying to solve this.

maparent:
For some reason I get errors with supabase when trying to do a `supabase link`. Something is really off with supabase. So without the link, I deleted the migrations by hand in the supabase sql interface. (We need a better flow for this, but it would involve supabase link.)
AND that should have solved the problem, but I also cannot re-run the failed step to check that. (Or rather I can ask, but the step is not re-run.)
What I suggest: Re-do a push, (maybe whitespace) and we'll see if that fixes the validation issue since I've deleted the migrations.

maparent:
(Or we could ignore the failing check, but that's probably bad practice.)

mdroidian (changes requested):
Supabase build failing.  As @maparent suggested, try re-pushing to trigger it again.

sid597:
@maparent @mdroidian build is passing now

mdroidian (approved):

PR #316, opened 2025-08-03
Roam: Eng-512 update node upsertdelete to include concepts

sid597:
## Summary by CodeRabbit

* **Refactor**
  * Improved orphaned node cleanup by switching from external API calls to direct database queries, resulting in more efficient and reliable operations.

* **Chores**
  * Removed obsolete API endpoints related to discourse node retrieval and deletion.


sid597:
@coderabbitai full review

apps/roam/src/utils/cleanupOrphanedNodes.ts
```diff
@@ -1,39 +1,29 @@
...
-    }
-
-    const baseUrl =
-      getNodeEnv() === "development"
-        ? "http://localhost:3000/api/supabase"
-        : "https://discoursegraphs.com/api/supabase";
+    const allNodeInstanceIds = await supabaseClient
+      .from("Concept")
+      .select("source_local_id")
+      .not("schema_id", "is", null);
```
maparent, line 12: Puzzled; this constraint means that you're looking at node instances, and not schemas. At line 16 you mention node schemas, I suspect that's just a typo. But can't schemas also be orphaned? Why restrict to instances?

apps/roam/src/utils/cleanupOrphanedNodes.ts
```diff
@@ -60,57 +50,71 @@ const getNonExistentRoamUids = (nodeUids: string[]): string[] => {
...
+  uids: string[],
+  spaceId: number,
+  supabaseClient: SupabaseClient,
+): Promise<void> => {
   try {
-    const context = await getSupabaseContext();
-    if (!context) {
-      console.error("Failed to get Supabase context");
-      return 0;
+    const { data: contentData, error: contentError } = await supabaseClient
```
maparent, line 59: Hmmm... disappointed that supabase does not allow delete on join here.

apps/roam/src/utils/cleanupOrphanedNodes.ts
```diff
@@ -60,57 +50,71 @@ const getNonExistentRoamUids = (nodeUids: string[]): string[] => {
...
-        spaceId: context.spaceId,
-        uids,
-      }),
-    });
+    if (contentIds.length > 0) {
+      const { error: conceptError } = await supabaseClient
+        .from("Concept")
+        .delete()
+        .in("represented_by_id", contentIds)
+        .eq("is_schema", false);
```
maparent, line 76: Again why restrict to instances?

maparent (changes requested):
ok. Curious about why only removing instances and not schemas. Otherwise lgtm.

maparent:
One of my comments fell outside the review because I'm still learning how GitHub does reviews...

apps/roam/src/utils/cleanupOrphanedNodes.ts
sid597: Yeah for schemas I would have a separate pr, this task was pending for weeks and was originally only for node instances. I will have the schema  pr this week which would also include upserting node schema and instances

apps/roam/src/utils/cleanupOrphanedNodes.ts
sid597: replied above

apps/roam/src/utils/cleanupOrphanedNodes.ts
maparent: And in that other PR you would essentially... remove the condition? That seems like busywork.

maparent: Actually, you have a point, we may want to think a bit harder about the consequences of removing a schema. What happens to orphaned instances? 
We'll get a bunch of schema-less concepts. (cascade set null on schema reference.) Should they be removed? I think yes. That would be a change on the delete cascade. And that's worth another PR.
So ok, let it stand. lgtm.

maparent (approved):

sid597:
cc @mdroidian I merged this pr without checking that you haven't reviewed it, I thought only maparent is going to review. I can revert the merge

sid597:
cc @mdroidian I merged this pr without checking that you have'nt reviewed it, I thought only maparent is going to review. I can revert the merge

mdroidian:
> cc @mdroidian I merged this pr without checking that you have'nt reviewed it, I thought only maparent is going to review. I can revert the merge 

All good

PR #318, opened 2025-08-04
ENG-675 correction to roam client and test

maparent:
Two 
1. packages/ui/src/lib/supabase/client should not use dbDotEnv, but injected variables from the build process
   This made the compile step fail when the client was included.
3. There was a discrepancy between getSupabaseContext and the cucumber step mimicking it.

## Summary by CodeRabbit

* **Chores**
  * Updated environment variable handling for connecting to Supabase.
  * Made internal adjustments to how spaces are created or fetched when opening a plugin.


maparent:
@coderabbit review

mdroidian (approved):

PR #319, opened 2025-08-05
ENG 696 Diagrams to show possible implementations of reference

maparent:
This is documenting how the references to discourse node instances would look in the database.

Some more context:

1. I think that content links are a pure content concern, and should cover (say) arbitrary hyperlinks
2. In theory, identifying concepts occurring inside content may not go through an explicit hyperlink, and may be a matter of interpretation. As such it makes sense for it to go from content directly to concept.
3. In the (common) case of a hyperlink to content that represents a concept, we may either materialize both or infer the occurence. I am still thinking about the complexity and performance tradeoffs here.
4. I believe it is worth it to materialize the hyperlink anchor in the database, but again there are complexity tradeoffs.

This documentation is not meant to be a definitive answer to those questions but to have a support to discuss options.


maparent:
did a few changes to harmonize conventions with ENG-704

mdroidian:
Deferred for now.

PR #321, opened 2025-08-07
ENG-706 Diagram to document the database model

maparent:
Add a diagram to the documentation showing how roam objects relate to database objects.


mdroidian (approved):
Good stuff!  Dense, but definitely helpful. 🙌

As discussed, this will be even more beneficial with the addition of the `json` examples of Discourse Node Types and Discourse Nodes in the admin/dev panel, as well as a higher level simple diagram/flow chart (something like a tldraw sketch, see below) of the 1:1 roam terms <-> db tables/columns.

example
image

*This was created by https://teach.tldraw.com/.  Just use to reference the idea, content may be erroneous 😅

PR #322, opened 2025-08-07
Dependency updates so we're close to latest create-turbo.

maparent:
Checked turbo build and deploy.


mdroidian (commented):
As discussed I'd like to leave the repo as is for a bit while we have some rather large additions (tldraw upgrade for roam, tldraw addition for obsidian) and see how those play out before upgrading/changing react versions.  Our current build state is still rather 'touch and go', unfortunately.

Some of that is documented here: [ENG-725: Continued React/React Type Issues (specifically in apps/roam)](https://linear.app/discourse-graphs/issue/ENG-725/continued-reactreact-type-issues-specifically-in-appsroam)

PR #326, opened 2025-08-08
[ENG-624] TLDraw Obsidian asset store

trangdoan982:
https://www.loom.com/share/133f76d44b0b464192512721dd51f771?sid=bb166032-0937-4aa8-9c0c-d26f474bae5b

## Summary by CodeRabbit

* **New Features**
  * Introduced integrated asset management for drawings, allowing users to upload, cache, and resolve images and videos directly within the app.
  * Assets are now stored and tracked in a dedicated attachments folder, linked to your markdown files.

* **Bug Fixes**
  * Improved reliability when switching or closing files by ensuring proper cleanup of asset resources.

### DIFF
--- a/apps/obsidian/src/components/TldrawView.tsx
+++ b/apps/obsidian/src/components/TldrawView.tsx
@@ -125,15 +125,19 @@ export class TldrawView extends TextFileView {
 
   private createReactRoot(entryPoint: Element, store: TLStore) {
     const root = createRoot(entryPoint);
-    this.assertInitialized();
+    if (!this.file) throw new Error("TldrawView not initialized: missing file");
+    if (!this.assetStore)
+      throw new Error("TldrawView not initialized: missing assetStore");
+    if (!this.store)
+      throw new Error("TldrawView not initialized: missing store");
 
     root.render(
       <React.StrictMode>
         <TldrawPreviewComponent
           store={store}
           plugin={this.plugin}
-          file={this.file!}
-          assetStore={this.assetStore!}
+          file={this.file}
+          assetStore={this.assetStore}
         />
       </React.StrictMode>,
     );
--- a/apps/obsidian/src/components/TldrawView.tsx
+++ b/apps/obsidian/src/components/TldrawView.tsx
@@ -75,12 +75,15 @@ export class TldrawView extends TextFileView {
       return;
     }
 
-    this.assetStore = new ObsidianTLAssetStore(`tldraw-${file.path}`, {
-      app: this.app,
-      file,
-    });
+    this.assetStore = new ObsidianTLAssetStore(
+      `tldraw-${encodeURIComponent(file.path)}`,
+      {
+        app: this.app,
+        file,
+      },
+    );
 
-    this.setStore(store);
+    await this.setStore(store);
   }
 
   private createStore(fileData: string): TLStore | undefined {
@@ -114,11 +117,9 @@ export class TldrawView extends TextFileView {
     if (!this.file) return;
 
     if (!this.assetStore) {
-      console.error("Asset store is not set");
+      console.warn("Asset store is not set");
       return;
     }
-
-    console.log("assetStore", this.assetStore);
 
     root.render(
       <React.StrictMode>
@@ -133,7 +134,7 @@ export class TldrawView extends TextFileView {
     return root;
   }
 
-  protected setStore(store: TLStore) {
+  protected async setStore(store: TLStore) {
     if (this.store) {
       try {
         this.store.dispose();
@@ -144,7 +145,7 @@ export class TldrawView extends TextFileView {
 
     this.store = store;
     if (this.tldrawContainer) {
-      void this.refreshView();
+      await this.refreshView();
     }
   }
 
--- a/apps/obsidian/src/components/TldrawView.tsx
+++ b/apps/obsidian/src/components/TldrawView.tsx
@@ -75,15 +75,10 @@ export class TldrawView extends TextFileView {
       return;
     }
 
-    // Create asset store
     this.assetStore = new ObsidianTLAssetStore(`tldraw-${file.path}`, {
       app: this.app,
       file,
     });
-
-    // Set the asset store on the TLStore
-    // store.props.assets = {};
-    // store.props.assets = this.assetStore;
 
     this.setStore(store);
   }
@@ -184,7 +179,6 @@ export class TldrawView extends TextFileView {
     this.onUnloadCallbacks = [];
     callbacks.forEach((cb) => cb());
 
-    // Clean up asset store
     if (this.assetStore) {
       this.assetStore.dispose();
       this.assetStore = undefined;
@@ -217,7 +211,6 @@ export class TldrawView extends TextFileView {
       this.store = undefined;
     }
 
-    // Clean up asset store
     if (this.assetStore) {
       this.assetStore.dispose();
       this.assetStore = undefined;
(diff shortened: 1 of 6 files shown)


trangdoan982:
@coderabbitai full review

apps/obsidian/src/components/TldrawView.tsx
```diff
@@ -6,11 +6,13 @@ import { TLStore } from "tldraw";
 import React from "react";
 import DiscourseGraphPlugin from "~/index";
 import { processInitialData, TLData } from "~/utils/tldraw";
+import { ObsidianTLAssetStore } from "~/utils/asset-store";
 
 export class TldrawView extends TextFileView {
   plugin: DiscourseGraphPlugin;
   private reactRoot?: Root;
   private store?: TLStore;
+  private assetStore?: ObsidianTLAssetStore;
```
mdroidian: Is this optional?  If not, remove the `?`.  (same with `reactRoot` and `store`)

apps/obsidian/src/utils/asset-store.ts
```diff
@@ -0,0 +1,217 @@
+import { App, TFile } from "obsidian";
```
mdroidian: filename camelCase

apps/obsidian/src/utils/asset-store.ts
```diff
@@ -0,0 +1,217 @@
...
+import { TLAsset, TLAssetStore, TLAssetId, TLAssetContext } from "tldraw";
+import { JsonObject } from "@tldraw/utils";
+
+const ASSET_PREFIX = "obsidian.blockref.";
+type BlockRefAssetId = `${typeof ASSET_PREFIX}${string}`;
+type AssetDataUrl = string;
+
+interface AssetStoreOptions {
```
mdroidian: type

apps/obsidian/src/utils/asset-store.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+    this.#resolvedAssetDataCache.set(blockRefAssetId, url);
+  }
+
+  constructor(options: AssetStoreOptions) {
+    this.#app = options.app;
+    this.#file = options.file;
+  }
+
+  async storeAsset(asset: TLAsset, file: File): Promise<BlockRefAssetId> {
```
mdroidian: `_asset`
`_` denotes that we will not be using it.

apps/obsidian/src/utils/asset-store.ts
```diff
@@ -0,0 +1,217 @@
...
+      // Non-media (e.g., text/markdown, application/*): let custom shapes decide.
+      // Return null so default media shapes won't attempt to render it.
+      return null;
+    } catch (error) {
+      console.error("Error resolving asset:", error);
+      return null;
+    }
+  }
+
+  async remove(assetIds: TLAssetId[]): Promise<void> {
```
mdroidian: _

apps/obsidian/src/components/TldrawViewComponent.tsx
```diff
@@ -118,10 +120,14 @@ export const TldrawPreviewComponent = ({
...
   }, [currentStore, saveChanges]);
 
-  const handleMount = useCallback((editor: Editor) => {
-    editorRef.current = editor;
-    editor.setCurrentTool("select");
-  }, []);
+  const handleMount = useCallback(
+    (editor: Editor) => {
+      editorRef.current = editor;
+      editor.setCurrentTool("select");
```
mdroidian: you can use `initialState` on the `<Tldraw>` component instead of this.

apps/obsidian/src/utils/asset-store.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  async resolve(asset: TLAsset, ctx: TLAssetContext): Promise<string | null> {
+    try {
+      const assetSrc = asset.props.src;
+      if (!assetSrc?.startsWith("asset:")) return assetSrc ?? null;
+
+      const assetId = assetSrc.split(":")[1] as BlockRefAssetId;
+      if (!assetId) return null;
+
+      const mimeType =
```
mdroidian: tl;dr
check `asset.type`

---

`TLAsset`'s type is
`export declare type TLAsset = TLBookmarkAsset | TLImageAsset | TLVideoAsset;`

If you look into those, they extend `TLBaseAsset`

```typescript
export declare interface TLBaseAsset<Type extends string, Props> extends BaseRecord<'asset', TLAssetId> {
    type: Type;
    props: Props;
    meta: JsonObject;
}
```

Which lists `type`.  I think this is `image` or `video` etc.  Let's see if we can check that instead of going the long way around with this mimeType stuff.

apps/obsidian/src/components/TldrawViewComponent.tsx
```diff
@@ -118,10 +120,14 @@ export const TldrawPreviewComponent = ({
...
+      currentStore.props.assets = assetStore;
```
mdroidian: We are storing `currentStore` in state, so we shouldn't mutate it.
Regardless, I believe `assets` should be part of `createTLStore` https://tldraw.dev/reference/editor/createTLStore

mdroidian (changes requested):
Good stuff, a few changes requested.

Couple additional things to list here

- I noticed the image gets added to the markdown with `!` which "embeds" it.  After a few images, this will get quite noisy.  Let's remove the `!`.  A user can preview with a modifier+hover if they want.

- I'd love to test this myself, but unfortunately I can't load it.  I think this is related to the `css` issue we discussed.

image

apps/obsidian/src/utils/asset-store.ts
trangdoan982: I'm thinking of doing mimeType because when we have DiscourseNodeShape, the asset type will be an md file, which is not an inherent asset type for Tldraw -> we need to resolve this asset type differently

mdroidian: Hmm, even in that case it's not clear why you need to use mimeType instead of just type as a filter in order to resolve said type.

trangdoan982: I store as mimeType because `file.Type` is actually mimeType. I think we would if else the type anyway so there's no need to change mimeType -> simpler fileType? did I get your question right?

mdroidian: You are correct.  Using type would just be more direct, less lines of code, easier to grok/modify later, etc, etc.

trangdoan982:
@mdroidian i just pushed a fix for the styling issue, can you check if it works on your local now?

mdroidian:
> @mdroidian i just pushed a fix for the styling issue, can you check if it works on your local now?

No change for me.  I even tried deleting `node_modules` / `package-lock.json` just to see if there was something funky there.

We can debug it together tomorrow.

apps/obsidian/src/components/TldrawView.tsx
```diff
@@ -102,23 +115,32 @@ export class TldrawView extends TextFileView {
...
-    if (!this.file) return;
+    this.assertInitialized();
 
     root.render(
       <React.StrictMode>
         <TldrawPreviewComponent
           store={store}
           plugin={this.plugin}
-          file={this.file}
+          file={this.file!}
```
mdroidian: We should avoid `!` whenever possible.  It basically just bypasses type safety.

If you inline the `assertInitialized()`, the `!` is not necessary.

mdroidian (approved):
🚀

PR #317, opened 2025-08-04
Prep Repo for Tldraw Upgrade

mdroidian:
This will be a draft PR that is broken down into multiple smaller PRs with the goal of preparing the repo for `apps/roam` tldraw upgrade and to help move our repo back to a healthier state.

In this PR, the following changes have been made:
- update `apps/roam` `package-lock.json` https://github.com/DiscourseGraphs/discourse-graph/issues/70
- update `apps/roam` to React 18
  - override `roamjs-components` and `blueprintjs`
- update `apps/website` / `package/ui` to React 19
  - apply minor refactors to match React 19 types
- clean install without `--legacy-peer-deps` ( rm `node_modules`, rm `package-lock`, `npm i` to rebuild `package-lock`)
- move/edit tldraw patches to match new `node_modules`
- modify `postinstall` to apply root packages 

## Caveats
- blueprintjs components still need to be patched to allow `children`, this will show a type error currently
- `npm ls react` still shows many errors, mostly from blueprintjs.  We may fork blueprintjs in the future to solve this.
- moving forward we will need to isolate `packages/ui` to React 19.  `apps/roam` will not use `packages/ui`.
- we will also need to move any utils from `packages/ui` to `packages/util` so they can be used across `apps/*` ([ENG-633](https://linear.app/discourse-graphs/issue/ENG-633/refactor-move-some-functions-to-packagesutils))

### Setup
To test this with your app/workflow, you can either clone to test folder or remove `node_modules`:

- `git clone` the repo to a new temp folder
- checkout `eng-617-reproduce-the-idenpm-issues`

OR

- remove all `node_modules` (from root and each `app/*` and `package/*` if they exist)

THEN 

- remove root `package-lock.json`
- run `npm i`
- check your regular workflow/files for issues

### Test
Let me know if after `npm i` 
- errors are thrown
- `package-lock.json` shows changes from this PR's `package-lock.json`
- any issues in your workflow/files

## Summary by CodeRabbit

* **New Features**
  * Added a PowerShell script to automate cleaning and reinstalling Node.js dependencies.

* **Refactor**
  * Updated several React components to simplify how references (refs) are handled.
  * Improved client-side loading for analytics tracking to enhance performance.
  * Streamlined async page rendering in documentation sections for better maintainability.

* **Chores**
  * Updated dependency versions for React, Next.js, and related type definitions across multiple packages.
  * Adjusted dependency overrides to ensure compatibility with specific React versions.
  * Cleaned up metadata in patch files for better clarity.

PR #331, opened 2025-08-09
ENG-717 update apps/website and package/ui to use React 19

mdroidian:
Base #317


mdroidian:
combined [ENG-718: clean install without `--legacy-peer-deps`](https://linear.app/discourse-graphs/issue/ENG-718/clean-install-without-legacy-peer-deps) and apps/roam patches to pass all build/checks.

PR #340, opened 2025-08-14
Eng 732 left side panel for suggestive mode

sid597:
## Summary by CodeRabbit

* **New Features**
  * Suggestive Mode overlay button on page references to toggle a contextual Suggestions panel.
  * Collapsible Suggestions panel with header controls, hover highlights, filters, and actions to open or add related pages/blocks.
  * Split-view panel manager with minimize/restore, close-all, and panel persistence across reloads.
  * New settings UI: personal toggle for Suggestive Mode, a Suggestive Mode settings page, per-node "Suggestive Rules" and Page Groups management.
  * Visual highlight styling for linked blocks (.dg-highlight).


sid597:
@coderabbitai full review

sid597:
@coderabbitai full review

sid597:
@mdroidian @maparent 

I think the review would be easier to do if done in 3 passes each relating to one of the flows:
- Setting and extracting from personal setting, render overlay
  - `initializeObserversAndListeners.ts` 
  -  `pageRefObserverHandler.ts`
  -  `SuggestiveModeOverlay.tsx`

- Sidebar `PanelManager.tsx`
  (I refer the individual suggestiveMode component that we open in the sidebar as panel)
  - Goal here is to manage  panels corresponding to different nodes. 
  - We should be able to:
     - If no panel is open, and user clicks on the button for some node, create the sidebar and add the panel
     -  Close all the nodes with the topbar Close icon
     - Minimise the sidebar, open it to see all the previous opened panels
     - Keep track of original width of roam main body, when panel is opened make the main body full and when closed, revert to original width.
     - If user hovers over a panel, highlight the corresponding overlay icon in the block and vice-a-versa.
     - Any new panel gets added to top of the sidebar instead of at the bottom
     
- Individual panels
  - Related Files
    - `DiscourseSuggestionsPanel.tsx`
       - Keeps the skeleton of the panel: Topbar with "Close" and "Minimise" icon
    - `SuggestionsBody.tsx`
      - Meaty part of the code, select pages  or groups, ask Hyde to get sorted suggestive list, render it and be able to filter it. Use all pages functionality to search all the pages (Note: we use roam query because supabase does not have the node.type information therefore we cannot get nodes of a specific types directly)
   - `hyde.ts`: lint fix
   - `useDiscourseData.ts`: utility function to get data needed per instnce

mdroidian:
@sid597 looks like there is a lot of UI here.  Could you create a loom video that details and walks through it.

sid597:
https://www.loom.com/share/fb19b2910b2344729da94b674afa9c8d?sid=e3727226-764f-4b2a-9ae2-ab773f1b62c5

@mdroidian I discovered 3 bugs, the last one related to hyde search (but it got introduced after some latest merges I will fix them but I think this video covers the what UI actions relate to what part of the pr.

maparent:
Cannot leave a line note outside of review scope it seems? Anyway hyde.ts:63, please use `nextApiRoot` from packages/utils.

maparent:
Ok... I know I need to dive into the code, but I have tons of ux questions which I realize are partly outside of the scope of this PR... and yet not. I will try to spend more time reviewing the code itself, but I went on a rant re my thoughts on how to do the suggestion panel. Roam [web](https://roamresearch.com/#/app/discourse-graphs/page/g6YZNuDOn) and [app](roam://#/app/discourse-graphs/page/g6YZNuDOn).

apps/roam/src/components/PanelManager.tsx
```diff
@@ -0,0 +1,613 @@
...
+    }
+  } else {
+    panelRoot = document.createElement("div");
+    panelRoot.id = PANEL_ROOT_ID;
+    roamBodyMain.insertBefore(panelRoot, articleWrapper);
+  }
+
+  let panelsContainer = document.getElementById(PANELS_CONTAINER_ID);
+  if (!panelsContainer) {
+    panelsContainer = document.createElement("div");
```
maparent, line 337: Curious: Why a div and not a React element for the panelContainer? Then openPanels would be part of its state, and it may be easier to register to. Or am I missing something?

apps/roam/src/utils/hyde.ts
```diff
@@ -350,100 +352,100 @@ export const getAllPageByUidAsync = async (): Promise<[string, string][]> => {
   return pages;
 };
 
-export const extractPagesFromChildBlock = (
```
maparent, line 354: detail: lines 347-401 look like great roam utilities outside of hyde and could live in their own file.

apps/roam/src/utils/useDiscourseData.ts
```diff
@@ -0,0 +1,146 @@
...
+    const relatedNodeType = discourseNode.type;
+
+    return validRelations.flatMap((relation) => {
+      const isSelfSource = relation.source === relatedNodeType;
+      const isSelfDestination = relation.destination === relatedNodeType;
+
+      let targetNodeType: string;
+      let currentRelationLabel: string;
+
+      if (isSelfSource) {
```
maparent, line 94: It looks like we're skipping the case where a node type could be both source and destination of a given relation.
(Also, is the `"*"` case a possibility for `relation.source` or `.destination` here?)

apps/roam/src/utils/useDiscourseData.ts
```diff
@@ -0,0 +1,146 @@
...
+    );
+    const types = Array.from(
+      new Set(
+        validRelations.flatMap((relation) => [
+          relation.source,
+          relation.destination,
+        ]),
+      ),
+    );
+    return hasSelfRelation ? types : types.filter((type) => type !== selfType);
```
maparent, line 146: I am not sure that we can always filter out the selfType, at least not if there's a symmetric relation.

apps/roam/src/utils/useDiscourseData.ts
```diff
@@ -0,0 +1,146 @@
...
+  } catch (error) {
+    console.error(`Error getting overlay info for ${tag}:`, error);
+    return {
+      results: [],
+      refs: 0,
+    };
+  }
+};
+
+export const useDiscourseData = (tag: string) => {
```
maparent: nit: I would not mind a more descriptive name. "use" is a bit generic: Why? For which purpose?
This function (and file) seem to (1) prepare node filters (2) for the suggestion panel (3) using the discourse data. 
I may be wrong on first part, correct me at need, but this is what (imho) is not obvious to the reader and should be made explicit. 
(I would also like to see the other two in comments, but I know that we have a policy of being sparse on comments.)

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -0,0 +1,426 @@
...
+}) => {
+  const {
+    results: existingResults,
+    discourseNode,
+    uniqueRelationTypeTriplets,
+    validTypes,
+    allNodes,
+  } = useDiscourseData(tag);
+  const allPages = useMemo(() => getAllPageNames(), []);
+  const cachedState = suggestionsCache.get(blockUid);
```
maparent, line 54: Shouldn't this be a component state?

maparent (changes requested):
Ok. First, there is a lot that I propose that is about how to use React (more state!) and I'm not a React expert, feel very free to push back and correct me if I'm missing some advanced React patterns.
Second, I have more UX comments on the panel that are much more pointed, not reinvent-everything, but that may still be out-of-scope for this PR. Let me know. https://roamresearch.com/#/app/discourse-graphs/page/Npc55KjGO
The only really blocking comments are on the body of the useDiscourseData function.

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -76,6 +77,28 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
           </>
         }
       />
+      <Checkbox
+        defaultChecked={
+          extensionAPI.settings.get("suggestive-mode-overlay") as boolean
+        }
+        onChange={(e) => {
+          const target = e.target as HTMLInputElement;
+          extensionAPI.settings.set("suggestive-mode-overlay", target.checked);
```
mdroidian: @sid597 do you see these eslints in your IDE?
image

If so, let's try our best to solve them before submitting the PR.  If you have issues, let me know and we can try to work them out together.

If you don't see these lints, definitely let me know so we can fix that.

mdroidian (changes requested):
> ... I discovered 3 bugs

@sid597 tag me again once the bugs are fixed. 👍

sid597:
@mdroidian The bugs were:

- When the left side panel is open we are not able to select the width of the main view, we have an observer to look if left side panel is open and change the main view width accordingly. Now I am not sure if this is a bug, wdyt.

- Second bug was related to the 429 api error: This is solved, Matt created a new account and an api key in 1password, I copied over the password to vercel and in `env.local` so now this is working as well

- I mentioned 3 bugs but can't find the third one even after watching the recording a few times.

apps/roam/src/utils/hyde.ts
sid597: yeah keeping it here for now, if @mdroidian knows of some similar existing function then we will replace that and maybe move the others to separate file.

mdroidian:
1) Left side panel should not override roam native main width setting.  I think this would be confusing to the user.  (unless we have some sort of "full screen" button or feature which conveys the fact that it is overriding the native setting).

apps/roam/src/utils/useDiscourseData.ts
sid597: > (Also, is the "*" case a possibility for relation.source or .destination here?)

Not sure most probably not

apps/roam/src/utils/useDiscourseData.ts
sid597: Are we aiming to solve this or have this now?

maparent: Are you asking if we should handle the symmetric relationship case now? I would say yes. Even if we don't have this case now, it's a likely case, and we should handle it well; otherwise we'll wonder why it does not work down the line.

sid597:
> 1. Left side panel should not override roam native main width setting.  I think this would be confusing to the user.  (unless we have some sort of "full screen" button or feature which conveys the fact that it is overriding the native setting).

It overrides it for the time left side panel is open once its closed we revert back to what user had previously set, I think its an improvement in ux i.e the width setting is not usable in left sidebar mode, we can use better heuristics to set the main width but I do think we need to do it automatically without user intervention

mdroidian:
Ok, I'll trust your judgement here.

apps/roam/src/utils/useDiscourseData.ts
sid597: Do you mean if there a symmetric node we should not include it? Currently its included e.g in the following example 

image


if we run this for Hypothesis node, we have 4 relations

Hypothesis Opposes Result
Hypothesis Opposes Hypothesis (symmetric case)
Claim           Supports Hypothesis
Question     Informs   Hypothesis


```
validRelations

[
    {
        "id": "xji263xjq",
        "label": "supports",
        "source": "_CLM-node",
        "destination": "PN8NyQv3q",
        "complement": "supported by",
        "triples": ...
        ]
    },
    {
        "id": "EXWazGXlq",
        "label": "Informs",
        "source": "_QUE-node",
        "destination": "PN8NyQv3q",
        "complement": "",
        "triples": []
    },
    {
        "id": "gnS--omLA",
        "label": "Opposes",
        "source": "PN8NyQv3q",
        "destination": "OV5a_YFlJ",
        "complement": "Opposed By",
        "triples": ....
    },
    {
        "id": "N_wNv9EMx",
        "label": "Opposes",
        "source": "PN8NyQv3q",
        "destination": "PN8NyQv3q",
        "complement": "",
        "triples": []
    }
]

validTypes

[
    "_CLM-node",
    "PN8NyQv3q", //Hypothesis node
    "_QUE-node",
    "OV5a_YFlJ"     // Result node
]

```

apps/roam/src/components/SuggestionsBody.tsx
sid597: If the user accidentally closes the panel and reopen it they would have to do the process all over again, by keeping the map outside we are saving it (at the cost of performance) so its there when the user opens it again in the same session. 

To tackle the performance I think adding a LRU cache kinda thing would be good and we will have best of both worlds. 
Good catch thanks!

apps/roam/src/utils/useDiscourseData.ts
maparent: No, we should definitely include  symmetric relations. The issue is with the filter in this line, you exclude the type of the `self` node, so the nodes of that type won't show up in the suggestions.
That type should only be excluded _if_ there are no symmetric relationships, so you have to check for them first.

apps/roam/src/components/SuggestionsBody.tsx
maparent: Ah! So it's a time/memory tradeoff. Hmmm... 
1. I think the page cache as a whole should be held in a WeakRef, so it gets garbage-collected at need.
2. Even if we hold the cache in a WeakRef outside the React panel, I would still tend to _also_ copy it into the component state so that changes to the values get reflected through the normal react mechanism while the panel is active.

maparent: Second thoughts: It could be a LRU of weakrefs, I guess, but not sure if that's worth the extra complexity.
Also.... store the date of computation, we don't want to use a very stale query result. Ideal would be to know that new changes invalidate the query, but that is complex and out of scope. 
We _might_ consider a global cache service that takes changes into account, at some point, but that would be a project of its own.

apps/roam/src/utils/useDiscourseData.ts
maparent: The latest commit looks better, thank you!

apps/roam/src/components/SuggestionsBody.tsx
maparent: Note: I said that tracking changes were out of scope in my last comment; but the idea to use WeakRef I see as doable in scope. That said, I don't want to block this PR on this. If you decide to punt, maybe just add a TODO so we remember this?

maparent (approved):
As mentioned, I would like to see allNodes in the state of SuggestionsBody, possibly based on a WeakRef cache, but I won't block on this. My main issue was addressed.

mdroidian:
@CodeRabbit full review

apps/roam/src/components/settings/PageGroupPanel.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+const PageGroupsPanel = ({
+  uid,
+  initialGroups,
+}: {
+  uid: string;
+  initialGroups: PageGroup[];
+}) => {
+  const [pageGroups, setPageGroups] = useState<PageGroup[]>(initialGroups);
+
```
mdroidian, line 19: .

apps/roam/src/components/settings/PageGroupPanel.tsx
```diff
@@ -0,0 +1,206 @@
...
+      <div className="flex flex-col gap-2">
+        {/* Add Group */}
+        <div className="flex items-center gap-2">
+          <AutocompleteInput
+            value={newGroupName}
+            setValue={setNewGroupName}
+            placeholder="Page group name…"
+            options={[]}
+          />
+          <Button
```
mdroidian, line 125: fix alignment (center vertically)

image

apps/roam/src/components/settings/PageGroupPanel.tsx
```diff
@@ -0,0 +1,206 @@
...
+                <AutocompleteInput
+                  key={getAutocompleteKey(group.uid)}
+                  value={getPageInput(group.uid)}
+                  placeholder="Add page…"
+                  setValue={(v) => setPageInput(group.uid, v)}
+                  options={getAllPageNames()}
+                  maxItemsDisplayed={50}
+                />
+              </div>
+              <Button
```
mdroidian, line 172: fix alignment (center vertically)

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -0,0 +1,444 @@
...
+                    ))}
+                </Menu>
+              }
+            >
+              <Tooltip content="Add pages from a group">
+                <Button icon="folder-open" small text="Add Group" />
+              </Tooltip>
+            </Popover>
+          )}
+          <Button
```
mdroidian, line 236: Most of our buttons use the same or similar padding, border-radius, etc.  This button seems to have them backwards

image

vs
image
image

image

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -57,9 +59,37 @@ export const endSyncTask = async (
...
+        id: "discourse-embedding-error",
+        content: "Failed to complete discourse node embeddings sync",
+        intent: "danger",
+        timeout: 5000,
+      });
+    } else {
+      if (status === "complete") {
+        renderToast({
+          id: "discourse-embedding-complete",
+          content: "Successfully completed discourse node embeddings sync",
```
mdroidian, line 73: This renders even when I don't have docker running and have console errors, so the status === "complete" is not the right condition to check for success.

eg
`instrumentNetwork.js:1  POST http://localhost:3000/api/embeddings/openai/small net::ERR_CONNECTION_REFUSED`
`upsertNodesToSupabaseAsContentWithEmbeddings: Embedding service failed – Failed to fetch`

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -0,0 +1,444 @@
...
+          </div>
+          <Tooltip
+            content={
+              selectedPages.includes(currentPageInput)
+                ? "Page already added"
+                : "Add page"
+            }
+            disabled={!currentPageInput}
+          >
+            <Button
```
mdroidian, line 197: Weird spacing here.

image

tuck this closer to the input, and align the button to the right.

apps/roam/src/components/settings/SuggestiveModeSettings.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <FlagPanel
+          title="Include Current Page Relations"
+          description="Include relations from pages referenced on the current page"
+          order={0}
+          uid={settings.suggestiveMode.includePageRelations.uid}
+          parentUid={effectiveSuggestiveModeUid}
+          value={settings.suggestiveMode.includePageRelations.value}
+        />
+
+        {/* TODO: if "Include Current Page Relations" is checked "Include Parent and Child Blocks"
```
mdroidian, line 65: If we aren't finishing this TODO in this PR, then add a link to it's linear, just so we know it's accounted for.

apps/roam/src/components/DiscourseSuggestionsPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+  onClose: () => void;
+  shouldGrabFromReferencedPages: boolean;
+  shouldGrabParentChildContext: boolean;
+}) => {
+  const [isOpen, setIsOpen] = useState(true);
+
+  const toggleHighlight = useCallback(
+    (on: boolean) => {
+      document
+        .querySelectorAll(`[data-dg-block-uid="${blockUid}"]`)
```
mdroidian, line 30: be more specific.  `data-dg-block-uid` could refer to any block-uid in Discourse Graph's extension for any feature/module.

apps/roam/src/components/DiscourseSuggestionsPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+        </Navbar.Group>
+      </Navbar>
+      <Collapse
+        isOpen={isOpen}
+        keepChildrenMounted={true}
+        transitionDuration={150}
+      >
+        <div
+          className={Classes.CARD}
+          style={{ flexGrow: 1, overflowY: "auto", padding: "6px" }}
```
mdroidian, line 105: tailwind

apps/roam/src/components/DiscourseSuggestionsPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+          paddingRight: 0,
+          display: "flex",
+          flexWrap: "nowrap",
+          alignItems: "center",
+        }}
+      >
+        <Navbar.Group align={Alignment.LEFT} style={{ flex: 1, minWidth: 0 }}>
+          <Navbar.Heading
+            className="truncate"
+            style={{
```
mdroidian, line 62: tailwind

apps/roam/src/components/DiscourseSuggestionsPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+        .querySelectorAll(`[data-dg-block-uid="${blockUid}"]`)
+        .forEach((el) => el.classList.toggle("dg-highlight", on));
+    },
+    [blockUid],
+  );
+
+  return (
+    <Card
+      data-dg-block-uid={blockUid}
+      style={{
```
mdroidian, line 39: tailwind

apps/roam/src/components/DiscourseSuggestionsPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+        display: "flex",
+        flexDirection: "column",
+        padding: "8px",
+      }}
+      onMouseEnter={() => toggleHighlight(true)}
+      onMouseLeave={() => toggleHighlight(false)}
+      className="discourse-suggestions-panel"
+    >
+      <Navbar
+        style={{
```
mdroidian, line 50: tailwind

apps/roam/src/components/DiscourseSuggestionsPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+              cursor: "pointer",
+            }}
+            onClick={() => setIsOpen((prev) => !prev)}
+          >
+            {tag}
+          </Navbar.Heading>
+        </Navbar.Group>
+        <Navbar.Group
+          align={Alignment.RIGHT}
+          style={{
```
mdroidian, line 75: tailwind

apps/roam/src/components/PanelManager.tsx
```diff
@@ -0,0 +1,593 @@
...
+);
+
+const openPanels = new Map<string, PanelState>();
+let isContainerMinimized = false;
+let navigationObserver: MutationObserver | null = null;
+
+const cssEscape = (value: string) =>
+  window.CSS && window.CSS.escape
+    ? window.CSS.escape(value)
+    : value.replace(/[^a-zA-Z0-9_\-]/g, (c: string) => `\\${c}`);
```
mdroidian, line 116: image

apps/roam/src/components/PanelManager.tsx
```diff
@@ -0,0 +1,593 @@
...
+const SidebarHeader = ({
+  onMinimize,
+  onCloseAll,
+}: {
+  onMinimize: () => void;
+  onCloseAll: () => void;
+}) => (
+  <Navbar className="rounded-t shadow-none">
+    <Navbar.Group align={Alignment.LEFT}>
+      <Navbar.Heading className="text-[13px] font-semibold">
```
mdroidian, line 86: let's make this background transparent, the white bg comes from blueprint and feels off

mdroidian, line 86: This is not responsive.  But making it responsive can (read should) be deferred to another PR because of the size of this one. 😁

mdroidian: image

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -0,0 +1,444 @@
...
+  Spinner,
+  Tag,
+  Tooltip,
+} from "@blueprintjs/core";
+import AutocompleteInput from "roamjs-components/components/AutocompleteInput";
+import getAllPageNames from "roamjs-components/queries/getAllPageNames";
+import openBlockInSidebar from "roamjs-components/writes/openBlockInSidebar";
+import { Result } from "roamjs-components/types/query-builder";
+import { useExtensionAPI } from "roamjs-components/components/ExtensionApiContext";
+import { performHydeSearch } from "../utils/hyde";
```
mdroidian, line 20: ~

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -0,0 +1,444 @@
...
+const SuggestionsBody = ({
+  tag,
+  blockUid,
+  shouldGrabFromReferencedPages,
+  shouldGrabParentChildContext,
+}: {
+  tag: string;
+  blockUid: string;
+  shouldGrabFromReferencedPages: boolean;
+  shouldGrabParentChildContext: boolean;
```
mdroidian, line 44: If we are just using these when performing the Hyde search, no need to pass them down.  Just grap then when the search is performed and we can reduce the complexity of all this prop drilling.  Getting the setting is a minor query.

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -0,0 +1,444 @@
...
+  };
+
+  const toggleOverlayHighlight = (nodeUid: string, on: boolean) => {
+    document
+      .querySelectorAll(`[data-dg-block-uid="${nodeUid}"]`)
+      .forEach((el) => el.classList.toggle("dg-highlight", on));
+  };
+
+  useEffect(() => {
+    const storedGroups = extensionAPI?.settings.get(
```
mdroidian, line 121: fyi, we have get/set util functions for this now

they'll probably be updated soon with the settings change, so no need to update in this code.  just an fyi

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -0,0 +1,444 @@
...
+    selectedPages,
+    useAllPagesForSuggestions,
+    hydeFilteredNodes,
+    activeNodeTypeFilters,
+  ]);
+
+  return (
+    <div onMouseDown={(e) => e.stopPropagation()}>
+      <div className="mt-2">
+        <label
```
mdroidian, line 163: blueprint `<Label>`

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -0,0 +1,444 @@
...
+                <Button icon="folder-open" small text="Add Group" />
+              </Tooltip>
+            </Popover>
+          )}
+          <Button
+            text="Find"
+            icon="search-template"
+            intent={Intent.PRIMARY}
+            onClick={() => {
+              const performSearch = async () => {
```
mdroidian, line 241: looks like this is used below, so DRY, and extract it to one function.

apps/roam/src/components/SuggestiveModeOverlay.tsx
```diff
@@ -0,0 +1,126 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
mdroidian, line 1: .

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -0,0 +1,444 @@
+// THIS IS NEW CODEBASE
```
mdroidian, line 1: Check if this is actually necessary.
If it's only affecting one (or a few) lines, add it directly there.
Otherwise it we could miss actual lints from this.

apps/roam/src/components/SuggestiveModeOverlay.tsx
mdroidian: Check if this is actually necessary.
If it's only affecting one (or a few) lines, add it directly there.
Otherwise it we could miss actual lints from this.

apps/roam/src/styles/discourseGraphStyles.css
```diff
@@ -177,3 +177,8 @@ div.roamjs-discourse-drawer div.bp3-drawer {
 .bp3-popover-wrapper + .referenced-node-autocomplete {
   margin-top: 1rem;
 }
+
+.dg-highlight {
```
mdroidian, line 181: `dg-highlight` is too ambiguous.  let's rename this so it is more clear what it is doing.

apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
mdroidian, line 1: Check if this is actually necessary.
If it's only affecting one (or a few) lines, add it directly there.
Otherwise it we could miss actual lints from this.

apps/roam/src/components/settings/NodeConfig.tsx
mdroidian, line 1: Check if this is actually necessary.
If it's only affecting one (or a few) lines, add it directly there.
Otherwise it we could miss actual lints from this.

apps/roam/src/components/settings/PageGroupPanel.tsx
mdroidian, line 1: Check if this is actually necessary.
If it's only affecting one (or a few) lines, add it directly there.
Otherwise it we could miss actual lints from this.

mdroidian: @sid597 There are many more these.  Please go through and check them all.  I checked a few and most seemed unnecessary. 
- [x] they have all been checked

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -16,14 +16,16 @@ import {
...
 import { fetchEmbeddingsForNodes } from "./upsertNodesAsContentWithEmbeddings";
 import { convertRoamNodeToLocalContent } from "./upsertNodesAsContentWithEmbeddings";
+import { getRoamUrl } from "roamjs-components/dom";
+import { render as renderToast } from "roamjs-components/components/Toast";
 import type { DGSupabaseClient } from "@repo/database/lib/client";
 import type { Json, CompositeTypes } from "@repo/database/dbTypes";
 
 type LocalContentDataInput = Partial<CompositeTypes<"content_local_input">>;
 type AccountLocalInput = CompositeTypes<"account_local_input">;
+const { createClient } = require("@repo/database/lib/client");
```
mdroidian, line 28: @maparent where are we at with the `require` vs `import` here?

apps/roam/src/utils/useDiscourseData.ts
```diff
@@ -0,0 +1,156 @@
...
+  } catch (error) {
+    console.error(`Error getting overlay info for ${tag}:`, error);
+    return {
+      results: [],
+      refs: 0,
+    };
+  }
+};
+
+export const useDiscourseNodeFilters = (tag: string) => {
```
mdroidian, line 53: This name and file are confusing to me.  Is this only used in `SuggestionsBody`?  What is `DiscourseData` supposed to represent?  It is awfully ambiguous.  And it doesn't match `useDiscourseNodeFilters`.  Also, what are `DiscourseNodeFilters`?

mdroidian (changes requested):
It started out as nits but turned into more.  Plus the CodeRabbit suggestions.

At this point I think we might want to break this down into smaller PR's.  Or merge it and and create a list of linear tickets to do immediately after merge to address the fixes.  Because this is quite unwieldly.   I'll let you decide how to move forward.

Also, there are some merge conflicts that need to be addressed,

Image


Finally, I tried to get it to work locally or via `https://discoursegraphs.com/releases/...` but neither worked for me.  Here are the errors from `.../releases/...`

Image

apps/roam/src/utils/syncDgNodesToSupabase.ts
mdroidian: As in, there were a lot of changes between `packages/database` going from `module` back to `commonjs` and back.  Should this be a `require` or `import`?

apps/roam/src/components/PanelManager.tsx
sid597: so current architecture is that we have shared data, different components (overlay) modify the shared data and we render the ui using the shared data. 

making this a react component we would then have to maintain the data twice, once in the local component and other in the shared data. Actually we don't have to maintain 2 sources of data, react component can just subscribe to the shared data, we have this architecture already in codebase for settings. 

I found it easier to do it in this way, can be done in react way

apps/roam/src/index.ts
```diff
@@ -123,6 +123,8 @@ export default runExtension(async (onloadArgs) => {
   document.addEventListener("input", discourseNodeSearchTriggerListener);
   document.addEventListener("selectionchange", nodeCreationPopoverListener);
 
+  await initializeSupabaseSync();
+
```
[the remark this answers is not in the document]
sid597: but this is a critical step and imo it is right to fail so we can catch and report, it is not a sideeffect it is main functionality when enabled

sid597:
New prs created out of these, review points resolved, some that did not get resolved are carried over to new prs

PR #341, opened 2025-08-15
[ENG-599] Discourse node shape

trangdoan982:
I don't have a Loom demo here because this PR mostly define the shape and its utility functions. For a demo of using the shape, see the follow up PR for creating new node shape

## Summary by CodeRabbit

* **New Features**
  * Added a new “Discourse Node” shape to canvases, showing a linked note’s title, type, and color.
  * Shape auto-updates when note metadata changes.

* **Bug Fixes**
  * Prevents the canvas from mounting when required assets aren’t available, avoiding crashes.
  * More reliable resolution of linked notes from block references, with improved error handling.

* **Refactor**
  * Consolidated link and metadata utilities to reduce duplication and improve consistency.


trangdoan982:
@coderabbitai full review

apps/obsidian/src/utils/shapes/DiscourseNodeShape.tsx
```diff
@@ -0,0 +1,172 @@
+import { BaseBoxShapeUtil, HTMLContainer, T, TLBaseShape } from "tldraw";
```
mdroidian: There's lots of different ways to organize things, but one rule of thumb I use is "If it ends in `tsx`, it isn't a util" 😄

A negative of doing something like `utils/shapes` would be that it is kind of ambiguous.  What does "shape" mean here?  In this moment, we know it means "tldraw shape", but in 4 months it might not be that obvious to us.

I've been trying to think of a hierarchy for this to help make these decisions, and it goes something like this:
- if the function(s) are just used in a single location, co-locate in that file
- if the function(s) are used for only a specific component, co-locate in `components/{feature}` folder
- if the function(s) are used in multiple components in this `app/{app}`, store in `app/{app}/utils`
- if the function(s) are used in multiple `apps`/`package`, move to `package/{whateverIsApplicable}`

apps/obsidian/src/utils/shapes/DiscourseNodeShape.tsx
```diff
@@ -0,0 +1,172 @@
...
+  }, [app, canvasFile, src, plugin]);
+
+  return (
+    <div
+      style={{
+        backgroundColor: nodeColor,
+      }}
+      className="box-border flex h-full w-full flex-col items-start justify-center rounded-md border-2 p-2"
+    >
+      <h1 style={{ margin: 0, fontSize: 16 }}>{title}</h1>
```
mdroidian: Let's try not to mix and match class/style.  Use tailwind where we can and use style only when we can't use tailwind.  Classes allow users to more easily add their own customizations.  And it's easier to read if we are use them consistently.

apps/obsidian/src/components/TldrawView.tsx
```diff
@@ -106,7 +106,11 @@ export class TldrawView extends TextFileView {
         return;
       }
 
-      const { store } = processInitialData(data, assetStore);
+      const { store } = processInitialData(data, assetStore, {
+        app: this.app,
+        canvasFile: this.file!,
```
mdroidian: Is there a specific reason we need to use `!` here?  `!` is generally unsafe and should be avoided whenever possible and only used as a last resort.

apps/obsidian/src/components/TldrawViewComponent.tsx
```diff
@@ -36,6 +42,15 @@ export const TldrawPreviewComponent = ({
   const saveTimeoutRef = useRef<NodeJS.Timeout>();
   const lastSavedDataRef = useRef<string>("");
 
+  const customShapeUtils = [
+    ...defaultShapeUtils,
+    DiscourseNodeUtil.configure({
```
mdroidian, line 43: Interesting choice to use `.configure` here.  As I understand it, `.configure` is used to pass optional configuration that alters the default behavior of a shape.  But here it is being used to pass required context for every instance of the shape.  Another way would be to use a factory function to create the util.

apps/obsidian/src/utils/shapes/discourseNodeShapeUtils.ts
```diff
@@ -0,0 +1,61 @@
+import type { App, TFile } from "obsidian";
+import type DiscourseGraphPlugin from "~/index";
+import type { DiscourseNode } from "~/types";
+import { resolveLinkedFileFromSrc } from "~/utils/assetStore";
+
+export type FrontmatterRecord = Record<string, unknown>;
+
+export const getLinkedFileFromSrc = async (
```
mdroidian: Is the only purpose of this function to call another function?  Why not just call `resolveLinkedTFile...()`?

apps/obsidian/src/utils/shapes/DiscourseNodeShape.tsx
```diff
@@ -0,0 +1,172 @@
...
+  canvasFile: TFile;
+  plugin: DiscourseGraphPlugin;
+  src: string | null;
+}) => {
+  const [title, setTitle] = useState<string>("...");
+  const [nodeTypeName, setNodeTypeName] = useState<string>("");
+  const [nodeColor, setNodeColor] = useState<string>("transparent");
+  const [linkedFile, setLinkedFile] = useState<TFile | null>(null);
+
+  useEffect(() => {
```
mdroidian: I’d be cautious about introducing state and function calls directly inside the shape render since this can have significant performance implications.

It may be worth reconsidering which values truly need to live in React state versus being stored in the shape itself. For example, nodeType—does this value actually change often, or is it effectively static once set? If it’s static, we could avoid fetching and storing it in state entirely.  Same with nodeColor.

The same applies to title. While it makes sense to refresh the value from props when the shape loads (to ensure accuracy), persisting it in state will trigger unnecessary re-renders. Since title is unlikely to change frequently, it might be more efficient to keep it in the shape data rather than local state.

apps/obsidian/src/utils/shapes/DiscourseNodeShape.tsx
```diff
@@ -0,0 +1,172 @@
...
+}: {
+  app: App;
+  canvasFile: TFile;
+  plugin: DiscourseGraphPlugin;
+  src: string | null;
+}) => {
+  const [title, setTitle] = useState<string>("...");
+  const [nodeTypeName, setNodeTypeName] = useState<string>("");
+  const [nodeColor, setNodeColor] = useState<string>("transparent");
+  const [linkedFile, setLinkedFile] = useState<TFile | null>(null);
```
mdroidian: linkedFile is not used

apps/obsidian/src/utils/shapes/DiscourseNodeShape.tsx
```diff
@@ -0,0 +1,172 @@
...
+  getNodeTypeIdFromFrontmatter,
+  getNodeTypeById,
+  FrontmatterRecord,
+} from "./discourseNodeShapeUtils";
+import { DiscourseNode } from "~/types";
+
+export type DiscourseNodeShape = TLBaseShape<
+  "discourse-node",
+  {
+    w: number;
```
mdroidian: We should store color/nodeType here

apps/obsidian/src/utils/tldraw.ts
```diff
@@ -7,9 +13,13 @@ import {
 } from "~/constants";
 import DiscourseGraphPlugin from "~/index";
 import { checkAndCreateFolder, getNewUniqueFilepath } from "./file";
-import { Notice } from "obsidian";
+import { App, Notice, TFile } from "obsidian";
```
mdroidian: remove unused references

mdroidian (changes requested):
Good stuff so far.  I think the biggest blocker is the useState/functions in the rendering of the shape.  Let's try to store as much in shape props as possible to avoid performance bottlenecks.

apps/obsidian/src/utils/shapes/DiscourseNodeShape.tsx
trangdoan982: oooh these are helpful. Thank you

apps/obsidian/src/components/TldrawViewComponent.tsx
trangdoan982: I think it's [official API](https://tldraw.dev/reference/editor/ShapeUtil#options) to parameterize a shape.

apps/obsidian/src/utils/shapes/DiscourseNodeShape.tsx
trangdoan982: yes i agree that having each shape to be an independent React component will cause too much re-rendering.
But I also don't think we should have the node information be static. This creates discrepancies between vault info and canvas. In the future when we sync up the relations, I think we'll need to deal with dynamic rendering of node information anyway.
I was researching a bit and found that we can use a nodeDataStore to keep track of node data information separately (nodeType color, title, frontmatter), so that the data fetching and component rendering could be separated. I pushed up the proposed solution but also happy to talk through this in our 1:1 tomorrow

mdroidian: I think we could go quite a long way using shape props (for color, nodetype, title, etc).  Color and nodeType forsure can be stored there, because they are unlikely to change often (or ever).<br><br>The next question would be: how often do we expect something like title to change?  And what is acceptable.  For how infrequently something like title changes *while on canvas*, seems like this could just be updated in props as well.  Is there some reason this can't be the case?  What problem would reaching for something like nodeDataStore solve that shape props can't?  What does tldraw do for their shapes (I imagine this works quite well for them)?

mdroidian: > But I also don't think we should have the node information be static. This creates discrepancies between vault info and canvas. In the future when we sync up the relations, I think we'll need to deal with dynamic rendering of node information anyway.

Could you speak more to what information "shouldn't be static".<br>In terms of nodes/relations: color, type, text all seem to change infrequently and can be handled by shape props and the built in tldraw functions for updating this data.<br><br>What other data would you expect to need a different mechanism to manage it?

apps/obsidian/src/components/TldrawViewComponent.tsx
mdroidian: Are these options ever going to be optional?  Or are they basically always required each time this shape is instantiated?

apps/obsidian/src/utils/shapes/DiscourseNodeShape.tsx
trangdoan982: For me, the biggest thing is building out the relationship mechanism, because the relations need to be fed live from the file's frontmatter. And frontmatter will change quite frequently. I understand type and color might be changed infrequently, but for title, I do expect it to be tweaked quite often (or at least I do when I apply DG).

apps/obsidian/src/components/TldrawViewComponent.tsx
trangdoan982: they are required each time this shape is instantiated. I don't think ShapeUtil.configure them here would make them optional. Out of all the props available for ShapeUtils I think we can only put these params in shapeUtils.options, bcz the other props are `migrations, props, type, editor`

mdroidian: We could think outside of ShapeUtils for these types of required parameters.

apps/obsidian/src/utils/shapes/DiscourseNodeShape.tsx
mdroidian: The relationship will be a wholly separate shape. 

Let's try putting title, nodeType, node color, and blockRef in shape props initially.  Let's see how far we can take it. Or note where/why props won't work for this.

mdroidian:
How's this coming along?

trangdoan982:
@mdroidian my bad i forgot to push my changes before going out this weekend. this simplified version is now ready for review

apps/obsidian/src/components/canvas/shapes/DiscourseNodeShape.tsx
```diff
@@ -0,0 +1,206 @@
...
+    );
+  }
+
+  indicator(shape: DiscourseNodeShape) {
+    return <rect width={shape.props.w} height={shape.props.h} />;
+  }
+
+  async getFile(
+    shape: DiscourseNodeShape,
+    ctx?: { app: App; canvasFile: TFile },
```
mdroidian: Is ctx ever optional?

apps/obsidian/src/components/canvas/shapes/DiscourseNodeShape.tsx
```diff
@@ -0,0 +1,206 @@
...
+  ): Promise<FrontmatterRecord | null> {
+    const app = ctx?.app ?? this.options.app;
+    const file = await this.getFile(shape, ctx);
+    if (!file) return null;
+    return getFrontmatterForFile(app, file);
+  }
+
+  async getDiscourseNodeType(
```
mdroidian: no longer used

apps/obsidian/src/components/canvas/shapes/DiscourseNodeShape.tsx
```diff
@@ -0,0 +1,206 @@
...
+    ctx?: { app: App; canvasFile: TFile },
+  ): Promise<DiscourseNode | null> {
+    const frontmatter = await this.getFrontmatter(shape, ctx);
+    const nodeTypeId = getNodeTypeIdFromFrontmatter(frontmatter);
+    return getNodeTypeById(this.options.plugin, nodeTypeId);
+  }
+
+  async getRelations(
+    shape: DiscourseNodeShape,
+    ctx?: { app: App; canvasFile: TFile },
```
mdroidian: is ctx ever optional?

apps/obsidian/src/components/canvas/shapes/DiscourseNodeShape.tsx
```diff
@@ -0,0 +1,206 @@
...
+              type: "discourse-node",
+              props: {
+                ...shape.props,
+                title: "(unlinked)",
+              },
+            });
+            return;
+          }
+
+          editor.updateShape<DiscourseNodeShape>({
```
mdroidian: We should only update if title !== basename, otherwise we might flicker the shapes.

apps/obsidian/src/components/canvas/shapes/discourseNodeShapeUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+    null
+  );
+};
+
+export const getNodeTypeForLinkedSrc = async (
+  app: App,
+  plugin: DiscourseGraphPlugin,
+  canvasFile: TFile,
+  src?: string,
+): Promise<DiscourseNode | null> => {
```
mdroidian: image

mdroidian (approved):
Added some thoughts, and approved so you can be unblocked!

PR #327, opened 2025-08-08
Roam: ENG-709: Settings panel for suggestive mode

sid597:
## Summary by CodeRabbit

* **New Features**
  * Introduced "Suggestive Mode" and "Suggestive Rules" configuration tabs in the settings interface, providing new options for discourse suggestions.
  * Added interactive panels for managing suggestive rules, including template editing, embedding block references with live previews, and toggling related settings.
  * Enabled management of "page groups" for suggestions, allowing users to add, remove, and organize groups and associated pages.
  * Added controls to include or exclude current page relations and parent/child blocks in suggestion logic.

* **Enhancements**
  * Improved configuration experience with structured panels and real-time UI updates reflecting the latest settings.

https://www.loom.com/share/7218d88262e142098ab0b5622fa25272?sid=94b47b41-2e76-4874-8849-ce78c8cb4baa

PR #343, opened 2025-08-17
Roam: ENG-733 Glue Settings UI components to database operations

sid597:
## Summary by CodeRabbit

- New Features
  - End-to-end sync of Roam discourse nodes to the database, including embeddings and concepts.
  - Per-type node retrieval with richer node data (author, type, optional title).
  - Progress reporting during embedding fetches.

- Performance
  - Larger embedding batch size for faster processing.
  - Targeted updates by detecting node types with recent changes.

- Reliability
  - Improved orphaned data cleanup for nodes and schemas.
  - More robust error handling and task coordination during sync.

- Developer Experience
  - Clearer, configurable embedding model usage.


sid597:
@coderabbitai full review

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -1,48 +1,171 @@
...
-type ISODateString = string;
+export type DiscourseNodesSinceResult = {
+  pageNodes: RoamDiscourseNodeData[];
+  blockNodes: RoamDiscourseNodeData[];
+};
+
+export const getDiscourseNodeTypeWithSettingsBlockNodes = (
+  node: DiscourseNode,
+  sinceMs: number,
+  extensionAPI: OnloadArgs["extensionAPI"],
```
sid597, line 28: Will update this to use configTree when this is merged to main and rebase the UI branch off it

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -1,48 +1,171 @@
...
+        [?child :create/user ?user-eid]
+        [?user-eid :user/uid ?author_local_id]
+        [?child :edit/user ?eu]
+        [?eu :user/display-name ?author_name]
+        [or 
+         [(> ?childEditTime ?since)]
+         [(> ?nodeEditTime ?since)]]
+        ]`;
+
+  const blockNode = window.roamAlphaAPI.data.q(
```
mdroidian, line 67: use async https://roamresearch.com/#/app/developer-documentation/page/VJfOx4mgR

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
```diff
@@ -0,0 +1,168 @@
...
+export const upsertNodesToSupabaseAsContentWithEmbeddings = async (
+  roamNodes: RoamDiscourseNodeData[],
+  supabaseClient: DGSupabaseClient,
+  context: SupabaseContext,
+): Promise<void> => {
+  if (!context) {
+    console.error("No Supabase context found.");
+    return;
+  }
+  const { spaceId, userId } = context;
```
mdroidian: unused

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  if (roamNodes.length === 0) {
+    return;
+  }
+
+  let nodesWithEmbeddings: RoamDiscourseNodeData[];
+  try {
+    nodesWithEmbeddings = await fetchEmbeddingsForNodes(roamNodes);
+  } catch (error: any) {
```
mdroidian: Don't use any if you can avoid it

apps/roam/src/utils/fetchEmbeddingsForNodes.ts
```diff
@@ -19,13 +11,18 @@ type EmbeddingApiResponse = {
...
+): Promise<RoamDiscourseNodeData[]> => {
   const allEmbeddings: number[][] = [];
-  const allNodesTexts = nodes.map((node) => node.text);
+  const allNodesTexts = nodes.map((node) =>
+    node.node_title ? `${node.node_title} ${node.text}` : node.text,
+  );
 
   for (let i = 0; i < allNodesTexts.length; i += EMBEDDING_BATCH_SIZE) {
     const batch = allNodesTexts.slice(i, i + EMBEDDING_BATCH_SIZE);
+    console.log(
```
mdroidian: Put this behind a dev flag if want to console log it

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -1,48 +1,171 @@
-import getDiscourseNodes from "./getDiscourseNodes";
+import getDiscourseNodes, { DiscourseNode } from "./getDiscourseNodes";
```
mdroidian: add `/* eslint-disable @typescript-eslint/naming-convention */` to the top of the file

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -0,0 +1,398 @@
...
+  nodeTypeSince,
+} from "./getAllDiscourseNodesSince";
+import { cleanupOrphanedNodes } from "./cleanupOrphanedNodes";
+import {
+  getLoggedInClient,
+  getSupabaseContext,
+  SupabaseContext,
+} from "./supabaseContext";
+import { fetchEmbeddingsForNodes } from "./fetchEmbeddingsForNodes";
+import { LocalContentDataInput } from "@repo/database/inputTypes";
```
mdroidian: Are you seeing these errors in your IDE?
image

mdroidian (changes requested):
Let's try to clean up as much of the warning/errors as possible in a first round pass.

If you aren't seeing these, let me know
image

image

apps/roam/src/utils/syncDgNodesToSupabase.ts
maparent: Hrrrm. this used to work, and got broken with changes to tsconfig, package.json etc.
Two problems here:
1. inputTypes is missing from exports in database/package.json, easy to fix. (NOTE: ALL BELOW WAS A MISTAKE, see next message.)
3. Once this is done, the linter complains (rightfully) about roam being in commonJs, and trying to import an ecmaScript module. 

So... what are our options?

1. Use ecmaScript everywhere.  @mdroidian  I tried to just use ecmaScript for roam, and no easy success, i.e. it builds but does not run. Now, that was a minimal attempt, do we have reasons to believe whether or not it could be made to work?
2. use commonJS everywhere, which is a step back.
4. wrap our ecmascript common code in a commonJs wrapper. Found this [option](https://stackoverflow.com/a/75348391).
5. Use UMD everywhere. Rabbit hole for a legacy format imho
6. Not reuse any component from outside in roam. Sigh.
I think this is a blocker that needs to be addressed. Willing to spend more time on this.

maparent: Argh. Forgot to `npm install` in my workspace first, I my workspace was in the `main` state. So all I wrote (after the first 1.) does not apply, i.e. I do not see the second error, though I'm not sure why not. (I also tried rebasing on main, no difference.)
But... I saw those errors before and now, and I'm sure there are conditions that trigger them, so will leave this comment here for now.

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
```diff
@@ -0,0 +1,168 @@
+import { RoamDiscourseNodeData } from "./getAllDiscourseNodesSince";
+import { SupabaseContext } from "./supabaseContext";
+import { LocalContentDataInput } from "@repo/database/inputTypes";
+import { DGSupabaseClient } from "@repo/ui/lib/supabase/client";
+
+
+const EMBEDDING_BATCH_SIZE = 200;
+const API_URL = `https://discoursegraphs.com/api/embeddings/openai/small`;
```
maparent: I suggest using `import { nextApiRoot } from "@repo/ui/lib/execContext";` to get the start of this URL.

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
```diff
@@ -0,0 +1,168 @@
...
+      `fetchEmbeddingsForNodes: Fetching batch ${i / EMBEDDING_BATCH_SIZE + 1} of ${allNodesTexts.length / EMBEDDING_BATCH_SIZE}`,
+    );
+
+    const response = await fetch(API_URL, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ input: batch }),
+    });
+
+    if (!response.ok) {
```
maparent, line 60: At some point there were a lot of transient errors with openAI, and retrying was expected. Do you know if this is still the case?
One case where I know this would still be true is if we ever run into request throttling, in which case we probably want incremental backoff. I know we're far from that much usage, but that will become an issue with more adopters. Punting that should at least be a conscious decision.

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
```diff
@@ -0,0 +1,168 @@
...
+      throw new Error(
+        `API Error (${response.status}) processing batch ${
+          i / EMBEDDING_BATCH_SIZE + 1
+        }: ${errorData.error || "Failed to fetch embeddings"}`,
+      );
+    }
+
+    const data: EmbeddingApiResponse = await response.json();
+    if (!data || !Array.isArray(data.data)) {
+      throw new Error(
```
maparent, line 78: also may be worth retrying.

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -1,48 +1,171 @@
...
 import findDiscourseNode from "./findDiscourseNode";
+import { OnloadArgs } from "roamjs-components/types";
+import getDiscourseNodeFormatExpression from "./getDiscourseNodeFormatExpression";
 
-type RoamDiscourseNodeData = {
+type ISODateString = string;
+
+export type RoamDiscourseNodeData = {
```
maparent, line 9: ~Could it be simpler to use LocalContentDataInput directly? (I tried and no major obstacle. Actually used LocalContentDataInput | null)~
Actually, you're right, use this as a the roam function output, minus the vector. Then convert to `LocalContentDataInput` ~as the output of~ before sending to `fetchEmbeddingsForNodes`.

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -1,48 +1,171 @@
...
+          nodeTypesSet.has(node.type)
+        ) {
+          return null;
+        }
+        return {
+          ...entity,
+          type: node.type,
+        };
+      })
+      .filter((n): n is RoamDiscourseNodeData => n !== null),
```
maparent: Not sure I understand the syntax here,  how is it different from simply `(n) => n !== null`

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
```diff
@@ -0,0 +1,168 @@
...
+const EMBEDDING_BATCH_SIZE = 200;
+const API_URL = `https://discoursegraphs.com/api/embeddings/openai/small`;
+
+type EmbeddingApiResponse = {
+  data: {
+    embedding: number[];
+  }[];
+};
+
+export const fetchEmbeddingsForNodes = async (
```
maparent, line 41: This looks almost identical to the code in `fetchEmbeddingsForNode`, is there a reason you cannot merge them?

apps/roam/src/utils/fetchEmbeddingsForNodes.ts
```diff
@@ -19,13 +11,18 @@ type EmbeddingApiResponse = {
 };
 
 export const fetchEmbeddingsForNodes = async (
-  nodes: DiscourseGraphContent[],
-): Promise<DiscourseGraphContent[]> => {
+  nodes: RoamDiscourseNodeData[],
+): Promise<RoamDiscourseNodeData[]> => {
```
maparent: So I suggest using `LocalContentDataInput` as the *input and* output type here

apps/roam/src/utils/fetchEmbeddingsForNodes.ts
```diff
@@ -67,7 +64,7 @@ export const fetchEmbeddingsForNodes = async (
   }
   return nodes.map((node, i) => ({
     ...node,
-    model: "openai_text_embedding_3_small_1536",
```
maparent, line 70: and here wrap those two in the `embedding_inline`. (Annoyingly TS does not flag errors in the map.)

apps/roam/src/utils/syncDgNodesToSupabase.ts
sid597: Yeah and I don't know how to solve it the right way, these I think also lead to lint errors in `convertDgToSupabaseConcepts` function

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -0,0 +1,398 @@
...
+    if (!context) {
+      console.error("proposeSyncTask: Unable to obtain Supabase context.");
+      return {
+        lastUpdateTime: null,
+        spaceId: 0,
+        worker: "",
+        shouldProceed: false,
+      };
+    }
+    const worker = window.roamAlphaAPI.user.uid();
```
sid597, line 79: This also throws errors but also works

apps/roam/src/utils/fetchEmbeddingsForNodes.ts
sid597: removed

apps/roam/src/utils/getAllDiscourseNodesSince.ts
sid597: Oh ok I have been using `data.backend.q`

apps/roam/src/utils/fetchEmbeddingsForNodes.ts
```diff
@@ -19,13 +11,18 @@ type EmbeddingApiResponse = {
 };
 
 export const fetchEmbeddingsForNodes = async (
-  nodes: DiscourseGraphContent[],
-): Promise<DiscourseGraphContent[]> => {
+  nodes: RoamDiscourseNodeData[],
+): Promise<RoamDiscourseNodeData[]> => {
   const allEmbeddings: number[][] = [];
-  const allNodesTexts = nodes.map((node) => node.text);
+  const allNodesTexts = nodes.map((node) =>
```
maparent: And ideally that transformation should be part of a separate utility function that translates `RoamDiscourseNodeData` to `LocalContentDataInput`, setting the variant at the same time. Having this as a separate type-shifting function would ensure we'd never forget to do it.

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
maparent: (this and other comments will move to fetchEmbeddingsForNodes)

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
```diff
@@ -0,0 +1,168 @@
...
+    });
+
+    if (error) {
+      console.error(`upsert_content failed for batch ${idx + 1}:`, error);
+      throw error;
+    }
+  }
+};
+
+export const upsertNodesToSupabaseAsContentWithEmbeddings = async (
```
maparent, line 124: Exists here and in syncDgNodesToSupabase. Unify.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -0,0 +1,398 @@
...
+    v_space_id: spaceId,
+    v_creator_id: userId,
+    content_as_document: true,
+  });
+  if (error) {
+    console.error("upsert_content failed:", error);
+  }
+};
+
+export const convertDgToSupabaseConcepts = async (
```
maparent: This converts _and uploads_. I think the name should reflect this. (Or better still, have this function return the converted concepts, and another function that calls this and upload. It'll be easier to test the former.)

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -0,0 +1,398 @@
...
+      extensionAPI,
+    );
+    const supabaseClient = await getLoggedInClient();
+    const context = await getSupabaseContext();
+    if (!context) {
+      console.error("No Supabase context found.");
+      await endSyncTask(worker, "failed");
+      return;
+    }
+    await upsertNodesToSupabaseAsContentWithEmbeddings(
```
maparent, line 379: So ideally the nodes should be converted to `LocalContentDataInput` before calling this function and the next one, it will simplify both.

maparent (changes requested):
I want to see what it looks like with a better type boundary, and then I'll look again at some of the update logic (but that looks good overall.)

apps/roam/src/utils/syncDgNodesToSupabase.ts
maparent:
```
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -7,6 +7,7 @@
   "exports": {
     "./types.gen.ts": "./types.gen.ts",
     "./types.gen": "./types.gen.ts",
+    "./inputTypes": "./inputTypes.ts",
     "./dbDotEnv": "./dbDotEnv.ts"
   },
   "scripts": {
```

apps/roam/src/utils/syncDgNodesToSupabase.ts
maparent: I think you mean shows linting error? ("Throwing" suggests something else!) But yes, there is a mismatch here.

maparent: (I mean the type we receive is incomplete.)

sid597: yeah linting error, we don't have `roamAlphaAPI.user.uid` in the definition but roam supports it

Image

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
sid597: my bad

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
sid597: No I am not aware if its still the case. We can add this as a comment in code for later, wdyt @mdroidian

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
sid597: should we not do it on server side?

apps/roam/src/utils/fetchEmbeddingsForNodes.ts
sid597: done

apps/roam/src/utils/fetchEmbeddingsForNodes.ts
sid597: Done

apps/roam/src/utils/syncDgNodesToSupabase.ts
sid597: we need to know the discourse node schema type for the instance node see line 223 and the `RoamDiscourseNodeData` has that so we cannot pass the converted type

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
maparent: That does make sense. I suggest adding a linear task for that.

maparent:
Ok, thanks a lot for the deduplication and type boundaries, much clearer to read.
I wanted to test it, but I notice that `createOrUpdateDiscourseEmbedding` is not called anywhere yet. Do you see that happening in another PR? I guess I should add a button in the admin panel on my side :-)
A last concern: You decided to materialize all four kinds of concepts (node schema, node, relation schemas, relations.) I did not look into how you get the relation instance nodes, is `getDiscourseRelationTriples` exhaustive?
More important: I wonder if the underlying content is always present for relation concepts? I get the impression that the content upload is limited to nodes, but again did not look deep into the roam queries. If not, I say let's wait for the relation schemas.
(This is what I wanted to test!)
Otherwise, lgtm.

sid597:
@maparent  Yeah I was not sure if we want to include the relations now, I think we should leave it for now, I will remove it from the function, thanks. 

The function will be called from the UI component I will update the callers on that pr #327 

We want to merge this pr first then #344 then #327 then #340

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
sid597: [ENG-761: Retry getting embedding for batches on the api/embeddings/openai/small route](https://linear.app/discourse-graphs/issue/ENG-761/retry-getting-embedding-for-batches-on-the-apiembeddingsopenaismall)

maparent (approved):
Good. I'll trust you checked the runtime. LGTM.

sid597:
@maparent Now you can use the branch locally the upsert will automatically start on plugin load, also review the changes in this commit https://github.com/DiscourseGraphs/discourse-graph/pull/343/commits/e3f7352ff36ed87b171f21ad41a51ee369dc3e0a

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
```diff
@@ -0,0 +1,172 @@
...
+  nodes: RoamDiscourseNodeData[];
+  userId: number;
+}): LocalContentDataInput[] => {
+  return nodes.map((node) => {
+    const variant = node.node_title ? "direct_and_description" : "direct";
+    const text = node.node_title
+      ? `${node.node_title} ${node.text}`
+      : node.text;
+    return {
+      author_id: userId,
```
maparent: Actually... this is wrong, as it makes the current user the author of all nodes. In theory, you should not have to set this, since you set `author_local_id`; but in practice, the author_local_id may not have been materialized in the database, which may lead to the insert failing. So make sure to collect all author_ids and creator_ids, and to create platformAccounts (prob. with createAccountInSpace; I will create a bulk version of that.)

maparent:
And for the record : all my upsert functions are failing miserably, I forgot to commit. Working on this, it may go to the morning.

sid597:
@maparent it worked for me when I tested it a few times then I pushed the commit and asked you for review, now when I tried to test it again on a different graph after reading your exp. Its failing on my end also I am getting the error: 

```
Access to fetch at 'http://127.0.0.1:54321/functions/v1/create-space' from origin 'https://roamresearch.com' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

```
supabaseContext.ts:82 Error: Failed to create space
    at getSupabaseContext (supabaseContext.ts:65:36)
    at async getLoggedInClient (supabaseContext.ts:93:21)
    at async proposeSyncTask (syncDgNodesToSupabase.ts:66:28)
    at async createOrUpdateDiscourseEmbedding (syncDgNodesToSupabase.ts:316:53)
```


There is another type of bug that I am getting like 

```
upsert_content failed for batch 1: {code: '23502', details: null, hint: null, message: 'null value in column "author_id" of relation "Document" violates not-null constraint'}
```
This bug I don't know refers to what, checking  data that is passed on to upsert_content every data entry has the same `author_id` idk what this is refering to. 

All the bugs are quite confusing because I tried this in a personal graph which has 4 nodes to test it on small batch and it works (it also worked yesterday)


image


image

maparent:
Ok. First apologies: The commit was a red herring, I was tired yesterday.
The first error you get may be due to missing cors policies, I think this is a change in supabase, corrected in ENG-758, you may have to merge that into your branch.
The second error is what started me on looking deeper into this. Basically, you need to make sure the authors of the documents exist. I am going to add a small fix so that the authors are created if absent, that's currently missing; but it's not the real solution. Ideally we should upsert all authors/creators before we upsert the content. That's going to be so much more efficient than putting them inline every time.
It's likely you missed that error earlier on your test graph because there were not that many authors (besides yourself as the logged in user.)
Working on a bulk uploading function, that will help.

maparent:
Bulk functions pushed in #357

sid597:
Yeah the personal graph only has me as the user but the other graph `discourse-dev-graph` has many users and it worked on it yesterday, and I was remembering that we create the users if they don't exist, might me mis-remembering or smth.

mdroidian (approved):

maparent (approved):
Ok, this basically lgtm. One small nit before you push: There are a lot of imports that could be `import type {...} ...`, or at least `import {x, type y} ...`, fwiw.

apps/roam/src/utils/upsertNodesAsContentWithEmbeddings.ts
sid597: I did not see this comment before I need to learn the github ui better

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -1,48 +1,171 @@
...
+    [?node :edit/time ?nodeEditTime]
+    [?node :create/user ?user-eid]
+    [?user-eid :user/uid ?author_local_id]
+    [?node :edit/user ?eu]
+    [(get-else $ ?eu :user/display-name "Unknown-person") ?author_name]
+    [(> ?nodeEditTime ?since)]
+]`;
+
+  //@ts-ignore - backend to be added to roamjs-components
+  const allNodes = (await window.roamAlphaAPI.data.async.q(
```
mdroidian, line 114: @sid597 This should probably be `.backend`.   We can discuss at our next meeting.

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -1,48 +1,171 @@
...
+        [?child :create/user ?user-eid]
+        [?user-eid :user/uid ?author_local_id]
+        [?child :edit/user ?eu]
+        [?eu :user/display-name ?author_name]
+        [or 
+         [(> ?childEditTime ?since)]
+         [(> ?nodeEditTime ?since)]]
+        ]`;
+
+  const blockNode = window.roamAlphaAPI.data.q(
```
mdroidian, line 67: @sid597 This should probably be `.backend`.   We can discuss at our next meeting.

apps/roam/src/utils/getAllDiscourseNodesSince.ts
```diff
@@ -1,48 +1,171 @@
...
 
   const discourseNodes = getDiscourseNodes();
+  const nodeTypesSet = new Set(nodeTypes.map((nodeType) => nodeType.type));
+
+  result.push(
+    ...allNodes.flatMap((entity) => {
+      if (!entity.source_local_id) {
+        return [];
+      }
+      const node = findDiscourseNode(entity.source_local_id, discourseNodes);
```
mdroidian, line 127: @sid597 This will likely need to be a custom function.  `findDiscourseNode` will likely be a performance issue on large graphs.  We can discuss at our next meeting.

PR #347, opened 2025-08-18
[ENG-404] Fix multiple things about node search menu

trangdoan982:
https://www.loom.com/share/b8cf4939d1f549f0a792b88a286b9212

## Note:
- I ended up not using the Multi-select because the interface to select/ deselect all that it offers is fitting for our context. I think that having all the nodes available be stationary is a better UI than being added to a "selected list". but happy to switch if you think Multi-select is better
- This PR address these following Linear tickets: [ENG-404: Node summoning menu multi-select filter UI](https://linear.app/discourse-graphs/issue/ENG-404/node-summoning-menu-multi-select-filter-ui), [ENG-673: node summoning menu items should wrap text](https://linear.app/discourse-graphs/issue/ENG-673/node-summoning-menu-items-should-wrap-text), and widening the menu

## Summary by CodeRabbit

- New Features
  - New menu-driven “Filter by Type” with per-type selection and an “Only” action.
  - Bulk select/deselect all types.

- Improvements
  - Wider, scrollable filter panel with streamlined layout.
  - Results now support multiline display and keep the menu open for faster selection.
  - Enhanced keyboard navigation and active states.
  - Improved popover behavior and focus preservation during interactions.

- Chores
  - Updated dependency import style in a utility module; no user-facing changes.

### DIFF
--- a/apps/obsidian/src/utils/generateUid.ts
+++ b/apps/obsidian/src/utils/generateUid.ts
@@ -1,4 +1,4 @@
-import { nanoid } from "nanoid";
+import nanoid from "nanoid";
 
 const generateUid = (prefix = "dg") => {
   return `${prefix}_${nanoid()}`;
--- a/apps/roam/src/components/DiscourseNodeSearchMenu.tsx
+++ b/apps/roam/src/components/DiscourseNodeSearchMenu.tsx
@@ -10,11 +10,10 @@ import {
   MenuItem,
   Popover,
   Position,
-  Checkbox,
   Button,
   InputGroup,
-  Switch,
 } from "@blueprintjs/core";
+import { MultiSelect, ItemRenderer, ItemPredicate } from "@blueprintjs/select";
 import ReactDOM from "react-dom";
 import getUids from "roamjs-components/dom/getUids";
 import getTextByBlockUid from "roamjs-components/queries/getTextByBlockUid";
@@ -184,6 +183,11 @@ const NodeSearchMenu = ({
       .filter((type) => checkedTypes[type.type])
       .filter((type) => searchResults[type.type]?.length > 0);
   }, [discourseTypes, checkedTypes, searchResults]);
+
+  const selectedTypes = useMemo(
+    () => discourseTypes.filter((t) => checkedTypes[t.type]),
+    [discourseTypes, checkedTypes],
+  );
 
   const allItems = useMemo(() => {
     const items: {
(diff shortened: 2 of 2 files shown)


trangdoan982:
@coderabbitai full review

mdroidian (changes requested):
I couldn't get it to work.
https://www.loom.com/share/082393dbf75242eaadc0a4eadae9d921

Also, in the ticket it was mentioned "Changing the pattern to be "Opt in" vs "Opt out" where default is "all" if none are selected."  I believe we should still do this.  Currently it takes multiple clicks just to filter a single node type, which is likely the most often action in this instance.

Personally I find the checkboxes quite busy.  I still would prefer multi-select, but I'm happy to defer to @jsmorabito   It is really quite busy now.  If we keep the checkboxes, maybe they should go down the left hand side or something.

mdroidian:
Any questions regarding this one?  What is left before re-review?

trangdoan982:
no this is still wip. I will ping you when it's ready for re-review

trangdoan982:
https://www.loom.com/share/754db8895a5546e39a9ed71f361abc52 @mdroidian new UI fixed

apps/obsidian/src/utils/generateUid.ts
```diff
@@ -1,4 +1,4 @@
-import { nanoid } from "nanoid";
+import nanoid from "nanoid";
```
mdroidian, line 1: Interesting, I wonder why this diff is showing up.  Looks like it actually is 
`import nanoid from "nanoid";` in `main` (in my IDE and GitHub)

image

mdroidian:
@CodeRabbit full review

apps/roam/src/components/DiscourseNodeSearchMenu.tsx
```diff
@@ -175,6 +184,8 @@ const NodeSearchMenu = ({
       .filter((type) => searchResults[type.type]?.length > 0);
   }, [discourseTypes, checkedTypes, searchResults]);
 
+  // derived selection set from checkedTypes; no separate selected list needed
```
mdroidian: Is this describing `filteredTypes`?
Seems more accurate to say that it is derived from `discourseTypes`

Regardless, if you feel that i put it directly above variable and leave no spaces.
But only if it is adding additional context that the code itself doesn't make obvious.

mdroidian (approved):
I don't think this is blocking, but to considered for future UX.

There is a few pixels on the top, bottom, and right of "Only" that click the menu item rather than "Only".

A `css` fix would be to move the padding vertical to the middle element and extend the "Only" button to the edges, then recenter "Only" and ✔ vertically.

image

PR #348, opened 2025-08-19
ENG 581 Add ability to change size/font in Discourse Shape

mdroidian:
https://www.loom.com/share/926daecb4d43423fae9805e74170a369

In this PR i've added the ability to edit the size and fontFamily.
The problem is this sets the default fontFamily to tldraw `draw` and size to `m`.

`getDefaultProps()` is getting clobbered by something.  I remember this was an issue in the previous version as well.

Regardless, there's an issue with the migration on this, so it will take some more work.

## Summary by CodeRabbit

- New Features
  - Discourse nodes now support per-node font family and size, improving readability and styling.
  - New nodes created via export or query builder default to medium size with a hand-drawn font.

- Chores
  - Automatic migration updates existing canvases to apply default font and size to discourse nodes for consistency.


mdroidian:
@CodeRabbit full review

PR #353, opened 2025-08-21
ENG-758 Edge function stopped working, now needs CORS handling

maparent:
While testing the edge function, it stopped working in development. (May be due to a fresh npm install updating supabase?)
Adding CORS seems to solve the issue.

## Summary by CodeRabbit

- New Features
  - Enabled cross-origin access for the create-space API, including support for Roam Research, localhost, and Vercel preview domains.
  - Added full CORS handling (preflight and response headers) to reduce errors and improve reliability when calling the API from browsers.

- Chores
  - Added a new dependency to support function calls from the UI.

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -35,6 +35,7 @@
     "typescript": "5.5.4"
   },
   "dependencies": {
+    "@supabase/functions-js": "^2.4.5",
     "@supabase/supabase-js": "^2.52.0",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
--- a/apps/website/app/utils/llm/cors.ts
+++ b/apps/website/app/utils/llm/cors.ts
@@ -3,7 +3,7 @@ import { NextRequest } from "next/server";
 const allowedOrigins = ["https://roamresearch.com", "http://localhost:3000"];
 
 const isVercelPreviewUrl = (origin: string): boolean =>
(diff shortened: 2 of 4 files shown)


maparent:
@coderabbit review.

maparent:
@coderabbit review (try again)

maparent:
@coderabbit 
1. The supabase edge function cannot reuse code from elsewhere in the code base, so this cannot be unified in a utility function.
2. Supabase edge functions are a totally different api set from the vercel apis, and it's not abnormal that the cors requirements differ.
3. Security: Great catch.
4. Good idea, but punting for now.

maparent:
@coderabbit for the security issue: Already pushed.

packages/ui/package.json
```diff
@@ -35,6 +35,7 @@
     "typescript": "5.5.4"
   },
   "dependencies": {
+    "@supabase/functions-js": "^2.4.5",
```
mdroidian, line 38: Why are we adding Supabase functions to UI?

mdroidian (approved):

packages/ui/package.json
maparent: Because it's used in `src/lib/supabase/contextFunctions.ts`. It was being picked up from elsewhere but we did not know it was needed locally. I think the zed linter complained?

PR #357, opened 2025-08-23
ENG-764 Bulk account uploading

maparent:
https://linear.app/discourse-graphs/issue/ENG-764/bulk-account-uploading

## Summary by CodeRabbit

* **New Features**
  * Bulk account upsert: add or update multiple space members in one request.
  * Inline account support in document and content upserts, auto-creating/updating authors/creators as needed.
  * Backward-compatible single-account creation preserved.

* **Refactor**
  * Unified composite account input across APIs for consistent name/email/editor handling.
  * Client types updated to include the new account input.
  * Small change to embedding match response field ordering.

### DIFF
--- a/packages/database/src/dbTypes.ts
+++ b/packages/database/src/dbTypes.ts
@@ -673,6 +673,17 @@ export type Database = {
         Args: { p_account_id: number }
         Returns: boolean
       }
+      upsert_account_in_space: {
+        Args: {
+          local_account: Database["public"]["CompositeTypes"]["account_local_input"]
+          space_id_: number
+        }
+        Returns: number
+      }
+      upsert_accounts_in_space: {
+        Args: { accounts: Json; space_id_: number }
+        Returns: number[]
+      }
       upsert_concepts: {
         Args: { data: Json; v_space_id: number }
         Returns: number[]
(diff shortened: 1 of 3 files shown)


maparent:
@coderabbit review

maparent:
@coderabbit review again

maparent:
@coderabbit good catch for `space_id` vs `space_id_` in `upsert_accounts_in_space`.
The issue with ordering of `_local_content_to_db_content` is moot; that function does not do database operations. The database operations happen in the INSERT many lines afterwards.

Otherwise, when there is a PR involving `projects/database/supabase/schemas` and `projects/database/supabase/migrations`, no need to identify errors twice, as long as the two are consistent. Just point out the errors in `projects/database/supabase/schemas` only.

mdroidian (approved):
@maparent What was the specific use case you were envisioning here?  When will we want to bulk upload accounts?  Is there a new client that is requesting this feature?  Or was there some other reason that prompted this work?

maparent:
The use case: The global node uploading (Sid's work on #343) failed when the accounts are not specified. There's a way to make it work (with inline accounts) but it's verbose. Pre-filling the accounts is way more efficient. Sorry I did not point to it, but most of the conversation is there.

mdroidian:
Do we intend to use `upsert_accounts_in_space`?  Or will the accounts just be automatically created if they don't already exist using `create_account_in_space`?

maparent:
the former will be more efficient.

PR #359, opened 2025-08-24
ENG-767 Roam tsc fixes

mdroidian:
`npx tsc` // `turbo check-types -F roam`
image

## Summary by CodeRabbit

- New Features
  - None

- Bug Fixes
  - Corrected option keys in results rendering to display pages/blocks as expected.
  - Ensured concept dependency merging is complete and consistent.
  - Improved error handling when identifying the current user to prevent silent failures.

- Refactor
  - Standardized module import approach for improved runtime compatibility.
  - Replaced external type references with local placeholders to stabilize builds during module migration.

PR #376, opened 2025-08-25
ENG-782 Allow CommonJS type imports in ESM

maparent:
https://linear.app/discourse-graphs/issue/ENG-782/allow-commonjs-type-imports-in-esm

Resolve mismatch between esmascript in database and commonjs in 

* Actually build database package, no cross-package reference.
* Revert most of #359, so we use imports instead of require.
* Silence the commonjs import errors on import type, which have no actual impact on type inference machinery.
* Mostly use relative paths within database.
* The resulting roam package builds and runs.

## Summary by CodeRabbit

- Refactor
  - Switched to ES module imports and centralized types for improved consistency and compatibility.
  - Updated build outputs to reference compiled artifacts for better runtime reliability.
- Chores
  - Adjusted dependency placement to dev-only where appropriate, reducing runtime footprint.
  - Tweaked TypeScript configuration to align with project standards and streamline builds.
- Potential Impact
  - Some previously exposed internal modules are no longer exported; consumers should rely on documented public APIs.
  - No functional changes to app behavior; improvements target stability, typing accuracy, and package hygiene.


maparent:
@coderabbit review

PR #377, opened 2025-08-26
ENG-785 Use commonjs in whole project

maparent:
https://linear.app/discourse-graphs/issue/ENG-785/use-commonjs-in-whole-project

*    Remove module from database, and adjust code accordingly
*    One exception gets kept as a mts file.
*    Ensure all projects' builds result in a dist folder.
*    Revert most of #359, so we use imports instead of require.

Note: At this point, cucumber tests are broken, still working on that.

## Summary by CodeRabbit

* Bug Fixes
  * Hypothetical node generation now returns only the node text, producing cleaner, more consistent results.

* Chores
  * Standardized module system and imports across packages.
  * Updated TypeScript configs and added build/type-check scripts to improve build reliability.
  * Removed explicit package transpilation from the website configuration.


maparent:
@coderabbit review

packages/database/cucumber.json
```diff
@@ -1,6 +1,6 @@
 {
   "default": {
-    "loader": ["ts-node/esm"],
-    "import": ["features/step-definitions/**/*.ts"]
+    "requireModule": ["ts-node/register"],
+    "require": ["features/step-definitions/**/*.ts"]
```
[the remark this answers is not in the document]
maparent: Interesting, but they do currently resolve. So not sure why this is needed?

packages/ui/package.json
```diff
@@ -17,6 +16,7 @@
   },
   "scripts": {
     "lint": "eslint .",
+    "build": "tsc",
```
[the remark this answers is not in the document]
maparent: Thank you for that one, most useful.

packages/ui/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "@repo/typescript-config/react-library.json",
   "compilerOptions": {
     "baseUrl": ".",
-    "rootDir": "."
+    "rootDir": ".",
+    "outDir": "dist"
   },
```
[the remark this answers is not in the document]
maparent: declaration is inherited, so not adding.
you propose files that do not exist as values for main and types. This package may not have a main, and that's fine.

maparent:
@mdroidian ok, this is functional. cucumber works. build, dev and check-types succeed locally.
What fails in deployment: tsc of packages is done with no-emit, so tsc of apps fails to find the declarations. Will give a look on how to fix that tomorrow morning, but otherwise it's ready for review.

maparent:
ok check-types was easy to solve, now stumped by deploy issue. Works locally (even after wiping all dist folders.)

maparent:
Ok, did fail locally after wiping dist after all. Not sure what I did wrong the first time. So... does deploy not come after a build?

packages/database/scripts/lint.ts
```diff
@@ -3,7 +3,7 @@ import { readdir, Dirent } from "node:fs";
 import { join, dirname } from "path";
 import { fileURLToPath } from "url";
```
mdroidian: unused

packages/database/src/dbDotEnv.ts
```diff
@@ -1,9 +1,15 @@
 import dotenv from "dotenv";
 import { readFileSync, existsSync } from "fs";
-import { join, dirname } from "path";
+import { join, dirname, basename } from "path";
 import { fileURLToPath } from "url";
```
mdroidian, line 4: unused

packages/database/package.json
```diff
@@ -3,23 +3,53 @@
...
+      ]
+    }
   },
   "scripts": {
     "init": "supabase login",
-    "dev": "supabase start && tsx scripts/createEnv.ts && supabase functions serve",
+    "dev": "supabase start && tsx scripts/createEnv.mts && supabase functions serve",
     "stop": "supabase stop",
-    "check-types": "tsc --noEmit --skipLibCheck",
+    "check-types": "tsc --emitDeclarationOnly --skipLibCheck",
```
mdroidian, line 47: Why do we want the check-types to emit declarations?  Aren't we just running a check here?

packages/ui/package.json
```diff
@@ -3,23 +3,45 @@
...
+      ]
+    }
   },
   "scripts": {
     "lint": "eslint .",
+    "build": "tsc",
     "lint:fix": "eslint . --fix",
     "generate:component": "turbo gen react-component",
-    "check-types": "tsc --noEmit --skipLibCheck",
+    "check-types": "tsc --emitDeclarationOnly --skipLibCheck",
```
mdroidian, line 44: Same thing here

packages/utils/package.json
```diff
@@ -3,11 +3,23 @@
...
+  "typesVersions": {
+    "*": {
+      "*": [
+        "dist/src/*.d.ts"
+      ]
+    }
   },
   "scripts": {
+    "build": "tsc",
+    "check-types": "tsc --emitDeclarationOnly --skipLibCheck",
```
mdroidian, line 22: same here

mdroidian:
> [Error](https://github.com/DiscourseGraphs/discourse-graph/actions/runs/17255970372/job/48967735839?pr=377#step:5:25): Cannot find module '/home/runner/work/discourse-graph/discourse-graph/node_modules/@repo/database/dist/src/dbDotEnv.js'

FYI, This is still being imported via `require`.  Is that the issue?

https://github.com/DiscourseGraphs/discourse-graph/blob/322d5ff2b8e2ccdb7124339c4fa9704407cf6e0e/apps/roam/scripts/compile.ts#L7

Search for `https://linear.app/discourse-graphs/issue/ENG-766/upgrade-all-commonjs-to-esm` to find all the other require references.

packages/database/cucumber.json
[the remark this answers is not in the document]
maparent: @coderabbit no that question was addressed to you. Why is this needed if the module resolves?

packages/database/src/dbDotEnv.ts
maparent: used at line 8.

packages/database/package.json
maparent: If the package declarations are not generated, the check of the apps, that refer to those declarations, fails.

packages/ui/package.json
maparent: same comment

packages/utils/package.json
maparent: same comment.

maparent:
> > [Error](https://github.com/DiscourseGraphs/discourse-graph/actions/runs/17255970372/job/48967735839?pr=377#step:5:25): Cannot find module '/home/runner/work/discourse-graph/discourse-graph/node_modules/@repo/database/dist/src/dbDotEnv.js'
> 
> FYI, This is still being imported via `require`. Is that the issue?

No, but it did ultimately point to the issue.

I asked if roam was built before deploy, and it is, through compile. But compile now uses dbDotEnv, and that is only built when the database is built.
I can think of three ways to fix this:
1. quick and dirty, already pushed: build the database before calling deploy.
2. Detect in scripts/compile that dbDotEnv is missing, and build it at need. (And then yes we'd use require.) This cross-build breaks encapsulation a bit, imho.
3. another option: Recode dbDotEnv as js directly, so we don't have to build it.

2 and 3 are more complex, I want your feedback on whether it's worth doing.

I went to import, and then back to require so I could give a better error message when the database is not built.

packages/database/cucumber.json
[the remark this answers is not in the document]
maparent: @coderabbit: you are getting into a rabbit hole, if I may say so. This works both locally and in deployment, don't try to explain why it cannot work.
As for why it works: the `packages/*/package.json` files have `"name": "@repo/..."` values, though this does indeed not follow directory structure. And the `apps/*/package.json` files refer to those packages by name, and hence have access to the corresponding exports.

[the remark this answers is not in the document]
maparent: @coderabbit: I am curious about where you saw this `tsc not found`. Do you have your own build environment? Again this is nowhere in deployment logs. Or are you hallucinating a deployment process?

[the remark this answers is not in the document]
maparent: @coderabbit thank you for letting me know you have an actual sandbox. And yes, if does not have (and fails to install) tsc, it does mean you cannot rely on it for this project.

mdroidian (approved):
> 1. quick and dirty, already pushed: build the database before calling deploy.

If deploy always requires this, then I think we could achieve this via turbo.json with something like `"dependsOn": ["@repo/database#build"],`  It would then be in a single place.

But I'll leave that up to you.

maparent:
I got your latest comment by mail but do not  see it here? Solved in #801.

mdroidian:
I think `npm i` actually solved it for me

packages/database/src/dbDotEnv.ts
mdroidian: @maparent `fileURLToPath` is unused
image

packages/database/package.json
mdroidian: @maparent 
Can you show me an example of what you are referring to?  This wasn't the case previously when I initially checked in `check-types`, and I just just changed them all to `noEmit` and it ran successfully for me

image

I also cd packages/database and ran `npm run check-types` successfully
image

maparent:
Process to reproduce the issue :
1. `rm -rf projects/*/dist apps/*/dist`
2. `turbo dev`
Process to avoid the issue: intercalate
`turbo build -F @repo/database`
Reason: roam expects to use dbDotEnv in its compiled form.
Solution: Probably make dbDotEnv into a js file? (cf ENG-799)

PR #355, opened 2025-08-22
ENG-760 adding "check-types" script to major apps/packages

mdroidian:
## Summary by CodeRabbit

* **Chores**
  * Standardized TypeScript type-checking scripts across apps and packages.
  * Improved developer workflow by enabling faster checks that skip library declaration validation.
  * Aligned existing scripts for consistency without affecting build or publish behavior.
  * Simplified database package preflight steps to focus solely on type validation.
  * No user-facing changes; functionality and runtime behavior remain unchanged.

PR #382, opened 2025-08-27
ENG-794 - Added a new script "check-schema"  to replace old check-types

mdroidian:
was replaced in https://github.com/DiscourseGraphs/discourse-graph/pull/355

### DIFF
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -15,6 +15,7 @@
     "dev": "supabase start && tsx scripts/createEnv.ts && supabase functions serve",
     "stop": "supabase stop",
     "check-types": "tsc --noEmit --skipLibCheck",
+    "check-schema": "npm run lint && supabase stop && npm run dbdiff",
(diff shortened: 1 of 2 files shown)


packages/database/package.json
```diff
@@ -15,6 +15,7 @@
     "dev": "supabase start && tsx scripts/createEnv.ts && supabase functions serve",
     "stop": "supabase stop",
     "check-types": "tsc --noEmit --skipLibCheck",
+    "check-schema": "npm run lint && supabase stop && npm run dbdiff",
```
maparent: npm run lint does both the code and database linting. Maybe instead do `tsx scripts/lint.ts` here, so we only do the latter.
Otherwise lgtm.

maparent (approved):

packages/database/package.json
mdroidian: @CodeRabbit can you update this code?

mdroidian: @Coderabbit ok, please go ahead

PR #384, opened 2025-08-27
ENG-796 Canvas Hide Block CSS Fix

mdroidian:
We were using attrs on the child block, but these were being created after render, so not being hidden.

## Summary by CodeRabbit

* **Bug Fixes**
  * Correctly hides Roam blocks when a canvas is present by applying visibility rules to the parent container, fixing inconsistent hiding in the main page and sidebar.
  * Reliably marks the canvas root container to ensure consistent show/hide behavior and reduce layout flicker.

* **Chores**
  * Removed unused insertion logic and debug logging.


mdroidian:
@coderabbit full review

PR #392, opened 2025-08-29
ENG-808 Repair turbo dev

maparent:
https://linear.app/discourse-graphs/issue/ENG-808/repair-turbo-dev

On an empty repo, `turbo dev` fails   because `turbo dev -F @repo/database` expects `@repo/utils` to be built.
Adding the turbo dependencies to that effect.
Also making the database a non-dev dependency of website.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -61,7 +61,7 @@
       "persistent": true,
       "inputs": ["$TURBO_DEFAULT$", ".env*"],
       "dependsOn": [
-        "@repo/database#build",
+        "@repo/utils#build",
         "@repo/ui#build",
         "@repo/database#build"
       ]
--- a/turbo.json
+++ b/turbo.json
@@ -59,7 +59,12 @@
       "passThroughEnv": ["OBSIDIAN_PLUGIN_PATH"],
       "cache": false,
       "persistent": true,
-      "inputs": ["$TURBO_DEFAULT$", ".env*"]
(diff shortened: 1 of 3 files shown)


turbo.json
```diff
@@ -59,7 +59,12 @@
       "passThroughEnv": ["OBSIDIAN_PLUGIN_PATH"],
       "cache": false,
       "persistent": true,
-      "inputs": ["$TURBO_DEFAULT$", ".env*"]
+      "inputs": ["$TURBO_DEFAULT$", ".env*"],
+      "dependsOn": [
+        "@repo/database#build",
+        "@repo/ui#build",
+        "@repo/database#build"
```
mdroidian: Does `"@repo/database#build"` need to be in here twice?

mdroidian (approved):

mdroidian (commented):
Actually we should probably try to find a different way to do this.

Looks like this makes `turbo dev -F roam` build ui which should not be required.  The @repo/packages build should be dependent on if they are used, and they need to be watched and rebuilt.

turbo.json
maparent: Good catch. The first one was meant to be utils.

maparent:
Actually, good point about recursive dependencies. I set it up so dev depends on recursive build, which is more targeted.

mdroidian:
@maparent package.lock needs to be updated from this PR.

PR #393, opened 2025-08-29
ENG-804 Load Settings Bug

maparent:
https://linear.app/discourse-graphs/issue/ENG-804/load-settings-bug

Restoring the previous jsx setting fixes the bug. 
Not sure about the semantics of that flag, interesting that it's different in roam vs all other react projects.

I also found out that the issue centres on the loop at Settings.tsx:194-201. Removing those lines also allows the panel to appear the first time. Replacing the content of the loop with something trivial does not help.

### DIFF
--- a/apps/roam/tsconfig.json
+++ b/apps/roam/tsconfig.json
@@ -11,6 +11,7 @@
     "baseUrl": ".",
     "outDir": "dist",
     "allowJs": false,
+    "jsx": "react",
     "noUncheckedIndexedAccess": false
   }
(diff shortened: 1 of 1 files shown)


mdroidian (approved):
@maparent what was the reason for removing it?

maparent:
It looked like an artifact of previous changes. The inherited value is jsx-roam, and that is from the turbo configuration, so I thought I would go with the official value, and the side-effect was not immediately obvious.

PR #395, opened 2025-08-30
Update README.md

jsmorabito:
Added light/dark cover image


mdroidian (approved):

mdroidian:
@jsmorabito I just realized that the Question in this banner is purple.  Whenever you have a sec, could we change it to the default #99890e?  This is the default color we use for `[[QUE]]`. 👍

PR #401, opened 2025-08-31
ENG-812 Update of database cli tools

maparent:
https://linear.app/discourse-graphs/issue/ENG-812/update-of-database-cli-tools

Updating supabase cli, vercel and cucumber.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2338,7 +2338,7 @@
       "version": "0.1.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@repo/types": "*",
+        "@repo/database": "*",
         "@repo/ui": "*",
         "@sindresorhus/slugify": "^2.2.1",
         "@supabase/ssr": "^0.6.1",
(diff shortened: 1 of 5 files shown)


mdroidian (changes requested):
We should probably upgrade our node version seeing as cucumber 12 dropped node 18 support ... 

https://github.com/DiscourseGraphs/discourse-graph/blob/82f0e22142588f764a5441e11dc00ddf878bfce2/package.json#L19

maparent:
> We should probably upgrade our node version seeing as cucumber 12 dropped node 18 support ...
> 
> https://github.com/DiscourseGraphs/discourse-graph/blob/82f0e22142588f764a5441e11dc00ddf878bfce2/package.json#L19

Absolutely. We are using 20 in GitHub workflows, and @types/node is always set to 20. I had not noticed this one.
Same commit or another one?

mdroidian (approved):

PR #402, opened 2025-09-01
Roam: ENG-798 Node Tags breaks Color Highlighter

sid597:
https://www.loom.com/share/0310aa0d89b843cc8eb66f42ff9a2f39?sid=bc015e12-05d3-44e7-b84a-a983524d95b5

## Summary by CodeRabbit

* **Bug Fixes**
  * Node tag popup now appears only for valid tags.
  * UI stays in sync after navigation and configuration updates, reducing stale or missing tag info.

* **Refactor**
  * Centralized caching of discourse nodes to improve responsiveness and consistency across the app.

### DIFF
--- a/apps/roam/src/utils/initializeObserversAndListeners.ts
+++ b/apps/roam/src/utils/initializeObserversAndListeners.ts
@@ -44,6 +44,7 @@ import {
   findBlockElementFromSelection,
 } from "~/utils/renderTextSelectionPopup";
 import { renderNodeTagPopupButton } from "./renderNodeTagPopup";
+import { formatHexColor } from "~/components/settings/DiscourseNodeCanvasSettings";
 
 const debounce = (fn: () => void, delay = 250) => {
   let timeout: number;
@@ -107,6 +108,9 @@ export const initObservers = async ({
         for (const node of getDiscourseNodes()) {
           if (tag.toLowerCase() === node.tag?.toLowerCase()) {
             renderNodeTagPopupButton(s, node, onloadArgs.extensionAPI);
+            if (node.canvasSettings?.color) {
+              s.style.color = formatHexColor(node.canvasSettings.color);
+            }
             break;
           }
         }
--- a/apps/roam/src/utils/initializeObserversAndListeners.ts
+++ b/apps/roam/src/utils/initializeObserversAndListeners.ts
@@ -102,7 +102,15 @@ export const initObservers = async ({
     className: "rm-page-ref--tag",
     tag: "SPAN",
     callback: (s: HTMLSpanElement) => {
-      renderNodeTagPopupButton(s, onloadArgs.extensionAPI);
+      const tag = s.getAttribute("data-tag");
+      if (tag) {
+        for (const node of getDiscourseNodes()) {
+          if (tag.toLowerCase() === node.tag?.toLowerCase()) {
+            renderNodeTagPopupButton(s, node, onloadArgs.extensionAPI);
+            break;
+          }
+        }
+      }
     },
   });
 
--- a/apps/roam/src/components/CreateNodeDialog.tsx
+++ b/apps/roam/src/components/CreateNodeDialog.tsx
@@ -11,6 +11,7 @@ import getDiscourseNodes, {
 } from "~/utils/getDiscourseNodes";
 import { getNewDiscourseNodeText } from "~/utils/formatUtils";
 import MenuItemSelect from "roamjs-components/components/MenuItemSelect";
+import createBlock from "roamjs-components/writes/createBlock";
 
 export type CreateNodeDialogProps = {
   onClose: () => void;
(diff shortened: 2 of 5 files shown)


sid597:
@coderabbitai full review

sid597:
* **#402** <a href="https://app.graphite.dev/github/pr/DiscourseGraphs/discourse-graph/402?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#414](https://github.com/DiscourseGraphs/discourse-graph/pull/414) <a href="https://app.graphite.dev/github/pr/DiscourseGraphs/discourse-graph/414?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#424](https://github.com/DiscourseGraphs/discourse-graph/pull/424) <a href="https://app.graphite.dev/github/pr/DiscourseGraphs/discourse-graph/424?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/DiscourseGraphs/discourse-graph/402?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -45,6 +45,16 @@ import {
 } from "~/utils/renderTextSelectionPopup";
 import { renderNodeTagPopupButton } from "./renderNodeTagPopup";
 
+let discourseNodes: DiscourseNode[] = [];
```
mdroidian: Could we reuse `configTreeRef.nodes` from `discourseConfigRef.ts` and `refreshConfigTree` from `refreshConfigTree.ts` for this?

mdroidian (changes requested):

apps/roam/src/utils/initializeObserversAndListeners.ts
sid597: Yes we can, have to modify the `configTreeRef.nodes` to return `tag` and `type` as well currently it only returns `text` and `children`

sid597: Why don't we extract all the node info of `DiscourseNode` type? then we would be able to use `configTreeRef.nodes` everywhere we need `getDiscourseNodes()` 
```
const configTreeRef: {
  tree: RoamBasicNode[];
  nodes: { [uid: string]: { text: string; children: RoamBasicNode[] } };
} = { tree: [], nodes: {} };

..... 

export type DiscourseNode = {
  text: string;
  type: string;
  shortcut: string;
  tag?: string;
  specification: Condition[];
  backedBy: "user" | "default" | "relation";
  canvasSettings: {
    [k: string]: string;
  };
  // @deprecated - use specification instead
  format: string;
  graphOverview?: boolean;
  description?: string;
  template?: InputTextNode[];
};

```

mdroidian: What don't we extract that you think we should extract?

sid597: Ok this was me being a noob all the node data is in children

sid597: Ok this was me being a noob all the node data is in children, and we should use it, too much context switching to remember the infra

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -102,7 +103,22 @@ export const initObservers = async ({
     className: "rm-page-ref--tag",
     tag: "SPAN",
     callback: (s: HTMLSpanElement) => {
-      renderNodeTagPopupButton(s, onloadArgs.extensionAPI);
+      const tag = s.getAttribute("data-tag");
+      if (tag) {
+        for (const [type, { text, children }] of Object.entries(
+          configTreeRef.nodes,
```
mdroidian: Again, this works, but you are duplicating work for yourself.
If you use `getDiscourseNodes`, this already does the work for pulling in `configTreeRef`, splicing it, and returning you the formatted value.

Then you wouldn't need the object.entries, getsettingsvaluesfromtree, and you could use the already defined DiscourseNode type instead of creating a new type.  I bet you could do this in ~5 lines and if we make changes to `tag` upstream, you would need to remember to change them here. 👍

mdroidian (changes requested):

apps/roam/src/utils/initializeObserversAndListeners.ts
mdroidian: Sorry, I should have pointed to you `getDiscourseNodes` directly in the first place.

sid597: Sorry I misunderstood your previous round of review comment, I thought we wanted to use `configTreeRef` directly because `getDiscourseNodes()` extracts all the data and not needed for this 


image

mdroidian: Yeah I misspoke, sorry.  `configTreeRef` is the entire config tree structure, with the config tree of the discourse node types as well.<br><br>`refreshConfigTree()` queries and updates `configTreeRef`<br>`getDiscourseNodes()` returns the formatted `DiscourseNodes[]` from `configTreeRef` (`discourse-graph/nodes/`)<br>`getFormattedConfigTree()` returns the formatted discourse settings (`roam/js/discourse-graph"`) from `configTreeRef`<br><br>Both `getDiscourseNodes()/getFormattedConfigTree()`should be quite fast as they are just JSON manipulation/formatting and not querying.  `refreshConfigTree()`is the one that queries.<br><br>I would recommend using `getDisocourseNodes()`so we tap into the `DiscourseNode[]` type and have a single source of truth for settings data/management.

sid597: sgtm

mdroidian (approved):
👍

PR #404, opened 2025-09-01
ENG-816 deploy should depend on build

maparent:
https://linear.app/discourse-graphs/issue/ENG-816/deploy-should-depend-on-build

The deploy script relies on the compile script, but this only ensures local dependency, not recursive building of packages. Adding the dependency at the turbo level.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -64,6 +64,7 @@
     },
     "deploy": {
       "cache": false,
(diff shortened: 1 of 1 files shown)


mdroidian:
Changes started here: https://github.com/DiscourseGraphs/discourse-graph/pull/377

mdroidian (approved):

PR #407, opened 2025-09-01
ENG-815 DbDotenv.d.ts

maparent:
https://linear.app/discourse-graphs/issue/ENG-815/treat-dbdotenv-like-a-normal-ts-file-rather-than-separate-js-and-dts

There are circular dependencies within `packages/database`,
especially of `stepdefs.ts` on `dbDotEnv`, which is why I made it a mjs module.

The alternative is to not compile `stepdefs.ts`; cucumber can use it by reference.
Then `dbDotEnv` can then be a `.mts` module. We then have to require it everywhere.
Ensure `stepdef` and scripts are type-checked, now that they're not compiled.

## Summary by CodeRabbit

- New Features
  - No user-facing changes.
- Refactor
  - Standardized how environment utilities are loaded across apps for consistency and reliability.
  - Added explicit typings to environment utilities to improve type safety.
- Chores
  - Updated package exports to use built artifacts for more predictable builds.
  - Switched type-checking to a dedicated lint configuration and streamlined compiler includes to source files.
  - Introduced a separate TypeScript config for lint/type emission across scripts and features.

Impact: Stability and build consistency improvements with no changes to user workflows.


maparent:
@coderabbit review

maparent:
@coderabbit good call on typesVersions. Making types explicit in each require is anti-dry, not going that way. I did find the `as typof import` syntax interesting, but does not work everywhere... may be worth investigating. Also thinking about the impact of the shim.

maparent:
As stated, this seems to work for me; curious how it'll work for you.

mdroidian (changes requested):
After removing and all `dist`, running `npm i` (because package-lock.json it is off), both `turbo check-types` and `turbo dev -F roam` fail

image

image

mdroidian:
Build is failing too

image

maparent:
@mdroidian Ok. I tried re-running this, and @repo/database had an old typescript@4.1, imposed by vercel, which does not pick up on some settings. 
Bizarrely, overriding typescript@5 has no impact.
I updated vercel and obtained typescript@4.9, which does run.
I think... given that this PR will have to be updated to pnpm shortly, I think we should sit on this, hopefully overrides will be more sensible.

maparent:
Rebased on pnpm branch, should be  good to go.

packages/database/package.json
```diff
@@ -6,17 +6,18 @@
   "exports": {
     "./lib/*": "./src/lib/*.ts",
     "./dbDotEnv": {
-      "types": "./types/dbDotEnv.d.ts",
-      "import": "./src/dbDotEnv.mjs",
-      "default": "./src/dbDotEnv.mjs"
+      "types": "./dist/src/dbDotEnv.d.mts",
+      "import": "./dist/src/dbDotEnv.mjs",
+      "require": "./dist/src/dbDotEnv.mjs",
```
maparent, line 11: This line annoys me; according to coderabbit, I should create a cjs wrapper and use it here.
Removing the line (as I did in the original PR) breaks the build. So may look into wrapping, even though things do work as-is.

mdroidian:
Decided together with MAP to cancel.  If it comes back up again, we'll re visit then.

PR #409, opened 2025-09-02
ENG-823 - Remove Discourse Graph Template link from homepage

mdroidian:
Will be combined with lab-template graph then re-added 
cc @joelchan


mdroidian (commented):
@CodeRabbit full review

PR #423, opened 2025-09-07
ENG-831 - Refactor tldraw maximize to stop canvas jumping around

mdroidian:
Before:
We were updating state which would re-render tldraw.  We also are currently not handling/saving the camera (at the moment).
So this would jump the camera to the original/last save state on minimize or maximize.  Very jarring and poor UX.

After:
Instead of state we manipulate the DOM directly.  Additionally, we use some of tldraw's built in utilities to to move the camera based on the new viewport.  It's not 100% but it is much less jarring.  We could spend a bit more time finessing it, but it is now diminishing returns.  Sidebar was not fully handled, also diminishing returns.

https://www.loom.com/share/9ff85d293fb14f11af0d7d321b9e6a6e

If we want to do sidebar:
- rm-sidebar-outline-wrapper = position:static
- .rm-sidebar-window = position:static;
- .rm-sidebar-window = rm margin left/right

and hope they don't have 2 canvas' open in sidebar.

## Summary by CodeRabbit

* **New Features**
  * Added a maximize/fullscreen toggle for the canvas with smoother transitions and automatic viewport adjustment.

* **Bug Fixes**
  * Canvas now reliably fills the screen when maximized and restores the previous layout on exit.
  * Prevents layout shifting or overlap in the main content and sidebar during maximize/unmaximize.

### DIFF
--- a/apps/roam/src/components/canvas/Tldraw.tsx
+++ b/apps/roam/src/components/canvas/Tldraw.tsx
@@ -134,11 +134,13 @@ const TldrawCanvas = ({ title }: { title: string }) => {
 
   const updateViewportScreenBounds = (el: HTMLDivElement) => {
     // Use tldraw's built-in viewport bounds update with centering
-    const rect = el.getBoundingClientRect();
-    appRef.current?.updateViewportScreenBounds(
-      new Box(rect.left, rect.top, rect.width, rect.height),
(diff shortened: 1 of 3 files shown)


mdroidian:
@coderabbit full review

PR #427, opened 2025-09-09
ENG-833 Isolate node query functions.

maparent:
https://linear.app/discourse-graphs/issue/ENG-833/isolate-node-query-functions

## Summary by CodeRabbit

* **New Features**
  * Browse and select node types within a space; view a node type's related nodes (schema ↔ instances).
  * List and filter nodes by type with optional nested details (content and document) shown when available.
  * Faster, more reliable node-type discovery and fetching with caching and improved query behavior.

* **Chores**
  * Updated development tooling (TypeScript) for the database package.

### DIFF
--- a/packages/database/src/lib/queries.ts
+++ b/packages/database/src/lib/queries.ts
@@ -12,8 +12,8 @@ export type NodeSignature = { id: number; name: string };
 
 export const nodeSchemaSignature: NodeSignature = { id: 0, name: NODE_SCHEMAS };
 
-type PDocument = Partial<Tables<"Document">>;
-type PContent = Partial<Tables<"Content">> & {
+export type PDocument = Partial<Tables<"Document">>;
+export type PContent = Partial<Tables<"Content">> & {
   Document: PDocument | null;
 };
 export type PConcept = Partial<Tables<"Concept">> & {
@@ -23,7 +23,7 @@ export type PConcept = Partial<Tables<"Concept">> & {
 
 const composeQuery = ({
   supabase,
-  spaceId = undefined,
+  spaceId,
   schemaName = NODE_SCHEMAS,
   conceptFields = ["id", "name"],
   contentFields = [],
@@ -116,7 +116,7 @@ export const DOCUMENT_FIELDS: (keyof Document)[] = [
 
 export const getNodes = async ({
   supabase,
-  spaceId = undefined,
+  spaceId,
   schemaName = NODE_SCHEMAS,
   conceptFields = CONCEPT_FIELDS,
   contentFields = CONTENT_FIELDS,
--- a/packages/database/src/lib/queries.ts
+++ b/packages/database/src/lib/queries.ts
@@ -47,6 +47,10 @@ type defaultQueryShape = {
 };
 
 // Utility function to compose a generic query to fetch concepts, content and document.
+//  - schemaDbIds = 0  → fetch schemas (is_schema = true)
+//  - schemaDbIds = n  → fetch nodes under schema with dbId n (is_schema = false, eq schema_id)
+//  - schemaDbIds = [] → fetch all nodes (is_schema = false, no filter on schema_id)
+//  - schemaDbIds = [a,b,...] → fetch nodes under any of those schemas
 const composeQuery = ({
   supabase,
   spaceId,
@@ -234,6 +238,10 @@ export const DOCUMENT_FIELDS: (keyof Document)[] = [
 // get all nodes that belong to a certain number of schemas.
 // This query will return Concept objects, and associated Content and Document,
 // according to which fields are requested. Defaults to maximal information.
+// Main call options:
+// • ALL schemas:              schemaLocalIds = "__schemas" (default)
+// • ALL nodes (instances):    schemaLocalIds = []
+// • Nodes from X,Y schemas:   schemaLocalIds = ["localIdX","localIdY",...]
 export const getNodes = async ({
   supabase,
   spaceId,
--- a/packages/database/src/lib/queries.ts
+++ b/packages/database/src/lib/queries.ts
@@ -20,12 +20,14 @@ export type NodeSignature = {
 export const nodeSchemaSignature: NodeSignature = {
   spaceId: 0,
   sourceLocalId: NODE_SCHEMAS,
-  name: "Node types",
+  name: "Node Types",
   dbId: 0,
 };
 
-let NODE_SCHEMA_CACHE: Record<string, NodeSignature | number> = {
-  NODE_SCHEMAS: nodeSchemaSignature,
+type CacheMissTimestamp = number;
+type CacheEntry = NodeSignature | CacheMissTimestamp;
+let NODE_SCHEMA_CACHE: Record<string, CacheEntry> = {
+  [NODE_SCHEMAS]: nodeSchemaSignature,
 };
 
 export type PDocument = Partial<Tables<"Document">>;
@@ -74,8 +76,10 @@ const composeQuery = ({
     query = query.eq("is_schema", true);
   } else {
     query = query.eq("is_schema", false);
-    if (Array.isArray(schemaDbIds)) query = query.in("schema_id", schemaDbIds);
-    else if (typeof schemaDbIds === "number")
+    if (Array.isArray(schemaDbIds)) {
+      if (schemaDbIds.length > 0) query = query.in("schema_id", schemaDbIds);
+      // else we'll get all nodes
+    } else if (typeof schemaDbIds === "number")
       query = query.eq("schema_id", schemaDbIds);
     else throw new Error("schemaDbIds should be a number or number[]");
   }
@@ -86,12 +90,12 @@ const composeQuery = ({
 export const getNodeSchemas = async (
   supabase: DGSupabaseClient,
   spaceId: number,
-  forceReload: boolean = false,
+  forceCacheReload: boolean = false,
 ): Promise<NodeSignature[]> => {
   let result = Object.values(NODE_SCHEMA_CACHE)
     .filter((x) => typeof x === "object")
     .filter((x) => x.spaceId === spaceId);
-  if (forceReload || result.length === 0) {
+  if (forceCacheReload || result.length === 0) {
     const q = composeQuery({ supabase, spaceId });
     const res = (await q) as PostgrestResponse<defaultQueryShape>;
     if (res.error) {
@@ -121,16 +125,16 @@ export const getNodeSchemas = async (
 
 const INTERVAL = 5000;
 
-// Utility function: Get the databaseIds from the schema's Uids, looking first in Cache then querying db.
-const getDbIds = async (
+// Utility function: Get the mapping of local ids to db Ids, looking first in Cache then querying db.
+const getLocalToDbIdMapping = async (
   supabase: DGSupabaseClient,
-  localUids: string[],
+  localLocalIds: string[],
   spaceId?: number,
 ): Promise<Record<string, number | null>> => {
   // assuming no collision with types' local Ids.
-  const partialResult: [k: string, v: number | NodeSignature | undefined][] =
-    localUids.map((x) => [x, NODE_SCHEMA_CACHE[x]]);
-  const getDbId = (x: NodeSignature | number | undefined) =>
+  const partialResult: [k: string, v: CacheEntry | undefined][] =
+    localLocalIds.map((x) => [x, NODE_SCHEMA_CACHE[x]]);
+  const getDbId = (x: CacheEntry | undefined) =>
     typeof x === "object" ? x.dbId : null;
   let dbIds: Record<string, number | null> = Object.fromEntries(
     partialResult.map(([k, v]) => [k, getDbId(v)]),
@@ -153,7 +157,7 @@ const getDbIds = async (
     if (Object.keys(NODE_SCHEMA_CACHE).length > 1) {
       // Non-empty cache, query selectively
       q = q
-        .in("Content.source_local_id", localUids)
+        .in("Content.source_local_id", localLocalIds)
         .not("Content.source_local_id", "is", null);
     } // otherwise populate the cache
     const res = (await q) as PostgrestResponse<defaultQueryShape>;
@@ -175,12 +179,12 @@ const getDbIds = async (
         ]),
       ),
     };
-    for (const uid of localUids) {
-      if (typeof NODE_SCHEMA_CACHE[uid] !== "object")
-        NODE_SCHEMA_CACHE[uid] = now;
+    for (const localId of localLocalIds) {
+      if (typeof NODE_SCHEMA_CACHE[localId] !== "object")
+        NODE_SCHEMA_CACHE[localId] = now;
     }
     dbIds = Object.fromEntries(
-      localUids.map((x) => [x, getDbId(NODE_SCHEMA_CACHE[x])]),
+      localLocalIds.map((x) => [x, getDbId(NODE_SCHEMA_CACHE[x])]),
     );
   }
   return dbIds;
@@ -233,27 +237,34 @@ export const DOCUMENT_FIELDS: (keyof Document)[] = [
 export const getNodes = async ({
   supabase,
   spaceId,
-  schemaUids = NODE_SCHEMAS,
+  schemaLocalIds = NODE_SCHEMAS,
   conceptFields = CONCEPT_FIELDS,
   contentFields = CONTENT_FIELDS,
   documentFields = DOCUMENT_FIELDS,
 }: {
   supabase: DGSupabaseClient;
   spaceId?: number;
-  schemaUids?: string | string[];
+  schemaLocalIds?: string | string[];
   conceptFields?: (keyof Concept)[];
   contentFields?: (keyof Content)[];
   documentFields?: (keyof Document)[];
 }): Promise<PConcept[]> => {
   let schemaDbIds: number | number[] = 0;
-  const uidsArray = typeof schemaUids === "string" ? [schemaUids] : schemaUids;
-  if (schemaUids !== NODE_SCHEMAS) {
-    const dbIdsMap = await getDbIds(supabase, uidsArray);
-    schemaDbIds = Object.values(dbIdsMap).filter((x) => x !== null);
-    if (schemaDbIds.length < schemaUids.length) {
+  const localIdsArray =
+    typeof schemaLocalIds === "string" ? [schemaLocalIds] : schemaLocalIds;
+  if (schemaLocalIds !== NODE_SCHEMAS) {
+    const dbIdsMapping = await getLocalToDbIdMapping(
+      supabase,
+      localIdsArray,
+      spaceId,
+    );
+    schemaDbIds = Object.values(dbIdsMapping).filter((x) => x !== null);
+    if (schemaDbIds.length < localIdsArray.length) {
       console.error(
-        "Some uids are not yet in database: ",
-        uidsArray.filter((uid) => dbIdsMap[uid] === undefined).join(", "),
+        "Some localIds are not yet in database: ",
+        localIdsArray
+          .filter((localId) => dbIdsMapping[localId] === null)
+          .join(", "),
       );
     }
   }
--- a/packages/database/src/lib/queries.ts
+++ b/packages/database/src/lib/queries.ts
@@ -1,0 +1,146 @@
+import { PostgrestResponse } from "@supabase/supabase-js";
+import type { Tables } from "../dbTypes";
+import { DGSupabaseClient } from "./client";
+
+type Concept = Tables<"Concept">;
+type Content = Tables<"Content">;
+type Document = Tables<"Document">;
+
+export const NODE_SCHEMAS = "Node types";
+
+export type NodeSignature = { id: number; name: string };
+
+export const nodeSchemaSignature: NodeSignature = { id: 0, name: NODE_SCHEMAS };
+
+type PDocument = Partial<Tables<"Document">>;
+type PContent = Partial<Tables<"Content">> & {
+  Document: PDocument | null;
+};
+export type PConcept = Partial<Tables<"Concept">> & {
+  Content: PContent | null;
+  schema_of_concept: { name: string } | null;
+};
+
+const composeQuery = ({
+  supabase,
+  spaceId = undefined,
+  schemaName = NODE_SCHEMAS,
+  conceptFields = ["id", "name"],
+  contentFields = [],
+  documentFields = [],
+}: {
+  supabase: DGSupabaseClient;
+  spaceId?: number;
+  schemaName?: string | string[];
+  conceptFields?: (keyof Concept)[];
+  contentFields?: (keyof Content)[];
+  documentFields?: (keyof Document)[];
+}) => {
+  let q = conceptFields.join(",\n");
+  if (schemaName !== NODE_SCHEMAS) q += ",\nschema_of_concept(name)";
+  if (contentFields.length > 0) {
+    q += ",\nContent (\n" + contentFields.join(",\n");
+    if (documentFields.length > 0) {
+      q += ",\nDocument (\n" + documentFields.join(",\n") + ")";
+    }
+    q += ")";
+  }
+  let query = supabase.from("Concept").select(q).eq("arity", 0);
+  if (spaceId !== null) query = query.eq("space_id", spaceId);
+  if (schemaName === NODE_SCHEMAS) {
+    query = query.eq("is_schema", true);
+  } else {
+    query = query.eq("is_schema", false).not("schema_of_concept", "is", null);
+    if (Array.isArray(schemaName))
+      query = query.in("schema_of_concept.name", schemaName);
+    else if (typeof schemaName === "string")
+      query = query.eq("schema_of_concept.name", schemaName);
+    else throw new Error("schemaName should be a string or string[]");
+  }
+  return query;
+};
+
+export const getNodeSchemas = async (
+  supabase: DGSupabaseClient,
+  spaceId: number,
+): Promise<NodeSignature[]> => {
+  const q = composeQuery({ supabase, spaceId });
+  const res = (await q) as PostgrestResponse<NodeSignature>;
+  if (res.error) {
+    console.error("getNodeSchemas failed", res.error);
+    return [nodeSchemaSignature];
+  }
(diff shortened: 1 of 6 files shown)


maparent:
@coderabbit review

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,146 @@
...
+}) => {
+  let q = conceptFields.join(",\n");
+  if (schemaName !== NODE_SCHEMAS) q += ",\nschema_of_concept(name)";
+  if (contentFields.length > 0) {
+    q += ",\nContent (\n" + contentFields.join(",\n");
+    if (documentFields.length > 0) {
+      q += ",\nDocument (\n" + documentFields.join(",\n") + ")";
+    }
+    q += ")";
+  }
```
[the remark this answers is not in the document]
maparent: @coderabbit: The '!inner' marker does not work on computed relationship.

maparent:
@sid597 let me know if these functions would do what you want them to

packages/database/src/lib/queries.ts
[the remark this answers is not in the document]
maparent: No, I have nothing better to suggest.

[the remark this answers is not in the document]
maparent: @coderabbit that was for you. I have no better syntax to suggest than the one I used.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,146 @@
+import { PostgrestResponse } from "@supabase/supabase-js";
+import type { Tables } from "../dbTypes";
+import { DGSupabaseClient } from "./client";
+
+type Concept = Tables<"Concept">;
+type Content = Tables<"Content">;
+type Document = Tables<"Document">;
+
+export const NODE_SCHEMAS = "Node types";
```
mdroidian: nit: maybe all lowercase to reduce possible confusion/mixup, eg `node types`
but if they are always typed it likely won't be an issue

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export type NodeSignature = { id: number; name: string };
+
+export const nodeSchemaSignature: NodeSignature = { id: 0, name: NODE_SCHEMAS };
```
mdroidian: Are all of these exports intended/expected to be used outside of this function/file?
If not, let's not export them until we need to.  That way it reduces the cognitive load when looking at this file.  If I see an export, I'm expecting that I (or someone) should/will be using it somewhere.  If it isn't exported, I can immediately know it is only used right here.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,146 @@
...
+  "space_id",
+  "source_local_id",
+  "url",
+  "created",
+  "metadata",
+  "last_modified",
+  "author_id",
+];
+
+export const getNodes = async ({
```
mdroidian, line 245: Perhaps we should put each query in it's own file.  This will further reduce cognitive load.  Right now, with multiple query functions and multiple other exported variables, types,  functions, it's not immediately clear what actually are the queries. One has to scroll down and read each export.

If we had
`getNodes.ts` and
`getNodeSchemas.ts` 

This would be immediately clear right in the IDE.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,146 @@
+import { PostgrestResponse } from "@supabase/supabase-js";
+import type { Tables } from "../dbTypes";
```
mdroidian, line 2: ~

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,146 @@
...
+export type PDocument = Partial<Tables<"Document">>;
+export type PContent = Partial<Tables<"Content">> & {
+  Document: PDocument | null;
+};
+export type PConcept = Partial<Tables<"Concept">> & {
+  Content: PContent | null;
+  schema_of_concept: { name: string } | null;
+};
+
+const composeQuery = ({
```
mdroidian, line 54: This is a +1 for Drizzle/Zod.
String concatenation is error-prone and hard to maintain.
Zod would give us validation.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,146 @@
...
+  supabase,
+  spaceId,
+  schemaName = NODE_SCHEMAS,
```
mdroidian: Is this referring to `DiscourseNode.text` or `DiscourseNode.type`?
`text` is the label.  `type` is the `uid`.

https://github.com/DiscourseGraphs/discourse-graph/blob/84d54a88f1e019c61a5a14d0a847befb009696f5/apps/roam/src/utils/getDiscourseNodes.ts#L14-L35

mdroidian (approved):
I won't block on this, but I **strongly** suggest:
- only exporting what is intended to be used by **majority** consumers.  (eg: consumers of `getNodes` and `getNodeSchemas`.  If something else needs to be exported later for some other use case, we will export it at that time.
- split the queries into their own files.  As I mentioned, there's a lot of cognitive overhead just trying to find the queries right now
- add documentation to the queries.  LLM's are really good at this :)

packages/database/src/lib/queries.ts
maparent: For the record: This is exposed in the UI, as the option to get types instead of a specific type; hence the casing.

packages/database/src/lib/queries.ts
maparent: They are all used in the admin panel.

packages/database/src/lib/queries.ts
maparent: These are utility shortcuts, I don't think they deserve a file.

packages/database/src/lib/queries.ts
maparent: This function introduces safe typing of column names, using dbTypes.
Drizzle-Zod would give us typing over direct Postgres access, but not over this Postgrest query string which has its own syntax. Now I agree string manipulation is not ideal.

packages/database/src/lib/queries.ts
maparent: ?

packages/database/src/lib/queries.ts
maparent: The text, not the Uid.
Would you prefer a uid search?

maparent:
\-1 on putting utility stubs in their own files, but I hear you on the cognitive overhead and will add comments. As for the generic function, @sid597 told me he wanted to test against a set of types, which is not accessible from the stubs. This is a good case of BIKINI (But I Know I Need It) against YAGNI. 

The real question: How do we refer to types? Db ids probably require a round-trip (but should be cached globally), names are unique in Roam by design but certainly not across spaces, so yes Uids is probably better. Will change.

Not merging until we have agreement on that last point.

maparent:
Ah I remember why I did not use Uid. It's more fragile because of the missing content, and requires another join (which has a small performance impact.) My initial implementation using database Ids was actually the best solution…

maparent:
Ok, I have thought of something better...

packages/database/src/lib/queries.ts
mdroidian: The convention is to use \~ to go from root instead of ..

packages/database/src/lib/queries.ts
mdroidian: Ah, then I'd capitalize Types `Node Types`to treat as a proper term/category.

mdroidian:
Host application `uid` for the Discourse Node Type would probably be the best best.  Single trip.  Stable reference.

maparent:
And another join. I've cached the roundtrip instead, this will mean I avoid two joins, I think it's worth it.

maparent:
This has changed enough that I think it's worth another review.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+    else throw new Error("schemaDbIds should be a number or number[]");
+  }
+  return query;
+};
+
+// Obtain basic data for all node schemas in a space, populating the cache.
+export const getNodeSchemas = async (
+  supabase: DGSupabaseClient,
+  spaceId: number,
+  forceReload: boolean = false,
```
mdroidian: nit: Consider a more explicit name, e.g. `skipCache`, which makes the intent clearer and avoids ambiguity.  It is not clear `forceReload` is talking about cache.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+  const previousMisses = Object.fromEntries(
+    partialResult.filter(([k, v]) => typeof v === "number"),
+  ) as Record<string, number>;
+  const numPreviousMisses = Object.values(previousMisses).length;
+  const now = Date.now();
+  const oldestMiss =
+    numPreviousMisses > 0 ? Math.min(...Object.values(previousMisses)) : now;
+  if (numMissing > numPreviousMisses || now - oldestMiss > INTERVAL) {
+    if (spaceId === undefined) {
+      console.error("Cannot populate cache without spaceId");
```
mdroidian: If spaceId is optional, this shouldn't be an error.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+): Promise<Record<string, number | null>> => {
+  // assuming no collision with types' local Ids.
+  const partialResult: [k: string, v: number | NodeSignature | undefined][] =
+    localUids.map((x) => [x, NODE_SCHEMA_CACHE[x]]);
+  const getDbId = (x: NodeSignature | number | undefined) =>
+    typeof x === "object" ? x.dbId : null;
+  let dbIds: Record<string, number | null> = Object.fromEntries(
+    partialResult.map(([k, v]) => [k, getDbId(v)]),
+  );
+  let numMissing = Object.values(dbIds).filter((x) => x === null).length;
```
mdroidian: use `const` when not reassigning.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+// Utility function: Get the databaseIds from the schema's Uids, looking first in Cache then querying db.
+const getDbIds = async (
+  supabase: DGSupabaseClient,
+  localUids: string[],
+  spaceId?: number,
+): Promise<Record<string, number | null>> => {
+  // assuming no collision with types' local Ids.
+  const partialResult: [k: string, v: number | NodeSignature | undefined][] =
+    localUids.map((x) => [x, NODE_SCHEMA_CACHE[x]]);
+  const getDbId = (x: NodeSignature | number | undefined) =>
```
mdroidian: Is the intent here to return `null` when a raw number is passed in? If so, the function name could be more explicit (e.g. `getDbIdFromNodeSignature`) to make it clear that it only extracts IDs from NodeSignature objects.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+  spaceId?: number;
+  schemaUids?: string | string[];
+  conceptFields?: (keyof Concept)[];
+  contentFields?: (keyof Content)[];
+  documentFields?: (keyof Document)[];
+}): Promise<PConcept[]> => {
+  let schemaDbIds: number | number[] = 0;
+  const uidsArray = typeof schemaUids === "string" ? [schemaUids] : schemaUids;
+  if (schemaUids !== NODE_SCHEMAS) {
+    const dbIdsMap = await getDbIds(supabase, uidsArray);
```
mdroidian: nit: this is actually a plain object, not a Map

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+  "author_id",
+];
+
+// get all nodes that belong to a certain number of schemas.
+// This query will return Concept objects, and associated Content and Document,
+// according to which fields are requested. Defaults to maximal information.
+export const getNodes = async ({
+  supabase,
+  spaceId,
+  schemaUids = NODE_SCHEMAS,
```
mdroidian: Is schemaUids referring to db Uids here?
If so, we have a cache issue.  How can we refresh the cache when a new node type is added?  I see there is a flag, but that means the consumer, when just wanting to retrieve nodes, will have to account for cache refreshing of schemas.

Also, if so, it seems like schemaUids is referring to local Uid's on ln 124, which is confusing if they are both referring to different things.

mdroidian: Also,  I guess this assumes/requires using `getNodeSchemas`?

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+};
+
+const INTERVAL = 5000;
+
+// Utility function: Get the databaseIds from the schema's Uids, looking first in Cache then querying db.
+const getDbIds = async (
+  supabase: DGSupabaseClient,
+  localUids: string[],
+  spaceId?: number,
+): Promise<Record<string, number | null>> => {
```
mdroidian, line 137: What is the expected value for `<string,number>` here?
Is number the dbId?  If so, what is `string`?

Let's put that in the comment as well.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+};
+
+export const nodeSchemaSignature: NodeSignature = {
+  spaceId: 0,
+  sourceLocalId: NODE_SCHEMAS,
+  name: "Node types",
+  dbId: 0,
+};
+
+let NODE_SCHEMA_CACHE: Record<string, NodeSignature | number> = {
```
mdroidian: It is not clear what `number` is supposed to be here.

Maybe something more clear like

```typescript
type CacheEntry = NodeSignature | CacheMissTimestamp;
type CacheMissTimestamp = number; // Timestamp when UID was not found in DB
```

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+  conceptFields?: (keyof Concept)[];
+  contentFields?: (keyof Content)[];
+  documentFields?: (keyof Document)[];
+}): Promise<PConcept[]> => {
+  let schemaDbIds: number | number[] = 0;
+  const uidsArray = typeof schemaUids === "string" ? [schemaUids] : schemaUids;
+  if (schemaUids !== NODE_SCHEMAS) {
+    const dbIdsMap = await getDbIds(supabase, uidsArray);
+    schemaDbIds = Object.values(dbIdsMap).filter((x) => x !== null);
+    if (schemaDbIds.length < schemaUids.length) {
```
mdroidian: `schemaUids` might be string, so you would be counting the length of that string

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+  "created",
+  "metadata",
+  "last_modified",
+  "author_id",
+];
+
+// get all nodes that belong to a certain number of schemas.
+// This query will return Concept objects, and associated Content and Document,
+// according to which fields are requested. Defaults to maximal information.
+export const getNodes = async ({
```
mdroidian, line 245: So to "get all nodes" it looks like we must first get all schemas uids and pass them all?  Ideally we could pass no schema uid's which would default to "get all nodes".  Currently it looks like passing no schema uid's will result in returning all schema's, which is what `getNodeSchemas` is for.

mdroidian (changes requested):
A few critical changes:
- [Cache key mismatch](https://github.com/DiscourseGraphs/discourse-graph/pull/427/files#r2342846929)
- [Length check bug](https://github.com/DiscourseGraphs/discourse-graph/pull/427#discussion_r2342809017)
- [NODE_SCHEMA_CACHE clearer types](https://github.com/DiscourseGraphs/discourse-graph/pull/427#discussion_r2342798699)

Then maybe more depending on answers to some of the questions.

Also, I'll call coderabbit for review, let's run through that as well.

mdroidian:
@CodeRabbit full review

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+
+export const nodeSchemaSignature: NodeSignature = {
+  spaceId: 0,
+  sourceLocalId: NODE_SCHEMAS,
+  name: "Node types",
+  dbId: 0,
+};
+
+let NODE_SCHEMA_CACHE: Record<string, NodeSignature | number> = {
+  NODE_SCHEMAS: nodeSchemaSignature,
```
mdroidian: `NODE_SCHEMAS` is defined as `export const NODE_SCHEMAS = "__schemas"`; but initialize the cache with a literal "NODE_SCHEMAS" key.  Then later it is read with `NODE_SCHEMA_CACHE[NODE_SCHEMAS]`, which looks up "__schemas". That entry won’t exist.

packages/database/src/lib/queries.ts
```diff
@@ -0,0 +1,274 @@
...
+  "space_id",
+  "arity",
+  "literal_content",
+  "reference_content",
+  "refs",
+  "is_schema",
+  "represented_by_id",
+];
+
+export const CONTENT_FIELDS: (keyof Content)[] = [
```
mdroidian, line 213: if the goal is all of the fields you could use 
`export const CONCEPT_FIELDS = Object.keys({} as Concept) as (keyof Concept)[];`

packages/database/src/lib/queries.ts
maparent: will use `forceCacheReload`, it's more than a skip.

packages/database/src/lib/queries.ts
maparent: It's ok to leave out spaceId if you know the cache is populated.

packages/database/src/lib/queries.ts
maparent: reassigned at line 182

packages/database/src/lib/queries.ts
maparent: The intent is to get dbId from local UIds, renaming to `getDbIdFromUids`. It extracts from a cache of NodeSignatureObjects, and maintains the cache; not sure how that's "only extracting". `FromNodeSignature` would read to me that the signatures are given, not maintained, no?

packages/database/src/lib/queries.ts
maparent: Map as in the concept: MAPping one set of values to another. The collision with the JS Map type is unfortunate, but I think the concept is prior to the implementation. I don't think calling it Object would help understand.
Now, I could call it Mapping, would that be clearer to you?

packages/database/src/lib/queries.ts
maparent: Actually decided to rename: getLocalToDbIdMapping

packages/database/src/lib/queries.ts
maparent: did that.

packages/database/src/lib/queries.ts
maparent: dbIds are numeric, I would not call them Uids. Decided to rename to localId for clarity.
This refers to local (eg roam) id, and the job of all that code is to maintain the cache transparently for you.
As such, you do not need to call getNodeSchemas beforehand., unless you want the list for other purposes (as the panel does.)

packages/database/src/lib/queries.ts
maparent: Hopefully clearer now.

packages/database/src/lib/queries.ts
maparent: That is indeed much clearer, thank you.

packages/database/src/lib/queries.ts
maparent: good catch!

packages/database/src/lib/queries.ts
maparent: Ouch. Curious how that did work...

packages/database/src/lib/queries.ts
maparent: Ah, I wish! Tried that and got an empty array.

packages/database/src/lib/queries.ts
maparent: Ah, did not account for the case where we'd want to get all nodes. 
Passing an  empty array would yield nothing, setting it to return all (non-schema) nodes.

packages/database/src/lib/queries.ts
maparent: English title case kills me...

packages/database/src/lib/queries.ts
maparent: Ah did this and it caused trouble. Actually we cannot use that convention in packages, because when a package file is transcluded from an app, ~ still refers to the app's root. (I tried adding it as a path in database/tsconfig, no go.)

maparent: SO we have to enforce the reverse convention in packages...

maparent: (I _could_ use `@repo/database/dbTypes` but that feels silly when `../dbTypes` is clearer and shorter.)

packages/database/src/lib/queries.ts
mdroidian: Then let's change it to log or warning, not error.

packages/database/src/lib/queries.ts
mdroidian: @maparent Why is this the case with packages/db but not packages/UI or packages/util

packages/database/src/lib/queries.ts
mdroidian: @maparent Where do you see this being reassigned?

image

image

packages/database/src/lib/queries.ts
mdroidian: @maparent 
> setting it to return all (non-schema) nodes.

Hmm, how so?

It looks like it is still the case that passing `schemaLocalIds` as `null` or `[]` would return all schemas and no nodes.

In both cases `composeQuery` is called with `schemaDbIds = 0` and sets ` .eq("is_schema", true)`

mdroidian (approved):
There's a few "re-opened" comments that should be address, but I'll approve so it can be merged after.

packages/database/src/lib/queries.ts
maparent: sorry it moved to line 186 with other changes.

packages/database/src/lib/queries.ts
maparent: Null is not a valid input value and should error on the map at line 136.
An empty array should map to another empty array (not 0) when calling composeQuery, we'll get is_schema=false (78) and no other condition, so all nodes.

maparent:
pushed a nit, waiting for your response to my latest comments.

packages/database/src/lib/queries.ts
mdroidian: I still don't see it.  Can you show me what you see?

Here's my IDE
image

Here's GitHub
image

maparent: Argh! I thought the let you targeted was dbIds, not numMissing. Corrected.

packages/database/src/lib/queries.ts
mdroidian: Sorry instead of `null` I meant not being passed.

There's a lot of mental gymnastics one has to do to figure out they should pass `[]` to get all nodes.
If we are going to keep the code like this, let's make that explicit in the comment.

mdroidian:
responded

packages/database/src/lib/queries.ts
mdroidian: Examples:

```
/**
 * ──────────────────────────────────────────────────────────────────────────────
 * getNodeSchemas
 *    - Query all node schemas for a space (and cache them)
 * ──────────────────────────────────────────────────────────────────────────────
 */
 ```
 ```
/**
 * ──────────────────────────────────────────────────────────────────────────────
 * getNodes
 *    - Query:
 *        • ALL schemas:              schemaLocalIds = "__schemas" (default)
 *        • ALL nodes (instances):    schemaLocalIds = []
 *        • Nodes from X,Y schemas:   schemaLocalIds = ["localA","localB",...]
 * ──────────────────────────────────────────────────────────────────────────────
 */
 ```
 ```
 /**
 * ──────────────────────────────────────────────────────────────────────────────
 * composeQuery
 * ──────────────────────────────────────────────────────────────────────────────
 *
 *  - schemaDbIds = 0  → fetch schemas (is_schema = true)
 *  - schemaDbIds = n  → fetch nodes under schema n (is_schema = false, eq schema_id)
 *  - schemaDbIds = [] → fetch all nodes (is_schema = false, no filter on schema_id)
 *  - schemaDbIds = [a,b,...] → fetch nodes under any of those schemas
 */
 ```

packages/database/src/lib/queries.ts
maparent: There is no `import { ... } from "~/..."` in either of those.

packages/database/src/lib/queries.ts
maparent: Note: I remember when you told me to not put so many comments to my functions!

mdroidian: Ideally the functions can be written and named so their use is clear and obvious.  But when they are overloaded like this, documentation is required.

packages/database/src/lib/queries.ts
mdroidian: Oh that's right, we used `@repo/*` until recently<br>[https://github.com/DiscourseGraphs/discourse-graph/pull/377/files#diff-afc3d44b435d66640d3a034fbae79c1b28ebcb18dfef03ede24a247aa0a3a083L1](https://github.com/DiscourseGraphs/discourse-graph/pull/377/files#diff-afc3d44b435d66640d3a034fbae79c1b28ebcb18dfef03ede24a247aa0a3a083L1)<br><br>I would prefer `@repo/*` to be closer to the convention, and therefor more uniform, easier autocomplete, and seamless file transfers.  Definitely in the case of `../../` but I won't block on `../`

mdroidian: Oh that's right, we used @repo/* until recently
https://github.com/DiscourseGraphs/discourse-graph/pull/377/files#diff-afc3d44b435d66640d3a034fbae79c1b28ebcb18dfef03ede24a247aa0a3a083L1

I would prefer @repo/* to be closer to the convention, and therefor more uniform, easier autocomplete, and seamless file transfers.  Definitely in the case of ../../ but I won't block on ../

PR #437, opened 2025-09-11
Eng 815 rebased, test

maparent:
There are circular dependencies within `packages/database`, especially of stepdefs.ts on dbDotEnv, which is why I made it a mjs module.

The alternative is to not compile stepdefs; cucumber can use it by reference. Then dbDotEnv can then be a mts module. We then have to require it everywhere. Ensure stepdef and scripts are type-checked, now that they're not compiled.


maparent:
works, reset the original branch, deleting this.

PR #442, opened 2025-09-15
ENG-842 change node creation behaviour from node tags

sid597:
https://www.loom.com/share/dbe9595a65ee4f08a0c6e91b574a38fd?sid=9b93e218-7339-4782-95fe-9033fee0407c

PR #443, opened 2025-09-15
ENG-836 make button like css for node tag

sid597:
image

## Summary by CodeRabbit

* **New Features**
  * Tag popups and canvas nodes now use an improved dynamic color theming that derives coordinated background, text, border, and contrast for better readability.
* **Style**
  * Visual refinements to tag popups: improved padding, rounded corners, subtle shadow, refined font sizing/weight, spacing, and cursor cues for clearer interaction.


sid597:
* **#443** <a href="https://app.graphite.dev/github/pr/DiscourseGraphs/discourse-graph/443?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DiscourseGraphs/discourse-graph/443?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#442** <a href="https://app.graphite.dev/github/pr/DiscourseGraphs/discourse-graph/442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sid597:
@coderabbitai full review

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -100,6 +101,39 @@ export const initObservers = async ({
...
+    const textColor =
+      backgroundLuminance > 0.5
+        ? baseLuminance > 0.4
+          ? tinycolor(hexColor).darken(40).toHexString()
+          : hexColor
+        : "#ffffff";
+
+    return {
+      text: textColor,
+
```
mdroidian: nit: remove unnecessary spaces

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -100,6 +101,39 @@ export const initObservers = async ({
     render: (b) => renderQueryBlock(b, onloadArgs),
   });
 
+  const generateTagColorScheme = (hexColor: string) => {
+    const base = tinycolor(hexColor);
+    const hsl = base.toHsl();
+
+    const backgroundColor = tinycolor({
```
mdroidian: We should be using an `npm` package to generate a WCAG compliant color based on the `hexColor`.  It looks like this is just setting a static color based on the initial `hexColor`, is that correct?

mdroidian (changes requested):
I think many colors will still fail using this.

image


Try the `contrast-color` package (we use it in DiscourseNodeUtil) or a package like it to generate the corresponding background that is WCAG compliant.

Make sure to test them with https://webaim.org/resources/contrastchecker/

packages/utils/src/getContrastingColor.ts
```diff
@@ -0,0 +1,46 @@
+import { Colord, colord, extend } from "colord";
```
mdroidian: `colord` is unused

apps/roam/src/components/canvas/DiscourseNodeUtil.tsx
```diff
@@ -354,7 +355,8 @@ export class BaseDiscourseNodeUtil extends ShapeUtil<DiscourseNodeShape> {
     const backgroundColor = formattedBackgroundColor
       ? formattedBackgroundColor
       : COLOR_PALETTE[paletteColor];
-    const textColor = ContrastColor.contrastColor({ bgColor: backgroundColor });
+    const textColor = getContrastingColor(colord(backgroundColor))
+      .secondary as string;
```
mdroidian: Image

mdroidian (changes requested):
While this does look good on our `discourse-graphs` graph

Image

The default shipped node colors turn out to be mostly black or white backgrounds :(

Image

Can you think of any tweaks we could make so that these are compliant colored backgrounds?

If not, a couple other options could be
- manually set the foreground/background for the default shipped nodes (not ideal)
- change the default shipped colors (also not ideal)

### Swap the main color

Another idea would be to use the node color (the color that is set in the node setting) as the background color instead of the text color.  Then calculate the text color based on that.

That way the fallback would be that the text color is black or white, not the background color.  This would be much cleaner than a bunch of black/white background tags.

I think we should do both:
- swap the main color to be background instead of text
- test `discourse-graphs` and the default shipped node colors and aim for 75% to not be black/white

Let me know what you think

sid597:
@mdroidian can we change the default text color? 

this is what the colors look like (in `plugin-testing-akamatsulab`) when we use the text color as background (swapped from before) 

image

and in default plugin 

image

mdroidian:
> @mdroidian can we change the default text color? 
> 
> this is what the colors look like (in `plugin-testing-akamatsulab`) when we use the text color as background (swapped from before) 
> 
> image
> 
> and in default plugin 
> 
> image
> 

This looks better.  Have you tried tweaking the algorithm at all?

sid597:
no the tweak we can do is make the steps smaller, currently we use `0.1` I changed it to `0.01` but it looked the same. I just swapped the text with background. 

By steps I mean if we don't find a contrasting color then either lighten or darken the current color by `x` amount smaller number means its more granular but more iterations

mdroidian:
[ENG-836 make button like css for node tag - 28 September 2025](https://www.loom.com/share/f7ff604746bd4f2989cbe991b0c8f9a6)

[https://github.com/mdroidian/color-contrast-button-test](https://github.com/mdroidian/color-contrast-button-test)

mdroidian (changes requested):
Looks like this PR needs a rebase and the new "pleasing" algorithm, is that correct @sid597 ?

sid597:
@mdroidian now using the pleasing algorithm but there are lint issues related to colord

```
import { colord, extend, type Colord } from "colord";
import a11yPlugin from "colord/plugins/a11y";
import mixPlugin from "colord/plugins/mix";
extend([a11yPlugin, mixPlugin]);

Unsafe call of an `error` type typed value.eslint[@typescript-eslint/no-unsafe-call](https://typescript-eslint.io/rules/no-unsafe-call)

....
const t = colord({ h: 0, s: 0, l: mid }); // neutral gray
const cr = bg.contrast(t);

Unsafe call of an `error` type typed value.eslint[@typescript-eslint/no-unsafe-call](https://typescript-eslint.io/rules/no-unsafe-call)
```

sid597:
@mdroidian I did not rename the function or removed the comments from the pleasing algorithm file I think for this case the all comments make sense.

package.json
```diff
@@ -30,6 +30,7 @@
     "tailwindConfig": "./packages/tailwind-config/tailwind.config.ts"
   },
   "dependencies": {
-    "posthog-js": "^1.261.8"
+    "posthog-js": "^1.261.8",
+    "colord": "^2.9.3"
```
mdroidian: Let's add this to the specific packages that use it.  That is the convention with turbo actually.  (we should move posthog-js too, but we can do that later).

Also, remove `contrast-color` from roam's package lock if possible.

mdroidian (approved):
Just move `colord` to apps/roam and packages/util and remove `contrast-color` then good to go! 👍

mdroidian:
Actually let's run the pleasing algo through code rabbit too.

@CodeRabbitai full review

package.json
sid597: Done 🙌🏽

sid597:
cc @mdroidian merging

PR #450, opened 2025-09-18
ENG-453 - filter out empty branches in OR queries

mdroidian:
This specifically targets when a branch references `is a` `someNode` where `someNode` doesn't exists.

image
image

Previous behavior was to silently error (console.error), returning no results.
image

## Summary by CodeRabbit

- Bug Fixes
  - Improved reliability of advanced filters when using OR conditions. The app now correctly ignores empty or invalid branches (e.g., references to non-existent items) without causing unexpected empty results.
  - More consistent query outcomes when some filter criteria are omitted or become invalid, reducing edge-case failures and improving result accuracy.
  - Enhanced stability in complex searches by ensuring only meaningful conditions are evaluated, minimizing unexpected behavior in filtered views.


mdroidian:
@CodeRabbit full review

PR #461, opened 2025-09-27
ENG-907 Update to supabase client library

maparent:
https://linear.app/discourse-graphs/issue/ENG-907/update-to-supabase-client-library

### DIFF
--- a/apps/website/app/utils/supabase/client.ts
+++ b/apps/website/app/utils/supabase/client.ts
@@ -10,5 +10,5 @@ export const createClient = () => {
   if (!url || !key) {
     throw new Error("Missing required Supabase environment variables");
   }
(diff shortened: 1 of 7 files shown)


maparent:
Our monthly updates, focusing on client libraries.
Bizarrely, trying to update the supabase CLI fails silently. I hope pnpm will solve that.

mdroidian (commented):
> Our monthly updates, focusing on client libraries. Bizarrely, trying to update the supabase CLI fails silently. I hope pnpm will solve that.

Could we put this in the main body of the PR next time.

With all of the GitHub apps, the body gets quite noisy, so the main information should be the first thing someone sees.

mdroidian (approved):
package-lock is not up to date.  Could you run `npm i` again before merging.

Image

PR #474, opened 2025-10-01
Eng 898 create database query functions for base dg queries

maparent:
Base database functions for dg queries. (WIP)

## Summary by CodeRabbit

- New Features
  - Expanded concept querying with filters for author, relation type, related node type, and specific related nodes.
  - Ability to fetch related concepts/relations; concept results now include schema_id.
- Database
  - Added relation access functions to traverse from a concept to its relations and vice versa.
- Tests
  - Introduced comprehensive BDD scenarios covering concept upsert and complex queries.
  - Added authenticated steps for querying nodes and upserting accounts within a space.
- Refactor
  - Streamlined test login/setup and query composition to support advanced filtering.


maparent:
@coderabbit review

packages/database/features/step-definitions/stepdefs.ts
```diff
@@ -3,6 +3,7 @@ import { Given, When, Then, world, type DataTable } from "@cucumber/cucumber";
 import { createClient } from "@supabase/supabase-js";
```
mdroidian, line 4: This file has quite a few lints.  Would love it if we could clean them up or disable them if they do not intend to be fixed.

packages/database/src/lib/queries.ts
```diff
@@ -58,24 +64,88 @@ const composeQuery = ({
   conceptFields = ["id", "name", "space_id"],
```
mdroidian: Same issue with this file as with `stepdefs.ts`  
This file has quite a few lints. Let's take some time to clean the lints up.

If we need to add additional rules, please suggestion them.
If they are unfixable, add the appropriate comments to disable them.

packages/database/features/contentAccess.feature
```diff
@@ -17,7 +17,7 @@ Feature: Content access
     And the user user3 opens the Roam plugin in space s1
     And the user user3 opens the Roam plugin in space s2
     And Document are added to the database:
-      | @id | _space_id | source_local_id | _author_id | created    | last_modified |
+      | $id | _space_id | source_local_id | _author_id | created    | last_modified |
```
mdroidian, line 20: Looks like this didn't get prettier formatted.  It is helpful to keep up the format with each pr/change because when it is missed it adds a lot of noise to the subsequent pr that applies it.

mdroidian: Actually it looks like this is the case with most of the files in this PR.

Please go through and ensure each file is formatted.

mdroidian (changes requested):
Ok, there's a lot going on here.  I've made this as `Request changes` just so it gets back to you and tagged in linear when re-requested for review.

Could you help me understand what the purpose of this PR is?

I see the title "create database query functions for base dg queries" and in the linear ticket there is a comment with list of 7 queries.  Yet in this PR I do not see any new queries nor any reference to how these changes are related or meant to apply to those 7 queries.

Is this just the base and specific queries will be created in the future?  If so, is there an additional linear ticket for that?  Or are the dev's supposed to use this base query and be able to construct all 7 required queries when they need to use them?  Is documentation on these additional parameters forthcoming?

If it is the case that this base query is to be used by other dev's to construct a query, could you give some examples of what that would look like for a few examples:
- discourse context: query all nodes in any relation with node x, and the relation label
- query all discourse nodes of type x
- query all discourse relations containing specific discourse node x

packages/database/features/step-definitions/stepdefs.ts
maparent: Did clean up.

packages/database/features/contentAccess.feature
maparent: Ok, I have to look into why prettier is not auto-applied.
In the case of the feature files, the zed formatter for gerkhin (the cucumber .feature file format) fights with prettier, and wants to right-align dates, which is... annoying but not wrong?
I will leave that part as-is for now to avoid churn.
But did apply prettier elsewhere.

maparent:
As to your main comment:
The examples for the requested queries are in the queryConcept.feature file.
There are a few combinations of factors that require inferring from the examples in the files, but each individual factor is represented. We can discuss later today.

packages/database/features/contentAccess.feature
mdroidian: Prettier should be applied on save.
When is the zed formatter for gerkhin applied?  If it is also applied on save, I would try to turn it off so we can have a single primary formatter across all files.

maparent: Yes, I'll have to dig into zed configurations. Trying to punt this.

maparent: got zed to auto-prettify ts files, which is good. It's still unwilling to use prettier on gherkin.

mdroidian: Do you need to add something like this to zed's settings file?<br>[https://www.npmjs.com/package/prettier-plugin-gherkin/v/0.0.1](https://www.npmjs.com/package/prettier-plugin-gherkin/v/0.0.1)

maparent:
@mdroidian  lmk if you feel the documentation is adequate now.

packages/database/doc/tests.md
```diff
@@ -0,0 +1,33 @@
+# Designing cucumber tests
```
mdroidian, line 1: The following files still add changes when they are saved:
- tests.md
- contentAccess.feature
- queryConcepts.feature
- README.md

Can you confirm this is the same for you?
Image

mdroidian (approved):
I still think there’s quite a bit of room for improvement from a DX perspective, but I won’t block on it.

For example — when I’m coding a component and need nodes of type x quickly, I’ll type in `getConcepts` (which already requires a small context switch), and then I’m greeted with this:

Image

At that point, I find myself asking: “Which params do I actually need to query nodes of type x?” To figure that out, I have to open `getConcepts` and read through a long block of text — sorting through relation-specific params just to find what I need. The new documentation definitely helps with this, but I still think we could simplify things quite a bit. It might make sense to split it into a few smaller, more focused queries down the road.

That said, this is totally fine as a starting point.

packages/database/doc/tests.md
maparent: Hmm... Somehow, Zed's enforced prettier settings don't agree with our prettier settings. Sigh.
Did apply prettier, will have to dig deeper into settings.

maparent:
Re the number of arguments: I understand that is not ideal to have 17 parameters, only 4 of which would be recurrent across versions, but think of the combinatorics of having small functions for any subset of the other 13!

mdroidian:
> Re the number of arguments: I understand that is not ideal to have 17 parameters, only 4 of which would be recurrent across versions, but think of the combinatorics of having small functions for any subset of the other 13!

Good point about avoiding combinatorics. I’m not advocating a pile of tiny functions—just clearer grouping and a couple of thin presets so the common queries don’t require reading the entire param wall.

This is also probably a good time to consider adding zod.

maparent:
Hmmm... I do not see the need for this but you do; I think this is worth a synchronous co-design session.
Or you could just add functions with less parameters that call the big function.
Also not sure Zod would help, but you may have something in mind that I don't see.

mdroidian:
Here's an example: https://github.com/DiscourseGraphs/discourse-graph/pull/488

Zod would help catch bad combinations.

We can discuss in the dev meeting today.

PR #476, opened 2025-10-03
ENG-932 - add {current user} to "with text" condition

mdroidian:
Intentionally left `targetOptions` blank, seems like it would just be confusing as it is the only option and 99% of input is raw text, but I'm 50/50 with this decision.

## Summary by CodeRabbit

- New Features
  - “With text” filters now support the {current user} tag, matching content that includes your name.
- Bug Fixes
  - Improved accuracy and consistency of “with text” filtering across plain text and regex cases, reducing false positives/negatives.
- Performance
  - Faster evaluations when using {current user} in “with text” filters by avoiding unnecessary pattern matching.


mdroidian:
@CodeRabbit full review

PR #477, opened 2025-10-03
ENG-931 - Node Menu fixes

mdroidian:
https://www.loom.com/share/4e4518added040f4a6dc232d1a3cb996

- fix new bug: cannot add node tag
- fix old bug: creating node doesn't open in sidebar
- rm body.click on escape so user can keep editing block
- and small refactor

## Summary by CodeRabbit

- Bug Fixes
  - Resolved timing issues that occasionally caused node creation or tag insertion to fail or appear delayed.
  - Eliminated focus-related glitches when creating nodes from the menu.
  - Improved consistency of Arrow Up/Down, Enter, and Escape behavior in keyboard navigation.
  - Shortcut changes now apply immediately without requiring a reload.

- Refactor
  - Streamlined internal timing and event handling to make actions more reliable and responsive, reducing intermittent UI hiccups during node creation and tag insertion.

### DIFF
--- a/apps/roam/src/components/DiscourseNodeMenu.tsx
+++ b/apps/roam/src/components/DiscourseNodeMenu.tsx
@@ -77,16 +77,18 @@ const NodeMenu = ({
           textarea.selectionStart,
           textarea.selectionEnd,
         );
-        setTimeout(async () => {
+
+        // Remove focus from the block to ensure updateBlock works properly
+        // https://github.com/RoamJS/query-builder/issues/286
+        document.body.click();
+
+        const createNodeAndUpdateBlock = async () => {
           const pageName = await getNewDiscourseNodeText({
             text: highlighted,
             nodeType: nodeUid,
             blockUid,
           });
-
-          if (!pageName) {
-            return;
-          }
+          if (!pageName) return;
 
           const currentBlockText = getTextByBlockUid(blockUid);
           const newText = `${currentBlockText.substring(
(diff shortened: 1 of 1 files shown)


mdroidian (commented):
y u no coderabit

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -77,40 +77,43 @@ const NodeMenu = ({
...
           )}[[${pageName}]]${currentBlockText.substring(textarea.selectionEnd)}`;
 
-          updateBlock({ text: newText, uid: blockUid });
-          posthog.capture("Discourse Node: Created via Node Menu", {
-            nodeType: nodeUid,
-            text: pageName,
-          });
-
-          createDiscourseNode({
+          await createDiscourseNode({
```
mdroidian, line 99: Having `createDiscourseNode` before `update` fixes the sidebar open issue.  `updateBlock` was actually creating a page too fast, so we were getting an error saying "this page was already created", thus not able to set the `pageuid` for `openinsidebar`

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -77,40 +77,43 @@ const NodeMenu = ({
           textarea.selectionStart,
           textarea.selectionEnd,
         );
-        setTimeout(async () => {
+
+        // Remove focus from the block to ensure updateBlock works properly
+        // https://github.com/RoamJS/query-builder/issues/286
+        document.body.click();
```
mdroidian, line 83: moved this to inside the `onSelect` because 
- it can be in a single place
- let's node tags be added without removing focus from the block
- allows us to `getFocusedBlock` in the future so that we can set the cursor after editing the block

mdroidian:
@coderabbitai full review

trangdoan982 (approved):
lgtm

PR #499, opened 2025-10-17
ENG-972 Missed last type regeneration

maparent:
https://linear.app/discourse-graphs/issue/ENG-972/missed-last-type-regeneration
I forgot to regenerate the types after my last migration changes. This is only generated code.


mdroidian (approved):
This has happened a few times now.  Could you add a linear tasks to your backlog to create a CI step to check for this, please?

PR #505, opened 2025-10-19
ENG-984 Add DiscourseContextOverlay support in ResultsTable

mdroidian:
simple v0 with the popup but this is the only way to async query the discourse relations/references to get the values.
`action:discourse:overlay`
image

## Summary by CodeRabbit

* **New Features**
  * Introduced an "overlay" suggestion option for discourse-related actions, allowing users to view and interact with context overlays alongside other interaction options.


mdroidian:
@CodeRabbitai full review

PR #481, opened 2025-10-08
ENG-947  dragging and/or clicking on section header crashes the sidebar

sid597:
## Summary by CodeRabbit

* **Bug Fixes**
  * Sidebar configuration now stays in sync when initial settings change, preventing stale layouts.
  * Clicking a section title without sub-items no longer triggers unintended navigation.
  * Chevron actions only apply when a section has sub-items, reducing accidental clicks.

* **Refactor**
  * Simplified sidebar item click logic to make interactions more predictable and consistent.

PR #506, opened 2025-10-19
ENG-974 Sidebar crashing test branch

mdroidian:
for testing only
https://discoursegraphs.com/releases/roam/eng-947-sidebar-crash-testing


mdroidian:
fixed in #481

PR #510, opened 2025-10-23
ENG-800 handle update embeddings button

sid597:
## Release Notes

* **New Features**
  * Added embedding management functionality to the settings panel

* **Improvements**
  * Enhanced search filtering capabilities with improved data retrieval
  * Optimized parallel data processing for better performance
  * Improved timestamp consistency with automatic fallback to creation date

### DIFF
--- a/apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
+++ b/apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
@@ -91,28 +91,6 @@ const DiscourseNodeSuggestiveRules = ({
           supabaseClient,
           context,
         );
-        const nodeBlockToLocalConcepts = blockNodesSince.map((node) => {
-          const localConcept = discourseNodeBlockToLocalConcept(context, {
-            nodeUid: node.source_local_id,
-            schemaUid: node.type,
-            text: node.node_title
-              ? `${node.node_title} ${node.text}`
-              : node.text,
-          });
-          return localConcept;
-        });
-
-        const { error } = await supabaseClient.rpc("upsert_concepts", {
-          data: nodeBlockToLocalConcepts,
-          /* eslint-disable-next-line @typescript-eslint/naming-convention */
-          v_space_id: context.spaceId,
-          /* eslint-disable-next-line @typescript-eslint/naming-convention */
-        });
-        if (error) {
-          throw new Error(
-            `upsert_concepts failed: ${JSON.stringify(error, null, 2)}`,
-          );
-        }
       }
     } finally {
       setIsUpdating(false);
--- a/apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
+++ b/apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
@@ -12,7 +12,7 @@ import TextPanel from "roamjs-components/components/ConfigPanels/TextPanel";
 import getSubTree from "roamjs-components/util/getSubTree";
 import { DiscourseNode } from "~/utils/getDiscourseNodes";
 import extractRef from "roamjs-components/util/extractRef";
-import { getDiscourseNodeTypeWithSettingsBlockNodes } from "~/utils/getAllDiscourseNodesSince";
+import { getAllDiscourseNodesSince } from "~/utils/getAllDiscourseNodesSince";
 import { upsertNodesToSupabaseAsContentWithEmbeddings } from "~/utils/syncDgNodesToSupabase";
 import { discourseNodeBlockToLocalConcept } from "~/utils/conceptConversion";
 import { getLoggedInClient, getSupabaseContext } from "~/utils/supabaseContext";
@@ -68,17 +68,18 @@ const DiscourseNodeSuggestiveRules = ({
     (e: React.ChangeEvent<HTMLInputElement>) => {
       const newValue = e.target.value;
       setEmbeddingRef(newValue);
+      node.embeddingRef = newValue;
     },
-    [],
+    [node],
   );
   const [isUpdating, setIsUpdating] = useState(false);
 
   const handleUpdateEmbeddings = async (): Promise<void> => {
     setIsUpdating(true);
     try {
-      const blockNodesSince = getDiscourseNodeTypeWithSettingsBlockNodes(
-        node,
-        0,
+      const blockNodesSince = await getAllDiscourseNodesSince(
+        new Date(0).toISOString(),
+        [node],
       );
       const supabaseClient = await getLoggedInClient();
       if (!supabaseClient) return;
@@ -131,11 +132,10 @@ const DiscourseNodeSuggestiveRules = ({
       <TextPanel
         title="Embedding Block Ref"
         description="Copy block ref from template which you want to be embedded and ranked."
-        order={0}
+        order={1}
         uid={node.embeddingRefUid || ""}
         parentUid={parentUid}
-        value={node.embeddingRef || ""}
-        defaultValue={node.embeddingRef || ""}
+        defaultValue={embeddingRef || ""}
         options={{
           placeholder: "((block-uid))",
           onChange: handleEmbeddingRefChange,
@@ -152,7 +152,7 @@ const DiscourseNodeSuggestiveRules = ({
       <FlagPanel
         title="First Child"
         description="If the block is the first child of the embedding block ref, it will be embedded and ranked."
-        order={1}
+        order={2}
         uid={node.isFirstChild?.uid || ""}
         parentUid={parentUid}
         value={node.isFirstChild?.value || false}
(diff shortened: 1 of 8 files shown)


sid597:
* **#510** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/510?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#527](https://github.com/DiscourseGraphs/discourse-graph/pull/527) <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/527?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#531](https://github.com/DiscourseGraphs/discourse-graph/pull/531) <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/531?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/510?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#521** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/521?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
```diff
@@ -109,11 +158,11 @@ const DiscourseNodeSuggestiveRules = ({
...
 
-      {/* TODO: Add a button to update embeddings in seperate PR */}
       <Button
         text="Update Embeddings"
         intent={Intent.NONE}
-        onClick={() => console.log("Not implemented")}
+        onClick={() => void handleUpdateEmbeddings()}
+        loading={isUpdating}
         className="w-52"
         disabled
```
mdroidian: FYI: This button is always disabled

mdroidian (changes requested):
I'm not sure what issue/problem this PR is solving.  The conversation in ENG-800 is quite long and 2 months old.

Could you update the body with a loom video or summarized context please?

sid597:
This pr is solving the problem of user wanting to update their embeddings by using a different strategy than  they initially opted for. 

We can just take the node titles of all discourse nodes and create embeddings for them, but this approach falls short for nodes like  `source nodes` so in addition to the node title user might want to add some other `description` from each source page. 

Currently we use the templates for each node to give user an option of which block from the template would they like to embed along side with the title. 
image

User selects the block uid from the template, and checks if for each such block on the corresponding page take the first child or not. 

Now at this point user has selected the uid, checked the box and its time to create embeddings for that node type. This is where this PR comes in, implementing what happens when the user clicks on the `Update Embeddings` button. 

- Get all the discourse nodes instances for this type of node since time `0`, now the nodes in graph can or cannot have the block that users want, why? maybe they were created before the template, or accidentaly deleted etc. so in that case we only extract the title.
   -  The implementation is in 2 steps, first we query and get all the discourse nodes of this type and store it in a map, key is the block-uid and value is the data. 
   Then in the next step we query for the discourse node of this type which also has the block data, and update the map again. 
   
   so at the end of this function we have all the nodes that don't have block data + the nodes that do have that data.

- Then update this new data as content and context to supabase, the nodes that have block data we mark them as `direct_and_description` variant and the ones that don't have block we mark them as `direct`.

mdroidian (changes requested):
Please run coderabbit first and make any necessary adjustments.  Then tag reviewers once that is complete. 👍

sid597:
@coderabbitai full review

sid597:
@mdroidian iirc I did it locally

apps/roam/src/index.ts
```diff
@@ -126,7 +127,18 @@ export default runExtension(async (onloadArgs) => {
   document.addEventListener("input", discourseNodeSearchTriggerListener);
   document.addEventListener("selectionchange", nodeCreationPopoverListener);
 
-  await initializeSupabaseSync();
+  const supabase = createClient();
+  if (supabase) {
+    const { data } = await supabase
```
maparent: This will fail unless you use a logged in client.

maparent:
Thinking through this. The variant is supposed to give all relevant information to describe how the content was obtained. So `direct_and_description` made sense when we only focused on `First Child`. Now that we have an embedding block ref, ideally that should be part of the variant information; otherwise someone could switch the block ref and the database would not know that the information is obsolete. <br>Now from your process description, I think it still works because you recompute the value for all of them. The outcome is that at most one of each variant can exist in the database, which I think is probably what we want at this point (esp. re your other question in discord.) But yeah, I'd like to extend the database variant model to include that information. Not a blocker of course.

apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
```diff
@@ -64,10 +68,56 @@ const DiscourseNodeSuggestiveRules = ({
...
+      if (!supabaseClient) return;
+
+      const context = await getSupabaseContext();
+      if (context && blockNodesSince) {
+        await upsertNodesToSupabaseAsContentWithEmbeddings(
+          blockNodesSince,
+          supabaseClient,
+          context,
+        );
+        const nodeBlockToLocalConcepts = blockNodesSince.map((node) => {
```
maparent: Not sure why a change of content needs to trigger a change of concept here? I suspect the whole stanza is unneeded.

maparent (changes requested):
two nits, otherwise lgtm.

apps/roam/src/index.ts
maparent: Realizing this is not just a sanity check, but probably used as a proxy for whether the embeddings were activated. But the Space entry is not accessible before login, and is created as a side-effect of login. So that information should be stored roam-side. (Having a sanity check for database access still makes sense...)

sid597:
@maparent what do you mean by 

>  Now that we have an embedding block ref, ideally that should be part of the variant information; otherwise someone could switch the block ref and the database would not know that the information is obsolete.

apps/roam/src/index.ts
sid597: Ah ok I missed that part, you are right, will fix it

apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
sid597: removed

apps/roam/src/index.ts
sid597: ok so now I making the change to check if either `data` exists in supabase or on roam-side, to make it backwards compatible with graphs already using supabase. 

How do we do login in prod for different graphs?

sid597: I am confused about the login part, this is my mental model about why it should work:

- Click `Generate & Upload All Node Embeddings` button
- which calls `createOrUpdateDiscourseEmbedding` button
- which calls `getLoggedInClient` 
- which calls `createLoggedInClient` 

So should this not create a client i.e login? Therefore if the user has not clicked the button implying that this graph was not initialized hence we don't run sync. 

If they click the button, they create a logged in client and then the space for this graph, so on further login it will be enabled for all other users as well because we create profiles for all the other users as well.

maparent: all this is correct. But in the code I pointed out, you are calling a createClient (Not logged in) and then testing if the space exists. The non-logged in client cannot see the space, so you will get an empty result no matter if the space exists or not.

sid597: oh I see, so using `getLoggedInClient`  will be correct  thing to do?

maparent:
Hmmm... 
Ok, I see you now use the getLoggedInClient. I'm not sure it's the best route.
I thought we had decided to avoid space creation unless the suggestive mode is enabled. So ideally, check for the setting **first**, and if it set just trigger the sync (which does its own checks for whether the client has access to supabase, creates the space, etc.) I don't think you need to check for the client at all in index.ts.

(Note: We also spoke of gating on existence of space password. Now thinking that's not such a great idea, since in the future existence of db space account will be gated not only on suggestive mode, but also on inter-graph.)

sid597:
Ok so removed supabase based checks, now only use global roam settings. After merge we would have explicitly clcik on the button to enable the functionality for all the current graphs using sync.

maparent (approved):
lgtm

apps/roam/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import { addStyle } from "roamjs-components/dom";
+import { addStyle, getRoamUrl } from "roamjs-components/dom";
```
mdroidian: unused

apps/roam/src/index.ts
```diff
@@ -32,6 +32,11 @@ import {
   setSyncActivity,
 } from "./utils/syncDgNodesToSupabase";
 import { initPluginTimer } from "./utils/pluginTimer";
+import { getLoggedInClient } from "./utils/supabaseContext";
```
mdroidian: unused

mdroidian (approved):
👍

PR #512, opened 2025-10-23
ENG-975: Propagate errors aggressively when upsert of content+embeddings fail.

maparent:
This will avoid creating orphan concepts.


maparent:
This is a minimalist version that solves ENG-975 issue with the least changes, until we agree on approach.

sid597 (approved):
LGTM

PR #518, opened 2025-10-26
ENG-1004 add a buffer, so we do not miss any change

maparent:
As described in [ENG-976](https://linear.app/discourse-graphs/issue/ENG-976/repair-sync-function), some roam changes happen after start of query. and if we query for changes starting at start of query, some will be skipped. The full solution would be [ENG-999](https://linear.app/discourse-graphs/issue/ENG-999/allow-sync-table-to-know-latest-timestamp-handled), but we can have a short-term fix with a simple time buffer.

### DIFF
--- a/apps/roam/src/utils/syncDgNodesToSupabase.ts
+++ b/apps/roam/src/utils/syncDgNodesToSupabase.ts
@@ -422,7 +422,8 @@ export const createOrUpdateDiscourseEmbedding = async (showToast = false) => {
     }
     claimed = true;
     const allUsers = await getAllUsers();
-    const time = (lastUpdateTime || DEFAULT_TIME).toISOString();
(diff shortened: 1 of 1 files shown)


sid597 (approved):
LGTM!

So we have eventually consistent system and I wonder re.

> I noticed this morning that the current strategy misses some nodes. We assume that, if we give the start of the task as point of reference, we will get all the modified nodes. But the query has a bit of a lag, so some nodes modified just after the query start are not included in the query result (see "skipped" in my gist, which I did see turn up in logs.)

that not having the latest nodes created just after loading the graph is fine because the assumption is that the sync takes all the nodes upto that point and uploads them. So what we are saying to the user is that "you have access to embeddings to all the nodes that were created just before you opened your instance of this graph" because new nodes could also be created by other users as the sync is going on, we don't provide any guarantees about consistency

PR #490, opened 2025-10-14
[ENG-869] Key figure

trangdoan982:
https://www.loom.com/share/dad6f94dbbf54d9693562f53a4dffe38

## Summary by CodeRabbit

* **New Features**
  * Added boolean field support in settings, including a “Key image” option for node types.
  * Discourse nodes can automatically display a key image (first image in the note) when enabled.
  * Nodes now auto-resize to fit title, subtitle, and image for improved readability.
  * Selecting an existing note preloads the key image when available for faster, smoother rendering.

PR #522, opened 2025-10-27
[ENG-869] Key figure

trangdoan982:
https://www.loom.com/share/11478ad987614ae4bdee0c0c61504fa8

Old PR comments can be found here: https://github.com/DiscourseGraphs/discourse-graph/pull/490

## Summary by CodeRabbit

* **New Features**
  * Added a boolean "key image" option to node-type settings.
  * Canvas nodes can preload and display a key image from linked files.
  * Node dimensions are now calculated dynamically from title, type and image.
  * Nodes auto-resize (and respect manual resizing) to better fit text and images.

### DIFF
--- a/apps/obsidian/src/utils/measureNodeText.ts
+++ b/apps/obsidian/src/utils/measureNodeText.ts
@@ -1,15 +1,29 @@
 import {
   MIN_NODE_WIDTH,
   MAX_NODE_WIDTH,
+  CONTAINER_PADDING,
+  CONTAINER_BORDER_WIDTH,
+  CONTAINER_BORDER_RADIUS,
+  TITLE_MARGIN,
+  TITLE_FONT_SIZE,
+  TITLE_LINE_HEIGHT,
+  TITLE_FONT_WEIGHT,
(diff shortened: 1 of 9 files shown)


trangdoan982:
@coderabbitai full review

trangdoan982:
@coderabbitai full review

apps/obsidian/src/utils/measureNodeText.ts
```diff
@@ -0,0 +1,74 @@
...
+export const measureNodeText = ({
+  title,
+  subtitle,
+}: {
+  title: string;
+  subtitle: string;
+}): { w: number; h: number } => {
+  // Create a container matching the actual component structure
+  const container = document.createElement("div");
+  container.style.setProperty("position", "absolute");
```
mdroidian, line 41: We should be careful about setting literal values here.  If they change in the component, the measurement will be off.  And it's probably a good assumption that they will change sooner than later. 😁

mdroidian (approved):
I would suggest not using literal values for the `measureNodeText`, but other than that, LGTM 👍

PR #530, opened 2025-11-03
[ENG-1026] Fix default node tags and improve lint for node type setting

trangdoan982:
Since our validator will discard any tag setting with `#` in it, we want to make sure our default values also reflect this constraint

### DIFF
--- a/apps/obsidian/src/components/NodeTypeSettings.tsx
+++ b/apps/obsidian/src/components/NodeTypeSettings.tsx
@@ -8,7 +8,7 @@ import { ConfirmationModal } from "./ConfirmationModal";
 import { getTemplateFiles, getTemplatePluginInfo } from "~/utils/templates";
 
 const generateTagPlaceholder = (format: string, nodeName?: string): string => {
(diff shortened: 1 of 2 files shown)


apps/obsidian/src/components/NodeTypeSettings.tsx
```diff
@@ -8,7 +8,7 @@ import { ConfirmationModal } from "./ConfirmationModal";
 import { getTemplateFiles, getTemplatePluginInfo } from "~/utils/templates";
 
 const generateTagPlaceholder = (format: string, nodeName?: string): string => {
-  if (!format) return "Enter tag (e.g., clm-candidate or #clm-candidate)";
```
maparent, line 11: Nit: given the field width, we cannot see the placeholder. I would broaden the field for that reason only.
Another nit, probably out of scope: You look for the ` - {content}` pattern, but the format validation logic only checks for the existence of `{content}`, without the dash.

maparent (approved):
A small nit, I would recommend making the tag field larger, but otherwise lgtm.

PR #534, opened 2025-11-05
ENG-1011 Create Function to Record Relation Triples in Roam

maparent:
https://linear.app/discourse-graphs/issue/ENG-1011/create-function-to-record-relation-triples-in-roam

## Summary by CodeRabbit

* **Chores**
  * Added utility functions to support internal system operations related to block and relation management.

### DIFF
--- a/apps/roam/src/utils/createReifiedBlock.ts
+++ b/apps/roam/src/utils/createReifiedBlock.ts
@@ -4,11 +4,15 @@ import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTit
 import setBlockProps from "./setBlockProps";
 import { getSetting } from "~/utils/extensionSettings";
 
-export const createReifiedBlock = async (
-  destinationBlockUid: string,
-  schemaUid: string,
-  parameterUids: Record<string, string>,
-): Promise<string> => {
+export const createReifiedBlock = async ({
+  destinationBlockUid,
+  schemaUid,
+  parameterUids,
+}: {
+  destinationBlockUid: string;
+  schemaUid: string;
+  parameterUids: Record<string, string>;
+}): Promise<string> => {
   // TODO/Question: Should we try to ensure uniqueness?
   const newUid = window.roamAlphaAPI.util.generateUID();
   await createBlock({
@@ -43,20 +47,24 @@ const getRelationPageUid = async (): Promise<string> => {
   return relationPageUid;
 };
 
-export const createReifiedRelation = async (
-  sourceUid: string,
-  relationBlockUid: string,
-  destinationUid: string,
-): Promise<string | undefined> => {
+export const createReifiedRelation = async ({
+  sourceUid,
+  relationBlockUid,
+  destinationUid,
+}: {
+  sourceUid: string;
+  relationBlockUid: string;
+  destinationUid: string;
+}): Promise<string | undefined> => {
   const authorized = getSetting("use-reified-relations");
   if (authorized) {
-    return await createReifiedBlock(
-      await getRelationPageUid(),
-      relationBlockUid,
-      {
+    return await createReifiedBlock({
+      destinationBlockUid: await getRelationPageUid(),
+      schemaUid: relationBlockUid,
+      parameterUids: {
         sourceUid,
         destinationUid,
       },
-    );
+    });
   }
 };
--- a/apps/roam/src/utils/createReifiedBlock.ts
+++ b/apps/roam/src/utils/createReifiedBlock.ts
@@ -4,56 +4,68 @@ import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTit
 import setBlockProps from "./setBlockProps";
 import { getSetting } from "~/utils/extensionSettings";
 
-export const queryForReifiedBlocks = async (
+const queryForReifiedBlocksUtil = async (
   parameterUids: Record<string, string>,
-  strict: boolean = true,
-): Promise<string | string[] | null> => {
-  // TODO: Check that the parameterUids keys correspond to the schema
+): Promise<[string, Record<string, string>][]> => {
   const paramsAsSeq = Object.entries(parameterUids);
   const query = `[:find ?u ?d
   :in $ ${paramsAsSeq.map(([k]) => "?" + k).join(" ")}
   :where [?s :block/uid ?u] [?s :block/props ?p] [(get ?p :discourse-graph) ?d]
   ${paramsAsSeq.map(([k]) => `[(get ?d :${k}) ?${k}]`).join(" ")} ]`;
-  const result = (await window.roamAlphaAPI.data.async.q(
+  return (await window.roamAlphaAPI.data.async.q(
     query,
     ...paramsAsSeq.map(([, v]) => v),
   )) as [string, Record<string, string>][];
-  if (strict) {
-    const resultF = result
-      .filter(([, params]) => Object.keys(params).length === paramsAsSeq.length)
-      .map(([uid]) => uid);
-    if (resultF.length > 0) return resultF[0];
-  } else {
-    return result.map(([uid]) => uid);
-  }
-  return null;
 };
 
-export const checkForReifiedBlock = async (
+export const queryForReifiedBlocks = async (
+  parameterUids: Record<string, string>,
+): Promise<string[]> => {
+  const results = await queryForReifiedBlocksUtil(parameterUids);
+  return results.map(([uid]) => uid);
+};
+
+export const strictQueryForReifiedBlocks = async (
   parameterUids: Record<string, string>,
 ): Promise<string | null> => {
-  const result = await queryForReifiedBlocks(parameterUids, true);
-  if (Array.isArray(result)) return result[0]; // this should never happen with strict
-  return result;
+  const result = await queryForReifiedBlocksUtil(parameterUids);
+  const numParams = Object.keys(parameterUids).length;
+  // post-filtering because cannot filter by number of keys in datascript
+  const resultF = result
+    .filter(([, params]) => Object.keys(params).length === numParams)
+    .map(([uid]) => uid);
+  if (resultF.length > 1) {
+    const paramsAsText = Object.entries(parameterUids)
+      .map(([k, v]) => `${k}: '${v}'`)
+      .join(", ");
+    console.warn(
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,63 @@
...
+const getRelationPageUid = async (): Promise<string> => {
+  if (relationPageUid === undefined) {
+    relationPageUid = getPageUidByPageTitle(RELATION_PAGE_TITLE);
+    console.log(relationPageUid);
+    if (relationPageUid === "") {
+      relationPageUid = await createPage({ title: RELATION_PAGE_TITLE });
+    }
+  }
+  return relationPageUid;
+};
```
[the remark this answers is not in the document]
maparent: it turns out this function is sync.
Thanks for picking up the log.

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,63 @@
...
+import setBlockProps from "./setBlockProps";
+import { getSetting } from "~/utils/extensionSettings";
+
+export const createReifiedBlock = async (
+  destinationBlockUid: string,
+  schemaUid: string,
+  parameterUids: Record<string, string>,
+): Promise<string> => {
+  // TODO/Question: Should we try to ensure uniqueness?
+  const newUid = window.roamAlphaAPI.util.generateUID();
```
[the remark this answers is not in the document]
maparent: none of this package will be activated without roamAlphaAPI, not a concern.
I think if there is an error in this code, I will let it percolate.
I was hesitant on TODO scope creep, but decided it's in scope.

maparent:
@coderabbit review last commit

maparent:
@coderabbit review last commit

maparent:
@coderabbit: setBlockProps does not need await. I am leaving the todo intentionally. Thanks for pushing me on api clarity.

mdroidian (changes requested):
Before doing a full review, could you please add the required loom video

> Done When
> Behavior is verified via a loom video by manually invoking the function and confirming correct output structure.

sid597:
A stub calling function would be helpful for review, maybe add it to `onChange` for the button `Reified Relation Triples` button?

sid597:
I spent some time on reviewing by adding the stub function, the function does creates the block and store the properties.  Then I went into rabbithole of what is going on and why, following threads of discussion and its all very interesting. 

image

sid597 (changes requested):
A video of expected input and output would be useful, not sure if I reproduced it right

maparent:
Here is the loom video as requested in the task. I thought those were only for UX PRs?
https://www.loom.com/share/2937c0cb53e042f48b9d874d9ca807d1

mdroidian:
> Here is the loom video as requested in the task. I thought those were only for UX PRs? 

Videos are required by default for UX PRs, but most PRs would benefit from including one as well. That said, if a task explicitly requests a video, the point is moot.

maparent:
Of course, just asking why the video was requested in this specific task.

mdroidian:
@maparent is this one ready for review now?  If so, please change the status from `Draft`.

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,118 @@
...
+  if (relationPageUid === undefined) {
+    relationPageUid = getPageUidByPageTitle(RELATION_PAGE_TITLE);
+    if (relationPageUid === "") {
+      relationPageUid = await createPage({ title: RELATION_PAGE_TITLE });
+    }
+  }
+  return relationPageUid;
+};
+
+export const createReifiedRelation = async ({
```
mdroidian: Why is this exported?

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,118 @@
...
+    );
+  }
+  return resultF.length > 0 ? resultF[0] : null;
+};
+
+export const createReifiedBlock = async ({
+  destinationBlockUid,
+  schemaUid,
+  parameterUids,
+  allowDuplicates = true,
```
mdroidian: What is a use case (based on the current task) for `allowDuplicates` true?

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,118 @@
...
+  const newUid = window.roamAlphaAPI.util.generateUID();
+  await createBlock({
+    node: {
+      text: newUid,
+      uid: newUid,
+    },
+    parentUid: destinationBlockUid,
+    order: "last",
+  });
+  setBlockProps(newUid, {
```
mdroidian: We should be able to set the props on createBlock

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,118 @@
...
+  :in $ ${paramsAsSeq.map(([k]) => "?" + k).join(" ")}
+  :where [?s :block/uid ?u] [?s :block/props ?p] [(get ?p :discourse-graph) ?d]
+  ${paramsAsSeq.map(([k]) => `[(get ?d :${k}) ?${k}]`).join(" ")} ]`;
+  return (await window.roamAlphaAPI.data.async.q(
+    query,
+    ...paramsAsSeq.map(([, v]) => v),
+  )) as [string, Record<string, string>][];
+};
+
+export const queryForReifiedBlocks = async (
```
mdroidian: Why is this exported?  What is the use case based on the existing task?

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,118 @@
...
+  parameterUids: Record<string, string>,
+): Promise<string[]> => {
+  const results = await queryForReifiedBlocksUtil(parameterUids);
+  return results.map(([uid]) => uid);
+};
+
+export const strictQueryForReifiedBlocks = async (
```
mdroidian: Why is this exported?  What is the use case based on the existing task?

apps/roam/src/utils/createReifiedBlock.ts
mdroidian: Actually, this isn't used anywhere currently, is it.  If it's not being used, let's remove it.

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,118 @@
...
+
+const queryForReifiedBlocksUtil = async (
+  parameterUids: Record<string, string>,
+): Promise<[string, Record<string, string>][]> => {
+  const paramsAsSeq = Object.entries(parameterUids);
+  const query = `[:find ?u ?d
+  :in $ ${paramsAsSeq.map(([k]) => "?" + k).join(" ")}
+  :where [?s :block/uid ?u] [?s :block/props ?p] [(get ?p :discourse-graph) ?d]
+  ${paramsAsSeq.map(([k]) => `[(get ?d :${k}) ?${k}]`).join(" ")} ]`;
+  return (await window.roamAlphaAPI.data.async.q(
```
mdroidian: should probably be `backend.q` to not lock UI

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,118 @@
...
+      .map(([k, v]) => `${k}: '${v}'`)
+      .join(", ");
+    console.warn(
+      `${resultF.length} results in strict query for {${paramsAsText}}`,
+    );
+  }
+  return resultF.length > 0 ? resultF[0] : null;
+};
+
+export const createReifiedBlock = async ({
```
mdroidian: Why is this exported?  When do you expect to use it?

apps/roam/src/utils/createReifiedBlock.ts
mdroidian: We should be able to set props in createBlock

apps/roam/src/utils/createReifiedBlock.ts
mdroidian: Actually, when would we need this strict query vs queryForReifiedBlocks?

apps/roam/src/utils/createReifiedBlock.ts
mdroidian: If there is no planned use case, please remove.

apps/roam/src/utils/createReifiedBlock.ts
mdroidian: If there is no planned use case, please remove.

mdroidian (approved):
Approved to not block, but quite a few things I am unclear on based on the task.

apps/roam/src/utils/createReifiedBlock.ts
maparent: Planned usecase is downstream, replacing with a comment for now.

apps/roam/src/utils/createReifiedBlock.ts
maparent: Leaving as comment for now.

apps/roam/src/utils/createReifiedBlock.ts
maparent: Right. On that note, it may be worth marking the backend as async in `roamjs-components`? I'll use the `Promise.resolve` workaround for now.

apps/roam/src/utils/createReifiedBlock.ts
maparent: Removing the export.

apps/roam/src/utils/createReifiedBlock.ts
maparent: ok

apps/roam/src/utils/createReifiedBlock.ts
maparent: Ok, simply porting this to backend took a while, because the query as written is refused by the backend. I have to add an extra layer of bindings, which has a performance impact. (~163ms, not huge but adds up in migration.)
I committed it, but I think it's fast enough when done in frontend that it would be better kept there.
More generally: I think we need to develop a policy of when a query should be sent to the backend or not. 
I don't think sending everything to the backend is the optimum point.

mdroidian (approved):

PR #536, opened 2025-11-06
ENG-1018 Relation migrations, with UX support

maparent:
This creates UI to trigger a migration, where reified relation objects are created from existing pattern relations in a graph.
The migration is idempotent, and can be re-run at will.

## Summary by CodeRabbit

* **New Features**
  * Tabbed navigation in admin panel with separate sections for Migration, Feature Flags, and Node List management
  * Relation migration tool with real-time progress tracking and result counters

* **Refactor**
  * Reorganized admin panel components for improved structure and maintainability

### DIFF
--- a/apps/roam/src/components/settings/AdminPanel.tsx
+++ b/apps/roam/src/components/settings/AdminPanel.tsx
@@ -282,6 +282,7 @@ const MigrationTab = (): React.ReactElement => {
             ? `${numRelations} already migrated`
             : "No migrated relations",
         );
+        // eslint-disable-next-line react-hooks/exhaustive-deps
         initial = false;
       }
     })();
--- a/apps/roam/src/components/settings/AdminPanel.tsx
+++ b/apps/roam/src/components/settings/AdminPanel.tsx
@@ -252,11 +252,12 @@ const MigrationTab = (): React.ReactElement => {
   const enabled = getSetting("use-reified-relations");
   const [useMigrationResults, setMigrationResults] = useState<string>("");
   const [useOngoing, setOngoing] = useState<boolean>(false);
+  const [useDryRun, setDryRun] = useState<boolean>(false);
   const doMigrateRelations = async () => {
     setOngoing(true);
     try {
       const before = await countReifiedRelations();
-      const numProcessed = await migrateRelations();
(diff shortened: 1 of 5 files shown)


mdroidian:
> The full migration work might also replace existing pattern relations by references to the new relation objects.

This is currently out of scope and not planned.

maparent:
@coderabbit review

mdroidian (changes requested):
Please add the required loom video.

maparent:
The loom video is in the linear task.

maparent:
sorry I missed that it was in draft mode.

mdroidian:
@maparent 
> This is preliminary work ...

Could you elaborate on what, if anything, is left out of this PR that is required in ENG-1018

> Note that there are still duplicates being created ...

Is this still an issue?

maparent:
No, that is solved. I consider this is ready.

apps/roam/src/components/settings/AdminPanel.tsx
```diff
@@ -211,6 +208,147 @@ const AdminPanel = (): React.ReactElement => {
...
+    void (async () => {
+      if (initial) {
+        const numRelations = await countReifiedRelations();
+        setMigrationResults(
+          numRelations > 0
+            ? `${numRelations} already migrated`
+            : "No migrated relations",
+        );
+        initial = false;
+      }
```
mdroidian, line 287: "Assignments to the 'initial' variable from inside React Hook useEffect will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside useEffect.eslintreact-hooks/exhaustive-deps"

mdroidian (approved):
👍

apps/roam/src/components/settings/AdminPanel.tsx
maparent: Thanks, that's interesting, but I think in this case I do want it to be reassigned. If we re-render the window, we do want to re-set the initial value of the migrationResults string.

mdroidian: This is the comment raised by the IDE.  If you plan on leaving that code in there, add the appropriate comment to ignore the warning.

PR #538, opened 2025-11-10
ENG-1012 Update Query Builder Relation Conditions to Use Reified Relation Triples

maparent:
https://linear.app/discourse-graphs/issue/ENG-1012/update-query-builder-relation-conditions-to-use-reified-relation

Update query builder relation conditions.

Notes: 
Right now, we use one of two query paths, according to the `use-reified-relations` setting. 
I had to make changes to the optimizer. (Otherwise queries with or-clause would just fail, or even crash.)
Performance is clearly better than before but a bit less than hoped for.

## Summary by CodeRabbit

* **New Features**
  * Admin panel redesigned with tabbed interface: "Node list" and "Experiments" sections for improved organization.
  * Experiment settings now toggleable via checkbox with persistent storage.
  * Reified relations support for managing complex node relationships with deduplication control.
  * Relation migration utility to transform existing relation data.

### DIFF
--- a/apps/roam/src/utils/createReifiedBlock.ts
+++ b/apps/roam/src/utils/createReifiedBlock.ts
@@ -5,10 +5,19 @@ import { getSetting } from "~/utils/extensionSettings";
 
 export const DISCOURSE_GRAPH_PROP_NAME = "discourse-graph";
 
+const SANE_ROLE_NAME_RE = new RegExp(/^[a-zA-Z][a-zA-Z0-9_-]*$/);
+
 const strictQueryForReifiedBlocks = async (
   parameterUids: Record<string, string>,
 ): Promise<string | null> => {
   const paramsAsSeq = Object.entries(parameterUids);
+  // sanitize parameter names
+  if (
+    Object.keys(parameterUids).filter((k) => !k.match(SANE_ROLE_NAME_RE)).length
+  )
+    throw new Error(
+      `invalid parameter names in ${Object.keys(parameterUids).join(", "),}`
+    );
   const query = `[:find ?u ?d
   :in $ ${paramsAsSeq.map(([k]) => "?" + k).join(" ")}
   :where [?s :block/uid ?u] [?s :block/props ?p] [(get ?p :${DISCOURSE_GRAPH_PROP_NAME}) ?d]
--- a/apps/roam/src/utils/createReifiedBlock.ts
+++ b/apps/roam/src/utils/createReifiedBlock.ts
@@ -5,13 +5,13 @@ import { getSetting } from "~/utils/extensionSettings";
 
 export const DISCOURSE_GRAPH_PROP_NAME = "discourse-graph";
 
-const SANE_ROLE_NAME_RE = new RegExp(/^[a-zA-Z][a-zA-Z0-9_-]*$/);
+const SANE_ROLE_NAME_RE = new RegExp(/^[\w\-]*$/);
 
 const strictQueryForReifiedBlocks = async (
   parameterUids: Record<string, string>,
 ): Promise<string | null> => {
   const paramsAsSeq = Object.entries(parameterUids);
-  // sanitize parameter names
+  // validate parameter names
   if (
     Object.keys(parameterUids).filter((k) => !k.match(SANE_ROLE_NAME_RE)).length
   )
@@ -81,7 +81,7 @@ const createReifiedBlock = async ({
 const RELATION_PAGE_TITLE = "roam/js/discourse-graph/relations";
 let relationPageUid: string | undefined = undefined;
 
-export const getOrCreateRelationPageUid = async (): Promise<string> => {
+const getOrCreateRelationPageUid = async (): Promise<string> => {
   if (relationPageUid === undefined) {
     relationPageUid = getPageUidByPageTitle(RELATION_PAGE_TITLE);
     if (relationPageUid === "") {
--- a/apps/roam/src/utils/createReifiedBlock.ts
+++ b/apps/roam/src/utils/createReifiedBlock.ts
@@ -72,7 +72,7 @@ const createReifiedBlock = async ({
 const RELATION_PAGE_TITLE = "roam/js/discourse-graph/relations";
 let relationPageUid: string | undefined = undefined;
 
-const getRelationPageUid = async (): Promise<string> => {
+export const getOrCreateRelationPageUid = async (): Promise<string> => {
   if (relationPageUid === undefined) {
     relationPageUid = getPageUidByPageTitle(RELATION_PAGE_TITLE);
     if (relationPageUid === "") {
@@ -82,8 +82,16 @@ const getRelationPageUid = async (): Promise<string> => {
   return relationPageUid;
 };
 
+export const getExistingRelationPageUid = (): string | undefined => {
+  if (relationPageUid === undefined) {
+    const uid = getPageUidByPageTitle(RELATION_PAGE_TITLE);
+    if (uid !== "") relationPageUid = uid;
+  }
+  return relationPageUid;
+};
+
 export const countReifiedRelations = async (): Promise<number> => {
-  const pageUid = await getRelationPageUid();
+  const pageUid = getExistingRelationPageUid();
   if (pageUid === undefined) return 0;
   const r = await window.roamAlphaAPI.data.async.q(
     `[:find (count ?c) :where [?p :block/children ?c] [?p :block/uid "${pageUid}"]]`,
@@ -103,7 +111,7 @@ export const createReifiedRelation = async ({
   const authorized = getSetting("use-reified-relations");
   if (authorized) {
     return await createReifiedBlock({
-      destinationBlockUid: await getRelationPageUid(),
+      destinationBlockUid: await getOrCreateRelationPageUid(),
       schemaUid: relationBlockUid,
       parameterUids: {
         sourceUid,
--- a/apps/roam/src/utils/createReifiedBlock.ts
+++ b/apps/roam/src/utils/createReifiedBlock.ts
@@ -16,7 +16,7 @@ const strictQueryForReifiedBlocks = async (
     Object.keys(parameterUids).filter((k) => !k.match(SANE_ROLE_NAME_RE)).length
   )
     throw new Error(
-      `invalid parameter names in ${Object.keys(parameterUids).join(", "),}`
+      `invalid parameter names in ${Object.keys(parameterUids).join(", ")}`,
     );
   const query = `[:find ?u ?d
   :in $ ${paramsAsSeq.map(([k]) => "?" + k).join(" ")}
--- a/apps/roam/src/utils/fireQuery.ts
+++ b/apps/roam/src/utils/fireQuery.ts
@@ -1,8 +1,8 @@
 import conditionToDatalog from "./conditionToDatalog";
 import type {
   PullBlock,
-  DatalogAndClause,
   DatalogClause,
+  DatalogAndClause,
 } from "roamjs-components/types";
 import compileDatalog from "./compileDatalog";
 import { getNodeEnv } from "roamjs-components/util/env";
@@ -107,7 +107,8 @@ const optimizeQuery = (
         if (Array.from(allVars).every((v) => capturedVariables.has(v))) {
           score = 10;
         } else {
-          score = 100002;
+          // downgrade disjunction and negation
+          score = c.type === "and-clause" ? 100002 : 100006;
         }
       } else if (c.type === "not-join-clause" || c.type === "or-join-clause") {
         if (c.variables.every((v) => capturedVariables.has(v.value))) {
@@ -156,6 +157,9 @@ const optimizeQuery = (
         .filter((v) => v.type === "variable")
         .forEach((v) => capturedVariables.add(v.value));
     }
+    // Question: Should we not consider all variables in a complex clause captured?
+    // const newVars = gatherDatalogVariablesFromClause(bestClause);
+    // newVars.forEach((v) => capturedVariables.add(v));
   }
   return orderedClauses;
 };
@@ -197,7 +201,7 @@ export const getDatalogQuery = ({
   const whereClauses = optimizeQuery(
     getWhereClauses({ conditions, returnNode }),
     new Set([]),
-  ) as DatalogClause[];
+  );
 
   const defaultSelections: {
     mapper: PredefinedSelection["mapper"];
--- a/apps/roam/src/utils/fireQuery.ts
+++ b/apps/roam/src/utils/fireQuery.ts
@@ -65,9 +65,9 @@ const firstVariable = (
 };
 
 const optimizeQuery = (
-  clauses: (DatalogClause | DatalogAndClause)[],
+  clauses: DatalogClause[],
   capturedVariables: Set<string>,
-): (DatalogClause | DatalogAndClause)[] => {
+): DatalogClause[] => {
   const marked = clauses.map(() => false);
   const orderedClauses: (DatalogClause | DatalogAndClause)[] = [];
   const variablesByIndex: Record<number, Set<string>> = {};
@@ -126,7 +126,8 @@ const optimizeQuery = (
             (a) => a.type !== "variable" || capturedVariables.has(a.value),
           )
         ) {
-          score = 1000;
+          // equality is almost as good as a binding
+          c.type == "pred-expr" && c.pred == "=" ? (score = 5) : (score = 1000);
         } else {
           score = 100004;
         }
@@ -156,10 +157,17 @@ const optimizeQuery = (
       bestClause.arguments
         .filter((v) => v.type === "variable")
         .forEach((v) => capturedVariables.add(v.value));
+    } else if (bestClause.type === "fn-expr") {
+      // A function expression acts as biding a variable to a unique function value
+      if (
+        bestClause.arguments.filter(
+          (a) => a.type === "variable" && !capturedVariables.has(a.value),
+        ).length === 0 &&
+        bestClause.binding.type === "bind-scalar" &&
+        bestClause.binding.variable.type === "variable"
+      )
+        capturedVariables.add(bestClause.binding.variable.value);
     }
-    // Question: Should we not consider all variables in a complex clause captured?
-    // const newVars = gatherDatalogVariablesFromClause(bestClause);
-    // newVars.forEach((v) => capturedVariables.add(v));
   }
   return orderedClauses;
 };
--- a/apps/roam/src/utils/registerDiscourseDatalogTranslators.ts
+++ b/apps/roam/src/utils/registerDiscourseDatalogTranslators.ts
@@ -447,18 +447,6 @@ const registerDiscourseDatalogTranslators = () => {
                 fn: "get",
                 arguments: [
                   { type: "variable", value: "reldata" },
-                  { type: "constant", value: ":hasSchema" },
-                ],
-                binding: {
-                  type: "bind-scalar",
-                  variable: { type: "variable", value: "relSchema" },
-                },
-              },
-              {
-                type: "fn-expr",
-                fn: "get",
-                arguments: [
-                  { type: "variable", value: "reldata" },
                   { type: "constant", value: ":sourceUid" },
                 ],
                 binding: {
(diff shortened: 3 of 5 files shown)


maparent:
@coderabbit review

mdroidian:
> I had to redefine `DatalogClause` to be able to use an equality predicate, will signal upstream.
> ( [RoamJS/roamjs-components#49](https://github.com/RoamJS/roamjs-components/issues/49) )
> I had to tweak the optimizer, this is a bit ad hoc but feels right. (Otherwise queries with or-clause would just fail.)

Could you elaborate on why these changes needed to be made?  I was under the assumption that all of the machinery already existed to simply switch over to the new reified relation triples.

Why, specifically, did you need to use an equality predicate?
We have many queries that already use or-clauses, is it the case that those don't fail but the ones you were trying did fail?

apps/roam/src/components/settings/AdminPanel.tsx
```diff
@@ -196,41 +210,87 @@ const AdminPanel = () => {
...
+                extensionAPI.settings.get("use-reified-relations") as boolean
+              }
+              onChange={(e) => {
+                const target = e.target as HTMLInputElement;
+                extensionAPI.settings.set(
+                  "use-reified-relations",
+                  target.checked,
+                );
+              }}
+              labelElement={
```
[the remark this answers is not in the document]
maparent: This works as-is, was modified somewhat, and follows a pattern used elsewhere in the code.

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -0,0 +1,126 @@
...
+import { getSetting } from "~/utils/extensionSettings";
+
+const queryForReifiedBlocksUtil = async (
+  parameterUids: Record<string, string>,
+): Promise<[string, Record<string, string>][]> => {
+  const paramsAsSeq = Object.entries(parameterUids);
+  const query = `[:find ?u ?d
+  :in $ ${paramsAsSeq.map(([k]) => "?" + k).join(" ")}
+  :where [?s :block/uid ?u] [?s :block/props ?p] [(get ?p :discourse-graph) ?d]
+  ${paramsAsSeq.map(([k]) => `[(get ?d :${k}) ?${k}]`).join(" ")} ]`;
```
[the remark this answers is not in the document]
maparent: Interesting point. I do not know that we have a good entry point to add parameters; but this could be useful. Worth discussing, probably in another scope. Meanwhile, I propose we check against a regex at input.

maparent:
First question: Why do I need equality?
Thank you for fixing it upstream, appreciated, it makes for less changes here.
Still illustrating the issue with data in the `discourse-graphs` graph, can be tried live.
Given that the uid of the `roam/js/discourse-graph/relations` page is `OjyVh_WEb`, we can ask for relations with a given source:

```
window.roamAlphaAPI.q(`[:find
  ?rel ?reldata
 :where
  [?relpage :block/uid "OjyVh_WEb"]
  [?rel :block/page ?relpage]
  [?rel :block/props ?relprops]
  [(get ?relprops :discourse-graph) ?reldata]
  [(get ?reldata :sourceUid) ?relSource]
  [(= ?relSource "vnvlZMNeF")]]`)
```

That was using equality. Without equality, the similar query

```
window.roamAlphaAPI.q(`[:find
  ?rel ?reldata
 :where
  [?relpage :block/uid "OjyVh_WEb"]
  [?rel :block/page ?relpage]
  [?rel :block/props ?relprops]
  [(get ?relprops :discourse-graph) ?reldata]
  [(get ?reldata :sourceUid) "vnvlZMNeF"]]`)
```

fails with the following message: `Cannot parse binding, expected (bind-scalar | bind-tuple | bind-coll | bind-rel)`

ETA:
This is also correctly enforced in 
https://github.com/RoamJS/roamjs-components/blob/b03819970e77f4c8a55506484b03a7a647d5a080/src/types/native.ts#L169C12-L169C26
where the `DatalogFnExpr.binding` expects a `DatalogBinding`, which does not include constants.

maparent:
Ok, here are experiments and reflexions on the clause optimizer in `fireQuery.ts`.
The experiments are based on a query triggered by discourse context on

https://roamresearch.com/#/app/plugin-testing-akamatsulab/page/VKvPktVnD

Here is an instance of a compiled function, with clauses ordered as I pass it to fireQuery.
I remove the function Uid prefix for clarity.

```
[:find
  (pull ?Evidence [:block/string :node/title :block/uid])
  (pull ?Evidence [:block/uid])
:where
  [?relpage :block/uid "L8UGMqGnz"]
  [?rel :block/page ?relpage]
  [?rel :block/props ?relprops]
  [(get ?relprops :discourse-graph) ?reldata]
  [(get ?reldata :hasSchema) ?relSchema]
  [(get ?reldata :sourceUid) ?relSource]
  [(get ?reldata :destinationUid) ?relTarget]
  [(= ?relTarget "W0oEwCbzb")]
  [?Evidence :block/uid ?relSource]
  (or [(= ?relSchema "WRCE-4nr9")]
      [(= ?relSchema "WRCE-4nr9")])
]
```

(works perfectly)

Here is the reordering done by the optimizer in its current form.

```
[:find
  (pull ?Evidence [:block/string :node/title :block/uid])
  (pull ?Evidence [:block/uid])
:where
  [?relpage :block/uid "L8UGMqGnz"]
  [?rel :block/page ?relpage]
  [?rel :block/props ?relprops]
  [(get ?relprops :discourse-graph) ?reldata]
  [?Evidence :block/uid ?relSource]
  (or [(= ?relSchema "WRCE-4nr9")]
      [(= ?relSchema "WRCE-4nr9")])
  [(get ?reldata :hasSchema) ?relSchema]
  [(get ?reldata :sourceUid) ?relSource]
  [(get ?reldata :destinationUid) ?relTarget]
  [(= ?relTarget "W0oEwCbzb")]
]
```

This fails with this error message:
```clojure.lang.ExceptionInfo: Insufficient bindings: #{?relSchema} not bound in [(= ?relSchema "WRCE-4nr9")] {:error :query/where, :form [(= ?relSchema "WRCE-4nr9")], :vars #{?relSchema}}```

Indeed, the or comes before the variable is bound.

My first attempt was a slightly hacky tinker: It is known in logic that or and not are more costly than and. (Line 111)

```
[:find
  (pull ?Evidence [:block/string :node/title :block/uid])
  (pull ?Evidence [:block/uid])
:where
  [?relpage :block/uid "L8UGMqGnz"]
  [?rel :block/page ?relpage]
  [?rel :block/props ?relprops]
  [(get ?relprops :discourse-graph) ?reldata]
  [?Evidence :block/uid ?relSource]
  [(get ?reldata :hasSchema) ?relSchema]
  [(get ?reldata :sourceUid) ?relSource]
  [(get ?reldata :destinationUid) ?relTarget]
  [(= ?relTarget "W0oEwCbzb")]
  (or [(= ?relSchema "WRCE-4nr9")]
      [(= ?relSchema "WRCE-4nr9")])
]
```

And indeed this works fine.

Today I made more experiments, and added a bit of code that considers the last variable of a function-expression as bound. (Line 161). But if I apply it without the or-reorder, I got this:

```
[:find
  (pull ?Evidence [:block/string :node/title :block/uid])
  (pull ?Evidence [:block/uid])
:where
  [?relpage :block/uid "L8UGMqGnz"]
  [?rel :block/page ?relpage]
  [?rel :block/props ?relprops]
  [(get ?relprops :discourse-graph) ?reldata]
  [(get ?reldata :hasSchema) ?relSchema]
  (or [(= ?relSchema "WRCE-4nr9")]
      [(= ?relSchema "WRCE-4nr9")])
  [(get ?reldata :sourceUid) ?relSource]
  [?Evidence :block/uid ?relSource]
  [(get ?reldata :destinationUid) ?relTarget]
  [(= ?relTarget "W0oEwCbzb")]
]
```

Crash. message: "took too long to run Please request less data or add limits. You might want to add limited recursion limits to pull specs: https://docs.datomic.com/pro/query/pull.html#recursive-specifications ."


If I do both, it fails even more brutally:

```
[:find
  (pull ?Evidence [:block/string :node/title :block/uid])
  (pull ?Evidence [:block/uid])
:where
  [?relpage :block/uid "L8UGMqGnz"]
  [?rel :block/page ?relpage]
  [?rel :block/props ?relprops]
  [(get ?relprops :discourse-graph) ?reldata]
  [(get ?reldata :hasSchema) ?relSchema]
  (or [(= ?relSchema "WRCE-4nr9")]
      [(= ?relSchema "WRCE-4nr9")])
  [(get ?reldata :sourceUid) ?relSource]
  [?Evidence :block/uid ?relSource]
  [(get ?reldata :destinationUid) ?relTarget]
  [(= ?relTarget "W0oEwCbzb")]
]
```

Crash. message: "java.lang.OutOfMemoryError: Java heap space"

I realized that my binding alteration forced the optimizer ot follow the order of my original binding clauses:

```
  [(get ?reldata :hasSchema) ?relSchema]
  [(get ?reldata :sourceUid) ?relSource]
  [(get ?reldata :destinationUid) ?relTarget]
```

Ideally this should be combined with giving a lower cost to equality (line 129) but without reordering the Or, I still get a crash.
```
[:find
  (pull ?Evidence [:block/string :node/title :block/uid])
  (pull ?Evidence [:block/uid])
:where
  [?relpage :block/uid "L8UGMqGnz"]
  [?rel :block/page ?relpage]
  [?rel :block/props ?relprops]
  [(get ?relprops :discourse-graph) ?reldata]
  [(get ?reldata :hasSchema) ?relSchema]
  (or [(= ?relSchema "WRCE-4nr9")]
      [(= ?relSchema "WRCE-4nr9")])
  [(get ?reldata :sourceUid) ?relSource]
  [?Evidence :block/uid ?relSource]
  [(get ?reldata :destinationUid) ?relTarget]
  [(= ?relTarget "W0oEwCbzb")]
]
```

Reordering this to put the hasSchema at the end helps, of course. But it would be interesting to be able to order those clauses according to the downstream cost... certainly out of scope now. I just put the relSchema at the end to make things easier.

In all cases: Those ors are useless in this case. (Not in all cases; I am quite sure there are cases where asking for a relation name will give more than one relation Id. But I did not find a good example of that.) Still, I simplified the queries by detecting the case of repeated Ids.

The current code has all the optimizations, which I believe are valid ; but of course we see that they have impact anad we should measure it on more than one sample.

**Comment by maparent** [comment]
Ah, one more note: You noted that the query results came one after another. The queries are launched in parallel, but I think there is queuing happening server-side which yields a linearity.

And a question: You said that using `data.async.q` would not be helpful to the UI. But I remember you asking Sid to do so, what was the rationale then? Or is it after he did that that you realized that it was not fully helpful?

**Comment by maparent** [comment]
This is now at the stage where it's functional and worth discussing. I do expect discussion to happen, there's a lot there.

**Comment by mdroidian** [comment]
> And a question: You said that using `data.async.q` would not be helpful to the UI. But I remember you asking Sid to do so, what was the rationale then? Or is it after he did that that you realized that it was not fully helpful?

Sorry, I cannot recall the specific instance you are referring too, and without the full context I would just be guessing.

**Review by mdroidian — CHANGES_REQUESTED** [review]
Please add the required loom video.

File: apps/roam/src/utils/createReifiedBlock.ts [hunk 3]
```diff
@@ -6,10 +6,19 @@ import { getSetting } from "~/utils/extensionSettings";
 
 export const DISCOURSE_GRAPH_PROP_NAME = "discourse-graph";
 
+const SANE_ROLE_NAME_RE = new RegExp(/^[a-zA-Z][a-zA-Z0-9_-]*$/);
```
mdroidian: Why are we validating via this regex?  I don't believe Roam uid's are required to start with a letter.  (can be a number, dash, underscore, maybe more)

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -6,10 +6,19 @@ import { getSetting } from "~/utils/extensionSettings";
...
+
 const queryForReifiedBlocksUtil = async (
   parameterUids: Record<string, string>,
 ): Promise<[string, Record<string, string>][]> => {
   const paramsAsSeq = Object.entries(parameterUids);
+  // sanitize parameter names
```
mdroidian: This is validating, not sanitizing.  (But we should probably sanitize instead of throw an error)

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -86,7 +95,7 @@ export const createReifiedBlock = async ({
 const RELATION_PAGE_TITLE = "roam/js/discourse-graph/relations";
 let relationPageUid: string | undefined = undefined;
 
-const getRelationPageUid = async (): Promise<string> => {
+export const getOrCreateRelationPageUid = async (): Promise<string> => {
```
mdroidian: Why is this exported?  What's the use case as it relates to to this specific task?

apps/roam/src/utils/fireQuery.ts
```diff
@@ -125,7 +126,8 @@ const optimizeQuery = (
             (a) => a.type !== "variable" || capturedVariables.has(a.value),
           )
         ) {
-          score = 1000;
+          // equality is almost as good as a binding
+          c.type == "pred-expr" && c.pred == "=" ? (score = 5) : (score = 1000);
```
mdroidian, line 130: > This comparison appears to be unintentional because the types '"<" | ">" | "re-matches" | "re-find" | "clojure.string/includes?" | "clojure.string/ends-with?" | "clojure.string/starts-with?"' and '"="' have no overlap.ts(2367)

apps/roam/src/utils/fireQuery.ts
```diff
@@ -107,7 +107,8 @@ const optimizeQuery = (
         if (Array.from(allVars).every((v) => capturedVariables.has(v))) {
           score = 10;
         } else {
-          score = 100002;
+          // downgrade disjunction and negation
+          score = c.type === "and-clause" ? 100002 : 100006;
```
mdroidian, line 111: Are we confident that these changes (here and at line 130/160) won’t impact other queries globally?
If not, that would be pretty concerning, since this part of the planner affects all clause ordering.

What safeguards could we put in place?

apps/roam/src/utils/registerDiscourseDatalogTranslators.ts
```diff
@@ -372,217 +409,586 @@ const registerDiscourseDatalogTranslators = () => {
...
-              forward,
-              source: relationSource,
-              destination: relationTarget,
-            }) => {
-              const sourceTriple = triples.find((t) => t[2] === "source");
-              const targetTriple = triples.find(
-                (t) => t[2] === "destination" || t[2] === "target",
+          if (!filteredRelations.length && !ANY_RELATION_REGEX.test(label))
+            return [];
+          console.log("source", source, "target", target);
```
mdroidian: rm console.log

apps/roam/src/utils/registerDiscourseDatalogTranslators.ts
```diff
@@ -372,217 +409,586 @@ const registerDiscourseDatalogTranslators = () => {
...
+              let forwardRelationIds = filteredRelations
+                .filter((r) => r.forward)
+                .map((r) => r.id);
+              forwardRelationIds = [...new Set(forwardRelationIds)];
+              const forwardRelationClauses = forwardRelationIds.map(
+                (id) =>
+                  ({
+                    type: "pred-expr",
+                    pred: "=",
+                    arguments: [
```
mdroidian, line 769: There are quite a few TS error, like this one:
> Conversion of type '{ type: "pred-expr"; pred: "="; arguments: ({ type: "variable"; value: string; } | { type: "constant"; value: string; })[]; }' to type 'DatalogClause' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type '{ type: "pred-expr"; pred: "="; arguments: ({ type: "variable"; value: string; } | { type: "constant"; value: string; })[]; }' is not comparable to type 'DatalogPredExpr'.
    Types of property 'pred' are incompatible.
      Type '"="' is not comparable to type '"<" | ">" | "re-matches" | "re-find" | "clojure.string/includes?" | "clojure.string/ends-with?" | "clojure.string/starts-with?"'.ts(2352)

Let's go ahead and clean those up.

apps/roam/src/utils/registerDiscourseDatalogTranslators.ts
```diff
@@ -372,217 +409,586 @@ const registerDiscourseDatalogTranslators = () => {
...
+                  variable: { type: "variable", value: "relSchema" },
+                },
+              },
+            ];
+            const typeOfSource = typeOfValue(source);
+            const typeOfTarget = typeOfValue(target);
+            const clauses: DatalogClause[] = [...relClauseBasis];
+
+            // todo: It could be a title or a node type.
+            console.log(
```
mdroidian: rm console.log

mdroidian (changes requested):

apps/roam/src/utils/createReifiedBlock.ts
maparent: That was a coderabbit suggestion. I think some sanitization is good, so I will broaden the regexp, but happy to remove if you feel it's not necessary.

apps/roam/src/utils/createReifiedBlock.ts
maparent: Calling it validate for now; I think that sanitization might create discrepancies with caller's expectations and more obscure errors, I think failing early is better.
That said, you also have a point that throwing an error without helping the user is not terribly useful either.
On the other other hand, I do think that user-side error handling is scope creep.
I mean... from my pov, none of this is known to happen, I was just trying to make the rabbit not complain. 
BUT if it happens, we want to know about it, and throwing errors should lead us to see something in birdseatbugs, so I take that as a win.

apps/roam/src/utils/createReifiedBlock.ts
maparent: I do not remember, I must have used it in a change that got reverted. removing.

apps/roam/src/utils/fireQuery.ts
maparent: This looks like you have the older version of `roamjs-components`?

apps/roam/src/utils/fireQuery.ts
maparent: As discussed: I am relatively confident that the changes do not affect most existing queries currently being generated.
It is known that the new queries fail without these changes, sometimes with a Java heap space issue. It is theoretically possible that some new such failures will happen. Initial tests do not show this, but I cannot be sure of coverage. 
The only 100% secure option I can see is actually create code that finds all query blocks in a given graph, and then runs them. That may be worth doing in all cases.

apps/roam/src/utils/registerDiscourseDatalogTranslators.ts
maparent: No ts error on my side, again I suspect you should update `roamjs-components`

apps/roam/src/utils/createReifiedBlock.ts
mdroidian: Roam probably handles this, we should let them do it with a simple `try`.  Unless there is a specific case in which we are arbitrarily creating UID's, it is all being controlled by them, no need for us to haphazardly add a check here.

apps/roam/src/utils/createReifiedBlock.ts
mdroidian: A simple try/catch would suffice.  Let Roam handle this, we use their functions to create UID's, create blocks, etc, etc.  They are in charge of what is validated and what isn't, not us.  We are just adding unnecessary complexity and risking false positive/negatives here.

apps/roam/src/utils/registerDiscourseDatalogTranslators.ts
```diff
@@ -372,217 +409,573 @@ const registerDiscourseDatalogTranslators = () => {
...
+                  type: "bind-scalar",
+                  variable: { type: "variable", value: "relSchema" },
+                },
+              },
+            ];
+            const typeOfSource = typeOfValue(source);
+            const typeOfTarget = typeOfValue(target);
+            const clauses: DatalogClause[] = [...relClauseBasis];
+
+            // todo: It could be a title or a node type.
```
mdroidian: What is this todo waiting for?  Does this still need to be completed for this PR?

mdroidian (approved):

apps/roam/src/utils/registerDiscourseDatalogTranslators.ts
maparent: Ah, the todo is a leftover, thanks for spotting it.

PR #539, opened 2025-11-10
ENG-1044 Experiments with different datalog query modalities

maparent:
https://linear.app/discourse-graphs/issue/ENG-1044/experiments-with-different-datalog-query-modalities

This contains code that would allow to use the slow vs fast async catalog queries, if it turns out to have an impact in accuracy after we have a proper testing harness.

### DIFF
--- a/apps/roam/src/utils/fireQuery.ts
+++ b/apps/roam/src/utils/fireQuery.ts
@@ -14,6 +14,7 @@ import predefinedSelections, {
 import { DEFAULT_RETURN_NODE } from "./parseQuery";
 import { DiscourseNode } from "./getDiscourseNodes";
 import { DiscourseRelation } from "./getDiscourseRelations";
+import type { json } from "./getBlockProps";
 import nanoid from "nanoid";
 
 export type QueryArgs = {
@@ -315,6 +316,29 @@ export const fireQuerySync = (args: FireQueryArgs): QueryResult[] => {
   }));
 };
 
+const PROP_NAME_RE = /:[a-zA-Z0-9_-]+\/[a-zA-Z0-9_-]+\b/g;
+const PULL_RE = /\(pull [^)]+\)/g;
+
+const renamePropsInResult = (
+  result: json | null,
+  mapping: Record<string, string>,
+): json | null => {
+  const rename = (x: json | null): json | null => {
+    if (Array.isArray(x)) return x.map(rename);
+    if (x === null || x === undefined) return x;
+    if (typeof x === "object") {
+      return Object.fromEntries(
+        Object.entries(x as object).map(([k, v]) => [
+          mapping[k] || k,
+          rename(v),
+        ]),
+      );
+    }
+    return x;
+  };
+  return rename(result);
+};
+
 const fireQuery: FireQuery = async (_args) => {
   const { isCustomEnabled, customNode, local, ...args } = _args;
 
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

maparent:
Copying linear comment:

I think this could go as-is, but I can think of two things you may want:

Copy in the migration panel to tell people to wait, and expect an unresponsive UI

We could even have a progression bar, as in the export?

Right now I'm using the local async function, and as I said did not run into issues. You seemed to expect issues in that case, may be worth digging into where you saw them, since right now cannot reproduce.

Not fully related: when the backend query fails and it retries locally… it is internal to Roam, but it would be nice to capture this and send it to posthog. We could substitute for the error handler, but that's invasive. Can you see a way I cannot?

maparent:
Loom video showing that local queries work fine (on test-plugin-akamatsulab2):
https://www.loom.com/share/f2ddf64a9d964027ada08980cc9a5dcc

mdroidian:
> Not fully related: when the backend query fails and it retries locally… it is internal to Roam, but it would be nice to capture this and send it to posthog. We could substitute for the error handler, but that's invasive. Can you see a way I cannot?

Sorry, I don't follow.  We could substitute what for the error handler?

mdroidian:
@CodeRabbitAI full review

mdroidian:
> Copy in the migration panel to tell people to wait, and expect an unresponsive UI

Yes.  But this can be a separate task/PR

> We could even have a progression bar, as in the export?

Probably not worth the effort

apps/roam/src/utils/fireQuery.ts
```diff
@@ -347,11 +371,31 @@ const fireQuery: FireQuery = async (_args) => {
...
     }
 
-    //@ts-ignore - todo add async q to roamjs-components
-    const queryResults = await window.roamAlphaAPI.data.backend.q(
-      query,
-      ...inputs,
-    );
+    let queryResults: unknown[][] = [];
+    if (local) {
+      // look for propNames in query. Could consider looking only in pull when that exists.
```
mdroidian: Could you clarify what this is looking for?  What does `propNames` refer to and/or why are we checking for it?

apps/roam/src/utils/fireQuery.ts
```diff
@@ -347,11 +371,31 @@ const fireQuery: FireQuery = async (_args) => {
...
+      const propNames = new Set(
+        [...query.matchAll(PROP_NAME_RE)].map((m) => m[0]),
+      );
+      const propNamesSub: Record<string, string> = Object.fromEntries(
+        [...propNames].map((n) => [n.split("/")[1], n]),
+      );
+      if (Object.keys(propNamesSub).length === propNames.size) {
+        // no name conflict, safe to use async query
```
mdroidian, line 402: Could you give an example of a name conflict that would be unsafe to use in async query?

apps/roam/src/utils/fireQuery.ts
```diff
@@ -347,11 +371,31 @@ const fireQuery: FireQuery = async (_args) => {
...
+        // BUT it returns non-namespaced names, so substitute prop names back
+        queryResults = await window.roamAlphaAPI.data.async.q(query, ...inputs);
+        queryResults = renamePropsInResult(
+          queryResults as json,
+          propNamesSub,
+        ) as unknown[][];
+      } else {
+        // more janky but safer
```
mdroidian: What is more janky?  Safer than what?

apps/roam/src/utils/fireQuery.ts
```diff
@@ -347,11 +371,31 @@ const fireQuery: FireQuery = async (_args) => {
...
+        queryResults = window.roamAlphaAPI.data.fast.q(query, ...inputs);
+      }
+    } else {
+      // eslint-disable-next-line @typescript-eslint/await-thenable
+      queryResults = await window.roamAlphaAPI.data.backend.q(query, ...inputs); // ts-ignore
```
mdroidian: What is the purpose of `// ts-ignore` here?

mdroidian (changes requested):
A few clarifying questions.  Also, I ran coderabbit to check the new changes, could you please address it's comment?

apps/roam/src/utils/fireQuery.ts
maparent: Ok. The rationale for this was that async queries, unlike backend (or fast) queries, return non-namespaced variable names for pull clauses. So I was checking for the potential of collision there. (`{log,version,window}/id; {graph,user}/settings; {block,user}/uid; {create,edit}/time`). BUT after some time spent improving that code, I realized that the `data.async.fast.q` has the same namespace characteristics as backend. So I could just use that and avoid a lot of code. BUT again... some queries fail in backend, maybe they also fail in fast? SO... proposing to keep the code around until we have a way to do global query testing.

apps/roam/src/utils/fireQuery.ts
maparent: pull variables that can collide: `{log,version,window}/id; {graph,user}/settings; {block,user}/uid; {create,edit}/time`

apps/roam/src/utils/fireQuery.ts
maparent: non-async makes for janky ui. Since backend queries sometimes fail (besides usage issues), and use fast, we may assume that fast queries are more fragile than non-fast.

apps/roam/src/utils/fireQuery.ts
maparent: ts was complaining about the unneeded await. Moot now that backend is async.

apps/roam/src/utils/fireQuery.ts
maparent: (Keeping the code around in a way that it is never run; async.fast is the simplest case, and using it for now.)

maparent:
@coderabbit review

mdroidian (changes requested):
I want to clarify intent here, since I think there was some understandable misalignment.

This PR originally proposed broader query changes, which were out of scope at the time. In the recent product meeting, when the reified relations migration came up and we discussed a `local` flag, what I meant specifically was:

> we can add a **`local` flag** so the reified-relations migration can run **local queries instead of backend queries**

I did **not** mean to reopen the broader query refactor or introduce additional execution paths.

---

### What needs to change in this PR

For this PR to land, it needs to be reduced to a very narrow scope:

* Add `local?: boolean`
* When `local === true`, use:

  ```ts
  roamAlphaAPI.data.async.fast.q
  ```
* Otherwise, preserve existing backend behavior

Everything else (slow paths, property renaming, regex inspection, future fallbacks, commented logic) should be removed.

---

### Why this matters

The additional logic introduces:

* dead or disabled paths
* speculative future handling
* unnecessary cognitive overhead for reviewers and maintainers

If we later find real issues with `fast.q`, we can address them in a **separate, focused PR** with concrete evidence.

---

### Summary

I’m aligned on the *use case*, and I acknowledge my wording in the meeting may have been ambiguous. That said, this PR still needs to be **significantly narrowed** before merge.

Please cut this down to the minimal `local` flag + `async.fast.q` change (≈10–20 LOC), then re-request review.

Alternatively, if you’d like to keep this as an experimental or exploratory effort, please leave this PR as draft and open a separate ticket specifically for introducing a `local` flag for the reified relations migration use case.

maparent:
I knew that's what you want. I'm pushing back in this case. We have bugs we don't understand yet, we can't track them that well, and until we do, we have actually reasons to believe that this currently dead code *might* turn out to be useful. You've said a few times it's just something that can live in git, but with our policy of squashing and killing branches, if I don't leave it in the code it's gone from git. (I could keep the code privately, but that means nobody else can discover that code.) I'm aware it's making your life a bit more difficult, but we shared the decision to delay the testing infrastructure, and until it exists there are decisions we do not know enough to take, and I think they include killing that code.

maparent:
That said… I agree that I can leave this as a draft in a separate PR.

maparent:
Made it into a separate PR as requested; I agree it's a cleaner separation of concerns. I will keep the branch around.

PR #554, opened 2025-11-18
[ENG-1052] Enable node tags via command key

trangdoan982:
https://www.loom.com/share/0c61409741e94d818c4ad3c87aa6f1e8

## Release Notes

* **New Features**
  * Configurable hotkey to open node tag suggestions while editing.
  * Interactive node tag suggestion panel with keyboard and mouse navigation to insert tags into your notes.


trangdoan982:
@coderabbitai full review

apps/obsidian/src/components/GeneralSettings.tsx
```diff
@@ -262,6 +274,36 @@ const GeneralSettings = () => {
...
+                e.preventDefault();
+                handleNodeTagHotkeyChange(e.key);
+              } else if (e.key === "Backspace") {
+                handleNodeTagHotkeyChange("");
+              }
+            }}
+            placeholder="\\"
+            maxLength={1}
+            className="setting-item-control"
+            style={{ width: "60px" }}
```
mdroidian: tailwind

apps/obsidian/src/index.ts
```diff
@@ -185,6 +188,40 @@ export default class DiscourseGraphPlugin extends Plugin {
         });
       }),
     );
+
+    // Register editor keydown listener for node tag hotkey
+    this.setupNodeTagHotkey();
+  }
+
+  private setupNodeTagHotkey() {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
```
mdroidian: 👀  Is there a specific reason we need this?  Ideally we can write the code that wouldn't need to comment out these errors, or if there is some way we can't avoid it, let's document why.

Generally errors like this are trying to tell us something important.

apps/obsidian/src/components/NodeTagSuggestModal.tsx
```diff
@@ -0,0 +1,335 @@
...
+      console.error("Error getting cursor position:", error);
+      return null;
+    }
+  }
+
+  private createPopover(): HTMLElement {
+    const popover = document.createElement("div");
+    popover.className = "node-tag-suggest-popover";
+    popover.style.cssText = `
+      position: fixed;
```
mdroidian: Would be nice to use tailwind wherever possible, when possible.  Tailwind has parts of a design system build directly into it by default.  If not, that's fine, just curious.

apps/obsidian/src/components/NodeTagSuggestModal.tsx
```diff
@@ -0,0 +1,335 @@
+import { Editor } from "obsidian";
```
mdroidian, line 1: Looks like we are mostly creating this from scratch?  Are there any existing obsidian components we can tie into without having to re-invent the wheel, cover all edge cases, etc?  List items, popovers, etc, etc.

apps/obsidian/package.json
```diff
@@ -16,6 +16,8 @@
   "author": "",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@codemirror/state": "^6.5.2",
```
mdroidian: Where is this being used?

mdroidian (approved):
Approved to not block, but I strongly recommend we understand and/or fix this issue:

>     // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access

apps/obsidian/src/components/NodeTagSuggestModal.tsx
trangdoan982: as far as i know there's no official API to create popovers at tooltip position. The popover API they have are the one that takes over the whole screen

PR #558, opened 2025-11-19
Eng 1062 optimize all relations

maparent:
Use a pseudo-relation for a query for all relations. Rework the algorithm for grouping results, which means adding some selections. 

https://www.loom.com/share/ced1e4cf67ec4acdb97db690431ee9a0

## Summary by CodeRabbit

* **New Features**
  * Added support for broader relation queries, including a new “any relation” mode.
  * Query results can now include extra fields for relation context and map them into richer output.

* **Bug Fixes**
  * Improved how discourse context results are grouped and returned, keeping related items together more reliably.
  * Made result mapping more robust when handling non-empty structured values.

* **Tests**
  * Added coverage for the new query variable handling and updated result grouping behavior.

### DIFF
--- a/apps/roam/src/utils/getDiscourseContextResults.ts
+++ b/apps/roam/src/utils/getDiscourseContextResults.ts
@@ -7,6 +7,9 @@ import getDiscourseRelations, {
   DiscourseRelation,
 } from "./getDiscourseRelations";
 import { Selection } from "./types";
+import { getSetting } from "./extensionSettings";
+import { ANY_RELATION_REGEX } from "./deriveDiscourseNodeAttribute";
+import { use } from "cytoscape";
 
 const resultCache: Record<string, Awaited<ReturnType<typeof fireQuery>>> = {};
 const CACHE_TIMEOUT = 1000 * 60 * 5;
@@ -55,6 +58,18 @@ const buildSelections = ({
       uid: window.roamAlphaAPI.util.generateUID(),
       label: "anchor",
       text: `node:${conditionUid}-Anchor`,
+    });
+  }
+  if (ANY_RELATION_REGEX.test(r.label)) {
+    selections.push({
+      uid: window.roamAlphaAPI.util.generateUID(),
+      label: "relationUid",
+      text: "hasSchema",
+    });
+    selections.push({
+      uid: window.roamAlphaAPI.util.generateUID(),
+      label: "effectiveSource",
+      text: "effectiveSource",
     });
   }
 
@@ -186,6 +201,10 @@ const getDiscourseContextResults = async ({
 
   const discourseNode = findDiscourseNode({ uid: targetUid });
   if (!discourseNode) return [];
+  const useReifiedRelations = getSetting<boolean>(
+    "use-reified-relations",
+    false,
+  );
   const nodeType = discourseNode?.type;
   const nodeTextByType = Object.fromEntries(
     nodes.map(({ type, text }) => [type, text]),
@@ -217,9 +236,24 @@ const getDiscourseContextResults = async ({
   });
 
   const relationsWithComplement = Array.from(uniqueRelations.values());
+  const queryRelations = useReifiedRelations
+    ? [
+        {
+          r: {
+            id: "null",
+            complement: "Has Any Relation To",
+            label: "Has Any Relation To",
+            triples: [],
+            source: "*",
+            destination: "*",
+          },
+          complement: false,
+        },
+      ]
+    : relationsWithComplement;
 
   const context = { nodes, relations };
-  const queryConfigs = relationsWithComplement.map((relation) =>
+  const queryConfigs = queryRelations.map((relation) =>
     buildQueryConfig({
       args,
       targetUid,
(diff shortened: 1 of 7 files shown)


apps/roam/src/utils/getDiscourseContextResults.ts
```diff
@@ -51,6 +53,18 @@ const buildSelections = ({
       text: `node:${conditionUid}-Context`,
     });
   }
+  if (ANY_RELATION_REGEX.test(r.label)) {
```
maparent: Detecting the "all relations" case

apps/roam/src/utils/getDiscourseContextResults.ts
```diff
@@ -211,9 +229,24 @@ const getDiscourseContextResults = async ({
   });
 
   const relationsWithComplement = Array.from(uniqueRelations.values());
+  const queryRelations = useReifiedRelations
```
maparent, line 237: Create a new case for all relations

apps/roam/src/utils/getDiscourseContextResults.ts
```diff
@@ -226,12 +259,49 @@ const getDiscourseContextResults = async ({
...
   );
 
-  const resultsWithRelation = await executeQueries(
+  let resultsWithRelation = await executeQueries(
     queryConfigs,
     targetUid,
     nodeTextByType,
     onResult,
   );
+  if (
```
maparent, line 276: Since we look at all relations at once, we have to re-group by relation type here.

apps/roam/src/utils/predefinedSelections.ts
```diff
@@ -285,6 +287,26 @@ const predefinedSelections: PredefinedSelection[] = [
     },
     suggestions: EDIT_BY_SUGGESTIONS,
   },
+  {
+    test: HAS_SCHEMA_TEST,
+    pull: () => {
+      return "?relSchema";
+    },
+    mapper: () => {
+      // not sure here?
```
maparent: The results are the same between both versions, so I think I got it right, but I'll admit to a bit of guesswork here.

mdroidian:
@maparent Devin flagged a few issues here. Could you take a look and address them, then re-tag me once they’re resolved?

apps/roam/src/utils/getDiscourseContextResults.ts
```diff
@@ -211,9 +229,24 @@ const getDiscourseContextResults = async ({
...
+            id: "null",
+            complement: "Has Any Relation To",
+            label: "Has Any Relation To",
+            triples: [],
+            source: "*",
+            destination: "*",
+          },
+          complement: false,
+        },
+      ]
```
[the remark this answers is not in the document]
maparent: using a constant for Id. I think that the type checker will catch new required fields.

maparent:
@mdroidian resolved the issues.

apps/roam/src/utils/getDiscourseContextResults.ts
```diff
@@ -211,9 +230,24 @@ const getDiscourseContextResults = async ({
   });
 
   const relationsWithComplement = Array.from(uniqueRelations.values());
+  const queryRelations = useReifiedRelations
+    ? [
+        {
+          r: {
+            id: ANY_RELATION_ID,
```
mdroidian, line 242: if you add the `id` outside `r` as well, it'll match the type `RelationWithComplement` cleanly

apps/roam/src/utils/getDiscourseContextResults.ts
```diff
@@ -211,9 +230,24 @@ const getDiscourseContextResults = async ({
...
+            complement: "Has Any Relation To",
```
mdroidian: reuse `ANY_RELATION_REGEX.source`

mdroidian (changes requested):
I like the high-level goal here: in reified-relations mode, discourse context should not have to run one query per relation type. A single “all relations touching this node” query, followed by regrouping, is the right direction.

I don’t think we should merge the current implementation shape as-is, though. It feels too brittle/indirect for a core discourse context path.

The main issue is that `hasSchema` and `effectiveSource` are being added as synthetic predefined selections so we can carry internal relation metadata out of the query. Predefined selections are part of the Query Builder/user-facing selection layer; these values are not real user selections, they are internal Datalog variables needed by this one caller. The placeholder `mapper` implementations returning `"?relSchema"` / `"?relSource"` are a signal that this is not really a selection abstraction.  That also forces `relSchema` and `relSource` to be preserved globally in the relation translator.

There are also some correctness issues in the regrouping:

- The complement handling looks backwards. For keys ending in `-true`, the code sets `isComplement: true` but uses `relation.r.label` for `label/text`; for `-false`, it uses `relation.r.complement`. Based on how `uniqueRelations` is built, and based on the existing `text = isComplement ? r.complement : r.label` logic, those should be flipped.
- `id` becomes the grouped key like `${relationUid}-${direction}` instead of the underlying relation schema id. That changes the meaning of the result id from relation id to implementation grouping key. Direction already has `isComplement`, so the grouped key should not escape as `id`.
- `onResult` behavior changed. Previously `executeQueries` received `onResult` and could call it per relation query as each query completed; now results are buffered, empty groups are filtered, and callbacks fire at the end. That may be intentional, but it is a behavior change outside just the reified-relations optimization and should be tested/called out.

There are a few ways we could take this. I’ll elaborate on two:

- Option 1: add a `findVariables` internal API.
- Option 2: special-case the discourse context query outside of the Query Builder translator.

I put together an example of Option 1 here: https://github.com/DiscourseGraphs/discourse-graph/pull/1160

That keeps the one-query approach and still reuses the existing `Has Any Relation To` translator, but it adds an internal `findVariables` API to `fireQuery` instead of adding fake predefined selections. So discourse context can explicitly ask for the scoped `relSchema` / `relSource` variables it needs, without exposing them as Query Builder selections.

Option 2 would be to special-case discourse context directly: since the discourse context overlay is the hot path here, we could build a direct reified-relation Datalog query for this flow and map the stored relation metadata into groups ourselves. That would be less reusable with Query Builder, but it may be the clearest implementation if this optimization is specifically for discourse context.

My recommendation is that we should not keep the fake predefined-selection approach. I’d be comfortable with either Option 1 from the example PR, or Option 2 as a dedicated discourse-context query path, but the current version is coupling too many layers in a brittle way.

maparent:
Ok. I did not read the PR yet, but I like option 1 as a direction and would be happy to take over from your PR. (I would start later today or even tomorrow morning.) If you feel you started it and want to finish it, please let me know.

apps/roam/src/utils/getDiscourseContextResults.ts
maparent: defined name as distinct from regex, and regex from name. I think it's cleaner than extracting from regex.

maparent:
Ok. Status: I corrected the nits.
In the case of onResults, the batching now happens only in the reified case, which should be fine. (This was originally raised by Devin [here](https://github.com/DiscourseGraphs/discourse-graph/pull/558#discussion_r3460641859).)
I looked at your branch, and it looks complete to me, nothing I can add. I would merge it in this one if you're ok with that process.
It seems to me that your branch handles most of the things you asked for, except for what I just did; Let me know if I missed something, but I think I'm done.

mdroidian:
@mdroidian pushed `cc45bf26` with the Option 1 direction from #1160.

Changes made:
- Added internal `findVariables` support to `fireQuery` for `relationUid` / `effectiveSource` metadata.
- Removed the synthetic `hasSchema` and `effectiveSource` predefined selections.
- Removed global `relSchema` / `relSource` preservation from the relation translator; discourse context now asks for the scoped internal variables explicitly.
- Updated reified regrouping to use the original `RelationWithComplement` metadata for `id`, label/text, target type, and complement direction.
- Kept reified `onResult` batching after regrouping so callback consumers receive correct relation labels.
- Added tests for internal find variables and reified regrouping across source-side and complement-side relations.

Verified locally:
- `pnpm --filter roam test`
- `pnpm --filter roam check-types`
- `pnpm --filter roam lint` (0 errors; existing warnings remain)

mdroidian:
@coderabbitai full review

PR #559, opened 2025-11-21
ENG-1019 Add Manual Relation Creation Flow from Source Node Discourse Context

maparent:
https://linear.app/discourse-graphs/issue/ENG-1019/add-manual-relation-creation-flow-from-source-node-discourse-context

## Summary by CodeRabbit

* **New Features**
  * Create Relation dialog to pick a relation type, search/select target pages, and create relations with success/error toasts.
  * Create Relation button added to context panel header and empty-state views for quick access.
  * Improved node-matching and format helpers to better find and interpret discourse nodes.

* **Style**
  * Updated callout icon styling and ensured popovers can display full width.

* **Refactor**
  * Context view now forces fresh results on open and uses a short delayed refresh after relation creation.


maparent:
@coderabbit review

maparent:
@coderabbit review

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,311 @@
...
+      formatToRegexpText(
+        nodesById[rel.forward ? rel.destination : rel.source].format,
+      ),
+    );
+    const re = RegExp(`^(${formats.join(")|(")})$`);
+    return allPages.filter((title) => title.match(re));
+  };
+  const [pageOptions, setPageOptions] = useState<string[]>(
+    getFilteredPageNames(relKeys[0]),
+  );
```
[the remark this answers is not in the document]
maparent: There are a lot of guards ensuring this never happens, notably checking for relData.length in extendProps ; and returning null at that stage means we cannot use the function in the renderOverlay (which expects a function that always returns a component.)

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,311 @@
...
+            </Label>
+          </div>
+          <div>
+            <AutocompleteInput
+              value={selectedTargetTitle}
+              setValue={getNodeFromTitle}
+              onBlur={getNodeFromTitle}
+              placeholder={"Search for a page..."}
+              options={pageOptions}
+            />
```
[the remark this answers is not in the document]
maparent: I need the same behaviour in either case. It is true that there is duplicate work in the select case. I added a guard thus:
`if (title === selectedTargetTitle) return;`
This branch seems to be called when appropriate.

maparent:
@coderabbit review

mdroidian (commented):
> First implementation. Still have to work on UI look, and to add the callout in the node page (it's currently in the discourse context.) But functionality is there.

Is this still the case?

maparent:
> > First implementation. Still have to work on UI look, and to add the callout in the node page (it's currently in the discourse context.) But functionality is there.
> 
> Is this still the case?

no, forgot to edit the description after un-drafting.

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,326 @@
+import React, { useState, useMemo } from "react";
+import { Dialog, Classes, Label, Button, Icon } from "@blueprintjs/core";
+import renderOverlay from "roamjs-components/util/renderOverlay";
+import { render as renderToast } from "roamjs-components/components/Toast";
+import MenuItemSelect from "roamjs-components/components/MenuItemSelect";
+import AutocompleteInput from "roamjs-components/components/AutocompleteInput";
+import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageUid";
+import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
+import getAllPageNames from "roamjs-components/queries/getAllPageNames";
+
```
mdroidian: .

apps/roam/src/styles/styles.css
```diff
@@ -114,6 +114,30 @@
...
+}
+
+.dg-detached-header .bp3-icon {
+  color: #c0d8e6;
+  background-color: #2c5f95;
+  padding: 1px;
+  margin-left: 0.7ex;
+}
+
+.full-popover-width .bp3-popover-target {
```
mdroidian: 1) This looks like it is targeting some generic BP3 component, so it will likely target more than just what you want it to and have negative consequenses.  Let's try to make it more specific with a class or id of the component you are targeting.
2) What is this meant to target?  If the class or id isn't obvious, let's leave a breadcrumb in the comment to be clear.

apps/roam/src/styles/styles.css
```diff
@@ -114,6 +114,30 @@
   border-bottom: 1px solid rgba(16, 22, 26, 0.15);
 }
 
+.roamjs-create-relation-dialog-body {
+  background-color: #e8eff3;
+}
+
+.dg-detached-header {
```
mdroidian: nit: most of this would be better as co-located inline tailwind classes.
One of the major benefits of why we do that is to enforce a design system.  Eg: padding 5px would be p-1 or p2 which would be 4px or 8px and align with more of our components.

apps/roam/src/utils/getDiscourseNodeType.ts
```diff
@@ -0,0 +1,48 @@
+import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageUid";
+import getDiscourseNodes, { type DiscourseNode } from "./getDiscourseNodes";
+import "core-js/proposals/regexp-escaping";
+import format from "date-fns/esm/fp/format/index.js";
```
mdroidian: unused

apps/roam/src/utils/getDiscourseNodeType.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// eslint-disable-next-line
```
mdroidian: Why is this necessary?  Using `any` should be avoided at all costs.

apps/roam/src/utils/getDiscourseNodeType.ts
```diff
@@ -0,0 +1,48 @@
+import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageUid";
+import getDiscourseNodes, { type DiscourseNode } from "./getDiscourseNodes";
+import "core-js/proposals/regexp-escaping";
```
mdroidian: Let's re-use existing solutions for this so the code is consistent: `getDiscourseNodeFormatExpression`.  Also, this is type safe and doesn't add additional, experimental, dependencies.

apps/roam/src/utils/getDiscourseNodeType.ts
```diff
@@ -0,0 +1,48 @@
...
+      console.warn(
+        `Multiple matches for ${title}: ${formatMatches.map((e) => e.format).join(", ")}`,
+      );
+    }
+    return formatMatches[0];
+  }
+  return null;
+};
+
+export const getDiscourseNodeTypeByUid = (
```
mdroidian: Unused, and functionality is already in findDiscourseNode().  Let's delete this.

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,326 @@
...
+  };
+  const [pageOptions, setPageOptions] = useState<string[]>(
+    getFilteredPageNames(relKeys[0]),
+  );
+
+  const identifyRelationMatch = (
+    targetTitle: string,
+  ): RelWithDirection | null => {
+    if (targetTitle.length === 0) return null;
+    const selectedTargetType = getDiscourseNodeTypeByTitle(
```
mdroidian: It looks like you already have `uid` at this point, so there is no need to create a new function (`getDiscoursenodeTypeByTitle`), we can use the existing `findDiscourseNode` function that is standard in the codebase.

apps/roam/src/utils/getDiscourseNodeType.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// eslint-disable-next-line
+const reEscape = (RegExp as any).escape as (str: string) => string;
+
+const SUB_RE = new RegExp(/\\\{\w+\\\}/g);
+
+export const formatToRegexpText = (format: string): string =>
+  reEscape(format.trim()).replaceAll(SUB_RE, ".*");
+
+export const getDiscourseNodeTypeByTitle = (
```
mdroidian: As noted in another comment, let's use `findDiscourseNode` to be consistent with the code base.

apps/roam/src/utils/getDiscourseNodeType.ts
```diff
@@ -0,0 +1,48 @@
+import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageUid";
```
mdroidian: It looks like we should be able to delete this file entirely. 👍

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,326 @@
...
+      targetTitle,
+      discourseNodes,
+    );
+    if (selectedTargetType === null) {
+      console.error("could not identify the target type");
```
mdroidian: What's the UX for the user on these console.errors and return null?  Is it a silent error?

Generally speaking, if we are reaching for `console.error` we should either do one or both of these:
1 - notify the team (right now we do that will `sendErrorEmail`)
2 - notify the user (right now we do that with `renderToast`)

But in production, `console.error` isn't actually that helpful to the average user, nor the dev that doesn't experience the error.

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,326 @@
...
+                onItemSelect={changeRelationType}
+                className="w-full"
+              />
+            </Label>
+          </div>
+          <div className="full-popover-width">
+            <AutocompleteInput
+              value={selectedTargetTitle}
+              setValue={getNodeFromTitle}
+              onBlur={getNodeFromTitle}
```
mdroidian, line 240: nit: we should just use `uid` here so we can skip a query (`getPageUidByPageTitle()`)
The standard pattern for dealing with pages (or any block query) is querying for `Result[]` which contains text (which `title`), and `uid`

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,326 @@
...
+      </div>
+    </Dialog>
+  );
+};
+
+const prepareRelData = (
+  targetNodeUid: string,
+  nodeTitle?: string,
+): RelWithDirection[] => {
+  nodeTitle = nodeTitle || getPageTitleByPageUid(targetNodeUid).trim();
```
mdroidian, line 269: As mentioned elsewhere, we could probably work directly with the `uid` here

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,326 @@
...
+  formatToRegexpText,
+} from "~/utils/getDiscourseNodeType";
+import getDiscourseNodes from "~/utils/getDiscourseNodes";
+
+export type CreateRelationDialogProps = {
+  onClose: () => void;
+  sourceNodeUid: string;
+};
+
+type RelWithDirection = DiscourseRelation & {
```
mdroidian, line 34: Do we have existing types for this in discourse context or elsewhere to reuse?  Or a similar pattern but inline?  It seems like we should.  If so, let's make sure to share/reuse the type.

mdroidian (changes requested):
Generally, good work. 👍

But I think we are introducing enough duplicate code here that is it worth going back and re-using existing types/function/patterns/etc.

I've listed a few below.  It looks like we can fully remove `getDiscourseNodeType.ts` and keep the codebase more consistent.

Lmk if you have any questions.

apps/roam/src/utils/getDiscourseNodeType.ts
maparent: I remember asking you if this existed and you said no? Anyhow, if something like this exists elsewhere I would appreciate being pointed to it, I looked and never saw it.

mdroidian: [DiscourseGraphs/discourse-graph#559](https://github.com/DiscourseGraphs/discourse-graph/pull/559#discussion_r2586450588)

apps/roam/src/components/CreateRelationDialog.tsx
maparent: Fair; was not aware of sendErrorEmail. Looking forward to having access to those emails myself, tbh. Or should we think of migrating those towards posthog errors?

apps/roam/src/styles/styles.css
maparent: It's targeting the inside of the component. I created the `.full-popover-width` marker class precisely to make the target specific, without side effects.

maparent: (but happy to comment of course)

mdroidian: Let's make the class more specific.  `.full-popover-width` could be referring to a lot of things.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: ?

mdroidian: Remove the space.

apps/roam/src/utils/getDiscourseNodeType.ts
maparent: It was for that code, but that code is gone.

apps/roam/src/utils/getDiscourseNodeType.ts
maparent: I had reserves about performance; but I realize the useful cases are fast-tracked. So I did replace the code. Only difference: I made a variant where the title is pre-loaded, which is the case for me, and avoids a fetch operation.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: But then it fetches the title from the uid. When I have the title, which always makes sense in my case, I think it's worth providing it.

apps/roam/src/utils/getDiscourseNodeType.ts
maparent: same comment.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: Ah! I find it improves intelligibility to separate library imports from local imports. 
I would submit it as a convention.
But removing for now.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: In this case, I do not have the Uid, since `AutocompleteInput` works on pure text, not a `uid->title` map. 
I _could_ build a `title->uid` map, which would improve performance here. Do you think it would be worth the trouble?

apps/roam/src/components/CreateRelationDialog.tsx
maparent: The title just gets fetched downstream.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: Nowhere I could find...

maparent:
@coderabbit review from 6b1522607d471c82fdf8980cb9a04f0a62ec77cc

maparent:
@coderabbit review last commit

maparent:
@coderabbit full review

apps/roam/src/utils/findDiscourseNode.ts
```diff
@@ -16,3 +19,25 @@ const findDiscourseNode = (uid = "", nodes = getDiscourseNodes()) => {
...
+    return discourseNodeTypeCache[uid];
+  }
+
+  const matchingNode = nodes.find((node) =>
+    matchDiscourseNode({ ...node, title }),
+  );
+
+  discourseNodeTypeCache[uid] = matchingNode || false;
+  return discourseNodeTypeCache[uid];
+};
```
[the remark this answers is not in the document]
maparent: No, the match logic is equivalent downstream, it makes sense to have a single cache.

apps/roam/src/utils/findDiscourseNode.ts
```diff
@@ -16,3 +19,25 @@ const findDiscourseNode = (uid = "", nodes = getDiscourseNodes()) => {
...
+  nodes?: DiscourseNode[];
+}): DiscourseNode | false => {
+  nodes = nodes || getDiscourseNodes();
+  if (typeof discourseNodeTypeCache[uid] !== "undefined") {
+    return discourseNodeTypeCache[uid];
+  }
+
+  const matchingNode = nodes.find((node) =>
+    matchDiscourseNode({ ...node, title }),
+  );
```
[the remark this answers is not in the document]
maparent: Again, this makes sense because downstream, the uid is mostly used to fetch the title. The uid here is mostly used to access the cache.

apps/roam/src/utils/findDiscourseNode.ts
```diff
@@ -16,3 +19,25 @@ const findDiscourseNode = (uid = "", nodes = getDiscourseNodes()) => {
   return discourseNodeTypeCache[uid];
 };
 export default findDiscourseNode;
+
+export const findDiscourseNodeByTitleAndUid = ({
```
mdroidian, line 23: We probably should just extend `findDiscourseNode` to allow for `title?` instead of creating a new function that does the same thing.  But we can clean that up later, I'd like to merge this ASAP.

apps/roam/src/components/CreateRelationDialog.tsx
mdroidian: `getFilteredRelations()`, but we can clean this up later.  I'd like to merge this ASAP.

apps/roam/src/components/CreateRelationDialog.tsx
mdroidian: Ideally, we should do this.  But in this case I'd like to merge this ASAP, so we can come back to it.

FYI, `LabelDialog` has an example of `AutocompleteInput` using `Result[]`, so it shouldn't be much trouble to adapt.

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,381 @@
...
+        } else {
+          renderToast({
+            id: `discourse-relation-error-${Date.now()}`,
+            intent: "warning",
+            content: <span>Failed to create relation</span>,
+          });
+        }
+        onClose();
+      })
+      .catch((error) => {
```
mdroidian: unused

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,381 @@
...
+  }
+  if (props === null) {
+    renderToast({
+      id: `discourse-relation-error-${Date.now()}`,
+      intent: "warning",
+      content: <span>No relation exists for {sourceNodeUid}</span>,
+    });
+  } else {
+    renderOverlay({
+      Overlay: CreateRelationDialog,
```
mdroidian, line 354: // eslint-disable-next-line @typescript-eslint/naming-convention

apps/roam/src/styles/styles.css
```diff
@@ -114,6 +114,17 @@
...
+.bp3-callout.invert-icon .bp3-icon {
+  color: #e1e8ed !important; /* @light-gray3 */
+  background-color: #106ba3; /* @blue2 */
+  padding: 1px;
+}
+
+/* Utility class that targets the width of the future popover element */
+.make-popover-full-width .bp3-popover-target {
+  width: 100%;
+}
```
mdroidian, line 126: This is still ambigious (even with the comment, what is "the future popover" referring to?).  As a dev, when I see this, I now have to copy/paste the class into search and look through the repo to find it, then figure out what component it is being used for.  Alternatively, if I named this `.create-relation-dialog-popover` or something similar, I could scan this css and get the context in which it applies, without having to break flow or do extra work.

apps/roam/src/styles/styles.css
```diff
@@ -114,6 +114,17 @@
   border-bottom: 1px solid rgba(16, 22, 26, 0.15);
 }
 
+.bp3-callout.invert-icon .bp3-icon {
```
mdroidian, line 117: nit: unneeded.  `intent: primary` on the `<Callout>` should get you most of the way there.

See `LabelDialog` example and Callout docs

https://blueprintjs.com/docs/versions/3/#core/components/callout
Image


LabelDialog
Image

apps/roam/src/components/CreateRelationDialog.tsx
mdroidian: [ENG-1092: Better Error Handling](https://linear.app/discourse-graphs/issue/ENG-1092/better-error-handling)

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,381 @@
...
+        }
+      },
+    );
+    if (candidateRelations.length === 0) {
+      // also should not happen
+      internalError("Could not find the relation");
+      return null;
+    }
+    if (candidateRelations.length !== 1) {
+      // This seems to happen... I need more data.
```
mdroidian, line 129: Seems like this should be traceable via chrome dev panel, no?

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,381 @@
...
+  };
+
+  const changeRelationType = (relName: string): void => {
+    setSelectedRelationName(relName);
+    setPageOptions(getFilteredPageNames(relName));
+    if (
+      selectedTargetUid !== undefined &&
+      identifyRelationMatch(selectedTargetTitle, selectedTargetUid) === null
+    ) {
+      setSelectedTargetUid(undefined);
```
mdroidian, line 189: The text does not clear if the relation is changed to an unsupported relation
https://www.loom.com/share/702625c392e74e699ebef2a07ae281f1

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,381 @@
...
+        onClose();
+      });
+  };
+
+  const changeRelationType = (relName: string): void => {
+    setSelectedRelationName(relName);
+    setPageOptions(getFilteredPageNames(relName));
+    if (
+      selectedTargetUid !== undefined &&
+      identifyRelationMatch(selectedTargetTitle, selectedTargetUid) === null
```
mdroidian, line 187: This will use the old `selectedRelationName` rather than the new `relName`, so won't clear the `uid`

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,381 @@
...
+    const uid = getPageUidByPageTitle(title);
+    if (uid.length === 0) {
+      setSelectedTargetUid(undefined);
+      return;
+    }
+    const relation = identifyRelationMatch(title, uid);
+    if (relation === null) {
+      setSelectedTargetUid(undefined);
+      return;
+    }
```
mdroidian, line 205: This is probably unecessary, because the value in `options` should already only be matches.  Futhermore, this would be poor UX in the case where it did happen.  Eg:
- user would select a title, but couldn't click "Confirm".
- user would click back into the input and the target title would still be available for selection.
- user would be stuck

apps/roam/src/components/CreateRelationDialog.tsx
mdroidian: Let's make a task to fix both this bug and the text not clearing (the line below this) bug after this PR is merged.

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -0,0 +1,381 @@
...
+    sourceNodeUid,
+    onClose,
+    relData,
+    sourceNodeTitle: nodeTitle,
+    selectedSourceType,
+  };
+};
+
+export const renderCreateRelationDialog = (
+  props: CreateRelationDialogProps | ExtendedCreateRelationDialogProps | null,
```
mdroidian, line 329: When is `CreateRelationDialogProps` used?

mdroidian: If we can remove this, we can remove the `as ExtendedCreateRelationDialogProps` assertions as well

mdroidian (approved):
I'm going to approve this so we can merge it and get it in the hands of the PI's, but there are a few bugs we need to create tasks for that should be addressed sooner than later.
- relation changing bugs
- "this seems to happen ... I need more data" follow up
- using `Result[]` in Autocomplete
- extending `findDiscourseNode` rather than duplicating the function

Let me know if you have any questions.  👍

apps/roam/src/components/CreateRelationDialog.tsx
mdroidian: If not, what more data do you need?  Could you elaborate?

apps/roam/src/utils/findDiscourseNode.ts
maparent: If we extend it, we have to change all the call points because it becomes a 3-argument function and needs to become a function with destructured arguments per the guidelines.
(I still think 3 is too little.)

apps/roam/src/components/CreateRelationDialog.tsx
maparent: But needed to make typescript happy.

maparent: ah sorry that was someplace else. correcting.

apps/roam/src/utils/findDiscourseNode.ts
mdroidian: This is why 3 is too much. 😎  If they were destructured already, the call points wouldn't need to be updated.

But yes, when we make changes like this, we do need to updated the places in where it is used.  This would be a minor change that could be mostly automated.

maparent: There is a performance cost to destructuration you know. 
And the reason to do it has to do with limitations of tooling, mostly.

apps/roam/src/styles/styles.css
maparent: Ok. I personally find verbs weird in CSS, but I don't care. 
As a matter of process: You did not like my initial name, said it was not clear enough for you. I have no problem with improving it. But basically, unless you include a counter-proposal, you make it a guessing game for me: What will you find clear? (given that I do not understand what you find unclear about it.)
So in that case, can I ask you to provide a counter-proposal? I can still counter-counter-propose at need.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: I need to do more experiments with more cases. I did a lot with the previous implementation, I need to re-do them with your implementation. The issue may be gone.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: Ah, that was intentional, I thought the text would come from an earlier attempt, so some part of the text may still be relevant. But clearing also makes sense, no objection. Ok adding it to the todo task.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: Agreed that this should never happen. It's a fallback to a fallback.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: Not used in this case, but someone may want to call the dialog without going through that button, and in that case they should be able to do so with minimal props.

mdroidian: The moment we start reaching for typescript assertions we should realize this is a red flag.  Instead, let's write the code in which we are tasked to write, and leave the code in the possible future use case for the possible person in the future to write.  Now you've put the burden on any future person who excepts a type safe codebase to catch this assertion if they make any changes to these types/components.

apps/roam/src/components/CreateRelationDialog.tsx
mdroidian: These should have been done before submitting for a review.  Please assign yourself a task in linear so we can track this and it doesn't get lost.

PR #561, opened 2025-11-23
ENG-1075 End-sync task bug

maparent:
https://linear.app/discourse-graphs/issue/ENG-1075/end-sync-task-bug
Problem: the pg_advisory lock in propose_sync_task was not the right approach, as the subsequent updates were not visible to other threads when they acquired the lock, and before the release. A more classic 'for update' solves the problem better.
We also noticed a stampede problem in that waiting clients would all start at the same time.

## Summary by CodeRabbit

* **Performance & Reliability**
  * Enhanced sync scheduling with optimized timing mechanisms to gracefully handle concurrent operations and prevent system resource conflicts when multiple processes reschedule.
  * Improved database task coordination with better timeout management, backoff recovery, and enhanced locking mechanisms for more reliable sync task execution.


maparent:
@coderabbit review

sid597 (approved):
Fixes what we discovered, LGTM

PR #562, opened 2025-11-23
ENG-1076 Variability in roam URLs

maparent:
https://linear.app/discourse-graphs/issue/ENG-1076/variability-in-roam-urls

## Summary by CodeRabbit

* **Bug Fixes**
  * Made Roam space URL construction deterministic, improving reliability when creating spaces and direct links.

### DIFF
--- a/apps/roam/src/utils/supabaseContext.ts
+++ b/apps/roam/src/utils/supabaseContext.ts
@@ -27,6 +27,8 @@ export type SupabaseContext = {
 
 let _contextCache: SupabaseContext | null = null;
 
+const ROAM_URL_PREFIX = "https://roamresearch.com/#/app/";
+
 const getOrCreateSpacePassword = () => {
   const settingsConfigPageUid = getPageUidByPageTitle(
     DISCOURSE_CONFIG_PAGE_TITLE,
@@ -54,9 +56,8 @@ export const getSupabaseContext = async (): Promise<SupabaseContext | null> => {
       const spacePassword = getOrCreateSpacePassword();
       const personEmail = getCurrentUserEmail();
       const personName = getCurrentUserDisplayName();
-      let url = getRoamUrl();
-      url = url.replace("/?server-port=3333#/", "/#/");
       const spaceName = window.roamAlphaAPI.graph.name;
+      const url = ROAM_URL_PREFIX + spaceName;
       const platform: Platform = "Roam";
       const spaceResult = await fetchOrCreateSpaceDirect({
         password: spacePassword,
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

maparent:
@coderabbit: I thought of parsing, but I was trying to avoid impacting new patterns in an unforeseen way, so I made the code very tailored to the one known case.

mdroidian:
Why do we need `url`?  Why not just use `platform` and `spaceName`?

apps/roam/src/utils/supabaseContext.ts
```diff
@@ -54,7 +54,8 @@ export const getSupabaseContext = async (): Promise<SupabaseContext | null> => {
       const spacePassword = getOrCreateSpacePassword();
       const personEmail = getCurrentUserEmail();
       const personName = getCurrentUserDisplayName();
-      const url = getRoamUrl();
+      let url = getRoamUrl();
+      url = url.replace("/?server-port=3333#/", "/#/");
```
sid597: I have seen a few examples of: 

`https://roamresearch.com/#/app/resultsgraph` 

`https://roamresearch.com/#/app/resultsgraph/search` 
`https://roamresearch.com/#/app/resultsgraph/graph`

being added to supabase but it should only be 
`https://roamresearch.com/#/app/resultsgraph`

maparent:
The assumption was that different spaces could be identified by a URL, which is thus the unique key. 
The name is intended for display, not as a canonical identifier for the space. (Though it should be unique so there is no name confusion.)
In the case of roam, all URLs are the same except one part, which we also use as the name; but it would be an unsafe assumption that all platform's URLs can have a single part that serves as a unique identifier.

mdroidian:
But for Roam, `window.roamAlphaAPI.graph.name` is what is unique, right?  So `url` provides no additional value, is that correct? (because it is `graph.name` + some additional characters).

maparent:
For Roam, true. For Intergraph, false.

mdroidian:
What will Obsidian use as `url`?

maparent:
As discussed: Obsidian should use a `urn:uuid:....` URN (misnamed as URL)
Since Obsidian graphs are not generally available on the Web.
I think it would be nice to have access to Obsidian URLs when exported, but currently out of scope.

Since there is more variation in Roam URLs than initially observed, I took a different approach, generating a canonical URL rather than trying to capture the variation.

maparent:
@coderabbit review

sid597 (approved):
LGTM

Issue #39, opened 2024-12-27
Optimize page load observers


mdroidian:
Specifically the `linkedReferencesObserver`, as it contains a roam query.

We should look into combining/optimizing this with the `pageTitleObserver` and upcoming features nanopub/GitHub publishing.

Many of these require checking if the page is a discourse node, which often can require a query.  If this is the case, we should only do this once and make the result available to the other functions/features.

mdroidian:
[ENG-1095: Optimize Discourse Node page observer check](https://linear.app/discourse-graphs/issue/ENG-1095/optimize-discourse-node-page-observer-check)

PR #564, opened 2025-11-24
ENG-1074 Prod duplicate node alert on page, closes ENG-1095

sid597:
duplicate node alert on page

add to node create dialog, only show small list, fix lint errors, fix bug to show in all open pages

## Summary by CodeRabbit

* **New Features**
  * Added vector similarity matching to identify and surface similar pages based on content
  * Integrated collapsible panel displaying matched pages with similarity scores
  * Users can navigate directly to matched pages from results
  * Feature respects Suggestive Mode setting

* **Refactor**
  * Consolidated reference handling and observer logic for improved code organization

https://www.loom.com/share/cb366c09942b4030a9933118ad281bc5


sid597:
* **#591** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/591?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#564** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/564?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/564?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sid597:
@coderabbitai full review

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -34,11 +35,21 @@ const CreateNodeDialog = ({
...
   const [title, setTitle] = useState(initialTitle);
+  const [debouncedTitle, setDebouncedTitle] = useState(initialTitle);
   const [selectedType, setSelectedType] =
     useState<DiscourseNode>(defaultNodeType);
   const [loading, setLoading] = useState(false);
   const inputRef = useRef<HTMLInputElement>(null);
 
+  useEffect(() => {
+    const handler = setTimeout(() => {
+      setDebouncedTitle(title);
```
mdroidian: Why is this now needed?

apps/roam/src/components/VectorDuplicateMatches.tsx
```diff
@@ -0,0 +1,192 @@
+import React, { useEffect, useState, useMemo } from "react";
+import { Collapse, Spinner, Icon } from "@blueprintjs/core";
+import { findSimilarNodesVectorOnly, type VectorMatch } from "../utils/hyde";
```
mdroidian: ~

apps/roam/src/components/VectorDuplicateMatches.tsx
```diff
@@ -0,0 +1,192 @@
...
+import { useNodeContext, type NodeContext } from "../utils/useNodeContext";
```
mdroidian: ~

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -136,6 +147,11 @@ const CreateNodeDialog = ({
               onChange={(e) => setTitle(e.currentTarget.value)}
               inputRef={inputRef}
             />
+            <VectorDuplicateMatches
+              pageTitle={debouncedTitle}
+              text={debouncedTitle}
```
mdroidian: Why pass two props for the same thing?

apps/roam/src/utils/hyde.ts
```diff
@@ -58,7 +58,7 @@ type EmbeddingFunc = (text: string) => Promise<EmbeddingVectorType>;
 
 type SearchFunc = (params: {
   queryEmbedding: EmbeddingVectorType;
-  indexData: CandidateNodeWithEmbedding[];
+  indexData: { uid: string; text: string }[];
```
mdroidian: use Result[]

apps/roam/src/utils/hyde.ts
```diff
@@ -173,7 +173,7 @@ const createEmbedding: EmbeddingFunc = async (
   }
 };
 
-const searchEmbeddings: SearchFunc = async ({
+export const searchEmbeddings: SearchFunc = async ({
```
mdroidian: Why is this exported? (it isn't used in any other file)

apps/roam/src/utils/hyde.ts
```diff
@@ -530,3 +530,61 @@ export const performHydeSearch = async ({
...
+          uid: item.roam_uid,
+          text: item.text_content,
+        },
+        score: item.similarity,
+      }),
+    );
+
+    return results;
+  } catch (error) {
+    console.error("Error in vector-only similar nodes search:", error);
```
mdroidian, line 561: What is the UX for the average user here?  Just a silent error? (regular users don't know what the console is)

Let's add a roamjs-components/components/Toast

apps/roam/src/utils/hyde.ts
```diff
@@ -530,3 +530,61 @@ export const performHydeSearch = async ({
...
+    });
+
+    if (error) {
+      console.error("Vector search failed:", error);
+      throw error;
+    }
+
+    if (!data || !Array.isArray(data)) return [];
+
+    const results: VectorMatch[] = data.map(
```
mdroidian: This can be similified to 

```typescript
    const results: VectorMatch[] = data.map((item) => ({
      node: {
        uid: item.roam_uid,
        text: item.text_content,
      },
      score: item.similarity,
    }));
```

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -83,6 +86,11 @@ export const initObservers = async ({
       const title = getPageTitleValueByHtmlElement(h1);
       const props = { title, h1, onloadArgs };
 
+      const uid = getPageUidByPageTitle(title);
+      if (isDiscourseNode(uid)) {
```
mdroidian: This is potentially a heavy query, so running it on every page should be made very carefully.

Additionally, it is already ran on every single page via Discourse Context, so this is duplicating the query.

This was originally noted here https://github.com/DiscourseGraphs/discourse-graph/issues/39 and planned to be optimized with nanopub/github which will need to do the same thing.

I've made this into a linear ticket and assigned you.  
[ENG-1095: Optimize Discourse Node page observer check](https://linear.app/discourse-graphs/issue/ENG-1095/optimize-discourse-node-page-observer-check)
This should be completed before ENG-1074 is merged.

I haven't looked into the difficulty of this, if you feel it is simple enough to include in this PR, so be it.

apps/roam/src/components/VectorDuplicateMatches.tsx
```diff
@@ -0,0 +1,192 @@
...
+    <div className="my-2 rounded border border-gray-200">
+      <div
+        className="flex cursor-pointer items-center justify-between p-2"
+        onClick={() => {
+          setIsOpen(!isOpen);
+        }}
+      >
+        <div className="flex items-center gap-2">
+          <Icon icon={isOpen ? "chevron-down" : "chevron-right"} />
+          <h5 className="m-0 font-semibold">Plain Vector Search Matches</h5>
```
mdroidian: "Vector Search" might not mean anything to many users.  Let's change this to "Possible Duplicates" from the [designs](https://linear.app/discourse-graphs/issue/DES-111/mock-up-duplicate-node-alert-in-note-header)

apps/roam/src/components/VectorDuplicateMatches.tsx
```diff
@@ -0,0 +1,192 @@
...
+              ))}
+            </ul>
+          )}
+        </div>
+      </Collapse>
+    </div>
+  );
+};
+
+export const renderPossibleDuplicates = (
```
mdroidian, line 162: Just FYI, we'll likely be combining this with a util fuction very soon in the future.

This util function will "handle title additions" for discourse context, github sync, nanopub, etc.

It was started here: https://github.com/DiscourseGraphs/discourse-graph/blob/1fad51301bb47157bb2d4812298f7abb591b728f/apps/roam/src/utils/handleTitleAdditions.ts

apps/roam/src/components/VectorDuplicateMatches.tsx
```diff
@@ -0,0 +1,192 @@
...
+                  <a
+                    onClick={() => {
+                      void handleSuggestionClick(match.node);
+                    }}
+                    className="min-w-0 flex-1 cursor-pointer break-words text-blue-600 opacity-70 hover:underline"
+                  >
+                    {match.node.text}
+                  </a>
+                  <span className="ml-2 shrink-0 text-xs tabular-nums text-gray-500">
+                    {match.score.toFixed(3)}
```
mdroidian: How useful is this to regular users?  Was it in the original designs?
It would make sense for our internal alpha/beta, but if we merge this PR it'll go to the Pilot Users, correct?

apps/roam/src/utils/useNodeContext.ts
```diff
@@ -0,0 +1,69 @@
...
+export const useNodeContext = (pageTitle: string): NodeContext | null => {
+  const [nodeContext, setNodeContext] = useState<NodeContext | null>(null);
+
+  useEffect(() => {
+    const discourseNodes = getDiscourseNodes();
+    const pageUid = getPageUidByPageTitle(pageTitle) || null;
+    let matchedNode: DiscourseNode | null = null;
+
+    if (pageUid) {
+      const found = findDiscourseNode(pageUid, discourseNodes);
```
mdroidian: Didn't we already run `isDiscourseNode` on page load (which calls both `getDiscourseNodes` and `findDiscourseNode`)?  We shouldn't run these again.

apps/roam/src/utils/useNodeContext.ts
```diff
@@ -0,0 +1,69 @@
...
+      if (found) {
+        matchedNode = found;
+      }
+    }
+
+    if (!matchedNode) {
+      matchedNode =
```
mdroidian: `matchDiscourseNode` is called by `findDiscourseNode` ... so this would be a third time that this is running.

apps/roam/src/utils/useNodeContext.ts
```diff
@@ -0,0 +1,69 @@
+import { useEffect, useState } from "react";
```
mdroidian: It is not clear to me that `useNodeContext` is actually needed.

Let's make sure we need this file and all of these queries.  It seems like we might only need `extractContentFromTitle`, which should be placed in a util file because a similar version is already used in getExportTypes (`getContentFromNodes`).

apps/roam/src/components/CreateNodeDialog.tsx
```diff
@@ -34,11 +35,21 @@ const CreateNodeDialog = ({
     discourseNodes[0];
 
   const [title, setTitle] = useState(initialTitle);
+  const [debouncedTitle, setDebouncedTitle] = useState(initialTitle);
```
mdroidian: This is only used in VectorDuplicateMatches, so let's move it (and the useEffect) there.

apps/roam/src/components/VectorDuplicateMatches.tsx
mdroidian: If we still need it for alpha/beta, let's find a way to make that explicit

apps/roam/src/utils/hyde.ts
```diff
@@ -530,3 +530,61 @@ export const performHydeSearch = async ({
...
+  threshold?: number;
+  limit?: number;
+}): Promise<VectorMatch[]> => {
+  if (!text.trim()) {
+    return [];
+  }
+
+  try {
+    const supabase = await getLoggedInClient();
+    if (!supabase) return [];
```
mdroidian, line 534: Looks like this returns a false negative for those who haven't set up in our database yet.  They might actually think we've searched and found nothing, but actually we haven't even searched.

mdroidian (changes requested):
Good start!  But I think this would be worth re-looking at `useNodeContext` because there is a lot going on there that seems unnecessary, which will result in longer page load times.

apps/roam/src/utils/hyde.ts
mdroidian: EG
image

apps/roam/src/components/VectorDuplicateMatches.tsx
sid597: yeah showing this will not be useful to users, removing it.

apps/roam/src/utils/hyde.ts
sid597: Good catch! We need to think through the database enabled features. Here we are checking if the client exists in supabase, but its wrong we need to check if the graph the embeddings are there in db. 

And to check that we have only single route of suggestive mode, so should we be searching for if suggestive mode is enabled only then enable this feature, we can do this now but for future it could be that users don't want to enable suggestive mode but want duplicate node alert. 

We need to decouple sync with suggestive mode setting until then we can use that.

sid597:
@coderabbitai full review

sid597:
@mdroidian I accidentally merged some changes intended for this branch  to the observer pr eng-1095 (#591), then to fix that I did some graphite stuff and it ruined #591 so I closed it, should i create a seperate pr for the observer or this one is fine?  

https://www.loom.com/share/f57d858ec5eb4663bbe0ad2286f9bd6f

mdroidian:
> @mdroidian I accidentally merged some changes intended for this branch to the observer pr eng-1095 (#591), then to fix that I did some graphite stuff and it ruined #591 so I closed it, should i create a seperate pr for the observer or this one is fine?
> 
> https://www.loom.com/share/f57d858ec5eb4663bbe0ad2286f9bd6f

I think this should be integrated with #614 , is that correct?

mdroidian (changes requested):
testing linear integration

apps/roam/src/components/VectorDuplicateMatches.tsx
```diff
@@ -0,0 +1,177 @@
...
+import extractContentFromTitle from "~/utils/extractContentFromTitle";
+import { handleTitleAdditions } from "~/utils/handleTitleAdditions";
+
+type VectorSearchParams = {
+  text: string;
+  threshold?: number;
+  limit?: number;
+};
+
+const vectorSearch = (params: VectorSearchParams) =>
```
mdroidian: If this is just to rename, you can import `findSimilarNodesVectorOnly as vectorSearch`.  No need to create a new function.

apps/roam/src/components/VectorDuplicateMatches.tsx
```diff
@@ -0,0 +1,177 @@
...
+    </div>
+  );
+};
+
+export const renderPossibleDuplicates = (
+  h1: HTMLHeadingElement,
+  title: string,
+  node: DiscourseNode,
+) => {
+  handleTitleAdditions(
```
mdroidian, line 167: Why can't we just call `handleTitleAdditions` in `initializeObserversAndListeners`?  Why is `renderPosisbleDuplicates` needed?

apps/roam/src/utils/extractContentFromTitle.ts
```diff
@@ -0,0 +1,28 @@
+import getDiscourseNodeFormatExpression from "./getDiscourseNodeFormatExpression";
+
+const extractContentFromTitle = (
```
mdroidian, line 3: This is only used in `VectorDuplicateMatches`, so let's co-locate it there.

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -85,24 +88,40 @@ export const initObservers = async ({
...
       const props = { title, h1, onloadArgs };
 
+      const isSuggestiveModeEnabled = getUidAndBooleanSetting({
+        tree: getBasicTreeByParentUid(
+          getPageUidByPageTitle(DISCOURSE_CONFIG_PAGE_TITLE),
+        ),
+        text: "(BETA) Suggestive Mode Enabled",
+      }).value;
+
+      const uid = getPageUidByPageTitle(title);
```
mdroidian: If you rebase, `findDiscourseNode` can now take `title`, so we can skip the additional call on each page.

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -85,24 +88,40 @@ export const initObservers = async ({
       const title = getPageTitleValueByHtmlElement(h1);
       const props = { title, h1, onloadArgs };
 
+      const isSuggestiveModeEnabled = getUidAndBooleanSetting({
```
mdroidian, line 91: Should this be it's own flag?

mdroidian (approved):
A few non-blocking suggestions

sid597:
Closes ENG-1095 or #591

apps/roam/src/components/VectorDuplicateMatches.tsx
sid597: ~yeah no specific reason, moving to initialize.. now~ its needed because the `initializeObserversAndListeners` is a `.ts` file and we need to render a component

apps/roam/src/utils/extractContentFromTitle.ts
sid597: but this seems like something that can be a common util, not sure where but its general enough so I think we should keep it in utils

apps/roam/src/utils/initializeObserversAndListeners.ts
sid597: Sorry, I don't follow, suggestive mode is already a flag in settings. Do you mean Duplicate node alert being a flag?

mdroidian: Yes

mdroidian: This should be able to be toggled per user.

PR #571, opened 2025-11-28
[ENG-1035] Fix the can't create relation type bug

trangdoan982:


apps/obsidian/src/components/RelationshipTypeSettings.tsx
```diff
@@ -84,7 +84,7 @@ const RelationshipTypeSettings = () => {
 
   const handleSave = async (): Promise<void> => {
     for (const relType of relationTypes) {
-      if (!relType.id || !relType.label || !relType.complement || !relType.color) {
+      if (!relType.id || !relType.label || !relType.complement) {
```
trangdoan982, line 87: this wrong validation mechanism caused users unable to create new relation (color is default "" and therefore will always raise this error)

mdroidian (approved):

PR #577, opened 2025-12-01
[ENG-870] Bug fix allow right click on canvas

trangdoan982:
https://www.loom.com/share/aad654710f5a413fa819dadf9252dfbd

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored right-click context menu behavior in the canvas editor so native and Radix UI menus open correctly.
  * Improved auto-save resilience: saves are now debounced, guarded against concurrent writes, and include error handling that recovers and reloads if a save fails.

### DIFF
--- a/apps/obsidian/src/components/canvas/TldrawViewComponent.tsx
+++ b/apps/obsidian/src/components/canvas/TldrawViewComponent.tsx
@@ -266,10 +266,35 @@ export const TldrawPreviewComponent = ({
     editorRef.current = editor;
     setIsEditorMounted(true);
 
+    // Handle right-click to ensure context menu opens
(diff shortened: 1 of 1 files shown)


trangdoan982:
@coderabbitai full review

trangdoan982:
@coderabbitai full review

mdroidian (approved):
Nice.

One thing I like to try with deep and tricky bugs like this is to have a vanilla version of tldraw ready to test with.  That is to say, tldraw embedded in an obsidian note (just like we do now), but with no extra discourse graph additions, just raw tldraw.

This makes it easier to narrow down where the bug is happening.

PR #584, opened 2025-12-04
ENG-1122 Replace or-clause with or-join-clause to fix variable binding in datalog queries

maparent:
https://linear.app/discourse-graphs/issue/ENG-1122/some-queries-now-broken
A new error is due to an or-clause being pushed after the variable it uses. This can be solved by replacing the or-clause by an or-join-clause; then the optimizer knows about the variables and takes them into account.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal condition handling structure for enhanced data consistency.

### DIFF
--- a/apps/roam/src/utils/conditionToDatalog.ts
+++ b/apps/roam/src/utils/conditionToDatalog.ts
@@ -446,7 +446,7 @@ const translator: Record<string, Translator> = {
       if (currentUserMatch) {
         return [
           {
-            type: "or-clause",
+            type: "or-join-clause",
             clauses: [
               {
                 type: "data-pattern",
@@ -465,6 +465,7 @@ const translator: Record<string, Translator> = {
                 ],
               },
             ],
+            variables: [{ type: "variable", value: `${source}-String` }],
           },
           {
             type: "pred-expr",
@@ -481,7 +482,7 @@ const translator: Record<string, Translator> = {
         const rePattern = regexRePatternValue(target);
         return [
           {
-            type: "or-clause",
+            type: "or-join-clause",
             clauses: [
               {
                 type: "data-pattern",
@@ -500,6 +501,7 @@ const translator: Record<string, Translator> = {
                 ],
               },
             ],
+            variables: [{ type: "variable", value: `${source}-String` }],
           },
           {
             type: "fn-expr",
@@ -528,7 +530,7 @@ const translator: Record<string, Translator> = {
 
       return [
         {
-          type: "or-clause",
+          type: "or-join-clause",
           clauses: [
             {
               type: "data-pattern",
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

mdroidian:
What are possible negatives of going from `or-clause` to `or-join-clause`?

mdroidian:
And what about all the other places `or-clause` is in the repo?  Will those still have the issue?

maparent:
Surprisingly, there is only a single other `or-clause` in the repo, and that's in my recent code, which does not exhibit the issue because there is no variable escaping. (I also remember I could not make it work as a `or-join-clause` at the time.)<br>Possible negatives: None I can think of? or-clause is just more explicit about shared variables.

mdroidian:
> Surprisingly, there is only a single other `or-clause` in the repo, and that's in my recent code, which does not exhibit the issue because there is no variable escaping. (I also remember I could not make it work as a `or-join-clause` at the time.)Possible negatives: None I can think of? or-clause is just more explicit about shared variables.

There are quite a few other references to `or-clause`
image

Surely these would be affected?

maparent:
I also thought of digging deeper into the optimizer to detect implicit variable escaping… I am not opposed to it, but I thought you would see it as more risky.

maparent:
most of those are about testing for / looking for or-clause, not declaring one. Look for `type: "or-clause"` only

mdroidian:
> I also thought of digging deeper into the optimizer to detect implicit variable escaping… I am not opposed to it, but I thought you would see it as more risky.

I would rather go back to the line that was changed that caused this error and explore deeper into **why** this line had to be changed.  As I recall, we could not recollect why the original change needed a different type of or-clause/join.  I think it would be worthwhile timeboxing exploring that deeper.  If it is the case that we could use the existing or-clause/join, then we wouldn't need to change the optimizer and go down the rabbit hole of fixing it.

maparent:
I gave an explanation at the time. (Disjunction is always slower than conjunction.) I recall your not being convinced by it. I still think this was the correct move. I think making the or into an or-join is also an improvement, which we should do no matter what; the fact that this sub-optimal query was exposed I see as a win. (That it was exposed through an error is not great.) <br>As I said, if you want to spend more time on this, I would go the opposite direction: Improve the optimizer further. I know you are loathe to do so, which is why I wanted to have a global test of existing queries. Certainly a better use of my time than reverting what I see as a necessary change.

mdroidian (approved):
If you are confident this will not cause further issues, feel free to merge.

maparent:
I am fully confident the change I proposed is not problematic. I hope no other similar issue turns up, but it's possible some of the code paths in `registerDiscourseDatalogTranslators.ts` are just not exercised yet, or someone might write an or-clause in the future. So I do believe solving this at the optimizer level would be good, which I think requires a global test function. But that is less urgent.

PR #586, opened 2025-12-05
ENG-1069: Move existing shortcuts to new global shortcuts

sid597:
## Solution

Implemented one-time migration that extracts starred pages from Roam's native `.starred-pages` DOM and adds them to Left Sidebar $\to$ Global-Section $\to$ Children in the graph configuration.

## Cases Handled

1.  **Fresh Install (New Graph)**
    * **Condition:** No "Favorites Migrated" flag, no Personal-Section exists
    * **Action:** Extract starred pages $\to$ Add to Global-Section $\to$ Set flag
    * **Result:** User's favorites preserved in new sidebar

2.  **Existing Install (Collaborative Graph)**
    * **Condition:** Any user has `{userId}/Personal-Section` under Left Sidebar
    * **Action:** Skip extraction, set "Favorites Migrated" flag
    * **Result:** No migration runs, existing configuration untouched

3.  **Fresh Install, Multiple Sessions**
    * **Condition:** Migration ran in previous session, graph reopened
    * **Action:** Check "Favorites Migrated" flag $\to$ early return
    * **Result:** No duplicate migration

4.  **Multi-User Graph (Late Joiner)**
    * **Condition:** User B joins after User A already configured plugin
    * **Action:** Detects User A's Personal-Section exists $\to$ skip migration
    * **Result:** No interference with existing setup

## Key Design Decisions

* **Graph-level flag:** "Favorites Migrated" stored under Left Sidebar (not per-user) ensures migration happens once per graph
* **Personal-Section heuristic:** Check for `*/Personal-Section` pattern to detect if plugin already in use by any user
* **Cache refresh:** Call `refreshConfigTree()` after creating flag to prevent duplicate blocks in same session
* **Performance:** Flag check is $O(1)$, expensive tree traversal only runs once per graph lifetime

## Technical Notes

* Migration runs in `mountLeftSidebar` before DOM is cleared
* Uses `getUidAndBooleanSetting` pattern consistent with existing config flags
* Async operation with proper `await` chains to ensure flag is created before subsequent calls

https://www.loom.com/share/fc639adb1f4341d68403a19094b6dd8c

## Summary by CodeRabbit

* **New Features**
  * Starred pages now automatically migrate into the Left Sidebar with duplicate title deduplication.
  * Added a migration status setting so the app tracks when migration has completed.

* **Bug Fixes / Reliability**
  * Sidebar mounting now runs migration reliably during startup so migrated favorites appear consistently without user action.

### DIFF
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -30,6 +30,7 @@ import type {
   LeftSidebarConfig,
   LeftSidebarPersonalSectionConfig,
 } from "~/utils/getLeftSidebarSettings";
+import type { BooleanSetting } from "~/utils/getExportSettings";
 import { createBlock } from "roamjs-components/writes";
 import deleteBlock from "roamjs-components/writes/deleteBlock";
 import getTextByBlockUid from "roamjs-components/queries/getTextByBlockUid";
@@ -407,13 +408,13 @@ const LeftSidebarView = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
   );
 };
 
-const migrateFavorites = async (starredPagesContainer: Element) => {
+const migrateFavorites = async () => {
   const configPageUid = getPageUidByPageTitle(DISCOURSE_CONFIG_PAGE_TITLE);
   if (!configPageUid) return;
 
   const config = getFormattedConfigTree().leftSidebar;
 
-  if (config.favoritesMigrated.value) return;
+  if ((config.favoritesMigrated as BooleanSetting).value) return;
 
   let leftSidebarUid = config.uid;
   if (leftSidebarUid) {
@@ -431,9 +432,12 @@ const migrateFavorites = async (starredPagesContainer: Element) => {
     }
   }
 
-  const titles = Array.from(starredPagesContainer.querySelectorAll(".page"))
-    .map((el) => el.textContent || "")
-    .filter((t) => t);
+  const results = window.roamAlphaAPI.q(`
+    [:find ?title 
+     :where [?e :page/sidebar]
+            [?e :node/title ?title]]
+  `);
+  const titles = (results as string[][]).map(([title]) => title);
 
   if (!leftSidebarUid) {
     const tree = getBasicTreeByParentUid(configPageUid);
@@ -505,10 +509,7 @@ export const mountLeftSidebar = async (
   const id = "dg-left-sidebar-root";
   let root = wrapper.querySelector(`#${id}`) as HTMLDivElement;
   if (!root) {
-    const existingStarred = wrapper.querySelector(".starred-pages");
-    if (existingStarred) {
-      await migrateFavorites(existingStarred);
-    }
+    await migrateFavorites();
     wrapper.innerHTML = "";
     root = document.createElement("div");
     root.id = id;
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -414,7 +414,7 @@ const migrateFavorites = async () => {
 
   const config = getFormattedConfigTree().leftSidebar;
 
-  if ((config.favoritesMigrated as BooleanSetting).value) return;
+  if (config.favoritesMigrated.value) return;
 
   let leftSidebarUid = config.uid;
   if (leftSidebarUid) {
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -38,6 +38,8 @@ import { Dispatch, SetStateAction } from "react";
 import { SettingsDialog } from "./settings/Settings";
 import { OnloadArgs } from "roamjs-components/types";
 import renderOverlay from "roamjs-components/util/renderOverlay";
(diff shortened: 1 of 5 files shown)


sid597:
* **#588** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/588?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#587** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/587?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#586** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/586?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/586?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sid597:
@coderabbitai full review

mdroidian:
@CodeRabbitAI full review

apps/roam/src/components/LeftSidebarView.tsx
```diff
@@ -407,20 +410,109 @@ const LeftSidebarView = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
...
 };
 
-export const mountLeftSidebar = (
+const migrateFavorites = async () => {
+  const configPageUid = getPageUidByPageTitle(DISCOURSE_CONFIG_PAGE_TITLE);
+  if (!configPageUid) return;
+
+  const config = getFormattedConfigTree().leftSidebar;
+
+  if ((config.favoritesMigrated as BooleanSetting).value) return;
```
mdroidian: You should be able to `config.favoritesMigrated.value`.  Just refresh ES-Lint if you are still getting an issue.  This will let you remove the `BooleanSetting` import as well.

apps/roam/src/components/LeftSidebarView.tsx
```diff
@@ -407,20 +410,109 @@ const LeftSidebarView = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
...
-  wrapper.innerHTML = "";
 
   const id = "dg-left-sidebar-root";
   let root = wrapper.querySelector(`#${id}`) as HTMLDivElement;
   if (!root) {
-    const existingStarred = wrapper.querySelector(".starred-pages");
-    if (existingStarred) {
-      existingStarred.remove();
-    }
+    await migrateFavorites();
```
mdroidian, line 512: Ideally we only run this when the left sidebar flag is turned `on`.  It is probably unncessary to run every load.

apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -239,7 +239,7 @@ export const initObservers = async ({
       const container = el as HTMLDivElement;
       if (isLeftSidebarEnabled) {
         container.style.padding = "0";
-        mountLeftSidebar(container, onloadArgs);
+        void mountLeftSidebar(container, onloadArgs);
```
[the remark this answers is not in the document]
mdroidian: This will likely only fire once, but might aswell use async/await @sid597

mdroidian (approved):
Nice.  Couple suggestions, but generally looks good 👍

apps/roam/src/components/LeftSidebarView.tsx
sid597: yeah and this will only run when the flag is turned on

mdroidian: But this runs on every load, correct?  That is not ideal.

sid597: on load we check if left sidebar is enabled, we run this observer on every load

```
  const leftSidebarObserver = createHTMLObserver({
    tag: "DIV",
    useBody: true,
    className: "starred-pages-wrapper",
    callback: (el) => {
      void (async () => {
        const isLeftSidebarEnabled = getUidAndBooleanSetting({
          tree: configTree,
          text: "(BETA) Left Sidebar",
        }).value;
        const container = el as HTMLDivElement;
        if (isLeftSidebarEnabled) {
          container.style.padding = "0";
          await mountLeftSidebar(container, onloadArgs);
        }
      })();
    },
  });
```

then this flag loads the function `mountLeftSidebar` and inside that we do the `migrateFavorites()` so this is guarded by the flag and only runs when this is enabled. <br><br>Oh do you mean we should only load this one time when the plugin is loaded first time with the setting enabled? 

this case is handled in the function 

```
const migrateFavorites = async () => {
  const configPageUid = getPageUidByPageTitle(DISCOURSE_CONFIG_PAGE_TITLE);
  if (!configPageUid) return;

  const config = getFormattedConfigTree().leftSidebar;

  if (config.favoritesMigrated.value) return;
```

so we return early if we see that migration has already happened once

mdroidian: The point is that running both `getPageUidByPageTitle()` and `getFormattedConfigTree()` on every load is unnecessary.  `migrateFavorites()` should only be run once when a user enables the sidebar and we migrate the favorites.  There are multiple API calls here that could be avoided.

PR #593, opened 2025-12-07
ENG-1102 Create reified relation via canvas

maparent:
This is a basic implementation of creation of reified relation through the canvas.
What is not handled yet: 
1. I think in this context I do not have to worry about inverse relations, is that accurate?
2. Should we destroy the relation if the user undoes the relation?
(In that case we should actually check if the relation preexisted, this is not simple.)
Is there a hook for undo actions?
Or is this out of scope?

## Release Notes

* **New Features**
  * Introduced a feature flag to control relation creation behavior. When enabled, relations are created using a new alternative method; when disabled, the original block-based approach continues. This provides flexibility for relation handling while maintaining complete backward compatibility with all existing data and workflows.

### DIFF
--- a/apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationUtil.tsx
+++ b/apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationUtil.tsx
@@ -63,6 +63,8 @@ import {
   shapeAtTranslationStart,
   updateArrowTerminal,
 } from "./helpers";
+import { getSetting } from "~/utils/extensionSettings";
+import { createReifiedRelation } from "~/utils/createReifiedBlock";
 import { discourseContext, isPageUid } from "~/components/canvas/Tldraw";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
 import { InputTextNode } from "roamjs-components/types";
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

mdroidian (commented):
Generally looks good.

Please retag for review when the loom video demonstrating the functionality is added.

RE:
> I think in this context I do not have to worry about inverse relations, is that accurate?

Could you elaborate?  What is the user action in which inverse relations are currently being handled?

> Should we destroy the relation if the user undoes the relation?

This is not currently handled, so this is out of scope.

maparent:
Good. 
> > I think in this context I do not have to worry about inverse relations, is that accurate?
> 
> Could you elaborate? What is the user action in which inverse relations are currently being handled?

That answers my question, thank you. I did not see such a user action, I thought I'd be faster checking with you in case I missed it.

> > Should we destroy the relation if the user undoes the relation?
> This is not currently handled, so this is out of scope.
Good. I think it's worth putting in the backlog, though.

mdroidian (changes requested):
> [Please retag for review when the loom video demonstrating the functionality is added.](https://github.com/DiscourseGraphs/discourse-graph/pull/593#pullrequestreview-3549668255)

maparent:
[Creating Relationships in the Canvas: A Quick Guide](https://www.loom.com/share/7400eb960a934b5da8f4fa0d6ad50b1d)

Sorry, it happened again: I add the loom video, it looks like it's gone, but it's actually in draft state.

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationUtil.tsx
```diff
@@ -540,58 +542,70 @@ export const createAllRelationShapeUtils = (
...
-                createBlock({ node, order, parentUid }).catch(() =>
-                  console.error(
-                    `Failed to create block: ${JSON.stringify(
-                      { node, order, parentUid },
-                      null,
-                      4,
-                    )}`,
+        if (getSetting("use-reified-relations")) {
+          await createReifiedRelation({
+            sourceUid: (source.props as DiscourseNodeShape["props"]).uid,
```
mdroidian: Avoid using assertions unless absolutely necessary.  And when absolutely necessary, a comment should be added as to why it is necessary.

Let's add a type guard instead, eg
```typescript
      isDiscourseNodeShape(shape: TLShape): shape is DiscourseNodeShape {
        const shapeUtil = this.editor.getShapeUtil(shape.type);
        return shapeUtil instanceof BaseDiscourseNodeUtil;
      }
```

mdroidian (changes requested):

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationUtil.tsx
maparent: Done. Is that really an issue, though? I think it would be caught by the node type guards at 544, and also note that I was following the same assertion pattern at line 586. 
I believe in defensive coding in many cases, but there's also a moment where it burdens the code unnecessarily.

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationUtil.tsx
```diff
@@ -494,6 +495,16 @@ export const createAllReferencedNodeUtils = (
   });
 };
 
+const asDiscourseNodeShape = (
```
mdroidian, line 498: nit: `isDiscourseNodeShape` would be the common pattern
https://www.typescriptlang.org/docs/handbook/2/narrowing.html?utm_source=chatgpt.com#using-type-predicates

This would avoid casting altogether.

I've added a task to [ENG-1158: disallow all type assertions](https://linear.app/discourse-graphs/issue/ENG-1158/disallow-all-type-assertions) so we can catch these as they happen.  `asDiscourseNodeShape` would be triggered by this.

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationUtil.tsx
mdroidian: It is an issue, yes.  Assertions make the codebase harder to reason about, weakens static guarantees, and increases the risk of runtime errors that TypeScript is intended to prevent.

> I think it would be caught by the node type guards at 544

If it is passed a different type of node with different props, this could pass 544 and error here.

> I believe in defensive coding in many cases, but there's also a moment where it burdens the code unnecessarily.

Sure, as with most things there is a spectrum with nuance.  If we can codify when the "moment where it burdens the code unnecessarily" then we can add it as a heuristic.

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationUtil.tsx
```diff
@@ -540,58 +553,82 @@ export const createAllRelationShapeUtils = (
...
+              sourceUid: sourceAsDNS.props.uid,
+              destinationUid: targetAsDNS.props.uid,
+              relationBlockUid: relation.id,
+            });
+          else {
+            void sendErrorEmail({
+              error: new Error(
+                "attempt to create a relation between non discourse nodes",
+              ),
+              type: "canvas-create-relation-non-dgn",
```
mdroidian, line 571: nit: in `sendErrorEmail`, `type` is used in the email subject and as a header.  So everywhere else we use a format like this: `type: "Export Dialog Failed",`

mdroidian (approved):

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationUtil.tsx
maparent: That's what I meant by "is it an issue", not whether you thought casting was bad, but whether this code path was likely to ever happen. I think that the heuristic should precisely be: This will never happen, so checking for it is burdening the code in two ways: bloat for an impossible code path, and getting people to worry about an impossible error.

maparent: ALSO the idea of disabling all type assertions?  (Combined with the rule that we should never have lint errors in new code)… I think it's an unrealistic requirement that is guaranteed to grind development to a halt. Yes, they should be minimized and justified, but sometimes it pragmatically needs to happen. As I said, i was following *your* pattern of doing it in line 586. <br>For the record: There are \~450 uses of the word "as" in the codebase right now.

mdroidian: > sometimes it pragmatically needs to happen

Agreed.  The rule would just make sure that the pragmatic reason is included.

> As I said, i was following your pattern of doing it in line 586. 

Yes, this code was checked in quite some time ago under different circumstances and should be changed.  A linter rule would have caught it 🙂.

maparent: Note: But I agree it's justified in this case, given what you said about passing 544. I wonder if I should make the test upstream, then?

maparent: I agree a linter rule would have caught it; but I doubt anybody uses "as" quite so casually. What I expect we'll see with a linter rule is every "as" being decorated with an eslint-ignore, not sure that's a gain.

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationUtil.tsx
maparent: Hmmm... I think it's useful to have a "slug" type. I could use automatic conversion to make it a title in email, would that help?

mdroidian: I would defer adding additional functionality to a separate task/PR.  It would make more sense to include it in the already in progress [ENG-1135](https://linear.app/discourse-graphs/issue/ENG-1135/replacing-sendmail-with-explicit-posthog-errors).

PR #595, opened 2025-12-08
ENG-1088 An affordance to delete reified relations

maparent:
https://linear.app/discourse-graphs/issue/ENG-1088/an-affordance-to-delete-reified-relations

This implements a "Delete relation" button near each discourse context search result (and in some cases query builder search results.) This is only visible if reified relations are on.

## Summary by CodeRabbit

* **New Features**
  * Added delete functionality for reified relations in results. A delete button now appears for each relation entry when the reified relations setting is enabled, allowing users to remove relations directly from the results view.

### DIFF
--- a/apps/roam/src/components/DiscourseContext.tsx
+++ b/apps/roam/src/components/DiscourseContext.tsx
@@ -269,9 +269,7 @@ const ContextTab = ({
       // TODO - always save settings, but maybe separate from root `parentUid`?
       preventSavingSettings
       parentUid={parentUid}
-      results={Object.values(results).map(
-        ({ target, complement, id, ...a }) => a as Result,
-      )}
+      results={Object.values(results).map(({ target, ...a }) => a as Result)}
       columns={columns}
       onRefresh={onRefresh}
       header={
--- a/apps/roam/src/components/results-view/ResultsTable.tsx
+++ b/apps/roam/src/components/results-view/ResultsTable.tsx
@@ -24,7 +24,6 @@ import DiscourseContextOverlay from "~/components/DiscourseContextOverlay";
 import { CONTEXT_OVERLAY_SUGGESTION } from "~/utils/predefinedSelections";
 import { getSetting } from "~/utils/extensionSettings";
 import { strictQueryForReifiedBlocks } from "~/utils/createReifiedBlock";
-import formatDistanceStrictWithOptions from "date-fns/esm/fp/formatDistanceStrictWithOptions/index.js";
 
 const EXTRA_ROW_TYPES = ["context", "discourse"] as const;
 type ExtraRowType = (typeof EXTRA_ROW_TYPES)[number] | null;
@@ -372,15 +371,19 @@ const ResultRow = ({
               ) : (
                 cell(key)
               )}
-              {useReifiedRel && r["ctxTargetUid"] !== undefined && (
-                <Button
-                  minimal
-                  icon="delete"
-                  className="float-right"
-                  title="Delete relation"
-                  onClick={onDelete}
-                ></Button>
-              )}
+              {useReifiedRel &&
+                typeof r["ctxTargetUid"] === "string" &&
+                typeof r["id"] === "string" &&
+                typeof r["complement"] === "number" &&
+                i === columns.length - 1 && (
+                  <Button
+                    minimal
+                    icon="delete"
+                    className="float-right"
+                    title="Delete relation"
+                    onClick={onDelete}
+                  ></Button>
+                )}
               {i < columns.length - 1 && (
                 <div
                   style={{
--- a/apps/roam/src/components/results-view/ResultsTable.tsx
+++ b/apps/roam/src/components/results-view/ResultsTable.tsx
@@ -7,6 +7,7 @@ import React, {
 } from "react";
 import { Button, HTMLTable, Icon, IconName } from "@blueprintjs/core";
 import { IconNames } from "@blueprintjs/icons";
+import { render as renderToast } from "roamjs-components/components/Toast";
 import { Column, Result } from "~/utils/types";
 import type { FilterData, Sorts, Views } from "~/utils/parseResultSettings";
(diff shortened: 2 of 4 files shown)


maparent:
@coderabbit review

apps/roam/src/utils/getDiscourseContextResults.ts
```diff
@@ -57,6 +58,12 @@ const buildSelections = ({
       text: `node:${conditionUid}-Anchor`,
     });
   }
+  if (getSetting<boolean>("use-reified-relations", false))
```
mdroidian: Selections are additional columns that are meant to be displayed.  Let's move this down to the `result` mapping on ln 96-99.

apps/roam/src/utils/registerDiscourseDatalogTranslators.ts
```diff
@@ -818,6 +818,7 @@ const registerDiscourseDatalogTranslators = () => {
               [
                 { from: source, to: source },
                 { from: target, to: target },
+                { from: "rel", to: "rel" },
```
mdroidian: If we move the relation-uid from selection to `result`, this is no longer needed, correct?

mdroidian (changes requested):
RE: cache
Discourse context is currently cached and we have an `ignoreCache` mechanism.  This should be called when we delete a relation.  `ResultsView` is passed `onRefresh` which is generally where the cache refresh is called, so take a look there.

If we are not adding/fixing that here, please create and link to the linear task in which it is planned on being completed so we can keep track of it.

apps/roam/src/utils/getDiscourseContextResults.ts
maparent: I would love _not_ to use selections, but I do not see another good hook to add to what gets fetch/pulled. By 96-99, the query has been sent, and `rel` is just not part of the result set. Maybe we need to discuss how to do this properly, it would require quite serious changes.

mdroidian: What about passing the source, target, and relation type id's and querying for the reified relation block id(s) that need to be deleted.

maparent: That's an extra query for no good reason… but I think it's a good workaround for now.

maparent: status:

To do this, I need the Uid of the focus of the DiscourseContext passed down.

I expected this to be set as parentUid, but actually this prop is ambiguous.

It is indeed the focus in the case of a DiscourseContext result;

But when the results are for a query block, the parentUid is actually the query block's uid, used for calculating settings.

What gets passed down from ResultsView to ResultsTable is taken from the settings:

ResultsView:1337, parentUid={settings.resultNodeUid}

So in the DiscourseContext case, this is empty, and the parentUid is lost.

I am quite sure that replacing that with parentUid is going to create trouble downstream.

I propose actually having two separate props: queryBlockUid and discourseContextFocusUid, or some such.

But that's a big change and I want to confirm.

mdroidian: We should be able to bypass `ResultsView` here and just pass the required `uid`  (source of discourse context, whether that be from a page discourse context or the discourse context overlay on a page reference) directly via the \`Result\` object.<br><br>I don't believe we'll need to change something that will affect non-discourse context queries globally.<br><br>I'd have to dig into the code to find exactly where, but if we need to pair program on this, let me know.

maparent: I think we should. Probably tomorrow after the dev meeting.

maparent:
@coderabbit review

apps/roam/src/components/results-view/ResultsTable.tsx
```diff
@@ -321,6 +372,15 @@ const ResultRow = ({
...
               )}
+              {useReifiedRel && r["ctxTargetUid"] !== undefined && (
+                <Button
+                  minimal
+                  icon="delete"
+                  className="float-right"
+                  title="Delete relation"
+                  onClick={onDelete}
+                ></Button>
+              )}
```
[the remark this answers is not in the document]
maparent: Good catch. I put it in the last cell.

apps/roam/src/components/results-view/ResultsTable.tsx
```diff
@@ -264,6 +269,52 @@ const ResultRow = ({
...
+      .catch((e) => {
+        // this one should be an internalError
+        console.error(e);
+        renderToast({
+          id: "delete-relation-error",
+          content: "Error searching for relation",
+          intent: "danger",
+        });
+      });
+  };
```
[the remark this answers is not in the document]
maparent: Hmm... when in ResultsTable, we may or may not have come from discourseContext, so ctxTargetUid is not guaranteed to exist at that point. By the time we're in onDelete, we came through the button; and I added the suggested type checks before displaying the button, so that should be enough. So not sure where that new type would be useful.

apps/roam/src/components/DiscourseContext.tsx
```diff
@@ -269,9 +269,7 @@ const ContextTab = ({
       // TODO - always save settings, but maybe separate from root `parentUid`?
       preventSavingSettings
       parentUid={parentUid}
-      results={Object.values(results).map(
-        ({ target, complement, id, ...a }) => a as Result,
-      )}
+      results={Object.values(results).map(({ target, ...a }) => a as Result)}
```
mdroidian, line 272: Why are these being removed?  Did you check if this has any regressions with the current pattern based discourse context?

mdroidian (approved):

apps/roam/src/components/DiscourseContext.tsx
maparent: They were mentioned here to be discarded; and it's data I need downstream, so I un-discard.
No impact on pattern-based discourse context.

PR #598, opened 2025-12-08
ENG-1101 Create reified relation via suggestion mode

maparent:
https://linear.app/discourse-graphs/issue/ENG-1101/create-reified-relation-via-suggestion-mode

Create a reified relation from the suggestion mode.

## Summary by CodeRabbit

* **New Features**
  * Added a toggleable setting for block creation workflows. When enabled, an alternative creation flow with validation and relation processing is used; when disabled, the original method is preserved. Enhanced error notifications improve user feedback.

### DIFF
--- a/apps/roam/src/components/SuggestionsBody.tsx
+++ b/apps/roam/src/components/SuggestionsBody.tsx
@@ -29,6 +29,7 @@ import { type RelationDetails } from "~/utils/hyde";
 import { getFormattedConfigTree } from "~/utils/discourseConfigRef";
 import { render as renderToast } from "roamjs-components/components/Toast";
 import { getSetting } from "~/utils/extensionSettings";
+import { USE_REIFIED_RELATIONS } from "~/data/userSettings";
 import { createReifiedRelation } from "~/utils/createReifiedBlock";
 
 export type DiscourseData = {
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

sid597:
quick question since I have been away from relations for some time,
Relations can be defined in different ways in the graph right? for e.g 
- if this node type  X has a child with reference to supoortedBy and that has another node of type Y then this is a relation
- if this is a node page type X and if any node type Y exists in this page that it is a relation

and a few more ways, right? 

if so then are we checking for that? From what i understand here we create a relation because the target node will be added as a child to parent node

maparent:
To expand on what I said in PM: We are not exactly checking for multiple patterns; but the idea is that, if we are using reified relations, we are just doing the reified relation and not the patterns. So no need to think about the many patterns in that case.
BUT yes, some pattern cases are not yet handled when reified relations are not being used. I saw correcting that as out-of-scope.

apps/roam/src/components/SuggestionsBody.tsx
```diff
@@ -303,10 +309,80 @@ const SuggestionsBody = ({
...
+      if (discourseNode === false) {
+        renderToast({
+          id: "suggestions-create-block-error",
+          content: "Could not identify type of source",
+          intent: "danger",
+          timeout: 5000,
+        });
+        return;
+      }
+      const selectedNodeType = findDiscourseNodeByTitleAndUid({
```
sid597: we can use `node.type`  its of type `type SuggestedNode = Result & { type: string };`

sid597 (changes requested):
One comment otherwise lgtm

sid597 (approved):
Image

PR #602, opened 2025-12-11
eng-1144: DiscourseContext UI disappears

maparent:
Distinguish the onRefresh and forceRefresh function. 
The cache bypass was not consistent, leading to too many calls.
* Deleted a seemingly unused `loadInBackground` argument in the `onRefresh` signature
* updateCount variable so we both benefit from caching and can break it.

## Summary by CodeRabbit

* **Refactor**
  * Improved refresh and cache-control interfaces so components can request fresh or cached data more explicitly.
  * Coordinated refresh timing tightened to deliver more consistent contextual results across views (including Kanban and results lists).
  * Minor UI markup cleanup in results table for more stable rendering.

### DIFF
--- a/apps/roam/src/components/DiscourseContext.tsx
+++ b/apps/roam/src/components/DiscourseContext.tsx
@@ -228,7 +228,7 @@ const ContextTab = ({
   r: DiscourseContextResults[number];
   groupByTarget: boolean;
   setGroupByTarget: (b: boolean) => void;
-  onRefresh: () => void;
+  onRefresh: (ignoreCache?: boolean) => void;
 }) => {
   const [subTabId, setSubTabId] = useState(0);
 
@@ -345,14 +345,17 @@ export const ContextContent = ({ uid, results }: Props) => {
     }));
   }, []);
 
-  const onRefresh = useCallback(() => {
-    setRawQueryResults({});
-    getDiscourseContextResults({
-      uid,
-      onResult: addLabels,
-      ignoreCache: true,
-    }).finally(() => setLoading(false));
-  }, [uid, setRawQueryResults, setLoading, addLabels]);
+  const onRefresh = useCallback(
+    (ignoreCache = true) => {
+      setRawQueryResults({});
+      void getDiscourseContextResults({
+        uid,
+        onResult: addLabels,
+        ignoreCache,
+      }).finally(() => setLoading(false));
+    },
+    [uid, setRawQueryResults, setLoading, addLabels],
+  );
 
   const delayedRefresh = () => {
     window.setTimeout(onRefresh, 250);
@@ -360,7 +363,7 @@ export const ContextContent = ({ uid, results }: Props) => {
 
   useEffect(() => {
     if (!results) {
-      onRefresh();
+      onRefresh(false);
     } else {
       results.forEach(addLabels);
       setLoading(false);
--- a/apps/roam/src/components/results-view/ResultsView.tsx
+++ b/apps/roam/src/components/results-view/ResultsView.tsx
@@ -38,7 +38,7 @@ import Charts from "./Charts";
 import Timeline from "./Timeline";
 import Kanban from "./Kanban";
 import MenuItemSelect from "roamjs-components/components/MenuItemSelect";
-import { RoamBasicNode } from "roamjs-components/types";
+import type { RoamBasicNode } from "roamjs-components/types/native";
 import { render as renderToast } from "roamjs-components/components/Toast";
 import { Column, Result } from "~/utils/types";
 import updateBlock from "roamjs-components/writes/updateBlock";
@@ -238,7 +238,7 @@ type ResultsViewComponent = (props: {
   preventSavingSettings?: boolean;
   onEdit?: () => void;
   onDeleteQuery?: () => void;
-  onRefresh: (loadInBackground?: boolean) => void;
+  onRefresh: (ignoreCache?: boolean) => void;
   globalFiltersData?: Record<string, Filters>;
   globalPageSize?: number;
   isEditBlock?: boolean;
@@ -1369,7 +1369,7 @@ const ResultsView: ResultsViewComponent = ({
                 <Kanban
                   data={allProcessedResults}
                   layout={layout}
-                  onQuery={() => onRefresh(true)}
+                  onQuery={() => onRefresh()}
                   resultKeys={columns}
                   parentUid={parentUid}
                   views={views}
--- a/apps/roam/src/utils/getDiscourseContextResults.ts
+++ b/apps/roam/src/utils/getDiscourseContextResults.ts
@@ -13,7 +13,7 @@ const CACHE_TIMEOUT = 1000 * 60 * 5;
 
 type BuildQueryConfig = {
   args: {
-    ignoreCache?: true;
+    ignoreCache?: boolean;
   };
   targetUid: string;
   fireQueryContext: {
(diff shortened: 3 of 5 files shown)


maparent:
@coderabbit review

apps/roam/src/components/DiscourseContext.tsx
```diff
@@ -345,17 +349,28 @@ export const ContextContent = ({ uid, results }: Props) => {
...
-  const onRefresh = useCallback(() => {
-    setRawQueryResults({});
-    getDiscourseContextResults({
-      uid,
-      onResult: addLabels,
-      ignoreCache: true,
-    }).finally(() => setLoading(false));
-  }, [uid, setRawQueryResults, setLoading, addLabels]);
+  // never set ignoreCache to true directly,
+  // as it may be negated by the useCallback cache.
```
mdroidian: `useCallback` does not “cache” results or arguments. It only memoizes the function reference (identity).

apps/roam/src/components/DiscourseContext.tsx
```diff
@@ -345,17 +349,28 @@ export const ContextContent = ({ uid, results }: Props) => {
...
+  const onRefresh = useCallback(
+    (ignoreCache?: boolean) => {
+      setRawQueryResults({});
+      void getDiscourseContextResults({
+        uid,
+        onResult: addLabels,
+        ignoreCache,
+      }).finally(() => setLoading(false));
+    },
+    [uid, setRawQueryResults, setLoading, addLabels, updateCount],
```
mdroidian: React Hook useCallback has an unnecessary dependency: 'updateCount'. Either exclude it or remove the dependency array.eslintreact-hooks/exhaustive-deps

mdroidian (changes requested):
`useCallback` shouldn't be the issue here.

Let's try to go back and identify what line of code introduced this regression.  From there, let's identify why that line of code is causing this behavior.  Then, let's find the solution.

I do not believe introducing another refresh function is the correct solution for this problem.

maparent:
@coderabbit review

maparent:
Ok, replaced the previous implementation by a different one. 
The root cause (introduced in ENG-1019) was adding the ignoreCache to the refreshButton; without which the refresh button was not working earlier. But onRefresh was also used at initial load, in the useEffect, and forcing a reload there is what made the UX glitch. So I still add ignoreCache as an optional parameter, I made the types consistent with that, and I set it to false in the useEffect.

maparent:
@mdroidian  A bit puzzled by the coderabbit suggestion here. TBH, I don't understand why static functions have to be in the arguments of `useCallback` or `useEffect`; but removing them (as coderabbit suggest) leads to an eslint complaint (`react-hooks/exhaustive-deps`). (Maybe because they're memoized functions and not functions?) Anyhow I think coderabbit is wrong, but there are react constraints I don't grok deeply enough, would appreciate your insight.

apps/roam/src/components/results-view/ResultsView.tsx
```diff
@@ -238,7 +238,7 @@ type ResultsViewComponent = (props: {
   preventSavingSettings?: boolean;
   onEdit?: () => void;
   onDeleteQuery?: () => void;
-  onRefresh: (loadInBackground?: boolean) => void;
```
mdroidian, line 241: `loadInBackground` is 'used' in QueryBuilder.tsx.  If we are changing this, should probably remove the `loadInBackground` reference.

apps/roam/src/components/results-view/ResultsView.tsx
```diff
@@ -1361,7 +1361,7 @@ const ResultsView: ResultsViewComponent = ({
                 <Kanban
                   data={allProcessedResults}
                   layout={layout}
-                  onQuery={() => onRefresh(true)}
+                  onQuery={() => onRefresh()}
```
mdroidian, line 1372: Could you add some context as to why this change was made, please?  Kanban quite seems unrelated to discourse context UI.

apps/roam/src/utils/getDiscourseContextResults.ts
```diff
@@ -176,7 +176,7 @@ const getDiscourseContextResults = async ({
   uid: string;
   nodes?: ReturnType<typeof getDiscourseNodes>;
   relations?: ReturnType<typeof getDiscourseRelations>;
-  ignoreCache?: true;
+  ignoreCache?: boolean;
```
mdroidian, line 179: This breaks CI

mdroidian (changes requested):
re: deps
React’s `exhaustive-deps` rule wants **every reactive value used inside an effect — including functions — in the dependency array** so React can re-run the effect with up-to-date values and avoid stale closures. [ref](https://react.dev/reference/eslint-plugin-react-hooks/lints/exhaustive-deps)

Taking a quick look, with the addition of `ignoreCache`, the `useEffect`/`onRefresh` could probably use a restructure.  Some thought should be put into the different scenarios and what the ramifications are.  

If this PR fixes the error and multiple scenarios in the UI have been tested (a loom video would **really** be great here), then this can probably be deffered.

maparent:
Loom video: https://www.loom.com/share/d0924cf5c6ab4743a92f805d91ba80d1?from_recorder=1&focus_title=1

apps/roam/src/components/results-view/ResultsView.tsx
maparent: Ok, I missed that one, thanks. And I agree with the quotation marks around 'used'. 
Two things I could do: Remove the loadInBackground reference, since it does so little; (and probably just not call setLoading, but also a decision);
OR add the parameter back, and have a signature with two parameters on the type everywhere.
Which do you prefer?

apps/roam/src/components/results-view/ResultsView.tsx
maparent: That is because I removed the loadInBackground flag, so I removed the one call instance which seemed to be using it. So it depends on what we decide to do with that flag, above.

apps/roam/src/components/results-view/ResultsView.tsx
mdroidian: Remove.  But keep the same behavior (which is `setLoading(true)` I believe)

apps/roam/src/components/results-view/ResultsView.tsx
mdroidian: ok, well "load in background" can be statically `true` , so this line can stay the way it is if we change `setLoading(true)` where `loadInBackground` was being used.

maparent: statically false, but otherwise agreed.

apps/roam/src/components/results-view/ResultsView.tsx
maparent: (the linear comment did not carry, copying) Done. onRefresh is called twice, this time with true (hence setLoading(true) as you say, which amounts to load in background being statically false, not true as you say in the other comment.) and in the DiscourseContext.tsx:366 useEffect, where it used to be false. I think this was to avoid a "loading" flicker on first display. I think this flicker is very short, accurately represents processing, and not worth another parameter to suppress.

maparent: Done. onRefresh is called twice, this time with true (hence `setLoading(true)` as you say, which amounts to load in background being statically false, not true as you say in the other comment.) and in the DiscourseContext.tsx:366 `useEffect`, where it used to be false. I think this was to avoid a "loading" flicker on first display. I think this flicker is very short, accurately represents processing, and not worth another parameter to suppress.

mdroidian (approved):

PR #605, opened 2025-12-12
ENG-1154: Reload graph when left sidebar is enabled

sid597:
## Release Notes

* **New Features**
  * Replaced Beta panel with an interactive checkbox to enable/disable the left sidebar feature
  * Added reload confirmation alert that appears when enabling the feature

### DIFF
--- a/apps/roam/src/components/settings/GeneralSettings.tsx
+++ b/apps/roam/src/components/settings/GeneralSettings.tsx
@@ -1,15 +1,27 @@
-import React, { useMemo } from "react";
+import React, { useMemo, useState } from "react";
 import TextPanel from "roamjs-components/components/ConfigPanels/TextPanel";
 import FlagPanel from "roamjs-components/components/ConfigPanels/FlagPanel";
 import { getFormattedConfigTree } from "~/utils/discourseConfigRef";
 import refreshConfigTree from "~/utils/refreshConfigTree";
 import { DEFAULT_CANVAS_PAGE_FORMAT } from "~/index";
+import { Alert, Checkbox, Intent } from "@blueprintjs/core";
+import Description from "roamjs-components/components/Description";
+import createBlock from "roamjs-components/writes/createBlock";
+import deleteBlock from "roamjs-components/writes/deleteBlock";
 
 const DiscourseGraphHome = () => {
   const settings = useMemo(() => {
     refreshConfigTree();
     return getFormattedConfigTree();
   }, []);
+
+  const [leftSidebarEnabled, setLeftSidebarEnabled] = useState(
+    settings.leftSidebarEnabled.value || false,
+  );
+  const [leftSidebarUid, setLeftSidebarUid] = useState(
+    settings.leftSidebarEnabled.uid,
+  );
+  const [isAlertOpen, setIsAlertOpen] = useState(false);
 
   return (
     <div className="flex flex-col gap-4 p-1">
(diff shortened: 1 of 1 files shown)


sid597:
* **#605** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/605?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/605?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sid597:
@coderabbitai full review

apps/roam/src/components/settings/GeneralSettings.tsx
```diff
@@ -30,14 +42,47 @@ const DiscourseGraphHome = () => {
...
+              setIsAlertOpen(true);
+            });
+          } else {
+            if (leftSidebarUid) {
+              void deleteBlock(leftSidebarUid);
+              setLeftSidebarUid(undefined);
+            }
+            setLeftSidebarEnabled(false);
+          }
+        }}
```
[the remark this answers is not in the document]
sid597: not sure wdyt @mdroidian

apps/roam/src/components/settings/GeneralSettings.tsx
```diff
@@ -30,14 +42,47 @@ const DiscourseGraphHome = () => {
...
         defaultValue={DEFAULT_CANVAS_PAGE_FORMAT}
       />
-      <FlagPanel
-        title="(BETA) Left Sidebar"
-        description="Whether or not to enable the left sidebar."
-        order={2}
-        uid={settings.leftSidebarEnabled.uid}
-        parentUid={settings.settingsUid}
-        value={settings.leftSidebarEnabled.value || false}
+      <Checkbox
```
mdroidian: `<FlagPanel>` has an `options: {onChange}` param.
Let's revert back to using flagpanel so we don't need to introduce all of this state.

apps/roam/src/components/settings/GeneralSettings.tsx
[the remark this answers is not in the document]
mdroidian: Mostly good ideas, but probably not that big of a deal considering the settings revamp that is coming soon.

mdroidian (changes requested):

apps/roam/src/components/settings/GeneralSettings.tsx
sid597: The flow we are going for here is,
- if unchecked, and user clicks on it
- show an alert
- if user confirms then create the block in roam otherwise dont

this is not possible to do with the FlagPanel it is eager, does not depend or wait for the onChange if its present

mdroidian: @sid597 Oh, no I don't think that is the case.  Regardless of if the user reloads now or later, the block should still be created.

EG: 
- users shows intent to turn on left sidebar
- we notify them that these changes won't take place until reload
- user could decide that they don't want to reload no, but they are happy with the changes taking place later
- BONUS: we provide an affordance to reload for them

The point is that we make them aware that the changes will take place after reload.  The fact that we provide the mechanism to reload for them is a bonus.

> if user confirms then create the block in roam otherwise dont

Also, that is not what this code is doing either.  This code is adding/deleting the block on `<Checkbox>` change, not on `<Alert>` confirm/cancel.

sid597: oh my bad very bad, we do the same in suggestive mode and i was conflating that to this. Yes i agree, this is reload only not giving a confirmation choice to user.

mdroidian (approved):

PR #606, opened 2025-12-12
ENG-1145 multiple-errors-running-migrate-all-relations

maparent:
https://linear.app/discourse-graphs/issue/ENG-1145/multiple-errors-running-migrate-all-relations
There are two issues: One is that the migration code is run when reified relations are active, and hence use the reified relations instead of the patterns as basis. Wrap the migration in a change of setting, which happens to be async (made changes accordingly)
The other issue is hammering the server: the check-for-duplicate query is short enough, and likely to be used often enough, that it's worth not using the server for it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitHub export authentication handling to avoid stuck states on auth errors.

* **Refactor**
  * Migration UI reliably re-enables controls after processing.
  * Reified-relations flow simplified and now runs by default.
  * Internal shape/coordinate computations made immutable for safety.

* **Chores**
  * Settings persistence and feature toggles use non-blocking calls that suppress unhandled promise rejections.

### DIFF
--- a/apps/roam/src/components/Export.tsx
+++ b/apps/roam/src/components/Export.tsx
@@ -199,7 +199,7 @@ const ExportDialog: ExportDialogComponent = ({
       if (response.status === 401) {
         setGitHubAccessToken(null);
         setError("Authentication failed. Please log in again.");
-        setSetting("oauth-github", "");
+        await setSetting("oauth-github", "");
         return { status: 401 };
       }
       return { status: response.status };
@@ -222,7 +222,7 @@ const ExportDialog: ExportDialogComponent = ({
   const addToSelectedCanvas = async (pageUid: string) => {
     if (typeof results !== "object") return;
 
-    let props: Record<string, unknown> = getBlockProps(pageUid);
+    const props: Record<string, unknown> = getBlockProps(pageUid);
 
     const PADDING_BETWEEN_SHAPES = 20;
     const COMMON_BOUNDS_XOFFSET = 250;
@@ -309,10 +309,10 @@ const ExportDialog: ExportDialogComponent = ({
       let minY = Number.MAX_SAFE_INTEGER;
 
       shapes.forEach((shape) => {
-        let rightX = shape.x + shape.w;
-        let leftX = shape.x;
-        let topY = shape.y;
-        let bottomY = shape.y - shape.h;
+        const rightX = shape.x + shape.w;
+        const leftX = shape.x;
+        const topY = shape.y;
+        const bottomY = shape.y - shape.h;
 
         if (rightX > maxX) maxX = rightX;
         if (leftX < minX) minX = leftX;
--- a/apps/roam/src/components/settings/AdminPanel.tsx
+++ b/apps/roam/src/components/settings/AdminPanel.tsx
@@ -254,7 +254,7 @@ const MigrationTab = (): React.ReactElement => {
   const [useOngoing, setOngoing] = useState<boolean>(false);
   const [useDryRun, setDryRun] = useState<boolean>(false);
   const [enabled, setEnabled] = useState<boolean>(
-    getSetting("use-reified-relations"),
+    getSetting("use-reified-relations", false),
   );
   const doMigrateRelations = async () => {
     setOngoing(true);
@@ -336,7 +336,7 @@ const FeatureFlagsTab = (): React.ReactElement => {
         onChange={(e) => {
           const target = e.target as HTMLInputElement;
           setUseReifiedRelations(target.checked);
-          setSetting("use-reified-relations", target.checked);
+          void setSetting("use-reified-relations", target.checked).catch();
         }}
         labelElement={
           <>
--- a/apps/roam/src/components/settings/AdminPanel.tsx
+++ b/apps/roam/src/components/settings/AdminPanel.tsx
@@ -250,10 +250,12 @@ const NodeListTab = (): React.ReactElement => {
 
 const MigrationTab = (): React.ReactElement => {
   let initial = true;
-  const enabled = getSetting("use-reified-relations");
   const [useMigrationResults, setMigrationResults] = useState<string>("");
   const [useOngoing, setOngoing] = useState<boolean>(false);
   const [useDryRun, setDryRun] = useState<boolean>(false);
+  const [enabled, setEnabled] = useState<boolean>(
+    getSetting("use-reified-relations"),
+  );
   const doMigrateRelations = async () => {
     setOngoing(true);
     try {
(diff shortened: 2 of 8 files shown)


maparent:
@coderabbit review

maparent:
@coderabbit review

maparent:
@coderabbit review

apps/roam/src/components/settings/AdminPanel.tsx
```diff
@@ -271,6 +273,7 @@ const MigrationTab = (): React.ReactElement => {
         `Migration failed: ${(e as Error).message ?? "see console for details"}`,
       );
     } finally {
+      setEnabled(true);
```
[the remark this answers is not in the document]
maparent: Enabled gates this function being called at all.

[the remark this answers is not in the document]
mdroidian: @maparent what is the purpose of `setEnabled(true)` here?  It isn't also changing the saved setting, only the state, so something seems amiss.

apps/roam/src/components/Export.tsx
```diff
@@ -199,7 +199,7 @@ const ExportDialog: ExportDialogComponent = ({
       if (response.status === 401) {
         setGitHubAccessToken(null);
         setError("Authentication failed. Please log in again.");
-        setSetting("oauth-github", "");
+        await setSetting("oauth-github", "");
```
mdroidian: Are these change related to the PR?

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -117,15 +112,12 @@ export const createReifiedRelation = async ({
...
-    return await createReifiedBlock({
-      destinationBlockUid: await getOrCreateRelationPageUid(),
-      schemaUid: relationBlockUid,
-      parameterUids: {
-        sourceUid,
-        destinationUid,
-      },
-    });
-  }
+  return await createReifiedBlock({
```
mdroidian, line 115: nit `createReifiedRelation` is now only used to call another function, so it should probably be removed.

apps/roam/src/utils/migrateRelations.ts
```diff
@@ -13,50 +13,56 @@ const MIGRATION_PROP_NAME = "relation-migration";
...
-      const migrationDataOrig = dgData[MIGRATION_PROP_NAME];
-      let migrationData: Record<string, json> =
-        migrationDataOrig !== null &&
-        typeof migrationDataOrig === "object" &&
-        !Array.isArray(migrationDataOrig)
-          ? migrationDataOrig
-          : {};
-      if (migrationData[uid] !== undefined) {
-        console.debug(`reprocessed ${key}`);
+  await setSetting("use-reified-relations", false); // so queries use patterns
```
mdroidian, line 17: This is a little sketchy.  Even though it is awaited, it isn't clear that their server/backend will be caught up to us immediately querying for this setting after changing it.  We could get stale data.

I understand that the code for this is pretty deeply integrated, so I would suggest adding a slight timeout for same measure.

mdroidian (changes requested):
This PR is quite confusing to me.

The body states that:
> There are two issues:
- "...migration code is run when reified relations are active..."
- "...other issue is hammering the server..."

It looks like this PR contains the solution for "...migration code is run when reified relations are active...".  It also contains many unrelated es-lint fixes?  And it does not contain the solution (or a ticket reference) to "...other issue is hammering the server..."

But when I look at the linear task, it is for something entirely different: specific errors that occurred when the migration was run.  And those specific errors were not mentioned/addressed:

```
{message: 'java.util.concurrent.ExecutionException: java.lang… Comparison method violates its general contract!'}
```

```
{message: 'clojure.lang.ExceptionInfo: Fail to visit list ran…FFFFFFFFFFFFFFFFFFFFFFFF" 0]], :val-range [:all]}'}
```

So there are two changes I think we should make here (the timeout and understanding what `setEnabled(true)` is doing).  Also, I want to make sure we properly address and track all of the issues listed.  For that reason, I am going to block this PR.

Could you please:
- create a ticket specifically for "...migration code is run when reified relations are active..."
- reassign this PR to said ticket (so that [ENG-1145](https://linear.app/discourse-graphs/issue/ENG-1145/multiple-errors-running-migrate-all-relations) does not get closed without being addressed)
- create a ticket for "...other issue is hammering the server..." and link to it

Or, if it is the case you believe these changes make it so that the errors no longer show up when doing the migration, please create a loom video doing the migration on a backup akamatsu graph show that this is the case.

apps/roam/src/components/settings/AdminPanel.tsx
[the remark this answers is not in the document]
maparent: TBH I'm a bit surprised it's needed; but before I added  it, the button remained disabled after the migration.

apps/roam/src/utils/createReifiedBlock.ts
maparent: `createReifiedBlock` is the generic function that I expect to keep using. The other one is syntactic sugar for a special case, even if it's the only case being used now.

maparent:
Quick note on unrelated lint corrections: Some of them are auto-applied on save. Undoing those specifically is a bit of a pain. I know you want to keep changes distinct but I don't think it should apply to autosave changes.
What I think we should do is apply some automated lint corrections globally at some point in a single PR.

apps/roam/src/utils/migrateRelations.ts
maparent: Good idea. I'd expect an awaited function to actually do the thing it promises, _and_ it seems to do so in my test, but we're not 100% sure, and a slight timeout won't matter much in a multi-minute operation.

apps/roam/src/components/settings/AdminPanel.tsx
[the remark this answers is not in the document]
mdroidian: I don't understand.  How could the button have "remained disabled" if it was just clicked?  Or are you saying the button was

* not disabled
* user clicks the button
* after `doMigrateRelations()` the button changed to `disabled`

If that is the case, then somewhere else in the code is responsible for changing that state.  Let's find that and change it.

apps/roam/src/utils/createReifiedBlock.ts
mdroidian: If it is just syntactic sugar it can be renamed on import.  It doesn't need a separate function.

mdroidian:
> Quick note on unrelated lint corrections: Some of them are auto-applied on save. Undoing those specifically is a bit of a pain. I know you want to keep changes distinct but I don't think it should apply to autosave changes. 

For files that include changes related to the PR, this absolutely makes sense.

For files that do not include changes related to the PR, it just adds confusion and is quite easy to undo.

> What I think we should do is apply some automated lint corrections globally at some point in a single PR.

Yes, I believe there is a backlog task for this.

maparent:
I think all the files have changes related to the PR. In many cases, it's just handling the fact that setSettings becomes async, which creates new lint errors.

PR #608, opened 2025-12-12
eng-1135 define internalError

maparent:
created an internalError handler that either logs to console.error or sends a report to posthog; and optionally shows a toast to the user. This should provide a pattern that combines many common use cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Centralized and standardized error reporting across the app, improving consistency of user-facing error notifications and reducing redundant console/email noise.

* **Chores**
  * Added a unified internal error handler and refactored components to use it for richer diagnostics and safer error flows.
  * Consolidated telemetry on extension load to include version/build context for better monitoring.

### DIFF
--- a/apps/roam/src/components/Export.tsx
+++ b/apps/roam/src/components/Export.tsx
@@ -54,7 +54,7 @@ import {
 import calcCanvasNodeSizeAndImg from "~/utils/calcCanvasNodeSizeAndImg";
 import { DiscourseNodeShape } from "~/components/canvas/DiscourseNodeUtil";
 import { MAX_WIDTH } from "~/components/canvas/Tldraw";
-import sendErrorEmail from "~/utils/sendErrorEmail";
+import internalError from "~/utils/internalError";
 import { getSetting, setSetting } from "~/utils/extensionSettings";
 
 const ExportProgress = ({ id }: { id: string }) => {
@@ -468,13 +468,12 @@ const ExportDialog: ExportDialogComponent = ({
         id: "query-builder-export-success",
       });
     } catch (e) {
-      const error = e as Error;
-      renderToast({
-        content: "Looks like there was an error. The team has been notified.",
-        intent: "danger",
-        id: "discourse-graphs-error",
+      internalError({
+        error: e as Error,
+        type: "export-error",
+        userMessage:
+          "Looks like there was an error. The team has been notified.",
       });
-      sendErrorEmail({ error, type: "Export Dialog Failed" }).catch(() => {});
     } finally {
       setLoading(false);
       onClose();
@@ -688,18 +687,13 @@ const ExportDialog: ExportDialogComponent = ({
                     setError(`Unsupported export type: ${exportType}`);
                   }
                 } catch (e) {
-                  const error = e as Error;
-                  renderToast({
-                    id: "export-error",
-                    content:
+                  internalError({
+                    error: e as Error,
+                    type: "export-error",
+                    userMessage:
                       "Looks like there was an error. The team has been notified.",
-                    intent: "danger",
-                  });
-                  sendErrorEmail({
-                    error,
-                    type: "Export Dialog Failed",
                     context: { activeExportType, filename, results },
-                  }).catch(() => {});
+                  });
                   setDialogOpen(true);
                   setError((e as Error).message);
                 } finally {
--- a/apps/roam/src/components/Export.tsx
+++ b/apps/roam/src/components/Export.tsx
@@ -470,7 +470,7 @@ const ExportDialog: ExportDialogComponent = ({
     } catch (e) {
       internalError({
         error: e as Error,
-        type: "export-error",
+        type: "Export Dialog Failed",
         userMessage:
           "Looks like there was an error. The team has been notified.",
       });
@@ -689,7 +689,7 @@ const ExportDialog: ExportDialogComponent = ({
                 } catch (e) {
                   internalError({
                     error: e as Error,
-                    type: "export-error",
+                    type: "Export Dialog Failed",
                     userMessage:
                       "Looks like there was an error. The team has been notified.",
                     context: {
--- a/apps/roam/src/components/Export.tsx
+++ b/apps/roam/src/components/Export.tsx
@@ -692,7 +692,16 @@ const ExportDialog: ExportDialogComponent = ({
                     type: "export-error",
                     userMessage:
                       "Looks like there was an error. The team has been notified.",
-                    context: { activeExportType, filename, results },
+                    context: {
+                      activeExportType,
+                      filename,
+                      resultsCount: Array.isArray(results)
+                        ? results.length
+                        : undefined,
+                      sampleUids: Array.isArray(results)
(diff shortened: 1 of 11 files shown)


maparent:
@coderabbit review

maparent:
@coderabbit review

maparent:
@coderabbit review

apps/roam/src/components/Export.tsx
```diff
@@ -688,18 +687,13 @@ const ExportDialog: ExportDialogComponent = ({
...
-                  });
-                  sendErrorEmail({
-                    error,
-                    type: "Export Dialog Failed",
                     context: { activeExportType, filename, results },
-                  }).catch(() => {});
+                  });
                   setDialogOpen(true);
                   setError((e as Error).message);
                 } finally {
```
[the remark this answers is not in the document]
maparent: Good catch on fullResults. setError is for display, should not leak anything.

maparent:
@coderabbit yes internalError does the string-to-error conversion internally.

apps/roam/src/utils/internalError.ts
```diff
@@ -0,0 +1,55 @@
...
+      error = new Error(error);
+    } else if (!(error instanceof Error)) {
+      try {
+        const serialized = JSON.stringify(error);
+        error = new Error(serialized);
+      } catch {
+        error = new Error(typeof error);
+      }
+    }
+    posthog.captureException(error, context);
```
mdroidian: also add sendErrorEmail so we continue to get the same error emails.

mdroidian (changes requested):
Generally looks good.

Just the one code request: keep `sendErrorEmail` in place, I don't want to remove that.

Also, could you please record a loom video demonstrating this new functionality?  Could you also include (either in the loom video or with text+screenshot) some information on where this can be found in posthog, as well as what it looks like/features it enables/etc.

Also note: could we try to keep the PR names in sync with the linear task names please?  They are generally more informative and when the PR is a different name it can be harder to locate if you are going from one platform to the other.  Thanks.

maparent:
Added back sendErrorEmail as an option. I could do a loom video, but honestly nothing much to show; there is no ux for sending an error yet. Should I repurpose the send Email button in the admin panel, esp. now that it does send an email?

But what you do need indeed is where to look for errors:

[https://us.posthog.com/project/113823/error_tracking](https://us.posthog.com/project/113823/error_tracking)

Now… there are now more errors there than I expected; it seems that enabling error tracking enabled autocapture. I disabled that now, and the error tracking panel still shows.

Two more things need to be added: 

1. Posthog complains about an outdated sdk; I'll make a separate update PR
2. In production, we should upload the JS code maps, so the stack traces are more legible. That has to be put in the build process. I think that is a separate task, and this code is still usable before that.

apps/roam/src/utils/internalError.ts
```diff
@@ -0,0 +1,69 @@
...
+import { getVersionWithDate } from "~/utils/getVersion";
+import getCurrentUserDisplayName from "roamjs-components/queries/getCurrentUserDisplayName";
+import sendErrorEmail from "~/utils/sendErrorEmail";
+
+const internalError = ({
+  error,
+  userMessage,
+  type,
+  context,
+  sendEmail, // true by default
```
mdroidian: This is only true for ln 48.

Use `sendEmail = true` to make the default explicit.  This aligns with [typescript patterns](https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters).

apps/roam/src/utils/internalError.ts
```diff
@@ -0,0 +1,69 @@
...
+      } catch {
+        error = new Error(typeof error);
+      }
+    }
+    posthog.captureException(error, context);
+    if (sendEmail !== false) {
+      sendErrorEmail({
+        // by now error is an Error but TS did not figure it out
+        error: error as Error,
+        type,
```
mdroidian, line 48: Not really a big fan of having the default email subject/headers changed.  This changes the grouping and search functionality.  Could we please keep the existing functionality?  If you'd like to add a specific "slug" for posthog, feel free add that/convert the existing titles.

apps/roam/src/utils/internalError.ts
```diff
@@ -0,0 +1,69 @@
...
+        context,
+      }).catch(() => {
+        console.error("Could not send error email", error, type, context);
```
mdroidian: nit: We should catch the error from `sendErrorEmail` and log that for more insight into the actual error.

mdroidian (changes requested):
> Now… there are now more errors there than I expected; it seems that enabling error tracking enabled autocapture. I disabled that now, and the error tracking panel still shows.

"... and the error tracking panel still shows."
What does the error tracking panel still show?  Does it still show all of the autocaptured errors?  We need to confirm that we can get only the errors we intentional report before merging this.

This is a good example of what a loom video would help catch before PR review.

maparent:
Rephrasing: The error tracking panel (as given by the URL) would be not only empty, but would not even show a empty table for errors; only a "please activate autocapture". I did activate, thinking it needed to also be activated in code, but the activation was somehow in the shared state? Anyhow, deactivated now, and the table with past error still shows, and I did not get back to the "Please activate" empty page.

maparent:
And since I (fortunately) cannot go back to the previous un-initialized state, I could not make a very useful loom video here.

mdroidian:
Can you confirm that we are only getting the errors we intentionally report?

maparent:
Hmm… actually, a few came through after I cut off the autocapture setting, so I'm puzzled. (Fewer than before though?) Some of them come from the website, looking into that, but not all of them. Doing some improvement on posthog setup while I'm there.

maparent:
Ok. I am looking at the errors that leaked through between my turning off autocapture on the server and now, and they come from other instances. I suspect that some of us started roam during those hours when autocapture was on, the posthog init caught that, and those roam instances will keep that setting until reloaded. Anyhow it started through action on the posthog website, and has nothing to do with the code here, which was not checked in. So I say there is no reason not to check this in.<br>I did push a few more changes: the adminPanel send email button now uses internalError; I moved some event parameters from the internalError function to posthog initialization.

maparent:
@coderabbit review

maparent:
Loom video: https://www.loom.com/share/e7c56253472e42a6863bcf7ed66f8ba5

maparent:
(btw I was wrong about there being website events; they are in a separate project, as appropriate.)

apps/roam/src/index.ts
```diff
@@ -39,13 +39,28 @@ import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTit
...
+      posthog.identify(userUid, {
+        graphName,
+      });
+      posthog.register({
+        version: version || "-",
+        buildDate: buildDate || "-",
+        graphName,
+      });
+      posthog.capture("Extension Loaded");
+    },
```
mdroidian, line 63: As long as we merge the ability to opt-out (and this is included in the opt-out) shortly after this and before Roam Depot release, then this should be fine.

mdroidian (approved):

PR #610, opened 2025-12-13
ENG-1155 use Result instead of UUID in CreateRelationDialog

maparent:
Refactor: Use Result[] in CreateRelationDialog

## Summary by CodeRabbit

* **Refactor**
  * Improved internal code organization of the Create Relation dialog component to enhance maintainability.

### DIFF
--- a/apps/roam/src/components/CreateRelationDialog.tsx
+++ b/apps/roam/src/components/CreateRelationDialog.tsx
@@ -73,13 +73,21 @@ const CreateRelationDialog = ({
     text: "",
     uid: "",
   });
-  const pageUidByTitle = Object.fromEntries(
-    window.roamAlphaAPI.data.fast.q(
-      "[:find ?t ?u :where [?p :node/title ?t] [?p :block/uid ?u]]",
-    ) as [string, string][],
+  const pageUidByTitle = useMemo(
+    () =>
+      Object.fromEntries(
+        window.roamAlphaAPI.data.fast.q(
+          "[:find ?t ?u :where [?p :node/title ?t] [?p :block/uid ?u]]",
+        ) as [string, string][],
+      ),
+    [],
   );
-  const allPages = Object.entries(pageUidByTitle).map(
-    ([text, uid]): Result => ({ uid, text }),
+  const allPages = useMemo(
+    () =>
+      Object.entries(pageUidByTitle).map(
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -71,33 +69,45 @@ const CreateRelationDialog = ({
...
-  const allPages = useMemo(() => getAllPageNames().sort(), []);
-  const getFilteredPageNames = (selectedRelationName: string): string[] => {
+  const [selectedTarget, setSelectedTarget] = useState<Result>({
+    text: "",
+    uid: "",
+  });
+  const pageUidByTitle = useMemo(
+    () =>
+      Object.fromEntries(
+        window.roamAlphaAPI.data.fast.q(
```
mdroidian, line 79: nit: use `data.async.fast.q` to future proof (note: it returns a promise)
https://roamresearch.com/#/app/developer-documentation/page/VJfOx4mgR

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -71,33 +69,45 @@ const CreateRelationDialog = ({
...
+          "[:find ?t ?u :where [?p :node/title ?t] [?p :block/uid ?u]]",
+        ) as [string, string][],
+      ),
+    [],
+  );
+  const allPages = useMemo(
```
mdroidian, line 85: nit: we don't need two useMemo's and multiple passes, eg

```typescript
const allPages = useMemo(() => {
  const rows = window.roamAlphaAPI.data.fast.q(
    "[:find ?t ?u :where [?p :node/title ?t] [?p :block/uid ?u]]",
  ) as [string, string][];

  return rows.map(([text, uid]): Result => ({ uid, text }));
}, []);
```

mdroidian (approved):

maparent:
Ah, sorry did not see the nits. I'll do them separately.

apps/roam/src/components/CreateRelationDialog.tsx
maparent: Actually, I use both variables separately, so I think it's worth having two useMemos.

PR #611, opened 2025-12-14
ENG-1156 use the pattern relation queries in migration

maparent:
The migration code is run when reified relations are active, and hence use the reified relations instead of the patterns as basis. Wrap the migration in a change of setting, which happens to be async (made changes accordingly)

This is a rename of the #606 PR, with a few bits sanded off.

### DIFF
--- a/apps/roam/src/components/settings/AdminPanel.tsx
+++ b/apps/roam/src/components/settings/AdminPanel.tsx
@@ -260,9 +260,7 @@ const MigrationTab = (): React.ReactElement => {
   const [useMigrationResults, setMigrationResults] = useState<string>("");
   const [useOngoing, setOngoing] = useState<boolean>(false);
   const [useDryRun, setDryRun] = useState<boolean>(false);
-  const [enabled, setEnabled] = useState<boolean>(
-    getSetting("use-reified-relations", false),
-  );
+  const enabled = getSetting("use-reified-relations", false);
   const doMigrateRelations = async () => {
     setOngoing(true);
     try {
@@ -280,7 +278,6 @@ const MigrationTab = (): React.ReactElement => {
         `Migration failed: ${(e as Error).message ?? "see console for details"}`,
       );
     } finally {
-      setEnabled(true);
       setOngoing(false);
     }
   };
--- a/apps/roam/src/components/settings/AdminPanel.tsx
+++ b/apps/roam/src/components/settings/AdminPanel.tsx
@@ -257,10 +257,12 @@ const NodeListTab = (): React.ReactElement => {
 
 const MigrationTab = (): React.ReactElement => {
   let initial = true;
-  const enabled = getSetting("use-reified-relations");
   const [useMigrationResults, setMigrationResults] = useState<string>("");
   const [useOngoing, setOngoing] = useState<boolean>(false);
   const [useDryRun, setDryRun] = useState<boolean>(false);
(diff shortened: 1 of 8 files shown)


maparent:
Note: This is essentially a past PR, so I'm not asking coderabbit again. The loom video shows that the timeout code is not broken. The setEnabled(true) is hard to justify but empirically needed for the button to behave well.

apps/roam/src/components/settings/AdminPanel.tsx
```diff
@@ -278,6 +280,7 @@ const MigrationTab = (): React.ReactElement => {
         `Migration failed: ${(e as Error).message ?? "see console for details"}`,
       );
     } finally {
+      setEnabled(true);
```
mdroidian: As mentioned [here](https://github.com/DiscourseGraphs/discourse-graph/pull/606#discussion_r2617375076), this needs to be addressed.

If it is the case that without `setEnabled(true)`:
- button is enabled
- user clicks button to run `doMigrateRelations()`
- button changes to disabled

This means something elsewhere is changing/affecting the state.

We should identify what is doing that and fix it at the cause.

Adding code like this will just make debugging harder and likely cause more errors / unknown behaviors.

mdroidian (changes requested):

apps/roam/src/components/settings/AdminPanel.tsx
maparent: Ok, reverted to previous version and the pattern I saw does happen anymore? I was puzzled before, even more now.

apps/roam/src/utils/migrateRelations.ts
```diff
@@ -13,50 +13,60 @@ const MIGRATION_PROP_NAME = "relation-migration";
...
-      const migrationDataOrig = dgData[MIGRATION_PROP_NAME];
-      let migrationData: Record<string, json> =
-        migrationDataOrig !== null &&
-        typeof migrationDataOrig === "object" &&
-        !Array.isArray(migrationDataOrig)
-          ? migrationDataOrig
-          : {};
-      if (migrationData[uid] !== undefined) {
-        console.debug(`reprocessed ${key}`);
+  // eslint-disable-next-line @typescript-eslint/await-thenable
```
mdroidian: We should take the time to fix the definition so that these ignores are not required.

mdroidian (approved):

apps/roam/src/utils/migrateRelations.ts
maparent: Yes, but the definition is in roamjs-components. Do you want me to open a PR there?

mdroidian: It is not in roamjs-components.  It is here [https://github.com/DiscourseGraphs/discourse-graph/blob/main/apps/roam/src/utils/extensionSettings.ts](https://github.com/DiscourseGraphs/discourse-graph/blob/main/apps/roam/src/utils/extensionSettings.ts)

maparent: oh, my mistake!

maparent: Ah, no it was another problem: My eslint got confused between frequent branch changes. The definition is fine, I just added the eslint-disable for spurious reasons. Removed.

mdroidian (approved):

PR #612, opened 2025-12-14
ENG-1157 Do not use a backend query in createReifiedNode

maparent:
https://linear.app/discourse-graphs/issue/ENG-1157/do-not-use-a-backend-query-in-createreifiednode

### DIFF
--- a/apps/roam/src/utils/createReifiedBlock.ts
+++ b/apps/roam/src/utils/createReifiedBlock.ts
@@ -21,15 +21,11 @@ const strictQueryForReifiedBlocks = async (
   const query = `[:find ?u ?d
   :in $ ${paramsAsSeq.map(([k]) => "?" + k).join(" ")}
   :where [?s :block/uid ?u] [?s :block/props ?p] [(get ?p :${DISCOURSE_GRAPH_PROP_NAME}) ?d]
-  ${paramsAsSeq.map(([k]) => `[(get ?d :${k}) ?_${k}] [(= ?${k} ?_${k})]`).join(" ")} ]`;
-  // Note: the extra _k binding variable is only needed for the backend query somehow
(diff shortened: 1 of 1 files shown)


maparent:
Again this is code that was already checked elsewhere.

apps/roam/src/utils/createReifiedBlock.ts
```diff
@@ -21,15 +21,11 @@ const strictQueryForReifiedBlocks = async (
...
-  ${paramsAsSeq.map(([k]) => `[(get ?d :${k}) ?_${k}] [(= ?${k} ?_${k})]`).join(" ")} ]`;
-  // Note: the extra _k binding variable is only needed for the backend query somehow
-  // In a local query, we can directly map to `[(get ?d :${k}) ?${k}]`
-  const result = await Promise.resolve(
-    window.roamAlphaAPI.data.backend.q(
-      query,
-      ...paramsAsSeq.map(([, v]) => v),
-    ) as [string, Record<string, string>][],
-  );
+  ${paramsAsSeq.map(([k]) => `[(get ?d :${k}) ?${k}]`).join(" ")} ]`;
```
mdroidian, line 24: Note: this line has removed `[(= ?${k} ?_${k})]`.  Because this seems unrelated to the task and there is no comment giving context as to why it was included, it is unclear if this is a mistake.

mdroidian (approved):

apps/roam/src/utils/createReifiedBlock.ts
maparent: If you look at the diff, I said in an earlier (erased) comment that this was needed by the backend and not by the frontend. Do you want me to reinstate the comment?

mdroidian: I missed that, sorry.

PR #624, opened 2025-12-18
ENG-1149 add try / catch to migration when querying for relation patterns

maparent:
https://linear.app/discourse-graphs/issue/ENG-1149/add-try-catch-to-migration-when-querying-for-relation-patterns
Added a try-catch to getRelationData, this will also affect export.
I used the new internalError, I think it's the simplest way to alert both the user and the developers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for relation data retrieval: failed relation fetches are now handled gracefully, returning empty results and surfacing clearer, user-facing error messages instead of causing unhandled exceptions. This reduces visible failures and provides more predictable behavior when relation data cannot be fetched.

### DIFF
--- a/apps/roam/src/utils/getRelationData.ts
+++ b/apps/roam/src/utils/getRelationData.ts
@@ -55,6 +55,7 @@ export const getRelationDataUtil = async (
             type: "Get relation data",
             userMessage: `Could not find relations of type ${s.label}`,
           });
+          return [];
         }
       }),
   ).then((r) => r.flat());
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

maparent:
I'm adding a loom video showing no error, which is not that illustrative... except that I have not broken the system.
https://www.loom.com/share/25f11d64030d45218afe71f9bcd7fe19
(I also tried with the akamatsulab migration)
I think that adding a fake error is a bit out of scope.

maparent:
@coderabbit review

mdroidian (approved):

PR #627, opened 2025-12-19
ENG-1164 Ability to add a canvas page to the clipboard to view the nodes on that canvas

maparent:
https://linear.app/discourse-graphs/issue/ENG-1164/ability-to-add-a-canvas-page-to-the-clipboard-to-view-the-nodes-on

Allow the references in a Canvas page to be accessible through the Canvas clipboard.
Did this through simply expanding getAllReferencesOnPage to handle canvas pages.
Curious if this would have side effects on Hyde? I expect Hyde to never deal with Canvas pages.
Also: Could a Canvas page also have non-canvas node references?
Also: Can canvas references include non-node pages? I do not (yet) filter them out. Would that be required?

## Summary by CodeRabbit

* **Refactor**
  * Reorganized internal utilities for improved code maintainability.
  * Minor formatting adjustments to query structures with no functional impact.

### DIFF
--- a/apps/roam/src/utils/getAllReferencesOnPage.ts
+++ b/apps/roam/src/utils/getAllReferencesOnPage.ts
@@ -1,0 +1,51 @@
+import normalizePageTitle from "roamjs-components/queries/normalizePageTitle";
+import type { Result } from "./types";
+import { isCanvasPage } from "./isCanvasPage";
+
+const getAllReferencesOnPage = async (title: string): Promise<Result[]> => {
+  title = normalizePageTitle(title);
+  let referencedPages: Array<[string, string]> = [];
+  if (isCanvasPage({ title })) {
+    const oldCanvasContent = window.roamAlphaAPI.data.fast.q(
+      `[:find ?uid ?title
+      :where
+      [?c :node/title "${title}"]
+      [?c :block/props ?props]
+      [(get ?props :roamjs-query-builder) ?rqb]
+      [(get ?rqb :tldraw) [[?k ?v]]]
+      [(get ?v :props) ?shape-props]
+      [(get ?shape-props :uid) ?uid]
+      [?n :block/uid ?uid]
+      [?n :node/title ?title]
+      ]`,
+    ) as Array<[string, string]>;
+    const newCanvasContent = window.roamAlphaAPI.data.fast.q(
+      `[:find ?uid ?title
+      :where
+      [?c :node/title "${title}"]
+      [?c :block/props ?props]
+      [(get ?props :roamjs-query-builder) ?rqb]
+      [(get ?rqb :tldraw) ?tldraw]
+      [(get ?tldraw :store) [[?k ?v]]]
+      [(get ?v :props) ?shape-props]
+      [(get ?shape-props :uid) ?uid]
+      [?n :block/uid ?uid]
+      [?n :node/title ?title]
+      ]`,
+    ) as Array<[string, string]>;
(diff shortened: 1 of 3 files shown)


maparent:
@coderabbit review

maparent:
Loom video: https://www.loom.com/share/3abd7ce868564f5aaa486e83e3cf40eb

maparent:
@coderabbit and @mdroidian 
I think that making type assertions at the boundary of two systems is warranted in general. 
Here, we know the output shape from the datalog query structure.

apps/roam/src/utils/getAllReferencesOnPage.ts
```diff
@@ -0,0 +1,53 @@
+import normalizePageTitle from "roamjs-components/queries/normalizePageTitle";
+
+const getAllReferencesOnPage = async (
+  pageTitle: string,
+): Promise<{ uid: string; text: string }[]> => {
+  pageTitle = normalizePageTitle(pageTitle);
+  let referencedPages: Array<[string, string]> = [];
+  if (pageTitle.startsWith("Canvas/")) {
```
mdroidian: Canvas page format is defined in settings, so this could change.  Let's make sure we grab the up to date format.

apps/roam/src/utils/getAllReferencesOnPage.ts
```diff
@@ -0,0 +1,53 @@
+import normalizePageTitle from "roamjs-components/queries/normalizePageTitle";
+
+const getAllReferencesOnPage = async (
+  pageTitle: string,
+): Promise<{ uid: string; text: string }[]> => {
```
mdroidian: nit: uid;text; is Result[], which is commonly used in the repo

apps/roam/src/utils/getAllReferencesOnPage.ts
```diff
@@ -0,0 +1,53 @@
...
+  pageTitle = normalizePageTitle(pageTitle);
+  let referencedPages: Array<[string, string]> = [];
+  if (pageTitle.startsWith("Canvas/")) {
+    const oldCanvasContent = window.roamAlphaAPI.data.fast.q(
```
mdroidian, line 9: nit: we should see if we can reuse some code with CanvasReferences

mdroidian (changes requested):
We need to grab the defined canvas page format, but looks good other than that.

> Curious if this would have side effects on Hyde? I expect Hyde to never deal with Canvas pages.

Should be fine.

> Also: Could a Canvas page also have non-canvas node references?

Currently, all references are dg node shapes.

> Also: Can canvas references include non-node pages? I do not (yet) filter them out. Would that be required?

Yes.  But this is already filtered out elsewhere.

apps/roam/src/utils/getAllReferencesOnPage.ts
maparent: Not really, because it's the opposite pattern (find canvases from node vs find nodes from canvas). What I could do if you want is create a utility function in CanvasReferences so both versions are colocated.

apps/roam/src/utils/getAllReferencesOnPage.ts
maparent: using isCanvasPage. Renamed the parameter to make that easier.

mdroidian (changes requested):
CI is failing.

maparent:
Looks like it was an artifact; check-types did pass locally. I just did a rebase on main/force-push, retriggered, and all is fine.

mdroidian (approved):

PR #628, opened 2025-12-20
ENG-1153 maximum update limit error on canvas after adding supports relation between evd and clm

maparent:
https://linear.app/discourse-graphs/issue/ENG-1153/maximum-update-limit-error-on-canvas-after-adding-supports-relation
The issue was that the reparent arrow function looking at non-arrow siblings in a way that did not take relations into account, leading to a reparenting cascade.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed index calculation logic when reparenting arrows to prevent conflicts with relation bindings.
  * Improved sibling selection and neighbor determination to correctly position arrows relative to adjacent elements.
  * Prevents visual overlap issues when repositioning bound relation arrows, ensuring proper ordering in diagram layout.

### DIFF
--- a/apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationBindings.tsx
+++ b/apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationBindings.tsx
@@ -231,7 +231,7 @@ function reparentArrow(editor: Editor, arrowId: TLShapeId) {
   let finalIndex: IndexKey;
 
   const relationIds = new Set(getDiscourseRelations().map((r) => r.id));
-  relationIds.add(arrow.type);
+  relationIds.add("arrow");
 
   // if the next sibling is also a bound arrow though, we can end up
   // all fighting for the same indexes. so lets find the next
--- a/apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationBindings.tsx
+++ b/apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationBindings.tsx
@@ -31,6 +31,7 @@ import {
   removeArrowBinding,
 } from "./helpers";
 import { AddReferencedNodeType } from "./DiscourseRelationTool";
+import getDiscourseRelations from "~/utils/getDiscourseRelations";
 
 export const createAllReferencedNodeBindings = (
   allAddReferencedNodeByAction: AddReferencedNodeType,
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

maparent:
Loom video: https://www.loom.com/share/045e028ea47740d9b073874b1c291ac5

maparent:
I think relations are kinds of arrows, so the variable rename is unnecessary.
The caching makes a lot of sense, there are many ways to do it and I would prefer for Micheal to indicate preference here.

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationBindings.tsx
```diff
@@ -229,29 +230,30 @@ function reparentArrow(editor: Editor, arrowId: TLShapeId) {
 
   let finalIndex: IndexKey;
 
+  const relationIds = new Set(getDiscourseRelations().map((r) => r.id));
+  relationIds.add(arrow.type);
```
mdroidian: Change this to `"arrow"` to include regular arrows.

You can test this by binding a regular arrow to the target shape and the error persists.

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationBindings.tsx
```diff
@@ -229,29 +230,30 @@ function reparentArrow(editor: Editor, arrowId: TLShapeId) {
...
 
   if (higherSiblings.length) {
     // there are siblings above the highest bound sibling, we need to
     // insert between them.
 
-    // if the next sibling is also a bound arrow though, we can end up
-    // all fighting for the same indexes. so lets find the next
-    // non-arrow sibling...
-    const nextHighestNonArrowSibling = higherSiblings.find(
-      (sibling) => sibling.type !== arrow.type,
```
mdroidian, line 246: Could we just change this to check for all arrow types?  Seems like that should solve it.

```typescript
      (sibling) => !relationIds.has(sibling.type),
```

So we could remove all other changes and it would just be a 3 line change.

```typescript
const relationIds = new Set(getDiscourseRelations().map((r) => r.id));
relationIds.add("arrow");
// ...
      (sibling) => !relationIds.has(sibling.type),
// ...
```

mdroidian (approved):
Nice find.

I've approved this, but we need to include type `arrow` otherwise the error still persists if there are multiple relations+regular arrows bound to the same shape.

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationBindings.tsx
maparent: Interesting, because that was following the original pattern, so maybe that was buggy too?

apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationBindings.tsx
maparent: That fails downstream, when we assume that the nextHighestNonArrowSibling exists at line 267.

mdroidian: Interesting, it didn't fail for me when I tested it.

PR #643, opened 2025-12-24
ENG-1208 Add local flag to fireQuery

maparent:
https://linear.app/discourse-graphs/issue/ENG-1208/add-local-flag-to-firequery
Problem: Battering the roam server with complex queries leads to errors.

In particular, this battering happens in the migration to materialize queries.

Solution: The alternative is to make some of the queries local instead, when we know we will make many queries.


maparent:
This is equivalent to #539 , minus the dead code. The same loom video should apply.

mdroidian (approved):

PR #656, opened 2025-12-29
[ENG-1177] Authenticate Obsidian user for publishing

trangdoan982:
https://www.loom.com/share/40904660f20b4a478019f86363d4c12f

## Summary by CodeRabbit

* **New Features**
  * (BETA) Added a Sync Mode toggle in General Settings to enable vault synchronization with a remote service.
  * New settings persisted: space password, account local ID, and syncModeEnabled.
  * When enabled, sync initialization runs on plugin load and after saving settings; users receive success and error notices.


trangdoan982:
@coderabbitai full review

apps/obsidian/src/utils/supabaseContext.ts
```diff
@@ -0,0 +1,142 @@
...
+        platform,
+      });
+
+      if (!spaceResult.data) {
+        console.error("Failed to create space");
+        return null;
+      }
+
+      const spaceId = spaceResult.data.id;
+      const userId = await fetchOrCreatePlatformAccount({
```
trangdoan982, line 94: I'm following Roam pattern here to create both Anonymous User and the real user (name here is vaultName instead of "Anonymous of space [x]". But I would say we don't need this for Obsidian since each vault is a space, and is owned by only one person? hence all the handling would go through only need to go through 1 account? I can just comment out this part of the code to delete this real user creation

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -0,0 +1,15 @@
...
+): Promise<void> => {
+  const context = await getSupabaseContext(plugin);
+  if (!context) {
+    throw new Error("Failed to initialize Supabase sync: could not create context");
+  }
+  console.log("Supabase sync initialized successfully", {
+    spaceId: context.spaceId,
+    userId: context.userId,
+  });
+};
```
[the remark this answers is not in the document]
trangdoan982: just the id so it should be fine

trangdoan982:
@coderabbitai full review

trangdoan982:
* **#665** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/665?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#656** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/656?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/656?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/obsidian/scripts/compile.ts
```diff
@@ -100,6 +119,15 @@ export const compile = ({
...
       external: external,
+      define: {
+        "process.env.SUPABASE_URL": dbEnv.SUPABASE_URL
+          ? `"${dbEnv.SUPABASE_URL}"`
+          : "null",
+        "process.env.SUPABASE_ANON_KEY": dbEnv.SUPABASE_ANON_KEY
+          ? `"${dbEnv.SUPABASE_ANON_KEY}"`
+          : "null",
+        "process.env.NEXT_API_ROOT": `"${dbEnv.NEXT_API_ROOT || ""}"`,
+      },
```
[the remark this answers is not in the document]
trangdoan982: similar to Roam

mdroidian (commented):
@maparent could you review this PR when you have a minute?

maparent:
Sorry did not see it was waiting for me. Will prioritize this.

apps/obsidian/src/utils/supabaseContext.ts
maparent: Hmm... Doesn't obsidian have a multi-user solution? And, in that case, can we detect which page was created/edited by whom? Is it in the metadata?
If so we would definitely still need the anonymous account. 
In all cases I think we had agreed to punt that.

packages/database/supabase/functions/create-space/index.ts
```diff
@@ -61,6 +61,12 @@ const spaceValidator = (space: SpaceCreationInput): string | null => {
     return "Missing or invalid platform.";
   if (!password || typeof password !== "string" || password.length < 8)
     return "password must be at least 8 characters";
+  if (platform === "Obsidian") {
```
maparent: -1 on platform tests in database functions.
What you're trying to do is extend this function to allow for single-owner platforms, which might apply to obsidian (tbd). Then let that be part of the input: if this information is given, we're creating a single-owner space, otherwise not.
Note: This may also imply creating utility functions to transform single-owner spaces into multi-owner spaces, eg if the user activates sharing on their obsidian vault; for this reason I'm not enthusiastic about this change, but I won't block on that, as long as it's not gated by platform, only by intent.
Another downside of doing things this way: We now have a user account without email, so no way to verify how their identity corresponds to another user another platform with the same email. Still thinking through the impact of that, but that could be blocking.
The way out is to have them create a DG account with supabase, which will verify the email, and then let them register themselves as owner of the vault. This should be separate from registering the vault with the database. It may be reasonable to move the space from anonymous to single-user in that scenario, but benefit unclear.

maparent:
Note: We agreed to revert back to roam-style anonymous users for now.

apps/obsidian/src/components/GeneralSettings.tsx
```diff
@@ -311,6 +333,23 @@ const GeneralSettings = () => {
         </div>
       </div>
 
+      <div className="setting-item">
```
trangdoan982: After we merge https://github.com/DiscourseGraphs/discourse-graph/pull/656, I'll move this setting to the FeatureFlagPanel

apps/obsidian/scripts/compile.ts
```diff
@@ -10,6 +11,19 @@ import autoprefixer from "autoprefixer";
 
 dotenv.config();
 
+// For local dev: Set SUPABASE_USE_DB=local and run `pnpm run genenv` in packages/database
```
maparent, line 14: Comment is fine, but let's also ensure it's automated in turbo.
The "with" statement at line 64-67 of `turbo.json` now also applies to obsidian. Not sure if there's a way to apply it to all apps? Otherwise just add an obsidian stanza.

apps/obsidian/src/components/canvas/shapes/DiscourseNodeShape.tsx
```diff
@@ -13,6 +13,7 @@ import {
   TLDefaultFontStyle,
   FONT_SIZES,
   FONT_FAMILIES,
+  toDomPrecision,
```
maparent: Is this from an unrelated PR?

apps/obsidian/src/utils/supabaseContext.ts
```diff
@@ -0,0 +1,155 @@
...
+  plugin: DiscourseGraphPlugin,
+): Promise<SupabaseContext | null> => {
+  if (contextCache === null) {
+    try {
+      const vaultName = plugin.app.vault.getName() || "obsidian-vault";
+
+      const spacePassword = await getOrCreateSpacePassword(plugin);
+      const accountLocalId = await getOrCreateAccountLocalId(plugin, vaultName);
+
+      const url = `space${accountLocalId}`;
```
maparent: I expect this to be a URI (url was a misnomer I guess.) We had said we could use a UUID-based URN, i.e. of the form    urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6
(following https://datatracker.ietf.org/doc/html/rfc4122)
I have no objection to using something that includes the vault name plus some random fluff instead; but let's at least make it look likely like a URN. (eg urn:x-obsidian:vaultname:uuid .... )
Actually since you're researching Obsidian sync: let's see if they have chosen an identifier themselves? Then we could align to it.

maparent (changes requested):
Good overall, one addition and one point may benefit from research.

maparent (commented):
Somehow that comment is pending in a separate review? Making sure you see it.

apps/obsidian/src/utils/supabaseContext.ts
trangdoan982: yes I found out how Obsidian creates the vaultID, used that instead.

apps/obsidian/scripts/compile.ts
trangdoan982: i will create a follow-up ticket for this: [ENG-1258: Automate env generation for Obsidian](https://linear.app/discourse-graphs/issue/ENG-1258/automate-env-generation-for-obsidian)

maparent: ah, well.

maparent:
Graphite seems to be changing the base again? Make sure the base remains main! (Again this is the base here in GitHub interface, nothing to do with git rebasing.)

apps/obsidian/src/utils/supabaseContext.ts
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * Gets the unique vault ID from Obsidian's internal API.
+ * @see https://help.obsidian.md/Extending+Obsidian/Obsidian+URI
+ */
+const getVaultId = (app: DiscourseGraphPlugin["app"]): string => {
+  return (app as unknown as { appId: string }).appId;
+};
+
+const canonicalObsidianUrl = (vaultId: string): string => {
+  return `obsidian://vault=${vaultId}`;
```
maparent: nit: could be simply `obsidian:<vaultId>` (the `vault=` part makes it less like an uri.

maparent (approved):

PR #659, opened 2025-12-31
ENG-1230 Backlinks in JSON-LD export

maparent:
https://linear.app/discourse-graphs/issue/ENG-1230/backlinks-in-json-ld-export
Add backlink information in json-ld export
Create a backlink relationship based on common elements (dcterms) and also reference to the document structure ontology (which allows describing roam block containment)
Accessorily: Added ontology info to dg-core, consolidated per-page calculations in a single function for better progress tracking.

Loom: https://www.loom.com/share/b9fc610d5d9349e0aff6af16dc801636

## Release Notes

* **New Features**
  * Added backlink functionality to display all pages that reference the current page.
  * Enhanced vocabulary schema to better support semantic relationships between content.
  * Improved JSON-LD export with comprehensive relationship and backlink tracking.
  * Refined progress tracking during data export operations.

### DIFF
--- a/apps/roam/src/utils/jsonld.ts
+++ b/apps/roam/src/utils/jsonld.ts
@@ -171,7 +171,7 @@ export const getJsonLdData = async ({
         .filter((x) => nodeSet.has(x));
       const r: Record<string, string | string[]> = {
         "@id": `pages:${uid}`, // eslint-disable-line @typescript-eslint/naming-convention
-        "@type": nodeType ?? "nodeSchema", // eslint-disable-line @typescript-eslint/naming-convention
+        "@type": nodeType, // eslint-disable-line @typescript-eslint/naming-convention
         title: text,
         content,
         modified: modified?.toJSON(),
--- a/apps/roam/src/utils/jsonld.ts
+++ b/apps/roam/src/utils/jsonld.ts
@@ -107,7 +107,10 @@ export const getJsonLdData = async ({
   nodeLabelByType: Record<string, string>;
   updateExportProgress: (progress: number) => Promise<void>;
 }): Promise<
-  Record<string, string | Record<string, string> | Record<string, string>[]>
+  Record<
+    string,
+    string | Record<string, string> | Record<string, string | string[]>[]
+  >
 > => {
   const roamUrl = canonicalRoamUrl();
   const getRelationData = () =>
--- a/apps/roam/src/utils/jsonld.ts
+++ b/apps/roam/src/utils/jsonld.ts
@@ -17,6 +17,7 @@ export const jsonLdContext = (baseUrl: string): Record<string, string> => ({
   prov: "http://www.w3.org/ns/prov#",
   sioc: "http://rdfs.org/sioc/ns#",
   dgb: "https://discoursegraphs.com/schema/dg_base",
+  dg: "https://discoursegraphs.com/schema/dg_core",
   subClassOf: "rdfs:subClassOf",
   title: "dc:title",
   label: "rdfs:label",
@@ -32,6 +33,7 @@ export const jsonLdContext = (baseUrl: string): Record<string, string> => ({
   relationDef: "dgb:RelationDef",
   relationInstance: "dgb:RelationInstance",
   inverseOf: "owl:inverseOf",
+  backlink: "dg:backlink",
   pages: `${baseUrl}/page/`,
 });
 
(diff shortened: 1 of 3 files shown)


maparent:
@coderabbit review

maparent:
Loom: https://www.loom.com/share/b9fc610d5d9349e0aff6af16dc801636

apps/website/public/schema/dg_core.ttl
```diff
@@ -3,11 +3,18 @@
 @prefix : <http://www.w3.org/2000/01/rdf-schema#> .
 @prefix dc: <http://purl.org/dc/elements/1.1/> .
 @prefix owl: <http://www.w3.org/2002/07/owl#> .
-@prefix vs: <http://www.w3.org/2003/06/sw-vocab-status/ns#> .
-@prefix sioc: <http://rdfs.org/sioc/ns#> .
 @prefix prov: <http://www.w3.org/ns/prov#> .
 @prefix dgb: <https://discoursegraphs.com/schema/dg_base#> .
 @prefix dg: <https://discoursegraphs.com/schema/dg_core#> .
+# http://purl.org/spar/po
+@prefix po: <http://www.essepuntato.it/2008/12/pattern#> .
```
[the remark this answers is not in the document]
maparent: You are right about the URL of the ontology itself, but all the terms of the ontology use the URL I gave as base, as can be observed if reading http://purl.org/spar/po.xml
So I will use this as base when defining terms.

mdroidian (changes requested):
I'm going to be more stringent with adherence to the PR workflows going forward.  Please take a minute to re-read the handbook if necessary.

Please add the loom video to the body of the PR then re-tag me for review.

apps/roam/src/utils/jsonld.ts
```diff
@@ -131,40 +136,56 @@ export const getJsonLdData = async ({
...
+        "@id": `pages:${uid}`, // eslint-disable-line @typescript-eslint/naming-convention
+        "@type": nodeType, // eslint-disable-line @typescript-eslint/naming-convention
+        title: text,
+        content,
+        modified: modified?.toJSON(),
+        created: date.toJSON(),
+        creator: displayName,
+      };
+      if (backlinks.length > 0) {
+        r["backlinks"] = backlinks.map((x) => `pages:${x}`);
```
mdroidian: defined as `backlink` not `backlinks`

apps/roam/src/utils/jsonld.ts
```diff
@@ -131,40 +136,56 @@ export const getJsonLdData = async ({
...
+        internalError({
+          error: `Unknown node type "${type as string}" for page "${text}"`,
+        });
+      }
+      const backlinks = (
+        await (window.roamAlphaAPI.data.backend.q(
+          `[:find ?uid
+        :where
+          [?page :block/uid "${uid}"]
+          [?block :block/page ?page]
```
mdroidian, line 176: That clause requires the target page to have at least one block, which might not always be the case.

apps/roam/src/utils/jsonld.ts
```diff
@@ -131,40 +136,56 @@ export const getJsonLdData = async ({
...
+          [or [?refBlock :block/refs ?block] [?refBlock :block/refs ?page]]
+          [?refBlock :block/page ?refPage]
+          [?refPage :block/uid ?uid]
+          ]`,
+        ) as Promise<Array<[string]>>)
+      )
+        .map((x) => x[0])
+        .filter((x) => nodeSet.has(x));
```
mdroidian, line 184: Why are we filtering here?  This means that queries that only return a subset of nodes will not return all backlinks.

apps/roam/src/utils/jsonld.ts
```diff
@@ -131,40 +136,56 @@ export const getJsonLdData = async ({
...
+      const { text, uid, type } = page;
+      const { date, displayName, modified } = getPageMetadata(text);
+      const nodeType = nodeSchemaUriByName[type as string];
+      if (!nodeType) {
+        internalError({
+          error: `Unknown node type "${type as string}" for page "${text}"`,
+        });
+      }
+      const backlinks = (
+        await (window.roamAlphaAPI.data.backend.q(
```
mdroidian, line 172: Potential performance concern running this on every single link.  Consider batching.

mdroidian (approved):
I believe there are a number of issues with this approach still, but I won't block on them.

More importantly, I don't believe this is satisfying the original request still.  Have you communicated this approach with Matt to see if it will satisfy his request?  Specifically 

> Can't distinguish between different link contexts

Image

Also, I don't see any node specific filtering

> I'm looking for "is a backlink, and is also a node"

apps/roam/src/utils/jsonld.ts
maparent: We're doing this on every page, not on every link.

apps/roam/src/utils/jsonld.ts
maparent: Precisely to ensure we only mention backlinks from discourse graph nodes (node-specific filtering, requested above.)

apps/roam/src/utils/jsonld.ts
maparent: Good point. I tried to do this with an or-join, but or-join seems irremediably broken: The results are always much more than either clause separately. 
I will do two queries.

apps/roam/src/utils/jsonld.ts
mdroidian: Sorry, that's what I meant.  This would still be in the thousands.

maparent:
I corrected or answered the line issues, but you are right to point out the bigger issue that this does not conform to the request, which I obviously misread. 
I think that the request as stated can only be addressed through the work on markdown, which I still have to document better. So it's indeed possible that this whole PR is dead in the water, and we should go back to my earlier approach. 
But I'm also uncomfortable with the request. I think that it falls way outside of the initial scope of "basic json-ld export", and does not conform to json-ld expectations. As such it's a request for an ad-hoc format, which I think we should take time to define better so it's compliant to some standard. In particular, the linkContext should be a RDF resource, and probably part of a frame definition. I will raise this with @mattakamatsu and will put this PR back in draft mode until then.

apps/roam/src/utils/jsonld.ts
mdroidian: That isn't what this is achieving though.  This export can be called (and often is) on arbitrary queries.  So if a user calls it on a subset of dgraph nodes, the backlinks will be inaccurate.

maparent:
@mattakamatsu as stated I think we need to decide what we're doing with your request. We can add this as-is for now if it's of any use, or not; but I think we need to define better what would be a proper json-ld-esque format for what you need (or decide it's an ad-hoc thing that we will provide as a separate download format.)

Ideally, the sections should be broken down in the JSON-LD export, so that links to sections would be referring to subobjects in the JSON-LD. I think this makes the most sense if some of the template is explicitly linked to a frame structure; but we can still break down before that.

mattakamatsu:
Please recommend or implement a way for us to include the back links in
our discourse graphs export. At this stage, this should be a quick fix: I
doubt anyone aside some Joel and I will use this in the near term

---
Matthew Akamatsu, PhD
Assistant Professor, University of Washington
Department of Biology
He/his
https://www.biology.washington.edu/people/profile/matthew-akamatsu

maparent:
I'll try to come up with something early next week.

PR #661, opened 2026-01-01
ENG-1232 Create group table and use it in space access functions

maparent:
https://linear.app/discourse-graphs/issue/ENG-1232/create-group-table-and-use-it-in-space-access-functions

This is the second stage: Group table used in space access.

https://www.loom.com/share/39fdfa9ca9db45bb880f6846cd630729

## Summary by CodeRabbit

* **New Features**
  * Introduced group membership system enabling users to be organized into groups with admin-based access control.
  * Enhanced access control with group-aware permissions for secure management of resources and user access.
  * Added group membership verification to support role-based access across the platform.

### DIFF
--- a/packages/database/src/dbTypes.ts
+++ b/packages/database/src/dbTypes.ts
@@ -532,6 +532,24 @@ export type Database = {
           },
         ]
       }
+      group_membership: {
+        Row: {
+          admin: boolean | null
+          group_id: string
+          member_id: string
+        }
+        Insert: {
+          admin?: boolean | null
+          group_id: string
+          member_id: string
+        }
+        Update: {
+          admin?: boolean | null
+          group_id?: string
+          member_id?: string
+        }
+        Relationships: []
+      }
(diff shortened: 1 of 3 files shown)


maparent:
@coderabbit review

maparent:
@coderabbit I changed the base, review from 1231 only.

maparent:
@coderabbit good point on the null uid, corrected. The member index is implicit in the primary key, and not needed otherwise. The group index would only be useful in the group_exists function, but new group insertion should be rare enough, I don't think the cost of maintaining the index is worth the speedup of this rare case. Am I missing another use case where the group index would be useful?

maparent:
@coderabbit I'm reading your analysis chain, but I did not get an answer to my question.

maparent:
Actually, you are right, group membership is a good enough reason. Adding the index.

mdroidian (approved):

PR #666, opened 2026-01-04
ENG-1189: Prod: Init block prop based schema

sid597:
On load, if the schema page or blocks do not exist but should exist, we create them.

Currently we have a function `printAllSettings` that logs the default values in the console. We will remove this after review and before merge.

I would recommend first loading the plugin locally, see the console log and then move on to the code review. 

## Summary by CodeRabbit

* **Chores**
  * Enhanced settings initialization infrastructure with improved configuration management and block property organization.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/666">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #669, opened 2026-01-05
ENG-1187: Prod: Implement get and set for block prop based settings

sid597:
https://www.loom.com/share/3c07a184d5d740a59410ee49d1fead0c  

## Summary by CodeRabbit

* **New Features**
  * Expanded settings management: unified handling for feature flags, global, personal, and node-specific settings with stronger validation and safer nested reads/writes.
* **Changes**
  * Relations data moved from index-based lists to keyed maps for more reliable access and iteration.
  * Added caching and cache invalidation for discourse nodes/relations to improve performance and robustness.
  * Added lightweight stubs/hooks to support left-sidebar personal settings initialization.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/669">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
+++ b/apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
@@ -4,7 +4,7 @@ import type { DiscourseRelationSettings } from "~/components/settings/utils/zodS
 // TODO: Delete the original default relations in data/defaultRelations.ts when fully migrated.
 const DEFAULT_RELATIONS_BLOCK_PROPS: Record<string, DiscourseRelationSettings> =
   {
-    "_INFF-rel": {
+    "_INFO-rel": {
       label: "Informs",
       source: "_EVD-node",
       destination: "_QUE-node",
@@ -81,7 +81,7 @@ const DEFAULT_RELATIONS_BLOCK_PROPS: Record<string, DiscourseRelationSettings> =
         },
       ],
     },
-    "_OPPP-rel": {
+    "_OPPO-rel": {
       label: "Opposes",
       source: "_EVD-node",
       destination: "_CLM-node",
--- a/apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
+++ b/apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
@@ -1,142 +1,140 @@
-import type { DiscourseRelationSettings } from "~/components/settings/utils/zodSchema";
 /* eslint-disable @typescript-eslint/naming-convention */ // This is for nodePosition keys
+import type { DiscourseRelationSettings } from "~/components/settings/utils/zodSchema";
 
 // TODO: Delete the original default relations in data/defaultRelations.ts when fully migrated.
-const DEFAULT_RELATIONS_BLOCK_PROPS: DiscourseRelationSettings[] = [
+const DEFAULT_RELATIONS_BLOCK_PROPS: Record<string, DiscourseRelationSettings> =
   {
-    id: "informs",
-    label: "Informs",
-    source: "_EVD-node",
-    destination: "_QUE-node",
-    complement: "Informed By",
-    ifConditions: [
-      {
-        triples: [
-          ["Page", "is a", "source"],
-          ["Block", "references", "Page"],
-          ["Block", "is in page", "ParentPage"],
-          ["ParentPage", "is a", "destination"],
-        ],
-        nodePositions: {
-          "0": "100 57",
-          "1": "100 208",
-          "2": "100 345",
-          source: "281 57",
-          destination: "281 345",
+    "_INFF-rel": {
+      label: "Informs",
+      source: "_EVD-node",
+      destination: "_QUE-node",
+      complement: "Informed By",
+      ifConditions: [
+        {
+          triples: [
+            ["Page", "is a", "source"],
+            ["Block", "references", "Page"],
+            ["Block", "is in page", "ParentPage"],
+            ["ParentPage", "is a", "destination"],
+          ],
+          nodePositions: {
+            "0": "100 57",
+            "1": "100 208",
+            "2": "100 345",
+            source: "281 57",
+            destination: "281 345",
+          },
         },
-      },
-    ],
-  },
-  {
-    id: "supports",
-    label: "Supports",
-    source: "_EVD-node",
-    destination: "_CLM-node",
-    complement: "Supported By",
-    ifConditions: [
-      {
-        triples: [
-          ["Page", "is a", "source"],
-          ["Block", "references", "Page"],
-          ["SBlock", "references", "SPage"],
-          ["SPage", "has title", "SupportedBy"],
-          ["SBlock", "has child", "Block"],
-          ["PBlock", "references", "ParentPage"],
-          ["PBlock", "has child", "SBlock"],
-          ["ParentPage", "is a", "destination"],
-        ],
-        nodePositions: {
-          "0": "250 325",
-          "1": "100 325",
-          "2": "100 200",
-          "3": "250 200",
-          "4": "400 200",
-          "5": "100 75",
-          "6": "250 75",
-          source: "400 325",
-          destination: "400 75",
+      ],
+    },
+    "_SUPP-rel": {
+      label: "Supports",
+      source: "_EVD-node",
+      destination: "_CLM-node",
+      complement: "Supported By",
+      ifConditions: [
+        {
+          triples: [
+            ["Page", "is a", "source"],
+            ["Block", "references", "Page"],
+            ["SBlock", "references", "SPage"],
+            ["SPage", "has title", "SupportedBy"],
+            ["SBlock", "has child", "Block"],
+            ["PBlock", "references", "ParentPage"],
+            ["PBlock", "has child", "SBlock"],
+            ["ParentPage", "is a", "destination"],
+          ],
+          nodePositions: {
+            "0": "250 325",
+            "1": "100 325",
+            "2": "100 200",
+            "3": "250 200",
+            "4": "400 200",
+            "5": "100 75",
+            "6": "250 75",
+            source: "400 325",
+            destination: "400 75",
+          },
         },
-      },
-      {
-        triples: [
-          ["Page", "is a", "destination"],
-          ["Block", "references", "Page"],
-          ["SBlock", "references", "SPage"],
-          ["SPage", "has title", "Supports"],
-          ["SBlock", "has child", "Block"],
-          ["PBlock", "references", "ParentPage"],
-          ["PBlock", "has child", "SBlock"],
-          ["ParentPage", "is a", "source"],
-        ],
-        nodePositions: {
-          "7": "250 325",
-          "8": "100 325",
-          "9": "100 200",
-          "10": "250 200",
-          "11": "400 200",
-          "12": "100 75",
-          "13": "250 75",
-          source: "400 75",
-          destination: "400 325",
+        {
+          triples: [
+            ["Page", "is a", "destination"],
+            ["Block", "references", "Page"],
+            ["SBlock", "references", "SPage"],
+            ["SPage", "has title", "Supports"],
+            ["SBlock", "has child", "Block"],
+            ["PBlock", "references", "ParentPage"],
+            ["PBlock", "has child", "SBlock"],
+            ["ParentPage", "is a", "source"],
+          ],
+          nodePositions: {
+            "7": "250 325",
+            "8": "100 325",
+            "9": "100 200",
+            "10": "250 200",
+            "11": "400 200",
+            "12": "100 75",
+            "13": "250 75",
+            source: "400 75",
+            destination: "400 325",
+          },
         },
-      },
-    ],
-  },
-  {
-    id: "opposes",
-    label: "Opposes",
-    source: "_EVD-node",
-    destination: "_CLM-node",
-    complement: "Opposed By",
-    ifConditions: [
-      {
-        triples: [
-          ["Page", "is a", "source"],
-          ["Block", "references", "Page"],
-          ["SBlock", "references", "SPage"],
-          ["SPage", "has title", "OpposedBy"],
-          ["SBlock", "has child", "Block"],
-          ["PBlock", "references", "ParentPage"],
-          ["PBlock", "has child", "SBlock"],
-          ["ParentPage", "is a", "destination"],
-        ],
-        nodePositions: {
-          "0": "250 325",
-          "1": "100 325",
-          "2": "100 200",
-          "3": "250 200",
-          "4": "400 200",
-          "5": "100 75",
-          "6": "250 75",
-          source: "400 325",
-          destination: "400 75",
+      ],
+    },
+    "_OPPP-rel": {
+      label: "Opposes",
+      source: "_EVD-node",
+      destination: "_CLM-node",
+      complement: "Opposed By",
+      ifConditions: [
+        {
+          triples: [
+            ["Page", "is a", "source"],
+            ["Block", "references", "Page"],
+            ["SBlock", "references", "SPage"],
+            ["SPage", "has title", "OpposedBy"],
+            ["SBlock", "has child", "Block"],
+            ["PBlock", "references", "ParentPage"],
+            ["PBlock", "has child", "SBlock"],
+            ["ParentPage", "is a", "destination"],
+          ],
+          nodePositions: {
+            "0": "250 325",
+            "1": "100 325",
+            "2": "100 200",
+            "3": "250 200",
+            "4": "400 200",
+            "5": "100 75",
+            "6": "250 75",
+            source: "400 325",
+            destination: "400 75",
+          },
         },
-      },
-      {
-        triples: [
-          ["Page", "is a", "destination"],
-          ["Block", "references", "Page"],
-          ["SBlock", "references", "SPage"],
-          ["SPage", "has title", "Opposes"],
-          ["SBlock", "has child", "Block"],
-          ["PBlock", "references", "ParentPage"],
-          ["PBlock", "has child", "SBlock"],
-          ["ParentPage", "is a", "source"],
-        ],
-        nodePositions: {
-          "7": "250 325",
-          "8": "100 325",
-          "9": "100 200",
-          "10": "250 200",
-          "11": "400 200",
-          "12": "100 75",
-          "13": "250 75",
-          source: "400 75",
-          destination: "400 325",
+        {
+          triples: [
+            ["Page", "is a", "destination"],
+            ["Block", "references", "Page"],
+            ["SBlock", "references", "SPage"],
+            ["SPage", "has title", "Opposes"],
+            ["SBlock", "has child", "Block"],
+            ["PBlock", "references", "ParentPage"],
+            ["PBlock", "has child", "SBlock"],
+            ["ParentPage", "is a", "source"],
+          ],
+          nodePositions: {
+            "7": "250 325",
+            "8": "100 325",
+            "9": "100 200",
+            "10": "250 200",
+            "11": "400 200",
+            "12": "100 75",
+            "13": "250 75",
+            source: "400 75",
+            destination: "400 325",
+          },
         },
-      },
-    ],
-  },
-];
+      ],
+    },
+  };
 
 export default DEFAULT_RELATIONS_BLOCK_PROPS;
--- a/apps/roam/src/components/settings/utils/accessors.ts
+++ b/apps/roam/src/components/settings/utils/accessors.ts
@@ -316,12 +316,6 @@ export const getGlobalSetting = <T = unknown>(
   }, settings) as T | undefined;
 };
 
-let relationsCache: DiscourseRelationSettings[] | null = null;
-
-export const invalidateRelationsCache = (): void => {
-  relationsCache = null;
-};
-
 export const setGlobalSetting = (keys: string[], value: json): void => {
   if (keys.length === 0) {
     internalError({
@@ -346,23 +340,15 @@ export const setGlobalSetting = (keys: string[], value: json): void => {
     keys: [TOP_LEVEL_BLOCK_PROP_KEYS.global, ...keys],
     value,
   });
-
-  if (keys[0] === "Relations") {
-    invalidateRelationsCache();
-  }
 };
 
 export const getAllRelations = (): DiscourseRelationSettings[] => {
-  if (relationsCache) return relationsCache;
-
   const settings = getGlobalSettings();
 
-  relationsCache = Object.entries(settings.Relations).map(([id, relation]) => ({
+  return Object.entries(settings.Relations).map(([id, relation]) => ({
     ...relation,
     id,
   }));
-
-  return relationsCache;
 };
 
 export const getPersonalSettings = (): PersonalSettings => {
@@ -459,12 +445,6 @@ export const getDiscourseNodeSetting = <T = unknown>(
   }, settings) as T | undefined;
 };
 
-let discourseNodesCache: DiscourseNodeSettings[] | null = null;
-
-export const invalidateDiscourseNodesCache = (): void => {
-  discourseNodesCache = null;
-};
-
 export const setDiscourseNodeSetting = (
   nodeType: string,
   keys: string[],
@@ -510,12 +490,9 @@ export const setDiscourseNodeSetting = (
   }
 
   setBlockPropAtPath(pageUid, keys, value);
-  invalidateDiscourseNodesCache();
 };
 
 export const getAllDiscourseNodes = (): DiscourseNodeSettings[] => {
-  if (discourseNodesCache) return discourseNodesCache;
-
   const results = window.roamAlphaAPI.data.fast.q(`
     [:find ?uid ?title (pull ?page [:block/props])
      :where
@@ -556,6 +533,5 @@ export const getAllDiscourseNodes = (): DiscourseNodeSettings[] => {
     }
   }
 
-  discourseNodesCache = nodes;
   return nodes;
 };
--- a/apps/roam/src/components/settings/utils/accessors.ts
+++ b/apps/roam/src/components/settings/utils/accessors.ts
@@ -1,0 +1,558 @@
+import getBlockProps, {
+  normalizeProps,
+  type json,
+} from "~/utils/getBlockProps";
+import setBlockProps from "~/utils/setBlockProps";
+import getBlockUidByTextOnPage from "roamjs-components/queries/getBlockUidByTextOnPage";
+import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
+import internalError from "~/utils/internalError";
+import { z } from "zod";
+
+import {
+  DG_BLOCK_PROP_SETTINGS_PAGE_TITLE,
+  DISCOURSE_NODE_PAGE_PREFIX,
+  TOP_LEVEL_BLOCK_PROP_KEYS,
+  FeatureFlagsSchema,
+  GlobalSettingsSchema,
+  PersonalSettingsSchema,
+  DiscourseNodeSchema,
+  getPersonalSettingsKey,
+  type FeatureFlags,
+  type GlobalSettings,
+  type PersonalSettings,
+  type DiscourseNodeSettings,
+  type DiscourseRelationSettings,
+} from "./zodSchema";
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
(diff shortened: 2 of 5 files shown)


sid597:
* **#745** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/745?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#684](https://github.com/DiscourseGraphs/discourse-graph/pull/684) <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/684?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#688](https://github.com/DiscourseGraphs/discourse-graph/pull/688) <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/688?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#689** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/689?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#682** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/682?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#669** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/669?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/669?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#666** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/666?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#719** <a href="https://app.graphite.com/github/pr/DiscourseGraphs/discourse-graph/719?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sid597:
@coderabbit ai full review

sid597:
@coderabbitai full review

apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
```diff
@@ -1,9 +1,8 @@
-import type { DiscourseRelationSettings } from "~/components/settings/utils/zodSchema";
 /* eslint-disable @typescript-eslint/naming-convention */ // This is for nodePosition keys
 
 // TODO: Delete the original default relations in data/defaultRelations.ts when fully migrated.
-const DEFAULT_RELATIONS_BLOCK_PROPS: DiscourseRelationSettings[] = [
-  {
+const DEFAULT_RELATIONS_BLOCK_PROPS = {
+  informs: {
     id: "informs",
```
mdroidian: If you are changing to a map, is this `id` still required?  Seems like duplicate information.

apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
```diff
@@ -1,9 +1,8 @@
-import type { DiscourseRelationSettings } from "~/components/settings/utils/zodSchema";
 /* eslint-disable @typescript-eslint/naming-convention */ // This is for nodePosition keys
 
 // TODO: Delete the original default relations in data/defaultRelations.ts when fully migrated.
-const DEFAULT_RELATIONS_BLOCK_PROPS: DiscourseRelationSettings[] = [
-  {
+const DEFAULT_RELATIONS_BLOCK_PROPS = {
+  informs: {
```
mdroidian: The id needs to be unique, so probably a uid.  And it should be shown in the example too as to not be confused with label.

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+  type PersonalSettings,
+  type DiscourseNodeSettings,
+  type DiscourseRelationSettings,
+} from "./zodSchema";
+
+export const getDiscourseNodePageTitle = (nodeLabel: string): string => {
+  return `${DISCOURSE_NODE_PAGE_PREFIX}${nodeLabel}`;
+};
+
+export const getDiscourseNodePageUid = (nodeLabel: string): string => {
```
mdroidian: probably unnecessary

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+  DiscourseNodeSchema,
+  getPersonalSettingsKey,
+  type FeatureFlags,
+  type GlobalSettings,
+  type PersonalSettings,
+  type DiscourseNodeSettings,
+  type DiscourseRelationSettings,
+} from "./zodSchema";
+
+export const getDiscourseNodePageTitle = (nodeLabel: string): string => {
```
mdroidian: unnecessary

mdroidian: also, why is this exported?

apps/roam/src/components/settings/utils/accessors.ts
mdroidian: also why is this exported?

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+
+export const getDiscourseNodePageUid = (nodeLabel: string): string => {
+  const pageTitle = getDiscourseNodePageTitle(nodeLabel);
+  return getPageUidByPageTitle(pageTitle);
+};
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === "object" && value !== null && !Array.isArray(value);
+
+export const getBlockPropsByUid = (
```
mdroidian: Same with this, it is unclear if this is exported because it is intended to be used elsewhere, or is it just for accessors.ts

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+    return allBlockProps;
+  }
+
+  const targetValue = keys.reduce((currentContext: json, currentKey) => {
+    if (
+      currentContext &&
+      typeof currentContext === "object" &&
+      !Array.isArray(currentContext)
+    ) {
+      const value = (currentContext as Record<string, json>)[currentKey];
```
mdroidian: why do we need an assertion here?

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+  keys: string[],
+  value: json,
+): void => {
+  if (!blockUid) {
+    console.warn("[DG:accessor] setBlockPropsByUid called with empty blockUid");
+    return;
+  }
+
+  if (keys.length === 0) {
+    setBlockProps(blockUid, value as Record<string, json>, false);
```
mdroidian: why is assertion needed here?

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+
+    if (
+      !currentContext[currentKey] ||
+      typeof currentContext[currentKey] !== "object" ||
+      Array.isArray(currentContext[currentKey])
+    ) {
+      currentContext[currentKey] = {};
+    }
+
+    return currentContext[currentKey] as Record<string, json>;
```
mdroidian: why is assertion needed here?

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+    return;
+  }
+
+  if (keys.length === 0) {
+    setBlockProps(blockUid, value as Record<string, json>, false);
+    return;
+  }
+
+  const currentProps = getBlockProps(blockUid);
+  const updatedProps = JSON.parse(JSON.stringify(currentProps || {})) as Record<
```
mdroidian: why is assertion needed here?

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+    }
+    return null;
+  }, allBlockProps);
+
+  return targetValue === null ? undefined : targetValue;
+};
+
+export const setBlockPropsByUid = (
+  blockUid: string,
+  keys: string[],
```
mdroidian, line 174: We could have this zod/type safe so the DX is easier and safer.

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+      const value = (currentContext as Record<string, json>)[currentKey];
+      return value === undefined ? null : value;
+    }
+    return null;
+  }, allBlockProps);
+
+  return targetValue === null ? undefined : targetValue;
+};
+
+export const setBlockPropsByUid = (
```
mdroidian: This function is overloaded and confusing:
- keys.length === 0 → treats value as a root props object (Record<string, json>) and merges it.
- keys.length > 0 → treats value as a leaf value and deep-sets it into the existing props.

That creates confusion because the signature suggests one operation (“set value at path”), but the empty-path case behaves like “merge an object into root,” and it requires the caller to know value must be an object in that branch.

Some cleaner options
Option A: Separate functions (most explicit)
 - mergeBlockProps(uid, partialProps)
- setBlockPropAtPath(uid, keys, value)

Option B: Keep one function, but make behavior explicit via API

- Require keys to be non-empty for path sets; throw or warn if empty.
- Provide a dedicated setAllProps(uid, props) / mergeRootProps(uid, props).

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
+import getBlockProps, { type json } from "~/utils/getBlockProps";
```
mdroidian: Please go through each function and only export the ones that are planned to be used

apps/roam/src/components/settings/utils/zodSchema.ts
```diff
@@ -201,7 +201,9 @@ export const GlobalSettingsSchema = z.object({
   "Left Sidebar": LeftSidebarGlobalSettingsSchema.default({}),
   Export: ExportSettingsSchema.default({}),
   "Suggestive Mode": SuggestiveModeGlobalSettingsSchema.default({}),
-  Relations: z.array(DiscourseRelationSchema).default(DEFAULT_RELATIONS_BLOCK_PROPS),
+  Relations: z
+    .record(z.string(), DiscourseRelationSchema)
+    .default(DEFAULT_RELATIONS_BLOCK_PROPS as unknown as Record<string, DiscourseRelationSettings>),
```
mdroidian: Is the unknown assertion required here?  Ideally we can define this specifically, no?

apps/roam/src/components/settings/utils/zodSchema.ts
```diff
@@ -267,9 +269,14 @@ export const getPersonalSettingsKey = (): string => {
   return cachedPersonalSettingsKey;
 };
 
+export const TOP_LEVEL_BLOCK_PROP_KEYS = {
+  featureFlags: "Feature Flags",
```
mdroidian, line 286: Again, this is not DRY.  Ideally we can set this once.  Currently we are having to set it twice.

apps/roam/src/components/settings/utils/accessors.ts
mdroidian: Make sure other functions that use `keys[]` don't also follow this pattern

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+  return targetValue === null ? undefined : targetValue;
+};
+
+export const setBlockPropsByUid = (
+  blockUid: string,
+  keys: string[],
+  value: json,
+): void => {
+  if (!blockUid) {
+    console.warn("[DG:accessor] setBlockPropsByUid called with empty blockUid");
```
mdroidian: Any console.warn that would result in unexpected behavior to the user should be switched to an `internalError()` so we track it and inform the user.

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+  }
+
+  setBlockPropsByUid(pageUid, keys, value);
+  invalidateDiscourseNodesCache();
+};
+
+export const getAllDiscourseNodes = (): DiscourseNodeSettings[] => {
+  if (discourseNodesCache) return discourseNodesCache;
+
+  const results = window.roamAlphaAPI.q(`
```
mdroidian: fast.q

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+     :where
+     [?page :node/title ?title]
+     [?page :block/uid ?uid]
+     [(clojure.string/starts-with? ?title "${DISCOURSE_NODE_PAGE_PREFIX}")]]
+  `) as [string, string][];
+
+  const nodes: DiscourseNodeSettings[] = [];
+
+  for (const [pageUid, title] of results) {
+    const blockProps = getBlockPropsByUid(pageUid, []);
```
mdroidian: you should be able to pull the block props with the query, avoiding multiple calls

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+    if (!blockProps) continue;
+
+    const result = DiscourseNodeSchema.safeParse(blockProps);
+    if (result.success) {
```
mdroidian, line 520: we should internalError()  to track these failures

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,515 @@
...
+
+  if (!settings) return undefined;
+
+  return keys.reduce<unknown>((current, key) => {
+    if (!isRecord(current) || !(key in current)) return undefined;
+    return current[key];
+  }, settings) as T | undefined;
+};
+
+let discourseNodesCache: DiscourseNodeSettings[] | null = null;
```
mdroidian: What is the thought behind having a cache just for discourse node settings?

mdroidian (changes requested):
Good start.

Main items to address:

* **Key-path semantics:** Several functions accept `keys: string[]`, but the meaning of `keys` differs by function (sometimes it’s a traversal path, sometimes the first element is treated as a lookup key, etc.). Please make this explicit at the call site—either by splitting functions (e.g., `setAtPath` vs `setRoot` / `getFromSettingsBlock` vs `getByUid`) or by renaming parameters (`path` vs `settingsKeys`) and adding docstrings/examples.

* **Type safety for keys:** Callers currently have to manually know the valid keys for each schema. Can we make this more Zod/TypeScript-safe (e.g., typed path helpers, exported key constants, or a small set of typed getters/setters per settings domain) to avoid stringly-typed access?

* **Type assertions:** There are a number of `as ...` assertions. Please reduce these where possible by tightening types and/or adding type guards. If an assertion is truly necessary, add a short comment explaining why it is required, and why it is safe.

* **Export surface:** It’s unclear which exports are intended as public API vs internal helpers. Please audit exports and keep only the functions intended for callers outside `accessors.ts` (or move internal helpers to non-exported scope).

* **Error reporting:** Any `console.warn` that indicates a user-impacting or unexpected state should go through `internalError()` so we can track it and surface appropriate messaging.

mdroidian:
@sid597 Also make sure you format with prettier

apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
sid597: Yeah this is not required

apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
sid597: Sorry not sure what you mean, do we this is default data so you mean we create a default uid and instead of `informs` use that as a key e.g `oierst3892` ?? 

I think for the default ones we can use the label as the uniqueness just like we do for discourse node e.g ` _CLM-node`

mdroidian: Yes, we can do the same as `_CLM-node`. `_CLM-node` is specifically formatted as a roam `uid`, so you could do something similar with `informs`, `supports`, and `opposes`.

apps/roam/src/components/settings/utils/accessors.ts
sid597: Yeah this was loaded tackling 2 cases in one, we don't need the `length===0` case for now, since no one is using it and might not even be needed in future and if it is we can create this function then

apps/roam/src/components/settings/utils/accessors.ts
sid597: you mean we should have the caching for other settings as well? Its for discourse node because we have to run a query and thats expensive so we cache it

apps/roam/src/components/settings/utils/accessors.ts
sid597: I added validation on the setter functions that call this for eg `setGlobalSetting`, `setPersonalSetting` etc.

apps/roam/src/components/settings/utils/zodSchema.ts
sid597: We need these as separate static schema because they will be used by the pullWatchers on setup time as well.

sid597:
Did the following changes:

- use `_INF-rel`, `_SUP-rel` etc as the key for default relations
- reduce usage of `as ....` to bare minimum 
- use `internalError`
- Add schema validation on setters, so each call to `setGlobalSetting`, `setPersonalSetting`, first validates the schema and then sets the value. We already do this on read but now also on set. Before this review I was going with , do the validation on the component side itself e.g FlagPanel component first checks the value its setting and only then we set the value the cons of that is iff a user calls the underneath function of `setGlobalSetting` with any wrong value type
- removed all the exports that are not sure of if needed later.
- Inside `geAllDiscourseNodes` function use `fast.q` and get block props in same call instead of mutliple calls

apps/roam/src/components/settings/utils/accessors.ts
mdroidian: Which query are you referring to?

apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
```diff
@@ -1,10 +1,9 @@
-import type { DiscourseRelationSettings } from "~/components/settings/utils/zodSchema";
 /* eslint-disable @typescript-eslint/naming-convention */ // This is for nodePosition keys
+import type { DiscourseRelationSettings } from "~/components/settings/utils/zodSchema";
 
 // TODO: Delete the original default relations in data/defaultRelations.ts when fully migrated.
-const DEFAULT_RELATIONS_BLOCK_PROPS: DiscourseRelationSettings[] = [
-  {
-    id: "informs",
+const DEFAULT_RELATIONS_BLOCK_PROPS: Record<string, DiscourseRelationSettings> = {
+  "_INF-rel": {
```
mdroidian: This is 8 chars instead of 9.  Roam UID's are 9 chars.

apps/roam/src/components/settings/utils/init.ts
```diff
@@ -218,6 +204,12 @@ export const initSchema = async (): Promise<InitSchemaResult> => {
 
   setTimeout(() => {
     printAllSettings(blockUids, nodePageUids);
+
+    // TODO: REMOVE stub calls after testing
```
mdroidian: .

apps/roam/src/components/settings/utils/zodSchema.ts
mdroidian: Could you define them and their schema relationship once?  EG:

```typescript
const staticTopLevelEntries = [
  {
    propKey: "featureFlags" as const,
    key: "Feature Flags",
    schema: FeatureFlagsSchema,
  },
  { 
    propKey: "global" as const, 
    key: "Global", 
    schema: GlobalSettingsSchema 
  },
] as const;
```

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,655 @@
...
+    if (
+      !currentContext[currentKey] ||
+      typeof currentContext[currentKey] !== "object" ||
+      Array.isArray(currentContext[currentKey])
+    ) {
+      currentContext[currentKey] = {};
+    }
+
+    return currentContext[currentKey] ;
+  }, updatedProps);
```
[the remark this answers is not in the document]
mdroidian: @sid597

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,655 @@
...
+
+export const getGlobalSettings = (): GlobalSettings => {
+  const { blockProps } = getBlockPropBasedSettings({
+    keys: [TOP_LEVEL_BLOCK_PROP_KEYS.global],
+  });
+
+  return GlobalSettingsSchema.parse(blockProps || {});
+};
+
+export const getGlobalSetting = <T = unknown>(keys: string[]): T | undefined => {
```
mdroidian: make sure you format with prettier

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,655 @@
...
+     [(clojure.string/starts-with? ?title "${DISCOURSE_NODE_PAGE_PREFIX}")]]
+  `) as [string, string, Record<string, json> | null][];
+
+  const nodes: DiscourseNodeSettings[] = [];
+
+  for (const [pageUid, title, rawProps] of results) {
+    if (typeof pageUid !== "string" || typeof title !== "string") continue;
+    const rawBlockProps = rawProps?.[":block/props"];
+    const blockProps = rawBlockProps ? normalizeProps(rawBlockProps) : undefined;
+    if (!blockProps || !isRecord(blockProps) || Object.keys(blockProps).length === 0) continue;
```
mdroidian: make sure you format with prettier

apps/roam/src/components/settings/utils/init.ts
```diff
@@ -4,13 +4,19 @@ import { createPage, createBlock } from "roamjs-components/writes";
...
+  stubSetLeftSidebarPersonalSections,
+  stubGetLeftSidebarPersonalSections,
+  getAllDiscourseNodes,
+} from "./accessors";
 import {
   DiscourseNodeSchema,
   getTopLevelBlockPropsConfig,
   getPersonalSettingsKey,
 } from "~/components/settings/utils/zodSchema";
 import { DG_BLOCK_PROP_SETTINGS_PAGE_TITLE, DISCOURSE_NODE_PAGE_PREFIX } from "./zodSchema";
```
mdroidian: make sure you format with prettier

mdroidian (changes requested):
A few comments to address then I'd like to walk through this with you 1:1.  Also check the merge conflict.

apps/roam/src/components/settings/utils/accessors.ts
sid597: line 466 in function `getAllDiscourseNodes()`

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,558 @@
...
+
+export const getAllRelations = (): DiscourseRelationSettings[] => {
+  if (relationsCache) return relationsCache;
+
+  const settings = getGlobalSettings();
+
+  relationsCache = Object.values(settings.Relations);
+
+  return relationsCache;
+};
```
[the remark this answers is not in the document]
sid597: This a great catch, we previously did have the id in default data, then I removed it because I was not remembering its usecase, but this was the case. GG

apps/roam/src/components/settings/utils/zodSchema.ts
mdroidian: Could we not do this or? (this is from [this comment](https://github.com/DiscourseGraphs/discourse-graph/pull/669#discussion_r2724722736))

apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
```diff
@@ -1,142 +1,140 @@
...
-          ["Block", "is in page", "ParentPage"],
-          ["ParentPage", "is a", "destination"],
-        ],
-        nodePositions: {
-          "0": "100 57",
-          "1": "100 208",
-          "2": "100 345",
-          source: "281 57",
-          destination: "281 345",
+    "_INFF-rel": {
```
mdroidian: _INFO-rel

apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
```diff
@@ -1,142 +1,140 @@
...
-          "2": "100 200",
-          "3": "250 200",
-          "4": "400 200",
-          "5": "100 75",
-          "6": "250 75",
-          source: "400 325",
-          destination: "400 75",
+      ],
+    },
+    "_OPPP-rel": {
```
mdroidian: _OPPO-rel

apps/roam/src/components/settings/utils/zodSchema.example.ts
```diff
@@ -182,9 +182,8 @@ const globalSettings: GlobalSettings = {
       },
     ],
   },
-  Relations: [
-    {
-      id: "relation-uid-1",
+  Relations: {
+    "relation-uid-1": {
```
mdroidian: update to the default relation id

apps/roam/src/components/settings/utils/zodSchema.example.ts
```diff
@@ -204,8 +203,7 @@ const globalSettings: GlobalSettings = {
         },
       ],
     },
-    {
-      id: "relation-uid-2",
+    "relation-uid-2": {
```
mdroidian: update to the default relation id

apps/roam/src/components/settings/utils/accessors.ts
mdroidian: `getAllDiscourseNodes()` (just a handlefull of pages based on a simple string match) shouldn't be a heavy query, let's not cache it

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,561 @@
...
+): T | undefined => {
+  const settings = getGlobalSettings();
+
+  return keys.reduce<unknown>((current, key) => {
+    if (!isRecord(current) || !(key in current)) return undefined;
+    return current[key];
+  }, settings) as T | undefined;
+};
+
+let relationsCache: DiscourseRelationSettings[] | null = null;
```
mdroidian: If this is relation types, this does not need to be cached.  It shouldn't be a heavy query.  Let's remove this for now.

mdroidian (changes requested):
Let's just remove the caches for now (shouldn't need them, so let's not make it more complex than we need to).  We'll add them back in if we find a need for them.

Then there's a few more comments to address.

mdroidian (approved):
Actually I'll approve this, but tag me before merging if there are any issues with implementing the requests.

apps/roam/src/components/settings/data/defaultRelationsBlockProps.ts
sid597: What am I doing bro!

apps/roam/src/components/settings/utils/zodSchema.ts
sid597: We now define the key-schema relationship once in staticTopLevelEntries, and  `getTopLevelBlockPropsConfig` derives from it, so the pairing is no longer duplicated.

  We still keep TOP_LEVEL_BLOCK_PROP_KEYS as a simple key-only export because when consumers (like accessors.ts)
  try to access .key or .schema from the combined entries, the Zod schema types in the array cause ESLint
  no-unsafe-member-access / no-unsafe-call errors. The key-only object isolates consumers from those complex types.

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -0,0 +1,537 @@
...
+    }
+  }
+
+  if (!pageUid) {
+    internalError({
+      error: `setDiscourseNodeSetting - could not find page for: ${nodeType}`,
+      type: "DG Accessor",
+    });
+    return;
+  }
```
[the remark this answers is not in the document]
sid597: fixed in https://github.com/DiscourseGraphs/discourse-graph/pull/682/changes/36fe06000563a5c780ef2b95bb34c4f7af2fd689

PR #671, opened 2026-01-05
[ENG 1241] Create new contentVariant

trangdoan982:
add "full" as contentVariant to signify that the row is of full content of the page, variant "direct" is for the title

### DIFF
--- a/packages/database/src/dbTypes.ts
+++ b/packages/database/src/dbTypes.ts
@@ -568,18 +568,6 @@ export type Database = {
         }
         Relationships: []
       }
-      result: {
-        Row: {
-          max: string | null
-        }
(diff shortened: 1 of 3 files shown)


packages/database/src/dbTypes.ts
trangdoan982, line 576: this was generated after `pnpm run migrate`, and the result seems to match with the latest main, so i'm not sure why it marks these as diff

maparent (approved):

packages/database/src/dbTypes.ts
maparent: Maybe the supabase cli tool version? Sometimes that creates variants. But I get the same result as you, so I think it's fine.

PR #679, opened 2026-01-08
eng-1262: allow for same id, different triples

maparent:
The source of the problem is that I failed if many relations had the same source, target and name. But relations can differ through their triple and have the same Id. I do not send an internal error in that innocuous case anymore.
Also, I do not fail the process even in that case but take the first relation.
https://www.loom.com/share/1c829585567c49fa873d51697847ffbb

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined relation matching error handling to reduce false error reporting when multiple similar relations exist, with improved fallback behavior for edge cases.

### DIFF
--- a/apps/roam/src/components/CreateRelationDialog.tsx
+++ b/apps/roam/src/components/CreateRelationDialog.tsx
@@ -141,7 +141,7 @@ const CreateRelationDialog = ({
       if (setOfRels.size > 1) {
         internalError({
           type: "Create Relation dialog",
-          error: `Too many relations between ${selectedTargetType.type} and ${selectedSourceType.type}: ${setOfRels.values().join(",")}`,
+          error: `Too many relations between ${selectedTargetType.type} and ${selectedSourceType.type}: ${[...setOfRels].join(",")}`,
         });
         // let it still fall through to the first
       }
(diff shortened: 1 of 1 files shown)


maparent:
@coderabbit review

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -133,13 +133,18 @@ const CreateRelationDialog = ({
...
     }
-    if (candidateRelations.length !== 1) {
-      // This seems to happen... I need more data.
-      internalError({
-        type: "Create Relation dialog",
-        error: `Too many relations between ${selectedTargetType.type} and ${selectedSourceType.type}: ${candidateRelations.map((r) => r.id).join(",")}`,
-      });
-      return null;
+    if (candidateRelations.length > 1) {
+      // Control for one very innocuous case: Many times the same relation, with different triples.
```
mdroidian, line 137: What does "Many times the same relation, with different triples" mean?  Could you give an example.

mdroidian: After some time I figured out what you meant.  This would be a good example of where you could use the Loom video to show an example of what specifically caused the error that you are fixing, so there less room for ambiguity with terms.  In this case it wasn't clear that "same relation" and "many relations" and "similar types of relations" were all referring to "multiple definitions of the same relation type."

Image

mdroidian (approved):

apps/roam/src/components/CreateRelationDialog.tsx
maparent: fair enough!

PR #680, opened 2026-01-09
ENG-1260 Permission denied when attempting to sync megacog nodes w/ database for embeddings

maparent:
https://linear.app/discourse-graphs/issue/ENG-1260/permission-denied-when-attempting-to-sync-megacog-nodes-w-database-for

The upsert_content fails in the case of content update. It probably never worked, this has nothing to do with the latest changes; we were mostly testing with inserts. 
It turns out this is not an issue if the update policy is defined with "using" instead of "with check". The latter syntax applies only after the change; leaving the before case unspecified, hence restrictive. "using" used alone applies both before and after the update. 

https://www.loom.com/share/00ac64c512004161905909ed5a45cfdd

## Summary by CodeRabbit

* **Chores**
  * Updated database access control policies to improve consistency and ensure proper enforcement of content modification and sharing permissions across the platform.


maparent:
@coderabbit review

mdroidian (approved):

PR #683, opened 2026-01-11
ENG-1160 Send JS maps to posthog for better backtraces

maparent:
https://linear.app/discourse-graphs/issue/ENG-1160/send-js-maps-to-posthog-for-better-backtraces

Generate the Roam plugin's JS map, and send to PostHog in deploy GitHub action. Uses the git commit SHA as version marker, or the version number for official versions.
The git commit SHA is unfortunately for a merge pseudo-commit, not the head of the branch in the case of branch PRs. This will be corrected after https://github.com/PostHog/posthog/pull/45772 lands.

Incidentally unify posthog versions to an updated version.
Note: This design implies that posthog is not injected in manual deploys. (Is that something we would want?)

https://www.loom.com/share/08403b4435d14f33bd9937d503be49f4

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/683">
  <picture>
    Open with Devin
  </picture>
</a>


maparent:
@coderabbit review

apps/roam/scripts/deploy.ts
```diff
@@ -23,6 +24,13 @@ const deploy = async () => {
     process.exit(1);
   }
 
+  try {
+    posthog();
+  } catch (error) {
+    console.error("Deployment failed on posthog:", error);
+    process.exit(1);
+  }
```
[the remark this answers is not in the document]
maparent: the posthog cli finds the git commit hash on its own.

apps/roam/scripts/posthog.ts
```diff
@@ -0,0 +1,23 @@
...
+  const { POSTHOG_CLI_ENV_ID, POSTHOG_CLI_TOKEN } = process.env;
+  if (POSTHOG_CLI_ENV_ID === undefined)
+    throw new Error("Missing posthog variable: POSTHOG_CLI_ENV_ID");
+  if (POSTHOG_CLI_TOKEN === undefined)
+    throw new Error("Missing posthog variable: POSTHOG_CLI_TOKEN");
+  let cmd = "posthog-cli sourcemap inject --directory dist --project Roam";
+  if (version !== undefined) {
+    cmd = `${cmd} --version ${version}`;
+  }
+  execSync(cmd);
```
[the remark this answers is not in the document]
maparent: I think the hard-coded directory is stable enough. I will sanitize version; it remains to see if this argument will be used (hopefully through publish.)

maparent:
@mdroidian this is not exactly ready for review, but next step should be collaborative.

maparent:
@mdroidian Here is the status:
I modified my code to use the posthog GitHub action, which has pros and cons.
Pro: It allows me to set the version properly on release.
Con:
I have to apply posthog injection between build and deploy.
Currently deploy triggers a build. Everything works if I remove this explicit dependency, and invoke build explicitly before deploy.
(I was hoping that the turbo cache would not rebuild if I did a build, posthog and deploy sequence, but even though the cache is replayed, that also resets the outputs to the previous build state so the posthog work is lost.)
The downside of that, of course, is that we have to remember about the explicit dependency.
I think that deploy is unlikely to be called manually at all, it is meant for CI/CD.
But another option would be to have two separate actions, buildAndDeploy vs deployWithoutBuilding (these are meant to be descriptions, actual names tbd.) so people doing deploy would get the usual behaviour.
OR I could revert to my original code, make posthog part of the deploy process. (In that case I would like to have a way to know I'm deploying as part of a publish process, so I could set a semantic version accordingly.)

Ah and we still need to do the GitHub linking.

.github/workflows/roam-main.yaml
```diff
@@ -38,5 +38,17 @@ jobs:
...
+        with:
+          directory: apps/roam/dist
+          env-id: ${{ secrets.POSTHOG_CLI_ENV_ID }}
+          cli-token: ${{ secrets.POSTHOG_CLI_TOKEN }}
+          project: Roam
+          version: ${{ github.sha }}
+
       - name: Deploy
-        run: npx turbo run deploy --filter=roam
+        run: npx turbo run deploy --filter=roam --ui stream
```
mdroidian: What is the purpose of `--ui stream` as it relates to posthog JS maps?

.github/workflows/roam-release.yaml
```diff
@@ -32,5 +32,23 @@ jobs:
...
+          path: apps/roam
+
+      - name: Inject & upload source maps to PostHog
+        uses: PostHog/upload-source-maps@v0.4.6
+        with:
+          directory: apps/roam/dist
+          env-id: ${{ secrets.POSTHOG_CLI_ENV_ID }}
+          cli-token: ${{ secrets.POSTHOG_CLI_TOKEN }}
+          project: Roam
+          version: ${{ steps.roam-version.outputs.current-version }}
```
mdroidian: Why is this using `npm-get-version` instead of `github.sha` like the other two?
And is there a way to do this via Node directly rather than using a dated `npm` package?

turbo.json
```diff
@@ -40,7 +42,7 @@
   ],
   "tasks": {
     "build": {
-      "dependsOn": ["@repo/database#genenv"],
+      "dependsOn": ["@repo/database#genenv", "^build"],
```
mdroidian: What problem is this solving?

mdroidian (changes requested):
> I think that deploy is unlikely to be called manually at all, it is meant for CI/CD.

I have called `deploy` manually myself several times.  It is useful when you need to test a `https://discoursegraphs.com/release/roam/` but are not ready or don't need to have a PR up.  So we should not remove this (or it needs to be replaced by something else).

> It allows me to set the version properly on release.

By version here to you mean package.json version or the github.sha?

maparent:
Status: The source maps are uploaded.
Of course a lot happens within Roam and is invisible to us, but it's still a gain.
The link to GitHub is not working as well as it should, for reasons I explained as a community question in 
https://posthog.com/docs/error-tracking/upload-source-maps/github-actions
I think this should not block us.

.github/workflows/roam-release.yaml
maparent: I am trying to get the publication version (eg 0.16) for official versions, partly to distinguish them from branch versions. The npm package was just the most convenient way I found to do this, open to alternatives.

.github/workflows/roam-main.yaml
maparent: Not related, opportunistic changing of turbo commands I'm touching otherwise. I think using the stream UI makes for nicer reports in github. I guess that could be a separate PR.

maparent (commented):
Some of my answers became part of a review? Still learning the Graphite UI. Hopefully will still show up close to original context.

.github/workflows/roam-release.yaml
mdroidian: Two concerns here:

1) Meaning/consistency: in this workflow `version` sometimes means a commit SHA (`github.sha`) and sometimes means the Roam package version. Can we rename outputs to make this unambiguous (e.g. `commitSha` vs `packageVersion`)?

2) Extra dependency: more of a nit, feels unnecessary. I figured since we already have Node, can't we just read the version directly?

turbo.json
mdroidian: What problem is this solving?

mdroidian (changes requested):
I'd like to understand the `turbo.json` change and see if we can clarify the `version` naming before continuing.

turbo.json
maparent: This is something I introduced while experimenting with the earlier approach and did not fully revert. Reverted now. 
I was also just surprised this was not the case already, but I now remember that we currently refer to libraries in their uncompiled state, so building them is not strictly necessary. That said... on looking deeper, I realized that we should probably introduce a dependency on the genTypes step. But also fully agreed that this is a totally different scope, which is why I reverted.

.github/workflows/roam-release.yaml
maparent: Extra dependency: removed. The only thing is that this is not obvious to test, so we'll see how it works at next release. The other mechanism was at least taken from someone who had tried it.

Meaning/consistency: version is the name of the input parameter to the GitHub command, no way to change that.

.github/workflows/roam-release.yaml
```diff
@@ -32,5 +32,21 @@ jobs:
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Build
+        run: npx turbo run build --filter=roam --ui stream
```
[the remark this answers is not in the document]
maparent: That is interesting, afaik this did not prevent the release from working; but we should check how well the official releases access the database.
Still feels right to correct it.
Strictly speaking, it's outside this PR's scope. Still correcting now.

mdroidian (changes requested):
This issue was not commented on or addressed:

https://github.com/DiscourseGraphs/discourse-graph/pull/683#discussion_r2724763800
> Meaning/consistency: in this workflow version sometimes means a commit SHA (github.sha) and sometimes means the Roam package version. Can we rename outputs to make this unambiguous (e.g. commitSha vs packageVersion)?

maparent:
In [https://github.com/DiscourseGraphs/discourse-graph/pull/683#discussion_r2726058142](<https://github.com/DiscourseGraphs/discourse-graph/pull/683#discussion_r2726058142>):<br>Meaning/consistency: version is the name of the input parameter to the GitHub command, no way to change that.

maparent:
Unless I misunderstand what you're asking for, in which case please clarify.

mdroidian (approved):
RE: Meaning/consistency

Ah, it didn't show up in my GitHub GUI.  At this rate, I really don't think GitHub will be here in 2027 ... 🙄

PR #693, opened 2026-01-13
ENG-1279 Discourse Context Overlay doesn't update number

maparent:
https://linear.app/discourse-graphs/issue/ENG-1279/discourse-context-overlay-doesnt-update-number

I added a way to update the number of the calling overlay.
Note it does not help with other overlays with the same node, which is a separate issue.

I also made the delays a bit longer, which helps with some discrepancies. 

https://www.loom.com/share/91b1b48e367349bc98dad6d2e80e6e98

## Summary by CodeRabbit

* **New Features**
  * Added support for an optional external overlay-refresh callback to coordinate updates between views.
* **Bug Fixes**
  * Increased delayed refresh timing to improve UI synchronization after changes.
  * Now signals refreshes to bypass cache after relation deletions to ensure consistent, up-to-date displays.

### DIFF
--- a/apps/roam/src/components/DiscourseContext.tsx
+++ b/apps/roam/src/components/DiscourseContext.tsx
@@ -31,6 +31,7 @@ export type DiscourseContextResults = Awaited<
 type Props = {
   uid: string;
   results?: DiscourseContextResults;
+  overlayRefresh?: () => void;
 };
 
 const ExtraColumnRow = (r: Result) => {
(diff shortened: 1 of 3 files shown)


maparent:
@coderabbit review

maparent:
@coderabbit review correction

maparent:
@coderabbit I'm still waiting for your review.

mdroidian (approved):
Thanks for calling out the other issues.

I wouldn't worry about the situation when the same overlay component for the same node doesn't show the correct results after change. That seems like quite the edge case that we can probably leave for now. 

And the visual issue ("weird supported by") is kinda wild, let's see if it happens often/gets more repeatable before tackling it.

👍

PR #706, opened 2026-01-15
[ENG-1178] Automatic local remote sync nodes relations

trangdoan982:
https://www.loom.com/share/f4d56e1519934b93b2a623a4e587d69c

https://www.loom.com/share/2116c7341cc8462aa5e03e0c4a017ec8

## Release Notes

* **New Features**
  * Obsidian plugin now automatically monitors file changes and synchronizes updates to the Discourse Graph database with debounced processing.
  * Enhanced node synchronization capabilities with support for tracking file renames, deletions, and content changes.
  * Improved change detection to identify and clean up orphaned nodes during synchronization.
  * Added support for Obsidian app origins in request validation.


trangdoan982:
@coderabbitai full review

apps/obsidian/src/utils/fileChangeListener.ts
```diff
@@ -0,0 +1,315 @@
...
+
+    console.log(`File renamed: ${oldPath} -> ${file.path}`);
+    this.queueChange(file.path, "title", oldPath);
+  }
+
+  /**
+   * Handle metadata changes (placeholder for relation metadata)
+   */
+  private handleMetadataChange(file: TFile): void {
+    if (!this.isDiscourseNode(file)) {
```
maparent, line 162: What of the case where a discourseNode stops being a discourseNode because someone removed the `nodeTypeId` was removed? 
Not sure you would have access to the old value in the metadataCache, OTH?
Also: What happens if someone changes or removes the nodeTypeId?

apps/obsidian/src/utils/fileChangeListener.ts
```diff
@@ -0,0 +1,315 @@
...
+      }
+
+      if (this.hasPendingOrphanCleanup) {
+        const deletedCount = await cleanupOrphanedNodes(this.plugin);
+        if (deletedCount > 0) {
+          console.debug(`Deleted ${deletedCount} orphaned node(s)`);
+        }
+      }
+
+      this.changeQueue.clear();
```
maparent: This comes after a lot of async work; new changes may have happened in the meantime and be lost. I suggest making the clear immediately after line 251, when the events are copied, so getting the queue is atomic. Then you'd want a try/catch, so you can put elements back if it fails... and then worry about recurring errors, or partial treatment...
Ok even better: I suggest dequeuing elements one by one as they are processed.

apps/obsidian/src/utils/fileChangeListener.ts
```diff
@@ -0,0 +1,315 @@
...
+  private handleFileCreate(file: TAbstractFile): void {
+    if (!(file instanceof TFile)) {
+      return;
+    }
+
+    if (!file.path.endsWith(".md")) {
+      return;
+    }
+
+    this.pendingCreates.add(file.path);
```
maparent, line 116: trying to understand why you add non-DNode files to pendingCreates. Is it because they might become DNodes? But then that would be handled in the handleMetadataChange, right?
I may be missing something but this does not feel needed.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -28,6 +28,95 @@ export type ObsidianDiscourseNodeData = {
...
+    if (error) {
+      console.error(
+        "Failed to get discourse node content from Supabase:",
+        error,
+      );
+      return [];
+    }
+
+    return (
+      data
```
maparent, line 58: You're getting both variants here (which I think is right) so you may want to pipe it through a set or deduplicate somehow.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -28,6 +28,95 @@ export type ObsidianDiscourseNodeData = {
...
+    if (contentDeleteError) {
+      console.error(
+        "Failed to delete content from Supabase:",
+        contentDeleteError,
+      );
+    }
+  } catch (error) {
+    console.error("Error in deleteNodesFromSupabase:", error);
+  }
+};
```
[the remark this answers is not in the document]
maparent: agreed this is worth looking at.
Also: Delete Documents?

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -66,6 +155,41 @@ type DiscourseNodeInVault = {
...
+  const merged = new Set<ChangeType>([...base, ...additional]);
+  return Array.from(merged);
+};
+
+const buildChangeTypesByPath = (
+  fileChanges: DiscourseNodeFileChange[],
+): Map<string, ChangeType[]> => {
+  const changeTypesByPath = new Map<string, ChangeType[]>();
+
+  for (const change of fileChanges) {
```
maparent: This looks like maybe it could be written more tersely using Array.reduce?
But I think it would be even simpler and clearer to do it at the moment of addition, in the fileChangeListeners.

maparent (changes requested):
I like the queue and debounce! But I can think of edge cases.
I think that handling changes in metadata especially may require more thinking than what I wrote in the comments.

apps/obsidian/src/utils/fileChangeListener.ts
trangdoan982: resolved during sync meeting: newly created node might not have nodeTypeId yet -> miss being in the queue. but agreed that in the future handleMetadataChangess will handle this case

apps/obsidian/src/utils/fileChangeListener.ts
trangdoan982: it will automatically get deleted in the next onLoad sync because the sync will pick up the orphanedNodes. since we can't have access to old value in metadataCache, so if we want to sync this live we'd need to store an index of all existing nodes and its past metadata, which I think is unnecessary for this case.

apps/obsidian/src/utils/fileChangeListener.ts
```diff
@@ -0,0 +1,310 @@
...
+          );
+          this.queueChange(file.path, "title", oldPath);
+        }
+      }
+      return;
+    }
+
+    console.debug(`File renamed: ${oldPath} -> ${file.path}`);
+    this.queueChange(file.path, "title", oldPath);
+  }
```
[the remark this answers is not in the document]
maparent: @trangdoan982 this seems to be unsolved?

apps/obsidian/src/utils/fileChangeListener.ts
maparent: I thought we resolved it, but not sure anymore. It is handled by handleMetadataChanges now, afaik. (In the future: of that file's lifecycle, not in a future version of the code, right?)
So I still don't see why pendingCreates exists.

maparent (changes requested):
There is one pending coderabbit issue, and I may be missing something but I still don't see why we need pendingCreates.

trangdoan982:
@maparent my bad i missed that one comment somehow. just addressed it. the pendingCreates is because when file is first created, the event is triggered, but to update the nodeTypeId to turn it into a Discourse node is an async function, so the pendingCreates make sure we actually process the files when it has the nodeTypeId in the frontmatter, not just when files are created.
but we also agreed that it should be eventually moved to handleMetadataChange, but pushing back this change until we have a clear mechanism to sync relations instances as well.

maparent (approved):
We agreed on why you need `pendingCreates`. Please just add a comment to the effect that, if someone creates a file, quits and relaunches obsidian, and then makes it a discourse node, it won't be caught until the next global sync operation (probably through another obsidian relaunch.)
But I think that's an acceptable limitation.
(The alternative is to hit the db every time we have a metadata change, which is also a bit much.)

PR #711, opened 2026-01-17
ENG-1310 Database change: Allow Concepts to exist without Content

maparent:
https://linear.app/discourse-graphs/issue/ENG-1310/database-change-allow-concepts-to-exist-without-content
Replace `Content.represented_by_id` by `Content.source_local_id`. This allows Concepts to exist without being rooted in a corresponding Content object.
We lose the corresponding foreign key, and we have to use the utility function in postgrest queries instead.
Allows to simplify some unrelated functions, such as orphan deletion.
Opportunistically replaced console.error by internalError in the orphanDeletion file.
Note that I left the represented_by_local_id variable in the input type, so older sync code would still work with the new database function.

https://www.loom.com/share/0892bb7c178b4be6a6e013b6e30d76ef

## Release Notes

* **Bug Fixes**
  * Standardized and improved error reporting across cleanup and deletion flows for clearer failure messages.

* **Refactor**
  * Switched concept identification to a string-based source_local_id and migrated related lookups and upserts to use (space_id + source_local_id).
  * Streamlined concept/content upsert and retrieval paths, reducing ambiguous references and improving data integrity.

### DIFF
--- a/apps/obsidian/src/utils/conceptConversion.ts
+++ b/apps/obsidian/src/utils/conceptConversion.ts
@@ -37,7 +37,7 @@ export const discourseNodeSchemaToLocalConcept = ({
   return {
     space_id: context.spaceId,
     name: node.name,
-    represented_by_local_id: node.id,
+    source_local_id: node.id,
     is_schema: true,
     author_local_id: accountLocalId,
     created: now,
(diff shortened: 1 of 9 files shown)


maparent:
@coderabbit review

maparent:
@coderabbit good call on the missing space_id check, corrected.
Not sure about the index: Do I not get better lookup performance by having the variable with the most variance at the beginning of the index? Would like references on this issue.

maparent:
@coderabbit I still think that indexing on the most diverse key first is more efficient in general, but I decided that going in the direction you suggest would allow me to drop the Concept_space index, and I'll gain enough on index maintenance cost to offset the small performance cost of key ordering.
Please review the latest changes.

mdroidian (approved):

PR #721, opened 2026-01-22
eng-1344 f10b upload obsidian relations and their schemas

maparent:
This uploads relationship schemas to supabase.

https://www.loom.com/share/ce2382d45c6741838fa6cf6b6f7e3d97

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/721" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/obsidian/src/utils/conceptConversion.ts
+++ b/apps/obsidian/src/utils/conceptConversion.ts
@@ -1,11 +1,15 @@
-/* eslint-disable @typescript-eslint/naming-convention */
 import type { TFile } from "obsidian";
-import type { DiscourseNode } from "~/types";
+import type {
+  DiscourseNode,
+  DiscourseRelation,
+  DiscourseRelationType,
+  RelationInstance,
+} from "~/types";
 import type { SupabaseContext } from "./supabaseContext";
+import type { DiscourseNodeInVault } from "./getDiscourseNodes";
 import type { LocalConceptDataInput } from "@repo/database/inputTypes";
 import type { ObsidianDiscourseNodeData } from "./syncDgNodesToSupabase";
 import type { Json } from "@repo/database/dbTypes";
-import DiscourseGraphPlugin from "..";
 
 /**
  * Get extra data (author, timestamps) from file metadata
@@ -14,6 +18,7 @@ const getNodeExtraData = (
   file: TFile,
   accountLocalId: string,
 ): {
+  /* eslint-disable @typescript-eslint/naming-convention */
   author_local_id: string;
   created: string;
   last_modified: string;
@@ -23,6 +28,7 @@ const getNodeExtraData = (
     created: new Date(file.stat.ctime).toISOString(),
     last_modified: new Date(file.stat.mtime).toISOString(),
   };
+  /* eslint-enable @typescript-eslint/naming-convention */
 };
 
 export const discourseNodeSchemaToLocalConcept = ({
@@ -34,8 +40,23 @@ export const discourseNodeSchemaToLocalConcept = ({
   node: DiscourseNode;
   accountLocalId: string;
 }): LocalConceptDataInput => {
-  const { description, template, id, name, created, modified, ...otherData } =
-    node;
+  const {
+    description,
+    template,
+    id,
+    name,
+    created,
+    modified,
+    importedFromRid,
+    ...otherData
+  } = node;
+  /* eslint-disable @typescript-eslint/naming-convention */
+  const literal_content: Record<string, Json> = {
+    label: name,
+    source_data: otherData,
+  };
+  if (template) literal_content.template = template;
+  if (importedFromRid) literal_content.importedFromRid = importedFromRid;
   return {
     space_id: context.spaceId,
     name,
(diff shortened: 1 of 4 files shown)


apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -53,6 +59,78 @@ export const discourseNodeSchemaToLocalConcept = ({
...
+}: {
+  context: SupabaseContext;
+  relationType: DiscourseRelationType;
+  accountLocalId: string;
+}): LocalConceptDataInput => {
+  const { id, label, complement, created, modified, ...otherData } =
+    relationType;
+  return {
+    space_id: context.spaceId,
+    name: label,
```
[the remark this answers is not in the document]
trangdoan982: i think the UX downstream mitigated this (you can't create the same RelationType from setting), so you should be good with this error

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -77,10 +163,58 @@ export const discourseNodeInstanceToLocalConcept = ({
...
+  context: SupabaseContext;
+  relationTypesById: Record<string, DiscourseRelationType>;
+  relationTriplesById: Record<string, DiscourseRelation>;
+  allNodesById: Record<string, DiscourseNodeInVault>;
+  relationData: RelationInstance;
+}): LocalConceptDataInput | null => {
+  const triple = relationTriplesById[relationData.type];
+
+  if (!triple) {
+    console.error("Missing triple id " + relationData.type);
```
trangdoan982: nit: I'd rename this to relationInstanceData for clarity

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -77,10 +163,58 @@ export const discourseNodeInstanceToLocalConcept = ({
...
+    return null;
```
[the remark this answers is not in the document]
trangdoan982: I think this comment is valid and should be addressed

trangdoan982 (changes requested):

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -50,6 +57,84 @@ export const discourseNodeSchemaToLocalConcept = ({
...
+    space_id: context.spaceId,
+    name: `${sourceName} -${label}-> ${destinationName}`,
+    source_local_id: id,
+    is_schema: true,
+    author_local_id: accountLocalId,
+    created: new Date(created).toISOString(),
+    last_modified: new Date(modified).toISOString(),
+    literal_content: {
+      roles: STANDARD_ROLES,
+      label,
```
trangdoan982: i think it's ok to upload these, but I just wonder what for? When we read relationType data upon import we still prefer to look up the discourseRelationType Concepts over this right?

apps/obsidian/src/utils/conceptConversion.ts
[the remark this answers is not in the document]
maparent: Hmmm... it's still a good point. It should be possible one day to distinguish relations of the same name coming from different sources.
What I'll do: I'll include the source uri in the name.
People can rename it later, as long as there's no collision, but it will give an "origin" to foreign schemas.

apps/obsidian/src/utils/conceptConversion.ts
maparent: Correct... it's me preferring to be exhaustive; and more important I think we want to pass them along in the import-export as it's part of the relation definition.

apps/obsidian/src/utils/conceptConversion.ts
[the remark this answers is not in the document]
maparent: As discussed: Yes, we'll keep things that way since we do name alignment.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -177,7 +181,23 @@ const getLastSchemaSyncTime = async (
...
+    .from("Concept")
+    .select("last_modified")
+    .eq("space_id", spaceId)
+    .eq("is_schema", false)
+    .gt("arity", 0)
+    .order("last_modified", { ascending: false })
+    .limit(1)
+    .maybeSingle();
+  return new Date((data?.last_modified || DEFAULT_TIME) + "Z");
+};
```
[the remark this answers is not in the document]
maparent: Arity is a column whose value is computed in the database.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -45,11 +67,109 @@ export const discourseNodeSchemaToLocalConcept = ({
...
+    destinationId,
+    created,
+    modified,
+    importedFromSpaceId,
+  } = relation;
+  const sourceName = nodeTypesById[sourceId]?.name ?? sourceId;
+  const destinationName = nodeTypesById[destinationId]?.name ?? destinationId;
+  const relationType = relationTypesById[relationshipTypeId];
+  if (!relationType)
+    throw new Error(`missing relation type ${relationshipTypeId}`);
```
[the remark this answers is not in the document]
trangdoan982: @maparent ooh i think this is a good flag and should be addressed

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -34,8 +40,23 @@ export const discourseNodeSchemaToLocalConcept = ({
...
+    template,
+    id,
+    name,
+    created,
+    modified,
+    importedFromRid,
+    ...otherData
+  } = node;
+  /* eslint-disable @typescript-eslint/naming-convention */
+  const literal_content: Record<string, Json> = {
```
trangdoan982, line 54: I'm thinking maybe we should have a stronger type system for Concepts.literal_content and Content.metadata. We already use metadata for Wikilink import (we store the full path of the file there). with different types that we upload here (Node, Relation, schemas), it might get messy really soon if we don't enforce the type and still read data from these json columns. It's not blocking for now, we can make this a separate ticket and create the corresponding literal_content type for each Node/Relation Instance-schema separately. or if you prefer to just add the Type in this PR for relation instance and schema that works too

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -66,21 +183,78 @@ export const discourseNodeInstanceToLocalConcept = ({
...
+    name: `[[${sourceNode ? sourceNode.file.basename : source}]] -${relationType.label}-> [[${destinationNode ? destinationNode.file.basename : destination}]]`,
+    source_local_id: relationInstanceData.id,
+    author_local_id: relationInstanceData.author,
+    schema_represented_by_local_id: type,
+    is_schema: false,
+    created: new Date(created).toISOString(),
+    last_modified: new Date(lastModified ?? created).toISOString(),
+    literal_content,
+    local_reference_content: {
+      source,
```
trangdoan982, line 255: relationType should also be here?

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -470,29 +490,102 @@ const convertDgToSupabaseConcepts = async ({
...
       context,
       nodeData: node,
       accountLocalId,
-    }),
-  );
+    });
+  });
+
+  const relationInstancesData = await loadRelations(plugin);
+  const relationsToLocalConcepts = Object.values(
```
trangdoan982: nit: relationInstanceToLocalConcepts for clarity

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -470,29 +490,102 @@ const convertDgToSupabaseConcepts = async ({
...
+        allNodesById,
+        relationInstanceData,
+      }),
+    )
+    .filter((n) => !!n);
 
   const conceptsToUpsert: LocalConceptDataInput[] = [
     ...nodesTypesToLocalConcepts,
+    ...relationTypesToLocalConcepts,
+    ...discourseRelationTriplesToLocalConcepts,
```
trangdoan982, line 607: yeah there IS a bug caused from what Devin mentioned about throwing errors before this mapping.
my local vault: has relationInstances from relations.json. but then I deleted the vault notes, yet the relations.json would still remain -> causing the `source` and `destination` in `local_reference_content` to be unresolvable. -> cause error `{"code":"P0002","details":null,"hint":null,"message":"query returned no rows"}` for `upser_concepts`. because of this nothing gets uploaded, including the schema

I do see this as a common error-prone case. This is the pattern in both Roam and Obsidian that we should address. I can take this on if you don't think it fits in this PR.

trangdoan982 (changes requested):

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -66,21 +183,78 @@ export const discourseNodeInstanceToLocalConcept = ({
...
+  if (!relationType) {
+    console.error("Missing relationType id " + type);
+    return null;
+  }
+  const sourceNode = allNodesById[source];
+  const destinationNode = allNodesById[destination];
+  if (
+    sourceNode?.frontmatter.importedFromRid ||
+    destinationNode?.frontmatter.importedFromRid
+  )
```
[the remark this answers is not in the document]
maparent: This will be addressed in a later pr.

apps/obsidian/src/utils/conceptConversion.ts
maparent: No, it's in `schema_represented_by_local_id`.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
maparent: This one is a real problem, probably unrelated to the devin issue.
I will reject relations to unknown nodes.

apps/obsidian/src/utils/conceptConversion.ts
maparent: This one I want to address later, it requires thought.

trangdoan982: yes non blocking for now

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -470,29 +490,103 @@ const convertDgToSupabaseConcepts = async ({
   const lastSchemaSync = (
```
trangdoan982: ok so on my local test there's a problem with using this lastSchemaSync to fetch relationType to sync:
- all the instances were uploaded but not the relation schemas
- i think the reason is that lastSchema sync is for node type. and since the local relation types were modified before we uploaded the node types, this isn't picked up by the system.
   + this could be the consequence of us shipping the publish nodes before shipping publishing relations in our dev flow. but since Holehouse lab will also use the publish nodes first we should use a different way to get lastSchemaSync separately for node  vs relation schema.

trangdoan982: here's the screenshot of my local test: no relationType or relationTripletSchema is uploaded:

Image

trangdoan982 (changes requested):

trangdoan982 (approved):

PR #722, opened 2026-01-23
[ENG-1295 & ENG-1350 & ENG-1297] F5 import and refresh published nodes

trangdoan982:
https://www.loom.com/share/410fb6cdb4aa4e9ba1a6a4058a69bb19

https://www.loom.com/share/69ae6c6ba43d4856b1adc1ab8ff4061e

https://www.loom.com/share/b02693ddf5cc4a9cba0aaf97986080c3

## Release Notes

* **New Features**
  * Import nodes from other spaces and groups with a guided selection interface
  * Refresh imported node content individually or all at once to sync with source
  * Refresh button added to individual imported files for on-demand updates

* **Improvements**
  * Imported files are now properly excluded from local sync operations

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/722">
  <picture>
    Open with Devin
  </picture>
</a>


trangdoan982:
@coderabbit full review

apps/obsidian/src/index.ts
```diff
@@ -263,10 +263,17 @@ export default class DiscourseGraphPlugin extends Plugin {
     try {
       this.createStyleElement();
 
-      let keysToHide: string[] = [];
+      const keysToHide: string[] = [];
 
       if (!this.settings.showIdsInFrontmatter) {
-        keysToHide.push("nodeTypeId");
+        keysToHide.push(
```
trangdoan982, line 275: [ENG-1350: Hide other frontmatters that are source of truths](https://linear.app/discourse-graphs/issue/ENG-1350/hide-other-frontmatters-that-are-source-of-truths)

apps/obsidian/src/utils/publishNode.ts
```diff
@@ -3,6 +3,58 @@ import type { default as DiscourseGraphPlugin } from "~/index";
 import { getLoggedInClient, getSupabaseContext } from "./supabaseContext";
 import { addFile } from "@repo/database/lib/files";
 import mime from "mime-types";
+import type { DGSupabaseClient } from "@repo/database/lib/client";
+
+const publishSchema = async ({
+  client,
```
trangdoan982, line 9: required for [ENG-1297: F11: Convert node type on import](https://linear.app/discourse-graphs/issue/ENG-1297/f11-convert-node-type-on-import)

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,669 @@
...
+ * and flat form { id, name, color, format, tag }.
+ */
+const parseSchemaLiteralContent = (
+  literalContent: unknown,
+  fallbackName: string,
+): Pick<DiscourseNode, "name" | "format" | "color" | "tag" | "template" | "keyImage"> => {
+  const obj =
+    typeof literalContent === "string"
+      ? (JSON.parse(literalContent) as Record<string, unknown>)
+      : (literalContent as Record<string, unknown>) || {};
```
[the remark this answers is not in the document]
trangdoan982: literalContent is jsonb from database so there shouldn't be error

apps/obsidian/TEST_PLAN_IMPORT.md
trangdoan982, line 1: ignore this file. mostly for Trang's local testing, will be removed after approved

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,742 @@
...
+  }
+
+  // No matching local nodeType: create one from literal_content and add to settings
+  const parsed = parseSchemaLiteralContent(
+    schemaData.literal_content,
+    schemaName,
+  );
+
+  const newNodeType: DiscourseNode = {
+    id: generateUid("node"),
```
trangdoan982: just put the nodeTypeId from the imported node here (parsed.id)

packages/database/src/dbTypes.ts
trangdoan982, line 1: changes in packages/database are from rebasing on top of MAP's branch. shouldn't be reviewed in this PR

apps/obsidian/src/components/ImportNodesModal.tsx
```diff
@@ -0,0 +1,373 @@
...
+          nodeIndex,
+        });
+      }
+    }
+
+    return (
+      <div>
+        <h3 className="mb-4">Select Nodes to Import</h3>
+        <p className="text-muted mb-4 text-sm">
+          {totalNodes} importable node(s) found. Select which nodes to import
```
maparent: Nit: Maybe special-case when there is zero importable node?

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+  return data || [];
+};
+
+export const getPublishedNodesForGroups = async ({
+  client,
+  groupIds,
+  currentSpaceId,
+}: {
+  client: DGSupabaseClient;
+  groupIds: string[];
```
maparent, line 33: Not sure why this is transmitted; we are getting all the ResourceAccess irrespective of group, and do not use the group later.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+  for (const ra of resourceAccessData) {
+    if (ra.source_local_id && ra.space_id) {
+      nodeKeys.add(`${ra.space_id}:${ra.source_local_id}`);
+    }
+  }
+
+  // Group nodes by space_id for batched queries
+  const nodesBySpace = new Map<number, Set<string>>();
+  for (const key of nodeKeys) {
+    const [spaceId, sourceLocalId] = key.split(":");
```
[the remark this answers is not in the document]
maparent: @trangdoan982 valid! Not all source_local_id will come from Obsidian, where we know the pattern.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+    const spaceIdNum = parseInt(spaceId || "0", 10);
+    if (!nodesBySpace.has(spaceIdNum)) {
+      nodesBySpace.set(spaceIdNum, new Set<string>());
+    }
+    nodesBySpace.get(spaceIdNum)!.add(sourceLocalId || "");
+  }
+
+  // Fetch Content with variant "direct" for all nodes, batched by space
+  const nodes: Array<{
```
maparent, line 86: I feel you could check for existing imports' last-modified first, to avoid re-loading the data from the database. (Not sure yet that information is stored reliably imported-side, but let's make sure it is.) BUT it's just the titles, so no real harm done.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+      `Error fetching node content (${variant}):`,
+      error || "No data",
+    );
+    return null;
+  }
+
+  return data.text;
+};
+
+export const fetchNodeMetadata = async ({
```
maparent: Unused as of yet. (I don't mind planning for the future, but curious where you expect to use this?)

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+        });
+
+        if (!fileName) {
+          console.warn(
+            `No direct variant found for node ${node.nodeInstanceId}`,
+          );
+          failedCount++;
+          processedCount++;
+          onProgress?.(processedCount, totalNodes);
+          continue;
```
maparent, line 1184: Optional: If there's an existing file, we could just assume the filename is unchanged and continue. TBH that is unlikely to happen, since we would not have created the file without a filename initially!

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+  }
+
+  let updatedContent = content;
+
+  // Helper to check if a link path matches an old path
+  const matchesOldPath = (linkPath: string, oldPath: string): boolean => {
+    // Exact match
+    if (linkPath === oldPath) return true;
+
+    // Match by filename (handles relative paths)
```
maparent: Ok, if I understand well, we are matching any name irrespective of path?
That scares me. I could see someone with a file structure `experiment <N>/result.jpg`...
Let's discuss and go through the cases here.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+  plugin: DiscourseGraphPlugin;
+  client: DGSupabaseClient;
+  sourceSpaceId: number;
+  rawContent: string;
+  filePath: string;
+}): Promise<{ file: TFile; error?: string }> => {
+  // 1. Create or update the file with the fetched content first
+  let file: TFile | null = plugin.app.vault.getFileByPath(filePath);
+  if (!file) {
+    file = await plugin.app.vault.create(filePath, rawContent);
```
maparent: Let's send the mtime, ctime of the file and use them in create/modify.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+      });
+
+      if (!fileContent) {
+        errors.push(`Failed to download file: ${filepath}`);
+        console.warn(`Failed to download file ${filepath} (hash: ${filehash})`);
+        continue;
+      }
+
+      // Save file to vault
+      await plugin.app.vault.createBinary(targetPath, fileContent);
```
maparent: Here also, we should get the ctime, mtime from the FileReference.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+      const sanitizedExt = ext ? `.${ext}` : "";
+      const sanitizedFileName = `${sanitizedName}${sanitizedExt}`;
+      let targetPath = `${assetsFolderPath}/${sanitizedFileName}`;
+
+      // Check if file already exists at target path (avoid duplicates)
+      if (await plugin.app.vault.adapter.exists(targetPath)) {
+        // File exists at expected path, reuse it
+        const file = plugin.app.vault.getAbstractFileByPath(targetPath);
+        if (file && file instanceof TFile) {
+          pathMapping.set(filepath, targetPath);
```
maparent: Here maybe check that the file wasn't modified since the fileRef's mtime. (Maybe also file length? You'd get that from the storage.info().
If it was, overwrite with db version.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+          // Search all files in the assets folder
+          const allFiles = plugin.app.vault.getFiles();
+          for (const vaultFile of allFiles) {
+            if (
+              vaultFile instanceof TFile &&
+              vaultFile.path.startsWith(assetsFolderPath) &&
+              vaultFile.basename === fileName &&
+              vaultFile.extension === fileExt
+            ) {
+              // Found a file with matching name in assets folder - likely the same file
```
maparent: This may be worth re-checking the mtime, file size and hash, in that order... To be discussed.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+};
+
+type ParsedFrontmatter = {
+  nodeTypeId?: string;
+  nodeInstanceId?: string;
+  publishedToGroups?: string[];
+  [key: string]: unknown;
+};
+
+const parseFrontmatter = (
```
maparent, line 929: Hmmm... should you have to do this at all? Tell me if I'm missing something, but the front matter is already parsed in the Concept's metadata json blob (which I noted you were not using.)
Otherwise... maybe save the file, and ask the Obsidian cache about the frontmatter? less sure about this.
If nothing else works, get an existing yaml parser module. 
Reinventing yaml parsing  is certainly wrong.

maparent (changes requested):
I'm not done, but I hate the idea of you having to wait for all my comments before you can work. I'll keep going.
First: Still very happy with the overall work. Behaviour of happy path is pretty much what we want.
Still looking at edge cases.

Another question I'd like for us to think about: The "imported From" and groupId are using opaque Ids. (The latter is on me!) 
It would be nice to have space name and group name in the frontMatter.
Is there a way to store both in the frontmatter that mostly shows the user-friendly name?
Of course we can store both as a map, or an array of pairs, but neither would look good.
We _might_ store something like <id>:name , which is easy to parse, and at least will include the name. What do you think?
Some improvements we can delay to other PRs (I created a task for node templates) but I think we should not store wrong things in the frontmatter if possible (though I already did that with groups!)
(Making the group names available now.)

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1207 @@
...
+  for (const [oldPath, newPath] of oldPathToNewPath.entries()) {
+    const newFile = app.metadataCache.getFirstLinkpathDest(
+      newPath,
+      targetFile.path,
+    );
+    if (newFile) {
+      oldPathToNewFile.set(oldPath, newFile);
+    }
+  }
+
```
trangdoan982: make sure that match on nodeTypeId before matching on Name

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -52,21 +58,150 @@ export const discourseNodeSchemaToLocalConcept = ({
   };
 };
 
+const STANDARD_ROLES = ["source", "destination"];
+
+export const discourseRelationTypeToLocalConcept = ({
```
maparent: The relation migration code had been removed and isolated in #721 ; please do not re-introduce it in this PR. (May be a merge error?).
ETA: I suspect this file should not be touched at all in this PR?

apps/obsidian/src/types.ts
```diff
@@ -61,4 +61,19 @@ export type BulkImportPattern = {
   enabled: boolean;
 };
 
+export type ImportableNode = {
```
maparent, line 64: This files's change also do not belong in this PR imho

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -11,8 +11,10 @@ import { default as DiscourseGraphPlugin } from "~/index";
 import { upsertNodesToSupabaseAsContentWithEmbeddings } from "./upsertNodesAsContentWithEmbeddings";
 import {
   orderConceptsByDependency,
-  discourseNodeInstanceToLocalConcept,
```
maparent, line 15: also includes a lot that was isolated in #721

maparent (changes requested):

apps/obsidian/src/types.ts
maparent: no that's wrong, sorry

apps/obsidian/src/utils/importNodes.ts
trangdoan982: we do need groupIds to filter the first layer of resourceAccessData (need to filter based on account_uid in groupIds)

apps/obsidian/src/utils/importNodes.ts
trangdoan982: yeah i'd say the complexity of the code increases without much performance gain i'll skip this

apps/obsidian/src/utils/importNodes.ts
maparent: Yes, but it's always based on all the groups you belong to, right? So those that have a group_id you belong to are the only ones you can see anyway, thanks to RLS. So the filter is a no-op. 
Actually, not quite: it makes it harder to make a resource accessible to an individual (vs group.) Not a use case we have in DB yet, tbh.
If we have a good reason to restrict to a subset of the groups you belong to, then that filter would make sense. But I don't see it as useful yet.
Actually... makes me realize we may not need to look at ResourceAccess at all! Just look for any resource (Document/Content/Concept), filtering on not current space as you do now, and count on the RLS to give you what you need, no more or less.
ETA: It goes a bit faster if you use the views (my_document, my_content, my_concept) instead, the RLS is then computed globally.

apps/obsidian/src/utils/importNodes.ts
maparent: Fair. Happy to move it to a future optimization task.

apps/obsidian/src/utils/importNodes.ts
trangdoan982: wait i thought createAt/modifedAt are only for actions from the owner's vault. What information about imported file do we need to update here?

maparent: We would want the imported files to have the same mtime/ctime as the original files. It makes it easier for users to know what has been updated when, in a Finder view. (Order by modified, etc.) 
Also, it would allow us to present that information to the user in the discourse context panel.
(We could put it in frontmatter, but that is wasteful when we have fine filesystem metadata at our disposal.)

maparent: (You may remember that option was mentioned during the Holehouse meeting, and they seemed to like that idea.)

maparent: Er... you put it in the frontmatter? I said I thought that was wasteful.
What I would want here is:
`plugin.app.vault.create(filePath, rawContent, {ctime: importedCreatedAt, mtime: importedModifedAt});`
assuming the other two were kept as timestamp numbers along the call chain and not stringified dates; you could pass them along as such. I think stringified dates is always a lossy option; Date object could be acceptable, but since everything here expects timestamp numbers let's skip to that.
This applies to all usages of `plugin.app.vault.create`, `plugin.app.vault.modify`, `plugin.app.vault.createBinary` in this file.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -37,7 +37,7 @@ export const discourseNodeSchemaToLocalConcept = ({
     node;
   return {
     space_id: context.spaceId,
-    name: name,
+    name: id ? `${name} (${id})` : name,
```
maparent: Ouch, this is scary, but I see where you're coming from.
OT1H, the risk of name collision is real as we're importing.
OTOH, this name is supposed to be user-facing eventually, and should not have IDs. 
Maybe... vault name in parenthesis?

maparent (changes requested):

apps/obsidian/src/utils/conceptConversion.ts
trangdoan982: agree to use vaultName

apps/obsidian/src/utils/importNodes.ts
trangdoan982: TODO: add the ctime mtime to the DiscourseContextView

apps/obsidian/src/utils/importNodes.ts
trangdoan982: i decided to not cover this case (where user move the asset files away) and just recreate the file instead. it's too risky/ involve too much assumptions about file name and not worth the data being saved

apps/obsidian/src/components/DiscourseContextView.tsx
```diff
@@ -45,6 +72,16 @@ const DiscourseContext = ({ activeFile }: DiscourseContextProps) => {
...
     }
+
+    const isImported = !!frontmatter.importedFromSpaceUri;
+    const sourceDates =
+      isImported && activeFile?.stat
+        ? {
+            createdAt: new Date(activeFile.stat.ctime).toLocaleString(),
+            modifiedAt: new Date(activeFile.stat.mtime).toLocaleString(),
+          }
+        : null;
```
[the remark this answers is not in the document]
trangdoan982: because we set the fileStat to correspond to the source fileStat, this is expected

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -0,0 +1,1386 @@
...
+          refLastModifiedMs > 0 && existingFile.stat.mtime > refLastModifiedMs;
+        if (!localModifiedAfterRef) {
+          setPathMapping(filepath, existingFile.path);
+          console.log(
+            `Reusing existing asset: ${filehash} -> ${existingFile.path}`,
+          );
+          continue;
+        }
+        overwritePath = existingFile.path;
+      }
```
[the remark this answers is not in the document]
trangdoan982: this is expected behavior

apps/obsidian/src/components/ImportNodesModal.tsx
```diff
@@ -0,0 +1,379 @@
...
+        groupIds,
+        currentSpaceId: context.spaceId,
+      });
+
+      const localNodeInstanceIds = getLocalNodeInstanceIds(plugin);
+
+      // Filter out nodes that already exist locally
+      const importableNodes = publishedNodes.filter(
+        (node) => !localNodeInstanceIds.has(node.source_local_id),
+      );
```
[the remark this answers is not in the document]
maparent: I think we can let that go; if I understand well it's about avoiding source_local_id collision. We'll revisit that when we distinguish local id and imported id.

maparent (approved):

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -290,6 +290,48 @@ export class QueryEngine {
...
+   * Returns the file if found, null otherwise
+   */
+  findExistingImportedFile = (
+    nodeInstanceId: string,
+    importedFromSpaceUri: string,
+  ): TFile | null => {
+    if (this.dc) {
+      try {
+        const safeId = nodeInstanceId.replace(/"/g, '\\"');
+       const dcQuery = `@page and nodeInstanceId = "${safeId}" and importedFromSpaceUri = "${importedFromSpaceUri.replace(/"/g, '\\"')}"`;
```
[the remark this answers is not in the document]
maparent: This is out of scope but feels real, we should track elsewhere.

maparent (approved):
Have a quick look at https://github.com/DiscourseGraphs/discourse-graph/pull/722#discussion_r2790502584
I think it's a different scope, but I could be wrong.
Take a decision, I approve either way.

apps/obsidian/src/components/ImportNodesModal.tsx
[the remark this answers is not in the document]
trangdoan982: yeah i think because source_local_id is uuvid7() so i think there's low to no chance of collision anyway.

PR #742, opened 2026-01-31
ENG-1331 Discourse relation panel does not reliably update when relations are created or deleted

maparent:
https://linear.app/discourse-graphs/issue/ENG-1331/discourse-relation-panel-does-not-reliably-update-when-relations-are

First, when the ignoreCache is sent to calculations, don't just ignore the cache but clear it. Otherwise another query can come in and  use the invalid cache.

Second, there were a few refresh functions that did not clear the cache.

https://www.loom.com/share/30db4346f43648ff8a3f18a5570b97fa

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/742">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/DiscourseContext.tsx
+++ b/apps/roam/src/components/DiscourseContext.tsx
@@ -280,7 +280,7 @@ const ContextTab = ({
             <CreateRelationButton
               sourceNodeUid={parentUid}
               onClose={() => {
-                window.setTimeout(onRefresh, 450, true);
+                window.setTimeout(onRefresh, 150, true);
               }}
             />
             <Switch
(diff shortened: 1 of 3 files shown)


mdroidian (approved):
This suggests a deeper rooted problem with an unstable prop or effect loop.  Based on the deep changes we have been making, it might be time to look at a refactor.  Also, the the flicker/re render after adding/deleting relations isn't ideal and likely related.

Just noting those observations here, considered out of scope.  Added to the backlog: [ENG-1380: Refactor the discourse context component.](https://linear.app/discourse-graphs/issue/ENG-1380/refactor-the-discourse-context-component)

PR #748, opened 2026-02-05
[ENG-1391] Migrate relations into vault relations.json as objects

trangdoan982:
https://www.loom.com/share/a87c77fe07034cce9daf5fd4b55ab2e3

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/748" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Release Notes

* **New Features**
  * Introduced a dedicated relations store for persistent relationship management, replacing frontmatter-based storage.
  * Automatic migration of existing frontmatter relations to the new relations store on plugin initialization.
  * Enhanced relationship tracking with versioning for improved consistency.

* **Refactor**
  * Restructured relationship management architecture to use a centralized JSON store instead of frontmatter links.


trangdoan982:
@coderabbitai full review

apps/obsidian/src/utils/relationsStore.ts
```diff
@@ -0,0 +1,380 @@
+import { normalizePath, TFile } from "obsidian";
+import { uuidv7 } from "uuidv7";
+import type DiscourseGraphPlugin from "~/index";
+import { ensureNodeInstanceId } from "~/utils/nodeInstanceId";
+import { checkAndCreateFolder } from "~/utils/file";
+
+const RELATIONS_DIR = "_discourse_graphs";
```
maparent: Curious: Why not reuse `nodesFolderPath`? Is the point to hide it from the user? (I would see that as a negative myself.)
-1 on creating a new directory, and -10 on creating a non-configurable new directory.
(But also not keen on adding to configuration, so... that amounts to reusing what's there.)

apps/obsidian/src/utils/relationsStore.ts
```diff
@@ -0,0 +1,380 @@
...
+const RELATIONS_FILE_NAME = "relations.json";
+const RELATIONS_FILE_VERSION = 1;
+const DEFAULT_AUTHOR = "local";
```
maparent: nit: maybe extra punctuation to make it less likely to collide with a random identifier from outside

apps/obsidian/src/utils/relationsStore.ts
```diff
@@ -0,0 +1,380 @@
...
+}
+
+export const getNodeInstanceIdForFile = async (
+  plugin: DiscourseGraphPlugin,
+  file: TFile,
+): Promise<string | null> => {
+  const cache = plugin.app.metadataCache.getFileCache(file);
+  const frontmatter = cache?.frontmatter;
+  if (!frontmatter?.nodeTypeId) {
+    return null;
```
maparent, line 186: Why return null? Should we not ensure in all cases?

apps/obsidian/src/utils/relationsStore.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  const removeFromFile = async (
+    f: TFile,
+    linkToRemove: string,
+  ): Promise<void> => {
+    await plugin.app.fileManager.processFrontMatter(f, (fm) => {
+      const raw = fm[relationTypeId] as unknown;
+      if (!raw) return;
+      const links = Array.isArray(raw) ? (raw as string[]) : [raw as string];
+      const filtered = links.filter((link) => String(link).trim() !== linkToRemove.trim());
```
maparent: This failed for me; the one link I have in my vault has the `.md` extension in the frontmatter link, i.e. `[[filename.md]]` as a value, so the filter misses it. 
Not sure how that value got in, but I doubt it's manual. Do you have an idea? Would we need to handle that case?

apps/obsidian/src/utils/relationsStore.ts
```diff
@@ -0,0 +1,380 @@
...
+  const pendingCleanups: Array<{
+    file: TFile;
+    targetFile: TFile;
+    relationTypeId: string;
+  }> = [];
+
+  for (const file of markdownFiles) {
+    const cache = plugin.app.metadataCache.getFileCache(file);
+    const frontmatter = cache?.frontmatter;
+    if (!frontmatter?.nodeTypeId) continue;
```
maparent, line 351: Maybe ensure anyway?

And another thought: If we already have relations, maybe only migrate node files that changed since the last-modify of that relation set? We could gain time at startup.

apps/obsidian/src/utils/relationsStore.ts
```diff
@@ -0,0 +1,380 @@
...
+export type AddRelationParams = {
+  type: string;
+  source: string;
+  destination: string;
+  author?: string;
+  importedFromSpaceId?: number;
+  publishedToGroupId?: string[];
+};
+
+export const addRelation = async (
```
maparent: I see you created a utility (for canvas) that checks for existence before creating. Should not that be the default, and available here? If there are points where it makes sense _not_ to verify for existence, maybe add a noCheck parameter, or make a separate function with an ominous name (addRelationNoCheck).

maparent (changes requested):
Some changes, but overall we'll be good to go.

apps/obsidian/src/utils/relationsStore.ts
trangdoan982: oh i thought of the underscore as something that signals a system folder. But yes for simplicity we can just use nodesFolderPath

apps/obsidian/src/utils/relationsStore.ts
trangdoan982: if it's null then we shouldn't guess the nodeTypeId here either. with nodeInstance we can ensure but I don't think we can for nodeTypeId. But all the flows that lead to this point should have nodeTypeId (either from the DiscourseContextView or from Canvas).

apps/obsidian/src/utils/relationsStore.ts
trangdoan982: same response as above i don't think we should guess nodeTypeId

apps/obsidian/src/utils/relationsStore.ts
maparent: my bad, I thought this was the instanceId! Sorry!

maparent (approved):

PR #787, opened 2026-02-16
ENG-1234 Show and hide nodes in the clipboard that are already on canvas

mdroidian:
https://www.loom.com/share/31d20f6525dc437ab769e74fe90ce705

## Summary
- add a clipboard-level "Show nodes on canvas" toggle, persisted per canvas and user in clipboard block props
- hide nodes already present on the active canvas when the toggle is off, with an explicit empty state
- optimize canvas membership checks by indexing shapes by uid for list rendering
- add a drop-time guard to prevent adding duplicate nodes while the toggle is off
- add clipboard options menu trigger to match the mock and place switch indicator on the right

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/787" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/canvas/Clipboard.tsx
+++ b/apps/roam/src/components/canvas/Clipboard.tsx
@@ -16,8 +16,11 @@ import {
   Icon,
   Intent,
   NonIdealState,
+  Popover,
+  Position,
   Spinner,
   SpinnerSize,
+  Switch,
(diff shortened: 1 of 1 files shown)


mdroidian:
@coderabbitai full review

PR #788, opened 2026-02-16
ENG-1378 automatic source appending for evd incorrectly pre pends

mdroidian:
https://www.loom.com/share/12276fa951e14a1d824e8da9e5f6ab7d

## Summary

  Fixes the EVD create flow so a prefilled Source is visibly locked and treated as an existing node, instead of being silently converted  into a “new source” input path.

  ## Problem

  In the create-node modal, the referenced Source could appear prefilled but still behave as unlocked. In that state, the input update path  could clear the referenced uid, which then caused submit logic to treat it as a new source and re-run source formatting.

  ## Root Cause

  FuzzySelectInput kept a separate local isLocked state that could drift from the actual source-of-truth (value.uid), and lock behavior  relied on prop-sync logic.

  ## What Changed

  - Refactored lock behavior to be derived, not stored:
      - isLocked = mode === "create" && Boolean(value?.uid)
      - File: apps/roam/src/components/FuzzySelectInput.tsx
  - Removed prop-sync lock mechanism and related props:
      - Removed initialIsLocked
      - Removed onLockedChange
      - Removed the useEffect that copied prop state into local lock state
      - File: apps/roam/src/components/FuzzySelectInput.tsx
  - Updated selection/clear behavior to drive lock purely via value.uid:
      - Selecting existing item (uid present) locks automatically
      - Clearing resets uid and unlocks automatically
      - File: apps/roam/src/components/FuzzySelectInput.tsx
  - Removed obsolete prop usage in dialog:
      - File: apps/roam/src/components/ModifyNodeDialog.tsx

  ## Why This Approach

  This removes duplicated state and makes lock behavior deterministic from a single source of truth. It is simpler, more React-idiomatic, and avoids stale lock transitions.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/788" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal state management in the fuzzy select component, reducing dependencies and redundant props while preserving all user-facing functionality and UI behavior.

### DIFF
--- a/apps/roam/src/components/FuzzySelectInput.tsx
+++ b/apps/roam/src/components/FuzzySelectInput.tsx
@@ -24,7 +24,7 @@ type FuzzySelectInputProps<T extends Result = Result> = {
   options?: T[];
   placeholder?: string;
   autoFocus?: boolean;
-  isLocked?: boolean;
+  isLocked: boolean;
 };
 
 const FuzzySelectInput = <T extends Result = Result>({
@@ -34,13 +34,12 @@ const FuzzySelectInput = <T extends Result = Result>({
   options = [],
   placeholder = "Enter value",
   autoFocus,
-  isLocked: isLockedProp,
+  isLocked,
 }: FuzzySelectInputProps<T>) => {
   const [query, setQuery] = useState<string>(() => value?.text || "");
   const [isOpen, setIsOpen] = useState(false);
   const [activeIndex, setActiveIndex] = useState(0);
   const [isFocused, setIsFocused] = useState(false);
-  const isLocked = isLockedProp ?? (mode === "create" && Boolean(value?.uid));
 
   const menuRef = useRef<HTMLUListElement>(null);
   const inputRef = useRef<HTMLInputElement>(null);
--- a/apps/roam/src/components/FuzzySelectInput.tsx
+++ b/apps/roam/src/components/FuzzySelectInput.tsx
@@ -24,6 +24,7 @@ type FuzzySelectInputProps<T extends Result = Result> = {
   options?: T[];
   placeholder?: string;
   autoFocus?: boolean;
+  isLocked?: boolean;
 };
 
 const FuzzySelectInput = <T extends Result = Result>({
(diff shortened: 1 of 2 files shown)


mdroidian:
@coderabbitai full review

mdroidian:
https://github.com/DiscourseGraphs/discourse-graph/pull/788/commits/c6b53e0e7b72f25e747c1eccec7c010226fffbc0 quick fixed:
https://www.loom.com/share/128efc61f58249d8bf61eca3264ec127

keyboard issue to be addressed here: [ENG-1316: When selecting node using enter, shouldn't focus on node select dropdown](https://linear.app/discourse-graphs/issue/ENG-1316/when-selecting-node-using-enter-shouldnt-focus-on-node-select-dropdown)

apps/roam/src/components/FuzzySelectInput.tsx
```diff
@@ -52,28 +52,31 @@ const FuzzySelectInput = <T extends Result = Result>({
 
   const handleSelect = useCallback(
     (item: T) => {
-      if (mode === "create" && item.uid && item.uid !== initialUid) {
```
trangdoan982, line 55: Oh I think we need another variable to detect if we're creating nodes in canvas.
Removing this and asserting `const isLocked = mode === "create" && Boolean(value?.uid);` makes it so that every new node creation in Canvas will have isLocked=true by default (because block uid is created first before the flow is finished)

Image

trangdoan982 (changes requested):

apps/roam/src/components/FuzzySelectInput.tsx
mdroidian: Nice catch.  Re-added `isLocked` and made it required.
https://www.loom.com/share/0403ecba18dd478995bfb9310a0a6aea

trangdoan982 (approved):
lgtm

PR #806, opened 2026-02-20
[ENG-1316] Fix focus when select node with keyboard flow

trangdoan982:
https://www.loom.com/share/e59fb5d8593e466aa07886daa193b4a0

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/806" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/FuzzySelectInput.tsx
+++ b/apps/roam/src/components/FuzzySelectInput.tsx
@@ -85,6 +85,19 @@ const FuzzySelectInput = <T extends Result = Result>({
         e.preventDefault();
         e.stopPropagation();
         if (isOpen && filteredItems[activeIndex]) {
+          // Suppress the upcoming keyup so it doesn't activate the next focused element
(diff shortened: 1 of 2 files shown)


mdroidian (changes requested):
I'd love to see if we could achieve this result without an additional `useEffect` or `keyup` listener.  Let's try to use the built in features of the components first and `useEffect` as a last resort.

Did you try setting `autoFocus` false on the `<Label>` containing the dropdown and the confirm `<Button>`?  Specifically, managing the value of `autoFocus` depending on the state, eg:
- if no content, focus on content
- if no source, focus on source
- if content+source, focus on confirm

Again, via `autoFocus`


It looks like there is a merge conflict now from https://github.com/DiscourseGraphs/discourse-graph/pull/788
Could you give this a try after fixing the merge conflict, with a short timebox (30 min).

PR #808, opened 2026-02-21
ENG-1461 Use full Rid instead of just spaceUri as marker of import

maparent:
https://linear.app/discourse-graphs/issue/ENG-1461/use-full-rid-instead-of-just-spaceuri-as-marker-of-import

This PR migrates imported node frontmatter from using `importedFromSpaceUri` to using `importedFromRid`, which contains both the original node's spaceUri and the original sourceLocalId, in the form of a koi-compliant Rid. 
(See [Rid specs](https://github.com/BlockScience/rid-lib)

Previously imported nodes are migrated accordingly.

Also, the sourceRid is added to the schema types, so we have provenance for schemas.

https://www.loom.com/share/b5dc1662b79e4b56a7ff0e3bf74b7039

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/808" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/obsidian/src/utils/rid.ts
+++ b/apps/obsidian/src/utils/rid.ts
@@ -15,9 +15,13 @@ export const spaceUriAndLocalIdToRid = (
 export const ridToSpaceUriAndLocalId = (
   rid: string,
 ): { spaceUri: string; sourceLocalId: string } => {
-  const m = rid.match(/^orn:(\w+).(\w+):(.*)\/([^/]+)$/);
+  const m = rid.match(/^orn:(\w+)\.(\w+):(.*)\/([^/]+)$/);
   if (m) {
     return { spaceUri: `${m[1]}:${m[3]}`, sourceLocalId: m[4]! };
+  }
+  const m2 = rid.match(/^orn:(\w+):(.*)\/([^/]+)$/);
+  if (m2) {
+    return { spaceUri: `${m2[1]}:${m2[2]}`, sourceLocalId: m2[3]! };
   }
   const parts = rid.split("/");
   const sourceLocalId = parts.pop()!;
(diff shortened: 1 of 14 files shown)


apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -668,7 +621,7 @@ const collectDiscourseNodesFromPaths = async (
       continue;
     }
 
-    if (frontmatter.importedFromSpaceUri) {
+    if (frontmatter.importedFromRid) {
       console.debug(`Skipping imported file: ${filePath}`);
       continue;
     }
```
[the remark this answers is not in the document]
maparent: We can assume the migration has happened here.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -828,6 +781,35 @@ export const cleanupOrphanedNodes = async (
   }
 };
 
+const migrateImportedFromFrontMatter = async (plugin: DiscourseGraphPlugin) => {
```
trangdoan982: i don't think we need this function. because the first "actual" use case will be from Holehouse which has no initial data. This is only if you care about our testing data being migration but for production I don't think this is necessary.

apps/obsidian/src/utils/rid.ts
```diff
@@ -0,0 +1,29 @@
+export const spaceUriAndLocalIdToRid = (
```
trangdoan982, line 8: nit: maybe here is a good place to document the structure/ link to the RID structure https://github.com/BlockScience/rid-lib

trangdoan982 (approved):
beside the unnecessary migration i think this is good to go

PR #817, opened 2026-02-23
ENG-1475 Export relations to imported nodes

maparent:
https://linear.app/discourse-graphs/issue/ENG-1475/export-relations-to-imported-nodes

https://www.loom.com/share/527b312dca4f441083f9ae25987e5f57

Code note: the main work was to send the original node's koi identifier in upsertConcept, and have that function resolve the identifier to the original node.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/817" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/obsidian/src/utils/publishNode.ts
+++ b/apps/obsidian/src/utils/publishNode.ts
@@ -198,7 +198,7 @@ export const publishNodeRelations = async ({
       resourceIds.add(triple.id);
     }
   });
-  if (!resourceIds) return;
+  if (resourceIds.size === 0) return;
   const publishResponse = await client.from("ResourceAccess").upsert(
     [...resourceIds.values()].map((sourceLocalId: string) => ({
       /* eslint-disable @typescript-eslint/naming-convention */
--- a/apps/obsidian/src/utils/publishNode.ts
+++ b/apps/obsidian/src/utils/publishNode.ts
@@ -182,7 +182,6 @@ export const publishNodeRelations = async ({
         (fm.publishedToGroups as string[]) || [];
       if (!publishedToGroups.includes(myGroup)) return;
     }
-    if (fm.importedFromRid) return; // temporary, should be removed after eng-1475
     relevantNodeTypeById[id] = fm.nodeTypeId as string;
   });
   relations.map((relation) => {
@@ -199,6 +198,7 @@ export const publishNodeRelations = async ({
       resourceIds.add(triple.id);
     }
   });
+  if (!resourceIds) return;
   const publishResponse = await client.from("ResourceAccess").upsert(
     [...resourceIds.values()].map((sourceLocalId: string) => ({
       /* eslint-disable @typescript-eslint/naming-convention */
--- a/apps/obsidian/src/utils/syncDgNodesToSupabase.ts
+++ b/apps/obsidian/src/utils/syncDgNodesToSupabase.ts
@@ -365,6 +365,7 @@ const buildChangedNodesFromNodes = async ({
   const changedNodes: ObsidianDiscourseNodeData[] = [];
 
   for (const node of nodes) {
+    if (node.frontmatter.importedFromRid) continue;
     const existingTitle = existingTitleMap.get(node.nodeInstanceId);
     const detectedChangeTypes = detectNodeChanges(
       node,
@@ -433,7 +434,6 @@ export const syncAllNodesAndRelations = async (
           context,
         });
 
-    console.log("changedNodeInstances", changedNodeInstances);
     console.debug(`Found ${changedNodeInstances.length} nodes to sync`);
 
     const accountLocalId = plugin.settings.accountLocalId;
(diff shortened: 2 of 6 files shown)


apps/obsidian/src/utils/publishNode.ts
```diff
@@ -182,23 +182,26 @@ export const publishNodeRelations = async ({
...
+          (triple.sourceId === relevantNodeTypeById[relation.destination] &&
+            triple.destinationId === relevantNodeTypeById[relation.source])),
     );
     if (triple) {
       resourceIds.add(relation.id);
       resourceIds.add(relation.type);
       resourceIds.add(triple.id);
     }
   });
+  if (!resourceIds) return;
```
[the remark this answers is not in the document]
trangdoan982: valid comment

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -423,7 +423,7 @@ export const syncAllNodesAndRelations = async (
     }
     console.debug("Supabase client:", supabaseClient);
 
-    const allNodes = await collectDiscourseNodesFromVault(plugin);
+    const allNodes = await collectDiscourseNodesFromVault(plugin, true);
 
     const changedNodeInstances = relationsOnly
       ? []
```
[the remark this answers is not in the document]
trangdoan982: agree with this bug. we shouldn't be syncing imported node to either Concept or Content table. should do filtering so that only relations of imported nodes are considered

trangdoan982 (changes requested):

packages/database/supabase/schemas/concept.sql
```diff
@@ -240,6 +240,49 @@ $$;
...
+END IF;
+SELECT id INTO source_id FROM public."Space" where url=uri;
+IF source_id IS NULL THEN
+    RETURN (null, concat('Cannot find ', uri))::public.accessible_resource;
+END IF;
+RETURN (source_id, source_local_id);
+END;
+$$;
+
+CREATE OR REPLACE FUNCTION public.rid_or_local_id_to_concept_db_id(rid VARCHAR, default_space_id BIGINT)
```
trangdoan982, line 272: i accidentally found a case where this might fail:
- if we clean up the orphaned node, then the stale relation instance won't be able to resolve to any db id -> cause this function to return null -> cause `concept.reference_content` -> violate non-null constraints -> unable to upsert ANY concepts.

i discovered this bug after having ran `supbase db reset`, keeping the vault the same. this means that locally there will be old imported files with frontmatter field `importFromRid`, that we can't resolve in db. Hence, i received this error:
VM209 plugin:@discou…raph/obsidian:55074 Initial sync failed: Error: upsert_concepts failed: {
  "code": "23502",
  "details": null,
  "hint": null,
  "message": "null value in column \"reference_content\" of relation \"Concept\" violates not-null constraint"
}


this shouldn't be blocking any syncing operation

trangdoan982 (changes requested):

packages/database/supabase/schemas/concept.sql
maparent: Ok, this is a bit of a headache. I corrected the function to allow partial failures; but it means we cannot use the lastSync time anymore!
I think... we need to look at error values, i.e. negative Ids returned by upsertConcept, and possibly remove the offending node/relations. Let's discuss 1-1.

maparent: Leaving a trail: we will think about error handling in another PR (ENG-1510.)
The issue of lastSync time may not be as pressing, since the nodes with an error probably cannot be inserted anyway.

packages/database/supabase/migrations/20260221193625_rid_functions.sql
```diff
@@ -0,0 +1,160 @@
...
+            name = db_concept.name,
+            description = db_concept.description,
+            author_id = db_concept.author_id,
+            created = db_concept.created,
+            last_modified = db_concept.last_modified,
+            schema_id = db_concept.schema_id,
+            literal_content = db_concept.literal_content,
+            is_schema = db_concept.is_schema,
+            reference_content = db_concept.reference_content
+        -- ON CONFLICT (space_id, name) DO NOTHING... why can't I specify two conflict clauses?
```
trangdoan982: =))) should this comment be left out?

packages/database/supabase/migrations/20260221193625_rid_functions.sql
```diff
@@ -0,0 +1,160 @@
+CREATE OR REPLACE FUNCTION public.rid_to_space_id_and_local_id(rid VARCHAR)
```
trangdoan982, line 1: why do we have 2 rid_to_space_id_and_local_id() in this migration and concept.sql as well?

trangdoan982 (approved):

packages/database/supabase/migrations/20260221193625_rid_functions.sql
maparent: it's ranting, but it's also signifying that this is a conflict that could happen and I need to handle it. I guess I could make that clearer...

packages/database/supabase/migrations/20260221193625_rid_functions.sql
maparent: Everything is always in both: one for from-scratch, one for from-last-state.

PR #827, opened 2026-02-25
ENG-1300 Publish relations when a node is published

maparent:
https://linear.app/discourse-graphs/issue/ENG-1300/publish-relations-when-a-node-is-published

https://www.loom.com/share/b7695fd01efb4e649848e4e362d17e91

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/827" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/obsidian/src/utils/conceptConversion.ts
+++ b/apps/obsidian/src/utils/conceptConversion.ts
@@ -288,6 +288,8 @@ const orderConceptsRec = ({
   remainder: { [key: string]: LocalConceptDataInput };
   processed: Set<string>;
 }): Set<string> => {
+  // Add to processed at the start to prevent cycles
+  processed.add(concept.source_local_id!);
   const relatedConceptIds = relatedConcepts(concept);
   let missing: Set<string> = new Set();
   while (relatedConceptIds.length > 0) {
@@ -310,7 +312,6 @@ const orderConceptsRec = ({
     }
   }
   ordered.push(concept);
-  processed.add(concept.source_local_id!);
   delete remainder[concept.source_local_id!];
   return missing;
 };
--- a/apps/obsidian/src/utils/importNodes.ts
+++ b/apps/obsidian/src/utils/importNodes.ts
@@ -11,7 +11,7 @@ import { spaceUriAndLocalIdToRid, ridToSpaceUriAndLocalId } from "./rid";
 
 export const getAvailableGroups = async (
   client: DGSupabaseClient,
-): Promise<{ group_id: string }[]> => {
+): Promise<string[]> => {
   const { data, error } = await client
     .from("group_membership")
     .select("group_id")
@@ -22,7 +22,7 @@ export const getAvailableGroups = async (
     throw new Error(`Failed to fetch groups: ${error.message}`);
   }
 
-  return data || [];
+  return (data || []).map((g) => g.group_id);
 };
 
 export const getPublishedNodesForGroups = async ({
--- a/apps/obsidian/src/utils/importNodes.ts
+++ b/apps/obsidian/src/utils/importNodes.ts
@@ -9,7 +9,7 @@ import type { DiscourseNode, ImportableNode } from "~/types";
 import { QueryEngine } from "~/services/QueryEngine";
 import { spaceUriAndLocalIdToRid, ridToSpaceUriAndLocalId } from "./rid";
 
-export const getAvailableGroups = async (
+export const getAvailableGroupIds = async (
   client: DGSupabaseClient,
 ): Promise<string[]> => {
   const { data, error } = await client
--- a/apps/obsidian/src/utils/publishNode.ts
+++ b/apps/obsidian/src/utils/publishNode.ts
@@ -79,7 +79,7 @@ const intersection = <T>(set1: Set<T>, set2: Set<T>): Set<T> => {
 export const publishNewRelation = async (
   plugin: DiscourseGraphPlugin,
   relation: RelationInstance,
-) => {
+): Promise<boolean> => {
   const client = await getLoggedInClient(plugin);
   if (!client) throw new Error("Cannot get client");
   const context = await getSupabaseContext(plugin);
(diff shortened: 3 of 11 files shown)


apps/obsidian/src/utils/relationsStore.ts
```diff
@@ -243,6 +243,23 @@ export const getFileForNodeInstanceId = async (
   return null;
 };
 
+export const getFileForNodeInstanceIds = (
```
trangdoan982, line 262: we might take advantage of datacore here. maybe develop this function in QueryEngine with this entire vault scan as fallback, but for optimization let's use datacore cache first to query files based on nodeInstanceIds

apps/obsidian/src/utils/publishNode.ts
```diff
@@ -56,6 +62,88 @@ const publishSchema = async ({
   }
 };
 
+export const publishNodeRelations = async ({
```
trangdoan982, line 151: hmm i think the biggest gap we have here is to only call publishNodeRelations when a node is published. i could see a case where:
- user published 2 nodes
- another user imported these 2 nodes
- much later, the original user would add a relation, and doesn't manually re-trigger the Publish flow, this relation will never get published

i don't quite have the solution for this problem yet, besides watching when relations.json change.
maybe we need a function that scans when relations.json was last modified, then potentially sync/ publish the relations that were created/modified after the last sync?

apps/obsidian/src/utils/publishNode.ts
```diff
@@ -56,6 +62,91 @@ const publishSchema = async ({
...
+      /* eslint-disable @typescript-eslint/naming-convention */
+      account_uid: myGroup,
+      source_local_id: sourceLocalId,
+      space_id: spaceId,
+      /* eslint-enable @typescript-eslint/naming-convention */
+    })),
+    { ignoreDuplicates: true },
+  );
+  if (publishResponse.error && publishResponse.error.code !== "23505")
+    throw publishResponse.error;
```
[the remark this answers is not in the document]
trangdoan982: worth addressing

apps/obsidian/src/utils/publishNode.ts
```diff
@@ -83,7 +166,32 @@ export const publishNode = async ({
   const existingPublish =
     (frontmatter.publishedToGroups as undefined | string[]) || [];
   const commonGroups = existingPublish.filter((g) => myGroups.has(g));
+  // temporary single-group assumption
   const myGroup = (commonGroups.length > 0 ? commonGroups : [...myGroups])[0]!;
+  return await publishNodeToGroup({ plugin, file, frontmatter, myGroup });
```
trangdoan982, line 251: 👍

trangdoan982 (changes requested):

apps/obsidian/src/utils/publishNode.ts
maparent: Ouch! Good point!

maparent: resolved in latest commit

apps/obsidian/src/utils/publishNode.ts
maparent: Re publishNodeRelations blocking publishNode: I am not sure it's an error. Basically, if we throw here, the user knows the node is not published, and may try again. Otherwise, the node will be said to be published, and we won't have another opportunity to know to publish the relations.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -66,21 +182,83 @@ export const discourseNodeInstanceToLocalConcept = ({
...
+
+  /* eslint-disable @typescript-eslint/naming-convention */
+  const literal_content: Record<string, Json> = {};
+  if (importedFromRid) literal_content.importedFromRid = importedFromRid;
+  return {
+    space_id: context.spaceId,
+    name: `[[${sourceNode.file.basename}]] -${relationType.label}-> [[${destinationNode.file.basename}]]`,
+    source_local_id: relationInstanceData.id,
+    author_local_id: relationInstanceData.author,
+    schema_represented_by_local_id: type,
```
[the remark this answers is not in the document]
maparent: This is intended.

apps/obsidian/src/utils/publishNode.ts
maparent: (Could fail silently after we create the scan for unpublished relations.)

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -403,41 +403,13 @@ const buildChangedNodesFromNodes = async ({
...
-  return buildChangedNodesFromNodes({
-    nodes: dgNodesInVault,
-    supabaseClient,
-    context,
-  });
-};
-
-export const createOrUpdateDiscourseEmbedding = async (
+export const syncAllNodesAndRelations = async (
   plugin: DiscourseGraphPlugin,
```
trangdoan982, line 407: nit: let's make this a parameter function {}:{} since it has more than 3 params

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -11,7 +11,7 @@ import { spaceUriAndLocalIdToRid, ridToSpaceUriAndLocalId } from "./rid";
 
 export const getAvailableGroups = async (
   client: DGSupabaseClient,
-): Promise<{ group_id: string }[]> => {
+): Promise<string[]> => {
```
trangdoan982, line 14: nit: i like this deconstruction. but maybe we should rename the function to getAvailableGroupIds for clarity?

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
maparent: It has exactly 3 params, not more.

trangdoan982 (approved):
Except the nit i think this lgtm. This handles the cases where relations are newly created, when a new node is published well.

we should still have a documentation somewhere of all the current cases that we're handling publishing relations right now. maybe we should add it in the official doc or as a README in the publish folder?!

maparent:
Re the documentation: I think it belongs in the main documentation.

PR #829, opened 2026-02-25
ENG-1407 Move Stored Relations toggle + migration flow to Personal Settings

maparent:
https://linear.app/discourse-graphs/issue/ENG-1407/move-stored-relations-toggle-migration-flow-to-personal-settings

https://www.loom.com/share/dd6bd0fa15a14dad9e8627f0c47e9a9a

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/829" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/settings/HomePersonalSettings.tsx
+++ b/apps/roam/src/components/settings/HomePersonalSettings.tsx
@@ -161,21 +161,15 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
           description="Transition to using stored relations instead of pattern-based relations"
           settingKeys={["Reified relation triples"]}
           value={storedRelations}
-          // eslint-disable-next-line @typescript-eslint/require-await
           onBeforeChange={async (checked) => {
             if (checked) {
-              countReifiedRelations()
-                .then((num: number) => {
-                  setNumExistingRelations(num);
-                  setActiveRelationMigration(
-                    num > 0
-                      ? RelationMigrationDialog.reactivate
-                      : RelationMigrationDialog.activate,
-                  );
-                })
-                .catch((error) => {
-                  internalError({ error });
-                });
+              const num = await countReifiedRelations();
+              setNumExistingRelations(num);
+              setActiveRelationMigration(
+                num > 0
+                  ? RelationMigrationDialog.reactivate
+                  : RelationMigrationDialog.activate,
+              );
             } else {
               setActiveRelationMigration(RelationMigrationDialog.deactivate);
             }
--- a/apps/roam/src/components/settings/HomePersonalSettings.tsx
+++ b/apps/roam/src/components/settings/HomePersonalSettings.tsx
@@ -155,28 +155,26 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
         />
       )}
 
-      <div id="stored-relation-flag">
-        <PersonalFlagPanel
-          title="Enable stored relations"
-          description="Transition to using stored relations instead of pattern-based relations"
-          settingKeys={["Reified relation triples"]}
-          value={storedRelations}
-          onBeforeChange={async (checked) => {
-            if (checked) {
-              const num = await countReifiedRelations();
-              setNumExistingRelations(num);
-              setActiveRelationMigration(
-                num > 0
-                  ? RelationMigrationDialog.reactivate
-                  : RelationMigrationDialog.activate,
-              );
-            } else {
-              setActiveRelationMigration(RelationMigrationDialog.deactivate);
-            }
-            return false;
-          }}
-        />
-      </div>
+      <PersonalFlagPanel
+        title="Enable stored relations"
+        description="Transition to using stored relations instead of pattern-based relations"
+        settingKeys={["Reified relation triples"]}
+        value={storedRelations}
+        onBeforeChange={async (checked) => {
+          if (checked) {
+            const num = await countReifiedRelations();
+            setNumExistingRelations(num);
+            setActiveRelationMigration(
+              num > 0
+                ? RelationMigrationDialog.reactivate
+                : RelationMigrationDialog.activate,
+            );
+          } else {
+            setActiveRelationMigration(RelationMigrationDialog.deactivate);
+          }
+          return false;
+        }}
+      />
 
       <PersonalFlagPanel
         title="Text selection popup"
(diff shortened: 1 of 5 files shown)


apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -186,6 +285,144 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
...
+            >
+              Deactivate
+            </Button>
+          </div>
+        </div>
+      </Dialog>
+      <Dialog
+        isOpen={activeRelationMigration === RelationMigrationDialog.activate}
+        onClose={() => {
+          setActiveRelationMigration(RelationMigrationDialog.none);
```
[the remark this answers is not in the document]
maparent: I think this is appropriate as-is. The toggle will have functioned.

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -186,6 +293,141 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
...
+                >
+                  Cancel
+                </Button>
+                <Button
+                  small
+                  intent={Intent.PRIMARY}
+                  onClick={() => {
+                    void setStoredRelations(true); // awaitable
+                    setActiveRelationMigration(RelationMigrationDialog.none);
+                  }}
```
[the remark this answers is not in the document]
maparent: Neither failure is likely, and not closing the dialog leaves the user stranded. but will still handle the failure.

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -21,20 +22,95 @@ import {
...
 const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
   const extensionAPI = onloadArgs.extensionAPI;
   const overlayHandler = getOverlayHandler(onloadArgs);
   const settings = useMemo(() => getFormattedConfigTree(), []);
-
+  const [activeRelationMigration, setActiveRelationMigration] =
+    useState<RelationMigrationDialog>(RelationMigrationDialog.none);
+  const [numExistingRelations, setNumExistingRelations] = useState<number>(0);
+  const [isOngoing, setOngoing] = useState<boolean>(false);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mdroidian: Is this still needed?  If so, why?

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -90,6 +166,37 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
...
         />
       )}
+
+      <div id="stored-relation-flag">
+        <PersonalFlagPanel
+          title="Enable stored relations"
+          description="Transition to using stored relations instead of pattern-based relations"
+          settingKeys={["Reified relation triples"]}
+          initialValue={getSetting<boolean>(USE_REIFIED_RELATIONS, false)}
+          // eslint-disable-next-line @typescript-eslint/require-await
```
mdroidian: Why do we need this?  Ideally we change the code so this isn't required.  If it is absolutely required, please explain why it is required.

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -21,20 +22,95 @@ import {
...
   const extensionAPI = onloadArgs.extensionAPI;
   const overlayHandler = getOverlayHandler(onloadArgs);
   const settings = useMemo(() => getFormattedConfigTree(), []);
-
+  const [activeRelationMigration, setActiveRelationMigration] =
+    useState<RelationMigrationDialog>(RelationMigrationDialog.none);
+  const [numExistingRelations, setNumExistingRelations] = useState<number>(0);
+  const [isOngoing, setOngoing] = useState<boolean>(false);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const settingStoredMigrationValue = useRef<boolean>();
```
mdroidian: Is there a reason we’re using `useRef` and direct DOM queries here?

This feels imperative and outside normal React patterns. Since the checkbox state can be controlled via React state, we could make this declarative and avoid `getElementById`, manual `.click()`, and ref flags entirely.

Proposed direction:

* Lift the setting into React state (single source of truth).
* Render the checkbox in controlled mode (`value={state}`).
* Persist changes inside the `onChange` handler.
* When migrations/dialogs need to toggle it, update state directly rather than manipulating the DOM.

This would:

* Remove the ref and DOM queries
* Simplify the mental model
* Keep behavior aligned with standard React patterns

If there’s a constraint preventing a controlled approach, it would be helpful to document it here.

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -186,6 +293,153 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
           }
         }}
       />
+      <Dialog
```
mdroidian, line 286: Let's use the common `<Dialog>` classes.  This should help the formatting and put the buttons on the right side (which would be consistent with all other dialogs).

EG:
`<div className={Classes.DIALOG_BODY}>`
`<div className={Classes.DIALOG_FOOTER}>`
`<div className={{Classes.DIALOG_FOOTER_ACTIONS}>`

Search `<Dialog` in the repo for examples.

mdroidian (changes requested):
Let's use the common `<Dialog>` classes to make the dialog more consistent.

If those classes don't fix the button sizing/formatting, please also make the dialog larger to accommodate either:
- single line buttons
- remove the whitespace on the right of the buttons

Image

EDIT: Or match with the proposed designs in the task:
image

apps/roam/src/components/settings/HomePersonalSettings.tsx
maparent: Ha! I did things this convoluted way because I did not realize I could pass the value along. Much simpler, thanks.

apps/roam/src/components/settings/HomePersonalSettings.tsx
maparent: Ah, I wrote the code in onChange (sync) first, then moved without rewriting. Fixed.

apps/roam/src/components/settings/HomePersonalSettings.tsx
maparent: Did that. I also made two of the dialogs wider to accommodate the text length. Unfortunately the only way I could do so was with a tailwind relative width command (`w-2/3`. 3/5 is not enough, nor is w-95.)
Is there a way to define a large width?

maparent: Another thing: I did not put the number of relations in a text field. It's not editable. I think that creating a form for this is a bit weird, but could do it if you think it's important.

mdroidian: If there isn't a tailwind class for it, use `style`.<br>The relations don't have to be in a text field, it's just the design to make it stick out.

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -90,6 +154,30 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
           }}
         />
       )}
+
+      <div id="stored-relation-flag">
```
mdroidian: this div is no longer needed

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -21,20 +22,83 @@ import {
...
 
 const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
   const extensionAPI = onloadArgs.extensionAPI;
   const overlayHandler = getOverlayHandler(onloadArgs);
   const settings = useMemo(() => getFormattedConfigTree(), []);
-
+  const [activeRelationMigration, setActiveRelationMigration] =
+    useState<RelationMigrationDialog>(RelationMigrationDialog.none);
+  const [numExistingRelations, setNumExistingRelations] = useState<number>(0);
+  const [isOngoing, setOngoing] = useState<boolean>(false);
```
mdroidian: setIsOngoing

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -186,6 +274,156 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
...
       />
+      <Dialog
+        isOpen={activeRelationMigration === RelationMigrationDialog.reactivate}
+        onClose={() => {
+          setActiveRelationMigration(RelationMigrationDialog.none);
+        }}
+        className="w-2/3"
+      >
+        <div className="p-4">
+          {isOngoing ? (
```
mdroidian: this should be inside `Classes.DIALOG_BODY` as well to maintain similar formatting

apps/roam/src/components/settings/HomePersonalSettings.tsx
```diff
@@ -186,6 +274,156 @@ const HomePersonalSettings = ({ onloadArgs }: { onloadArgs: OnloadArgs }) => {
...
+                }}
+              >
+                Deactivate
+              </Button>
+            </div>
+          </div>
+        </div>
+      </Dialog>
+      <Dialog
+        isOpen={activeRelationMigration === RelationMigrationDialog.activate}
```
mdroidian: If the only difference of the first and last <Dialog> are the buttons and showing the # of relations, then we could DRY by adding the conditional there instead of the whole dialog.

apps/roam/src/utils/migrateRelations.ts
```diff
@@ -4,18 +4,17 @@ import type { json } from "./getBlockProps";
 import setBlockProps from "./setBlockProps";
 import { getSetting, setSetting } from "./extensionSettings";
```
mdroidian, line 5: remove

apps/roam/src/utils/migrateRelations.ts
```diff
@@ -4,18 +4,17 @@ import type { json } from "./getBlockProps";
...
 import { USE_REIFIED_RELATIONS } from "~/data/userSettings";
```
mdroidian, line 6: remove

mdroidian (approved):
Mostly nits in the comments.

I also still don't love the buttons when the viewport is small, but we can defer this to responsive/mobile support which needs work across the app.

Lastly, re: Relations in text input.  As mentioned, it's not a text input, it's just the design.  Regardless of decision, it could use equal spacing.

Image

apps/roam/src/utils/migrateRelations.ts
maparent: Actually needed, the change was not done.

apps/roam/src/utils/migrateRelations.ts
maparent: needed.

mdroidian: Where is it needed?

image

mdroidian: Removed in https://github.com/DiscourseGraphs/discourse-graph/pull/852

PR #835, opened 2026-02-28
[ENG-1165] Rename node titles on page load if title has changed

trangdoan982:
https://www.loom.com/share/43a3acb67f294da8b69c8fc64a3beab7

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/835" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/canvas/Tldraw.tsx
+++ b/apps/roam/src/components/canvas/Tldraw.tsx
@@ -106,6 +106,7 @@ import ToastListener, { dispatchToastEvent } from "./ToastListener";
 import { CanvasDrawerPanel } from "./CanvasDrawer";
 import { ClipboardPanel, ClipboardProvider } from "./Clipboard";
 import internalError from "~/utils/internalError";
+import { syncCanvasNodeTitlesOnLoad } from "~/utils/syncCanvasNodeTitlesOnLoad";
 import { AUTO_CANVAS_RELATIONS_KEY } from "~/data/userSettings";
 import { getSetting } from "~/utils/extensionSettings";
 import { isPluginTimerReady, waitForPluginTimer } from "~/utils/pluginTimer";
@@ -932,6 +933,7 @@ const TldrawCanvasShared = ({
   return (
     <div
       className="roamjs-tldraw-canvas-container relative z-10 h-full w-full overflow-hidden rounded-md border border-gray-300 bg-white"
+      data-page-uid={pageUid ?? undefined}
       ref={containerRef}
       tabIndex={-1}
       onDragOver={handleDragOver}
@@ -1026,6 +1028,16 @@ const TldrawCanvasShared = ({
               // hack for "cannot change atoms during reaction cycle" bug
               installSafeHintingSetter({ app, title, pageUid });
               setHasMountedEditor(true);
+              void syncCanvasNodeTitlesOnLoad(
+                app,
+                allNodes.map((n) => n.type),
+                allRelationIds,
+              ).catch((err) => {
+                internalError({
+                  error: err instanceof Error ? err : new Error(String(err)),
+                  type: "Canvas: Sync node titles on load",
+                });
+              });
 
               app.on("change", (entry) => {
                 lastActionsRef.current.push(entry);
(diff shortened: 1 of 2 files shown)


apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -1026,6 +1028,16 @@ const TldrawCanvasShared = ({
               // hack for "cannot change atoms during reaction cycle" bug
               installSafeHintingSetter({ app, title, pageUid });
               setHasMountedEditor(true);
+              void syncCanvasNodeTitlesOnLoad(
+                app,
+                allNodes.map((n) => n.type),
+                allRelationIds,
+              ).catch((err) => {
+                internalError({
+                  error: err instanceof Error ? err : new Error(String(err)),
```
mdroidian: just pass `err` directly, `internalError` accepts unknown and handles it.

apps/roam/src/utils/syncCanvasNodeTitlesOnLoad.ts
```diff
@@ -0,0 +1,109 @@
...
+    `[:find ?uid (pull ?e [:node/title :block/string])
+      :in $ [?uid ...]
+      :where [?e :block/uid ?uid]]`,
+    uids,
+    /* eslint-disable-next-line @typescript-eslint/naming-convention */
+  )) as [string, { ":node/title"?: string; ":block/string"?: string }][];
+
+  const map = new Map<string, string>();
+  for (const [uid, pull] of results) {
+    const title = pull?.[":node/title"] ?? pull?.[":block/string"] ?? "";
```
mdroidian, line 23: What is the use case for `block/string`?  Block shapes are currently embedded, so they don't use text stored in shape props.

apps/roam/src/utils/syncCanvasNodeTitlesOnLoad.ts
```diff
@@ -0,0 +1,109 @@
...
+  const shapesToUpdate: { shape: DiscourseNodeShape; newTitle: string }[] = [];
+  const shapesToRemove: DiscourseNodeShape[] = [];
+
+  for (const shape of discourseNodeShapes) {
+    const uid = shape.props.uid;
+    const currentInRoam = uidToTitle.get(uid);
+    if (currentInRoam === undefined) {
+      shapesToRemove.push(shape);
+    } else if (
+      (shape.props.title ?? "").trim() !== (currentInRoam ?? "").trim()
```
mdroidian: Why are we trimming?  Roam titles can have spaces at the end.

apps/roam/src/utils/syncCanvasNodeTitlesOnLoad.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  if (shapesToRemove.length > 0) {
+    for (const shape of shapesToRemove) {
+      deleteNodeShapeAndRelations(editor, shape, relationIds);
+    }
+  }
+
+  if (shapesToUpdate.length > 0) {
+    editor.updateShapes(
```
mdroidian, line 99: Note: Ideally we check/resize the node as well based on the text (eg: change from "Some really long title abc 123" to "Short title" would leave a bunch of white space) (this is the current flow when nodes get added).  But it could be the case that the size has been manually set, in which case we wouldn't want to resize.  So we can leave this for now.

mdroidian (approved):
Some nits/thoughts.

apps/roam/src/utils/syncCanvasNodeTitlesOnLoad.ts
trangdoan982: yes we had the same discussion in Obsidian tldraw and agreed that size shouldn't be reset when text change

apps/roam/src/utils/syncCanvasNodeTitlesOnLoad.ts
trangdoan982: not sure what you mean? this query is for block string from roam not the tldraw shape?!

mdroidian: Why do you need to query `block/string`?  Where will it be used?

mdroidian: Discussed 2026-03-12

PR #847, opened 2026-03-02
ENG-1301: Import relations when a node is imported

trangdoan982:
https://www.loom.com/share/407aff42d138483d9c59e20d55a24aec

https://www.loom.com/share/f299d265128d4fe4b79db9c7542e7fcb

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/847" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


apps/obsidian/src/utils/importRelations.ts
```diff
@@ -0,0 +1,285 @@
...
+
+  const { data: concepts } = await client
+    .from("my_concepts")
+    .select("id, source_local_id")
+    .eq("space_id", spaceId)
+    .in("id", [sourceConceptId, destConceptId]);
+
+  if (!concepts || concepts.length < 2) {
+    return null;
+  }
```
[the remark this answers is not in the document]
trangdoan982: we don't allow creating relations like this so don't have to concern ourselves

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -424,3 +450,49 @@ export class QueryEngine {
...
+/**
+ * Returns info about imported nodes (from import/ folder only).
+ * Uses DataCore to query by path and metadata when available; otherwise iterates vault.
+ * - nodeKeys: "spaceId:nodeInstanceId" for each imported node
+ * - keyToRid: maps "spaceId:nodeInstanceId" -> importedFromRid
+ */
+export const getImportedNodesInfo = async ({
+  queryEngine,
+  plugin,
+  client,
```
maparent: I think mixing database and query engine is not ideal. Separate this function into a pure QE here and another function elsewhere (relationStore?) that uses those results to get spaceIds.
Also, decompose it so we don't hit the db once per file; but gather the spaceUrls from all files, and then do a single query that will give you the IDs from these. (I realize I'm guilty of the getSpaceNameIdFromRid function, but it was not used in a loop!)

apps/obsidian/src/utils/importRelations.ts
```diff
@@ -0,0 +1,285 @@
...
+    ...(plugin.settings.relationTypes ?? []),
+    newRelationType,
+  ];
+  await plugin.saveSettings();
+  return newRelationType.id;
+};
+
+/**
+ * Fetch relation instances from a remote space. Relation instances are concepts with
+ * is_schema=false and schema_id pointing to a relation type (arity=2).
```
maparent, line 168: That works, but it's simpler than that: the relations themselves have arity>0.

apps/obsidian/src/utils/importRelations.ts
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  return instances as unknown as RemoteRelationInstance[];
+};
+
+/**
+ * Resolve source and destination nodeInstanceIds from a relation instance's reference_content.
+ * reference_content has { source: conceptId, destination: conceptId }.
+ */
+const resolveRelationEndpoints = async ({
```
maparent: This function can be eliminated with a better query. I'll do that part.

apps/obsidian/src/utils/importRelations.ts
```diff
@@ -0,0 +1,285 @@
...
+
+  for (const rel of relationInstances) {
+    const endpoints = await resolveRelationEndpoints({
+      client,
+      spaceId,
+      referenceContent: rel.reference_content,
+    });
+    if (!endpoints) continue;
+
+    const sourceKey = `${spaceId}:${endpoints.sourceId}`;
```
maparent: This uses the current spaceId, not the endepoint's spaceId. I think it cannot work for imported nodes, but maybe I'm missing something? Let's discuss.

apps/obsidian/src/utils/importRelations.ts
```diff
@@ -0,0 +1,285 @@
...
+      continue;
+    }
+
+    const sourceRid = keyToRid.get(sourceKey);
+    const destRid = keyToRid.get(destKey);
+    if (!sourceRid || !destRid) continue;
+
+    if (!rel.schema_id) continue;
+
+    const { data: schemaConcept } = await client
```
maparent: Again worth factoring out! Take all schema_ids from all relations as as set, single query with in, then use that in the loop.

maparent (changes requested):
Some of this we'll look at together. Let's plan a joint session.

apps/obsidian/src/utils/importRelations.ts
```diff
@@ -0,0 +1,239 @@
...
+  created: string | null;
+  last_modified: string | null;
+  concepts_of_relation: ConceptInRelation[];
+};
+
+/**
+ * Map a remote relation type to local. Match by id first (use local if id exists with different label/complement),
+ * then by label, create if new.
+ */
+const mapRelationTypeToLocal = async ({
```
trangdoan982, line 37: also need to add the triple

packages/database/scripts/createGroup.ts
```diff
@@ -0,0 +1,176 @@
+/**
```
trangdoan982: this script is mostly for my local testing to create group quickly. i've been using it for a while and it doesn't touch any part of the codebase but happy to delete it as well

trangdoan982:
@maparent here are some big changes since the last time you reviewed:
- I addressed your comments on query optimization. Thank you for the notes!
- import the relation triple as well (not just relation type). the order is import node type -> relation type -> relation triplet. so the triplet will use whatever is resolved from previous steps
- add the "provisional" marker on the imported relation. but no UI change yet. all imported relations will have this field marked to `false` for now
- added importPreview page where we tell users how many schemas and relations are being imported implicitly.
image
image

- biggest change is how we resolve the endpoint of a relation. now every imported relation's source/dest will resolve to Rid -> this way we can parse both local and remote node using 1 method. my argument is that it's also good to know that the relation was created when the node was imported. but let me know what you think.

Cases I tested:
- import relations with nodes imported from the same vault
- Vault 1: import node from vault2 -> create a relation btw local node and imported node -> publish local node -> Vault 2 import node from vault 1, see that the relations were resolved correctly

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -319,6 +320,148 @@ export class QueryEngine {
...
+  };
+
+  /**
+   * Find a file by nodeInstanceId or importedFromRid (endpoint id).
+   * Tries DataCore getDiscourseNodeById and getFileByImportedFromRid when available;
+   * falls back to iterating files with nodeInstanceId and matching either field.
+   */
+  getFileByEndpoint = (endpointId: string): TFile | null => {
+    if (this.dc) {
+      const byId = this.getDiscourseNodeById(endpointId);
```
maparent, line 449: Small concern: I'm realizing that `getDiscourseNodeById` may return an imported node that coincides on `nodeInstanceId`. That is probably not intended, and we should probably make it explicitly into `getLocalNodeById` and test accordingly.  And it's also a problem in the fallback: maybe we should test `rid === endpointId || (!rid && id === endpointId)`

apps/obsidian/src/utils/importRelations.ts
```diff
@@ -0,0 +1,385 @@
...
+        ),
+      );
+      const sourceSchemaId = byConceptId[sourceData.id];
+      const destSchemaId = byConceptId[destData.id];
+      if (sourceSchemaId != null && destSchemaId != null) {
+        const { data: schemaRows } = await client
+          .from("my_concepts")
+          .select("id, source_local_id")
+          .in("id", [sourceSchemaId, destSchemaId]);
+        if (schemaRows && schemaRows.length === 2) {
```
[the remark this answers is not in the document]
maparent: Interesting; I am not sure we _want_ to support that but we don't exclude it yet.

apps/obsidian/src/utils/importPreview.ts
```diff
@@ -0,0 +1,473 @@
...
+      const schemaNumericId = [...relSchemaIdToLocalId.entries()].find(
+        ([, localId]) => localId === relTypeLocalId,
+      )?.[0];
+      if (schemaNumericId == null) continue;
+
+      const { data: schemaData } = await client
+        .from("my_concepts")
+        .select("name, literal_content, space_id")
+        .eq("id", schemaNumericId)
+        .maybeSingle();
```
maparent: One more! There's probably not many relation schemata, but there can be many relations. So it would be worth factoring that request out.

apps/obsidian/src/utils/importPreview.ts
```diff
@@ -0,0 +1,473 @@
...
+  const allEndpointConceptIds = [
+    ...new Set(
+      matchingRelations.flatMap((m) => [m.sourceConceptId, m.destConceptId]),
+    ),
+  ];
+
+  // Batch fetch schema_id for all endpoint concepts
+  const conceptIdToSchemaId = new Map<number, number>();
+  if (allEndpointConceptIds.length > 0) {
+    const { data: conceptSchemas } = await client
```
maparent, line 333: I think that's right but checking:  `allConceptEndpointIds` should be larger than `selectedNotes`, right?
I had a doubt because of the filter for local+imported nodes in L195.
(Otherwise I'd think about whether we could get the data for this by adding `id` to the select at line 86... but almost certainly overkill.)

apps/obsidian/src/utils/importPreview.ts
```diff
@@ -0,0 +1,473 @@
...
+      nodeTypeIdToName.get(destSchemaInfo.source_local_id) ??
+      destSchemaInfo.name;
+
+    const tripletKey = `${sourceSchemaInfo.source_local_id}:${relTypeLocalId}:${destSchemaInfo.source_local_id}`;
+    if (tripletSet.has(tripletKey)) continue;
+    tripletSet.add(tripletKey);
+
+    // Check if this triplet already exists in discourseRelations
+    // We need to check using the mapped local ids (match by id first, then by name)
+    const resolveLocalNodeTypeId = (
```
maparent, line 412: Not vital, since not hitting the db, but I'd have cached this, it's still a search per instance. Oh well.

maparent (changes requested):
Ok. Of my comments, two are real code change requests, and after that I'll approve.
Otherwise, thanks for all the great work with optimizations, and I like how you handled fallbacks within the QueryEngine.
I did some testing, that seems to work great.
In terms of behaviour... I'm actually not sure why we need the preview mode in most cases. We're asking people to confirm a choice they just made. There is one good reason: when we import a schema, because it has unfortunate consequences: We cannot edit or remove it. (Also unhappy with that.)
SO two recommendation: 
1. Make handling imported schemas a very high priority after this is merged.
2. Could we skip the validation screen when there is no schema import?
The latter is a design decision, and may need to be validated with Joel, Johnny and/or Michael. So not sure we should block merge on it. But I think it's worth asking: I really don't see the point of that validation step if there's no schema change.

maparent:
I'm having second thoughts. I agreed, and I think it was the right call, that imported schemata should not be edited in the short term. But I think that schemata (imported or not) should be delete-able, as long as they have no live instances. Again, this would do much to mitigate the consequences of importation. We'd need buy-on on that, and I'm willing to help to code it, but I think that importing un-delete-able schemata is a design mistake.
Whether we correct this in this PR or another is a judgment call.

apps/obsidian/src/services/QueryEngine.ts
maparent: Forget about this one, obviously it's unique anyway. Apologies.

apps/obsidian/src/utils/importPreview.ts
trangdoan982: yes it should >= selectedNodes. you want to add the check here instead of >0 check?

apps/obsidian/src/utils/importPreview.ts
trangdoan982: not sure what you mean here?

maparent (approved):

maparent:
I approved but the checks are all failing. More annoying, the validate check passes locally, and the error message is obviously wrong.
The lint issue is real, and has to do with createGroup.ts. You could fix or remove the file.
If you fix, it's asking about adding a variable to turbo, I would suppress that warning instead.

apps/obsidian/src/utils/importRelations.ts
```diff
@@ -0,0 +1,380 @@
...
+              sourceNodeTypeId: remoteSourceNodeTypeId,
+            });
+            mappedDestNodeTypeId = await mapNodeTypeIdToLocal({
+              plugin,
+              client,
+              sourceSpaceId: spaceId,
+              sourceSpaceUri: spaceUri,
+              sourceNodeTypeId: remoteDestNodeTypeId,
+            });
+          }
```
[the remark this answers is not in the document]
maparent: Actually this is worth fixing!

maparent:
The build error was fixed by rebasing on main. It seems the CI does its own flawed rebasing.

maparent:
Ok still approved but do address the graphite comment, and the lint issue.

trangdoan982:
yes thank you

PR #851, opened 2026-03-03
[ENG-1108] Canvas tool lock

trangdoan982:
https://www.loom.com/share/07b5bafb3bb54450bbe9a40862f9cd4a

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/851" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -932,6 +939,7 @@ const TldrawCanvasShared = ({
   return (
     <div
       className="roamjs-tldraw-canvas-container relative z-10 h-full w-full overflow-hidden rounded-md border border-gray-300 bg-white"
+      data-page-uid={pageUid}
```
trangdoan982: we need to add this back because i keep receiving error "No canvas page UID found"

mdroidian (changes requested):
It seems there was come confusion on what the task was requesting.  Unfortunately, the ticket should have been more clear, sorry.

If a user:
- selects dg tool
- toggle lock on
- creates a node

The dg tool/node should still be the current tool

If a user:
- selects dg tool
- toggles lock off (or leaves as default)
- creates a node

The current tool should be the `select` tool

---

This follows the same behavior as all other tldraw tools.

https://www.loom.com/share/265d543710924f548202e006d033452b

trangdoan982:
@mdroidian is this the same behavior we want for [ENG-1383: Stay at the relation arrow menu after creating a relation](https://linear.app/discourse-graphs/issue/ENG-1383/stay-at-the-relation-arrow-menu-after-creating-a-relation)? Based on the description in this ticket I understood that once the discourse relation tool is selected, it should stay locked there without doing any extra lock tool operation.

In this latest push I had discourse node and relation tools to have the same behavior, but let me know if you interpret it differently.

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -1099,6 +1106,18 @@ const InsideEditorAndUiContext = ({
   const toasts = useToasts();
   const msg = useTranslation();
 
+  // When the user selects the select tool, clear tool lock so we only lock while on discourse tools
+  const currentToolId = useValue(
```
mdroidian: Why are we doing this?  This changes the default tldraw behavior.

apps/roam/src/components/canvas/toolLock.ts
```diff
@@ -0,0 +1,19 @@
...
+/** When the user selects the select tool, clear tool lock so we only lock while on discourse tools. */
+export const unlockToolWhenSelect = (editor: Editor): void => {
+  if (editor.getCurrentToolId() === "select") {
+    editor.updateInstanceState({ isToolLocked: false });
+  }
+};
+
+/** Unlock the tool (e.g. when opening a dialog so the user is not stuck with Select + locked). */
+export const unlockTool = (editor: Editor): void => {
+  editor.updateInstanceState({ isToolLocked: false });
```
mdroidian: nit: what is the benefit of wrapping this in a function?

apps/roam/src/components/canvas/toolLock.ts
```diff
@@ -0,0 +1,19 @@
+import { Editor } from "tldraw";
+
+export const setCurrentToolToSelectIfUnlocked = (editor: Editor): void => {
+  if (!editor.getInstanceState().isToolLocked) {
+    editor.setCurrentTool("select");
+  }
+};
+
+/** When the user selects the select tool, clear tool lock so we only lock while on discourse tools. */
+export const unlockToolWhenSelect = (editor: Editor): void => {
```
mdroidian: note: this is only used in one place, so it is unnecessary to create a function for.  It also creates the wrong signal as per: [Guidelines](https://linear.app/discourse-graphs/document/project-handbook-dd139f6e26fc#pr-guidelines-9e4fee89) > Code Placement & Extraction Heuristic

mdroidian (changes requested):
https://www.loom.com/share/07aab48252a744f4ad016e00e619e7e1
 
- cannot edit discourse nodes
- changing default tldraw lock behavior unnecessarily (unlocking on choosing select tool)

trangdoan982:
https://www.loom.com/share/878e7f7432b645ceb310c914ed200d93


Here's my updated test with the bugs you pointed out. I couldn't reproduce the unable to edit bug. Can you try again?

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -818,6 +820,7 @@ const TldrawCanvasShared = ({
             ...position,
           },
         ]);
+        setCurrentToolToSelectIfUnlocked(app);
```
mdroidian: Why are we changing the tool state here?

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -1175,6 +1178,7 @@ const InsideEditorAndUiContext = ({
           },
         },
       ]);
+      setCurrentToolToSelectIfUnlocked(editor);
```
mdroidian: Why are we changing the tool here?  This diverges from the regular tldraw pattern.  eg
If tool = rectangle and I paste text that is not a dg shape, tool stays as rectangle.
If tool = rectangle and I paste text that is a dg shape, tool changes to select.

apps/roam/src/components/canvas/uiOverrides.tsx
```diff
@@ -156,6 +157,7 @@ export const getOnSelectForShape = ({
     initialText: string;
     imageUrl?: string;
   }) => {
+    editor.updateInstanceState({ isToolLocked: false });
```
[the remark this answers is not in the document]
mdroidian: @trangdoan982 Why are we changing the lock state here? Wthat problem is this solving?

mdroidian (changes requested):
There are some undesired changes that I am unclear why they are included in this PR.

Also, this devin comment is worth looking into: https://github.com/DiscourseGraphs/discourse-graph/pull/851#discussion_r2961617046

mdroidian (approved):

PR #860, opened 2026-03-05
ENG-1503: Replace getFormattedConfigTree consumers with direct helper calls

sid597:
https://www.loom.com/share/e1bb8573adfe49549013385f6391d21a

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/860" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **Refactor**
  * Restructured internal settings access patterns across configuration components, transitioning from legacy config tree retrieval to new helper-based utilities for left sidebar, suggestive mode, export, and general settings management. No user-facing changes; improvements to code maintainability and data flow architecture.

### DIFF
--- a/apps/roam/src/components/settings/ExportSettings.tsx
+++ b/apps/roam/src/components/settings/ExportSettings.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { getFormattedConfigTree } from "~/utils/discourseConfigRef";
+import { getExportSettingsAndUids } from "~/utils/getExportSettings";
 import {
   GlobalFlagPanel,
   GlobalNumberPanel,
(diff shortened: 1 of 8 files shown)


sid597:
@coderabbitai full review

mdroidian (approved):
I won't block on this, but most of these changes look like they use the same functions that `getFormattedConfigTree` was using. If so, how is this an improvement? I was under the impression the removal of `getFormattedConfigTree` was to consolidate under dual read. But these changes are effectively the same behavior/logic as previously?  It actually seems like these changes will make it more difficult to fully migrate (we'll have to look for many more smaller generic functions, rather than one function)

Can you help me understand what the purpose/benefit of these changes are?

apps/roam/src/components/settings/ExportSettings.tsx
```diff
@@ -8,9 +8,8 @@ import {
 } from "./components/BlockPropSettingPanels";
 
 const DiscourseGraphExport = () => {
-  const settings = getFormattedConfigTree();
-  const exportSettings = settings.export;
-  const parentUid = settings.export.exportUid;
+  const exportSettings = getExportSettingsAndUids();
```
sid597, line 15: So the idea is that depending on how we're going to clean up the functions, if we want to clean up by an individual component, we had to remove `getFormattedConfigTree`. This is because it could be used on multiple components in a single file but also because we cannot remove just a single. It could be the case that if we move a single key when we're trying to clean up that entire tree of components somewhere else in the repo could use that key, which would then break and go outside of the scope of what we're trying to clean up.

PR #890, opened 2026-03-13
[ENG-1502] Create Agent SKILLS to auto-update documentations once a feature is finished

trangdoan982:
https://www.loom.com/share/ffe8e854eb00455193b647705f1aed5d

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/890" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


.claude/skills/update-docs/SKILL.md
```diff
@@ -0,0 +1,11 @@
+---
```
mdroidian: Why is this in `.claude`?  Shouldn't this be available to all agents and be in `/skills/update-docs/SKILL.md`?

See: https://agentskills.io/specification

mdroidian: Actually, why is SKILL.md separate from update-docs.md?

Image

skills/update-docs.md
```diff
@@ -0,0 +1,126 @@
...
+### If running for the current branch
+
+Collect information from three sources:
+
+1. **Changed files** — Run `git diff main...HEAD --name-only` to identify what changed. Use file paths to determine the affected platform:
+   - `apps/obsidian/` → Obsidian docs
+   - `apps/roam/` → Roam docs
+   - `packages/` → Ask the dev which platform(s) are affected
+2. **Commit messages** — Run `git log main...HEAD --oneline` for a summary of what was done.
+3. **Linear ticket** — If a Linear ticket is linked (branch name or PR description), pull the title, description, and acceptance criteria for additional context.
```
mdroidian: How does the agent access linear?  I do not see explicit instructions here.
Is it assumed that the dev already has the linear MCP activated (and only activated for discourse-graphs?)
If so, let's make that explicit.

skills/update-docs.md
```diff
@@ -0,0 +1,126 @@
+# Update Docs
+
+Auto-update user-facing documentation when a feature or behavior change ships.
+
+## When to Run
+
+On demand. Two modes:
```
mdroidian: The SKILL.md says "Run after a feature branch is ready for PR or review" but here it says "On demand".  This seems to be conflicting from the agent perspective.

Does "on demand" mean that the dev calls it?  Or is it run automatically by the agent?

skills/update-docs.md
```diff
@@ -0,0 +1,126 @@
...
+Collect information from three sources:
+
+1. **Changed files** — Run `git diff main...HEAD --name-only` to identify what changed. Use file paths to determine the affected platform:
+   - `apps/obsidian/` → Obsidian docs
+   - `apps/roam/` → Roam docs
+   - `packages/` → Ask the dev which platform(s) are affected
+2. **Commit messages** — Run `git log main...HEAD --oneline` for a summary of what was done.
+3. **Linear ticket** — If a Linear ticket is linked (branch name or PR description), pull the title, description, and acceptance criteria for additional context.
+
+### If running for a merged PR
```
mdroidian: PR could be merged but they are still on the branch.  Maybe make this more explicit, eg: instead of "if running for a merged PR" something like "when an argument is passed".

skills/update-docs.md
```diff
@@ -0,0 +1,126 @@
+# Update Docs
```
mdroidian: Let's rename this folder to differentiate between "user facing docs" vs just "docs" or "dev docs".

File should be `SKILL.md`, eg: `/skills/{SKILL-NAME}/SKILL.md` so `update-docs.md` shouldn't exist.
See previous comment on `.claude/skills/update-docs/SKILL.md`

skills/update-docs.md
```diff
@@ -0,0 +1,126 @@
...
+  published: true
+  ---
+  ```
+- **Platform routing:** Docs are platform-specific by default. Only place docs in `sharedPages/` if explicitly specified.
+
+### Navigation Mapping
+
+If a **new page** is created, register it in the appropriate `docMap.ts` and `navigation.ts`:
+
+**docMap.ts** (maps slugs to content directories — only needed for shared pages; platform-specific pages use the default):
```
mdroidian: We could DRY this and point to dev docs.
That way we define it only in one place for both agents and devs.

This will be useful especially with the upcoming nextra/cms changes.

mdroidian (changes requested):
Good start, but I think it need some cleanup before we can merge.  See below:

skills/update-docs.md
trangdoan982: yes dev needs to have linear MCP activated. let me update

skills/update-docs.md
trangdoan982: dev calls it. I thought of every commit or when PR is created but that may be redundant

trangdoan982: let me update

skills/update-docs.md
trangdoan982: yeah i added it into `./claude` folder so that i can evoke it as a command. but ig this is more of a local dev flow. i added `./claude` into gitignore instead

skills/update-user-docs/references/navigation-mapping.md
```diff
@@ -0,0 +1,86 @@
+# Navigation Mapping
```
mdroidian, line 1: We should DRY this with CONTRIBUTING.md for a single source of truth

mdroidian (approved):

PR #900, opened 2026-03-18
[ENG-1271] Implement filter, search, sort designs in clipboard

trangdoan982:
https://www.loom.com/share/ca48bea531b44511ac951363dd996971

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/900" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/canvas/Clipboard.tsx
+++ b/apps/roam/src/components/canvas/Clipboard.tsx
@@ -992,9 +992,15 @@ const ClipboardPageSection = ({
             </div>
           ) : visibleGroupedNodes.length === 0 ? (
             <div className="rounded border border-dashed border-gray-200 p-2">
+
               {searchQuery || selectedNodeType !== "All"
-                ? "No nodes match the current filters."
+                ? showNodesOnCanvas
+                  ? "No nodes match the current filters."
+                  : "No nodes match the current filters, or matching nodes are already on canvas. Turn on \"Show nodes on canvas\" to view them."
+                : showNodesOnCanvas
+                ? "All nodes from this page are already on canvas."
                 : 'All nodes from this page are already on canvas. Turn on "Show nodes on canvas" to view them.'}
+
             </div>
           ) : (
             <div className="space-y-1">
--- a/apps/roam/src/components/canvas/Clipboard.tsx
+++ b/apps/roam/src/components/canvas/Clipboard.tsx
@@ -586,6 +586,10 @@ const ClipboardPageSection = ({
     [groupedNodes, showNodesOnCanvas, searchQuery, selectedNodeType],
   );
 
+  // Publish this page's distinct node types up to ClipboardPanel so it can
+  // build a unified filter dropdown across all open pages. When
+  // `showNodesOnCanvas` is off, only consider nodes not yet on the canvas so
+  // the filter options match what the user can actually act on.
   useEffect(() => {
     const candidateNodes = showNodesOnCanvas
       ? groupedNodes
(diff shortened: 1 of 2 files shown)


apps/roam/src/components/canvas/Clipboard.tsx
```diff
@@ -1119,7 +1168,7 @@ export const ClipboardPanel = () => {
         </h2>
         <div className="flex-shrink-0">
           <Button
-            icon={<Icon icon="minus" />}
+            icon={<Icon icon={isCollapsed ? "chevron-down" : "minus"} />}
```
trangdoan982: unrelated to this ticket but a quick UI fix

mdroidian: `chevron-down` usually means expanded.  Introducing this change now has two different meanings in the same component.  Let's remove this for now and add it as a FEE ticket.

Image

Image

Image

mdroidian (changes requested):
Let's match the UI designs more closely:

- Nodes should have their node colors (you can take this code from elsewhere in the project, eg Canvas Drawer)

Image

- Search on left, filters/sort on right
- Show "Clear Filters" but keep it disabled.  This is to be consistent with other components and help remove jank and it won't come as a surprise to users
- Filter button should have the button design as seen in the image

Image

- Remove search layout jank
https://www.loom.com/share/3547f91eff7e4ffa95da825a29184954

trangdoan982:
https://youtu.be/9B_Xv7wvz48

@mdroidian Loom is down so this is the recorded test for the new filter according to Johnny's new design and your feedback

apps/roam/src/components/canvas/Clipboard.tsx
```diff
@@ -534,26 +546,48 @@ const ClipboardPageSection = ({
   const groupedNodes = useMemo(() => {
     const groups: NodeGroup[] = discourseNodes.map((node) => {
       const shapes = shapesByUid.get(node.uid) ?? [];
+      const discourseNode = findDiscourseNode({ uid: node.uid });
```
mdroidian: `findDiscourseNode` can be a heavy query.  `Type` should be accessible when we get `node` so we should do this further up the chain to avoid `findDiscourseNode` inside of a `.map`

EG: `fetchDiscourseNodes` is already running `isDiscourseNode` which runs `findDiscourseNode`.  So let's modify `fetchDiscourseNodes` to retun `type` there and include it with `setDiscourseNodes`

mdroidian (changes requested):
Let's move getting the node `type` up the chain so that we don't duplicate the `findDiscourseNode` call

apps/roam/src/components/canvas/Clipboard.tsx
```diff
@@ -1090,6 +1135,48 @@ export const ClipboardPanel = () => {
...
+  }, []);
+
+  useEffect(() => {
+    if (
+      selectedNodeType !== "All" &&
+      !availableNodeTypes.includes(selectedNodeType)
+    ) {
+      setSelectedNodeType("All");
+    }
+  }, [availableNodeTypes, selectedNodeType]);
```
mdroidian, line 1184: Let's add a comment as to what problem this useEffect is solving.
EG: When would `availableNodeTypes` not include `selectedNodeType`, and why would we want to switch to `All` instead of showing no results?

apps/roam/src/components/canvas/Clipboard.tsx
```diff
@@ -537,23 +556,44 @@ const ClipboardPageSection = ({
...
+    [groupedNodes, showNodesOnCanvas, searchQuery, selectedNodeType],
   );
 
+  useEffect(() => {
+    const candidateNodes = showNodesOnCanvas
+      ? groupedNodes
+      : groupedNodes.filter((n) => n.shapes.length === 0);
+    const types = [...new Set(candidateNodes.map((n) => n.type))];
+    onNodeTypesChange(page.uid, types);
+  }, [groupedNodes, page.uid, onNodeTypesChange, showNodesOnCanvas]);
```
mdroidian, line 599: Let's add a comment as to what problem this useEffect is solving.

mdroidian (approved):

PR #902, opened 2026-03-19
ENG-948: Add blocks to left sidebar via block context menu

sid597:
Register block context menu commands so users can right-click any block and add it to a left sidebar section. On click, the block is added (dual-write to tree + block props), and the settings panel opens to the relevant tab with the target section auto-expanded.

https://www.loom.com/share/bf3d36dead044d1b854d223d9ec6500a

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/902" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added context menu commands to add blocks to personal sections in the left sidebar
  * Personal sidebar sections now support initial expansion state
  * Settings interface enhanced to track and manage section expansion state
  * Relevant sections automatically expand when blocks are added to them

### DIFF
--- a/apps/roam/src/components/settings/Settings.tsx
+++ b/apps/roam/src/components/settings/Settings.tsx
@@ -66,11 +66,13 @@ export const SettingsDialog = ({
   isOpen,
   onClose,
   selectedTabId,
+  expandedSectionUid,
 }: {
   onloadArgs: OnloadArgs;
   isOpen?: boolean;
   onClose?: () => void;
   selectedTabId?: TabId;
+  expandedSectionUid?: string;
 }) => {
   const extensionAPI = onloadArgs.extensionAPI;
   const settings = getFormattedConfigTree();
@@ -172,7 +174,11 @@ export const SettingsDialog = ({
             id="left-sidebar-personal-settings"
             title="Left sidebar"
             className="overflow-y-auto"
-            panel={<LeftSidebarPersonalSections />}
+            panel={
+              <LeftSidebarPersonalSections
+                expandedSectionUid={expandedSectionUid}
+              />
+            }
           />
           <SectionHeader className="text-lg font-semibold text-neutral-dark">
             Global Settings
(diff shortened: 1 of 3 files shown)


sid597:
@coderabbitai full review

apps/roam/src/components/settings/Settings.tsx
```diff
@@ -172,7 +174,11 @@ export const SettingsDialog = ({
             id="left-sidebar-personal-settings"
             title="Left sidebar"
             className="overflow-y-auto"
-            panel={<LeftSidebarPersonalSections />}
+            panel={
+              <LeftSidebarPersonalSections
+                expandedSectionUid={expandedSectionUid}
```
sid597, line 179: To directly open the panel where the uid was added

apps/roam/src/utils/registerCommandPaletteCommands.ts
```diff
@@ -301,4 +309,98 @@ export const registerCommandPaletteCommands = (onloadArgs: OnloadArgs) => {
...
+
+    for (const section of personalSections) {
+      if (!section.childrenUid) continue;
+
+      const sectionName = section.text.startsWith("((")
+        ? getTextByBlockUid(extractRef(section.text)) || section.text
+        : section.text;
+
+      window.roamAlphaAPI.ui.blockContextMenu.addCommand({
+        label: `DG: Left sidebar - Add to "${sectionName}" (personal)`,
```
mdroidian: Change to: `DG: Favorites - Add to "${sectionName}" section`

- change `(personal)` to `section
- change `Left sidebar` to `Favorites`

"Left sidebar" is ambiguous ([FEE-770: Rename "Left sidebar" to "Custom favorites" or something](https://linear.app/discourse-graphs/issue/FEE-770/rename-left-sidebar-to-custom-favorites-or-something)).  But "Favorites" is what the UI section is called, so that should be more clear.

Image

mdroidian (approved):
Just some thoughts: it might be unwanted and unnecessary to open the settings after this.  This is assuming a user wants to set an `alias`.  Also, it will likely be quite jarring, because there is no highlight/focus on the newly added block, so a user might not have any idea what they are really looking at or that something changed.  (And the default focus on the input box is a misdirection here too).

Personally, I would remove it, but I won't block on that.

apps/roam/src/utils/registerCommandPaletteCommands.ts
sid597: Re: opening settings, I think it is useful blocks generally much more content and the use cases for blocks in sidebar are more so for say daily notes, query block etc. and these blocks have content that does not convey intent so we would need alias. I think we should let it be like this until we get feedback on this but I think this is the right assumption to base this on for now, iirc Matt also mentioned in his notes and demo about the alias.

PR #907, opened 2026-03-22
[ENG-1545] Switch positions of node type and node title fields

trangdoan982:
https://youtu.be/X0MbEH5GjvU

## Summary
- **Reorder fields**: Content input now renders above Node Type selector, so users can immediately start typing a node title without picking a type first
- **Nullable Node Type**: Node Type dropdown starts empty when opened via command palette; queries all discourse node types in parallel and auto-detects the type when a node is selected
- **Lock Node Type on selection**: Node Type dropdown locks when a node is selected (unlocks when cleared)

## ENG-1316 regression safety
All focus fixes preserved: `autoFocus={!isContentLocked}` on Content, `autoFocus={false}` on Node Type Label, `popoverProps={{ openOnTargetFocus: false }}`, keyup handler in FuzzySelectInput, and confirm button `autoFocus={isContentLocked}` with key remount.

## Test plan
- [x] **Command palette flow:** Open modal → Content field focused, Node Type empty, all nodes shown → type to search → select a node → Node Type auto-populates and locks → Confirm focused
- [x] **Tab flow:** Content → Node Type → Source (if EVD) → Confirm
- [x] **Canvas flow:** Opens with pre-set nodeType → works as before (type not empty)
image

- [x] **ENG-1316 regression:** Select node via Enter → focus does NOT jump to Node Type dropdown
- [x] **Lock/unlock:** Select node → type locked → clear (X) → type unlocked, stays at detected value
- [x] **Evidence flow:** Select EVD type → Source field appears below Node Type
- [x] **New node creation:** Type a new name, select a type, confirm → node created correctly

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/907" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -505,6 +559,26 @@ const ModifyNodeDialog = ({
         style={{ pointerEvents: "all" }}
       >
         <div className={`${Classes.DIALOG_BODY} flex flex-col gap-4`}>
+          {/* Content Input */}
```
trangdoan982, line 543: [ENG-1556: Move Node Type dropdown below content input](https://linear.app/discourse-graphs/issue/ENG-1556/move-node-type-dropdown-below-content-input)

apps/roam/src/utils/registerCommandPaletteCommands.ts
```diff
@@ -189,19 +189,9 @@ export const registerCommandPaletteCommands = (onloadArgs: OnloadArgs) => {
...
-        id: "create-discourse-node-command-no-types",
-        content: "No discourse node types found in settings.",
-      });
-      return;
-    }
-
     renderModifyNodeDialog({
       mode: "create",
-      nodeType: defaultNodeType,
+      nodeType: "",
```
trangdoan982, line 194: [ENG-1554: Add support for empty Node Type dropdown](https://linear.app/discourse-graphs/issue/ENG-1554/add-support-for-empty-node-type-dropdown)

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -109,13 +109,16 @@ const ModifyNodeDialog = ({
       : allNodes.filter(excludeDefaultNodes);
   }, [includeDefaultNodes]);
 
-  const [selectedNodeType, setSelectedNodeType] = useState(() => {
+  const [selectedNodeType, setSelectedNodeType] = useState<
+    (typeof discourseNodes)[number] | null
```
trangdoan982: [ENG-1554: Add support for empty Node Type dropdown](https://linear.app/discourse-graphs/issue/ENG-1554/add-support-for-empty-node-type-dropdown)

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -224,11 +260,25 @@ const ModifyNodeDialog = ({
...
-  }, []);
+  }, [selectedNodeType, referencedNode, discourseNodes]);
+
+  const setValue = useCallback(
+    (r: Result) => {
+      setContent(r);
+      if (!selectedNodeType && r.uid) {
+        const detectedType = (r as Record<string, unknown>)
+          .discourseNodeType as string | undefined;
+        if (detectedType) {
```
trangdoan982, line 251: [ENG-1555: Lock Node Type dropdown when node is selected](https://linear.app/discourse-graphs/issue/ENG-1555/lock-node-type-dropdown-when-node-is-selected)

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -109,13 +109,16 @@ const ModifyNodeDialog = ({
       : allNodes.filter(excludeDefaultNodes);
   }, [includeDefaultNodes]);
 
-  const [selectedNodeType, setSelectedNodeType] = useState(() => {
+  const [selectedNodeType, setSelectedNodeType] = useState<
```
mdroidian, line 114: We have an explicit type for this: `DiscourseNode`.  Let's use that to be clear and direct.

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -109,13 +109,16 @@ const ModifyNodeDialog = ({
...
+    (typeof discourseNodes)[number] | null
+  >(() => {
+    if (!nodeType) return null;
```
mdroidian: Let's make `nodeType` optional if we are doing this.

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -160,6 +163,39 @@ const ModifyNodeDialog = ({
           if (contentRequestIdRef.current === req && alive) {
             setOptions((prev) => ({ ...prev, content: results }));
           }
+        } else {
+          // Query all discourse node types in parallel
+          const allResults = await Promise.all(
```
mdroidian: `discourse-graphs` has 20+ nodes, so this would send 20+ queries and hammer the server + take a very long time.  Let's replace this with the same pattern used in `getAllDiscourseNodesSince.ts:75-105` (extract into a reusable function called `getAllDiscourseNodeInstances` or similar

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -224,11 +260,25 @@ const ModifyNodeDialog = ({
...
-  const setValue = useCallback((r: Result) => {
-    setContent(r);
-  }, []);
+  }, [selectedNodeType, referencedNode, discourseNodes]);
+
+  const setValue = useCallback(
+    (r: Result) => {
+      setContent(r);
+      if (!selectedNodeType && r.uid) {
+        const detectedType = (r as Record<string, unknown>)
```
mdroidian: Is this assertion necessary?  It's already a `Result`

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -224,11 +260,25 @@ const ModifyNodeDialog = ({
...
+          .discourseNodeType as string | undefined;
```
mdroidian: Is this assertion necessary?

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -505,6 +559,26 @@ const ModifyNodeDialog = ({
         style={{ pointerEvents: "all" }}
       >
         <div className={`${Classes.DIALOG_BODY} flex flex-col gap-4`}>
+          {/* Content Input */}
+          <div className="w-full">
+            <Label>Content</Label>
```
mdroidian: This was pre-exisitng, but let's wrap the `FuzzySelectInput` with `Label` to match the styling of the Node Type Selector.  It's counter intuitive, but [matches blueprint](https://blueprintjs.com/docs/versions/3/#core/components/label)

apps/roam/src/components/ModifyNodeDialog.tsx
mdroidian: This was pre-exisitng, but let's wrap the `FuzzySelectInput` with `Label` to match the styling of the Node Type Selector.  It's counter intuitive, but [matches blueprint](https://blueprintjs.com/docs/versions/3/#core/components/label)

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -505,6 +559,26 @@ const ModifyNodeDialog = ({
...
+              }
+              mode={mode}
+              isLocked={isContentLocked}
+              autoFocus={!isContentLocked}
+            />
+          </div>
+
           {/* Node Type Selector */}
           <div className="flex w-full">
             <Label autoFocus={false}>
```
mdroidian: When tabbing to `Node Type`, it isn't showing the active state so a user doesn't know if they are on this field.

Image

mdroidian (changes requested):
Thank you for adding inline comments, but I'm not clear if this PR is solving all tickets, or these are intended to be some placeholder code.  Unfortunately the YouTube video was private, so I'm not sure if that question was answered there.

If this PR is solving all tickets, make sure to link the PR to the Linear tickets.

This biggest change request is querying for all nodes when no node type is selected.

trangdoan982:
@mdroidian this PR addresses all of the sub-tickets and the inline comments meant to point out the section that work on each sub-ticket. My bad I also made the video unlisted so you can see again. Same behavior except the focus now shows for the node type menu when user focuses there

Screenshot 2026-04-15 at 14 17 27

mdroidian:
> @mdroidian this PR addresses ...

Ok.  If this PR is ready for review again, please tag me as a reviewer/re-request review.

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -373,7 +408,7 @@ const ModifyNodeDialog = ({
         } else {
           formattedTitle = await getNewDiscourseNodeText({
             text: content.text.trim(),
-            nodeType: selectedNodeType.type,
+            nodeType: selectedNodeType!.type,
```
mdroidian: Avoid using `!` . It assumes `selectedNodeType` is always defined, but that invariant isn’t enforced and could break with upstream changes.

Either:

handle the null case (`?`. or fallback), or
assert explicitly with a guard/throw, or
tighten the type so selectedNodeType cannot be undefined at this point.

If you have to use `!` as a last resort, document why (either where typescript is failing to calculate properly, or the change to make it do so would be out of scope, etc)

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -384,7 +419,7 @@ const ModifyNodeDialog = ({
         // Create new discourse node
         const newPageUid = await createDiscourseNode({
           text: formattedTitle,
-          configPageUid: selectedNodeType.type,
+          configPageUid: selectedNodeType!.type,
```
mdroidian: Avoid using `!` . It assumes `selectedNodeType` is always defined, but that invariant isn’t enforced and could break with upstream changes.

Either:

handle the null case (`?`. or fallback), or
assert explicitly with a guard/throw, or
tighten the type so selectedNodeType cannot be undefined at this point.

If you have to use `!` as a last resort, document why (either where typescript is failing to calculate properly, or the change to make it do so would be out of scope, etc)

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -505,56 +540,68 @@ const ModifyNodeDialog = ({
...
+                  mode === "edit" || disableNodeTypeChange || isContentLocked
                 }
+                popoverProps={{ openOnTargetFocus: false }}
+                ButtonProps={{
+                  className:
+                    mode === "edit" || disableNodeTypeChange || isContentLocked
+                      ? "cursor-not-allowed opacity-50"
+                      : "",
+                  onFocus: (e) => {
+                    e.currentTarget.style.boxShadow = `0 0 0 2px ${Colors.BLUE3}, 0 0 0 4px ${Colors.BLUE3}4d`;
```
mdroidian: Could you elaborate on why we are adding custom styling to this button?  Should other buttons in dg follow this pattern?  What problem is this solving for "[ENG-1545] Switch positions of node type and node title fields"?

mdroidian (changes requested):
Generally speaking we want to avoid `!`—it bypasses null safety.

apps/roam/src/components/ModifyNodeDialog.tsx
trangdoan982: this change was to addressed your previous comment about the focus state for node type menu not showing.
This is generally because of how MenuItemSelect doesn't show focus status on the button (Query builder also has this issue). If we try setting it using `className`, it's added to the Select component and not the Button component. That's why I needed to add the style for onFocus in the ButtonProps. this style matches with Blueprint's CSS class `.bp3-button:focus`

mdroidian: I see.  It looks like this type of focus styling is pretty [standard for blueprint](<https://blueprintjs.com/docs/versions/3/#select/select-component>).



This indicates that something is not working as expected.  If so, we should document such a case and explain why we are using a non-standard fix.  Any idea what is causing it to not work in this case?<br><br>Also, could we just use `.bp3-button` then?  Or what is the reason that `.bp3-button:focus` isn't working?

trangdoan982: I haven't been able to find the root cause, but I think this might be Roam global css styling. Nothing in `<MenuItemSelect/>` or `<Select/>` would make the focus-ring behavior different from `<Input/>`. But I tried adding the <Select/> at various places in Roam and there's no focus either. This behavior also applies to wherever we used `<MenuItemSelect/>`, including Query builder, export modal. Even for native Roam components like the top left bar, when i change focus from the search bar -> buttons (they do have `.bp3-button` class), those buttons don't have focus-ring either. I can make a follow-up bug ticket to investigate this behavior but I think it's out of scope for this ticket.

mdroidian: @trangdoan982 <br>To find the root cause in the `css`, you can select an element via chrome dev tools



When we inspect the button vs a toggle in the roam settings, we can see the problem:



There is a global `*:focus { outline:none; }`<br><br>For a switch, blueprint has a greater specificity to override the generic `*` and apply an outline:



This is the pattern I would mimic in our `style.css` (including a reason for the addition) eg:

```
/* 
  Override button focus inside roamjs-canvas-dialog.

  Reason:
  - Blueprint defines a global `:focus` outline, but does NOT provide a
    button-specific focus rule with higher specificity.
  - If a global `outline: none` (or similar override) is applied elsewhere,
    buttons can lose their visible focus state.
  - This restores a proper focus outline for accessibility, scoped locally,
    matching Blueprint’s default focus styling intent.
*/
.roamjs-canvas-dialog .bp3-button:focus-visible {
  outline: 2px auto rgba(19,124,189,0.6);
  outline-offset: 2px;
}
```

trangdoan982: oh i missed that `focus-ring: none`. Your changes work I just pushed it up! thanks for helping me investigating this

mdroidian (approved):

PR #916, opened 2026-03-26
[ENG-1302] Fix bug canvas reload font style and size to "draw" and "m"

trangdoan982:
https://www.loom.com/share/59e9ecc209fc48bca4a22b817bbbf5a0

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/916" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


mdroidian (changes requested):
The problem that these two specific lines of code were fixing was that, if there was no size or no font family that is actually required in the new schema, the canvas won't load because it will fail on the schema check.

mdroidian (approved):

mdroidian:
> The problem that these two specific lines of code were fixing was that, if there was no size or no font family that is actually required in the new schema, the canvas won't load because it will fail on the schema check.

Discussed 2026-04-02, MG had 🧠💨

PR #928, opened 2026-03-30
ENG-797: Handle multiple canvases loaded at the same time

sid597:
https://www.loom.com/share/8af7d98fbab74f429a8c7a4e4982c260

Scope active canvas focus so clipboard, query builder actions, and shared context don't leak across simultaneously open canvases.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/928" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **Refactor**
  * Improved canvas focus and blur handling when working with multiple canvases
  * Enhanced action routing to support targeted canvas page selection
  * Disabled auto-focus behavior on Tldraw editor initialization

### DIFF
--- a/apps/roam/src/components/Export.tsx
+++ b/apps/roam/src/components/Export.tsx
@@ -65,7 +65,7 @@ import {
   createNodeShapeUtils,
   DiscourseNodeShape,
 } from "~/components/canvas/DiscourseNodeUtil";
-import { discourseContext, MAX_WIDTH } from "~/components/canvas/Tldraw";
+import { MAX_WIDTH } from "~/components/canvas/Tldraw";
 import internalError from "~/utils/internalError";
 import { getSetting, setSetting } from "~/utils/extensionSettings";
 import { isTLStoreSnapshot } from "./canvas/useRoamStore";
@@ -79,9 +79,7 @@ import {
   createAllReferencedNodeUtils,
 } from "./canvas/DiscourseRelationShape/DiscourseRelationUtil";
 import getDiscourseNodes from "~/utils/getDiscourseNodes";
-import getDiscourseRelations, {
-  DiscourseRelation,
-} from "~/utils/getDiscourseRelations";
+import getDiscourseRelations from "~/utils/getDiscourseRelations";
 import { AddReferencedNodeType } from "./canvas/DiscourseRelationShape/DiscourseRelationTool";
 import posthog from "posthog-js";
 
@@ -312,17 +310,6 @@ const ExportDialog: ExportDialogComponent = ({
     // TODO lots of this is reused in tldraw.tsx, use a function to avoid duplication
     if (!isTLStore) {
       const relations = getDiscourseRelations();
-      discourseContext.relations = relations.reduce(
-        (acc, r) => {
-          if (acc[r.label]) {
-            acc[r.label].push(r);
-          } else {
-            acc[r.label] = [r];
-          }
-          return acc;
-        },
-        {} as Record<string, DiscourseRelation[]>,
-      );
       const allRelations = relations;
       const allRelationIds = allRelations.map((r) => r.id);
       const allNodes = getDiscourseNodes(allRelations);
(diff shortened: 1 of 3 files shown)


sid597:
@coderabbitai full review

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -129,15 +129,38 @@ export type DiscourseContextType = {
...
   // { [Relation.Label] => DiscourseRelation[] }
   relations: Record<string, DiscourseRelation[]>;
-  lastAppEvent: string;
-  lastActions: HistoryEntry<TLRecord>[];
 };
 
 export const discourseContext: DiscourseContextType = {
   nodes: {},
   relations: {},
-  lastAppEvent: "",
```
sid597, line 139: Surface 1 handled

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -129,15 +129,38 @@ export type DiscourseContextType = {
...
-  lastActions: [],
+};
+
+let activeCanvasPageUid: string | null = null;
+let activeCanvasEditor: Editor | null = null;
+
+const setActiveCanvas = ({
```
sid597, line 146: Surface 3 handled, we track the active canvas and ensure only that canvas is focused, so only this canvas responds to the clipboard events

apps/roam/src/components/Export.tsx
```diff
@@ -312,17 +310,6 @@ const ExportDialog: ExportDialogComponent = ({
     // TODO lots of this is reused in tldraw.tsx, use a function to avoid duplication
     if (!isTLStore) {
       const relations = getDiscourseRelations();
-      discourseContext.relations = relations.reduce(
-        (acc, r) => {
```
sid597, line 316: Mutating this could effect the state for other open canvases??? we do set this up in tldraw.tsx line 455

This is for you Michael it seems correct to me but can give an authortative answer here

apps/roam/src/components/results-view/ResultsTable.tsx
```diff
@@ -229,7 +229,11 @@ const ResultRow = ({
       return (
         <Button
           {...buttonProps}
-          onClick={() => {
+          onClick={(event) => {
```
sid597, line 232: Pass this to the correct canvas

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -129,15 +129,38 @@ export type DiscourseContextType = {
...
   relations: Record<string, DiscourseRelation[]>;
-  lastAppEvent: string;
-  lastActions: HistoryEntry<TLRecord>[];
 };
 
 export const discourseContext: DiscourseContextType = {
   nodes: {},
   relations: {},
-  lastAppEvent: "",
-  lastActions: [],
```
sid597, line 140: needs to stay, because we use this to log on error

sid597, line 140: We use this to send log of last actions taken by user in case of error. Should not remove this.

apps/roam/src/components/Export.tsx
sid597: Needs to stay because this is a way to update the relations when we export to the canvas

mdroidian (changes requested):
As discussed in our meeting

Let's 
- remove the `useEffect` and do a closer solution to: https://tldraw.dev/examples/multiple
- re-add `lastActions` and make sure we test multiple canvas' against it
- re-add the export.tsx discourseContext.relations
- test `sending from Share Via in query builder only sends to the target canvas`
  - we tested live, seems fine
- we also noticed that `copy all` / `delete` was shared across canvas'
  - this is likely the same "focus" problem, but worth checking

mdroidian (approved):

PR #938, opened 2026-04-02
[ENG-1287] On canvas load, update key image

trangdoan982:
https://www.loom.com/share/1dc9445cdf5a4712af0864c579b99d10

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/938" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/roam/src/utils/syncCanvasNodesOnLoad.ts
+++ b/apps/roam/src/utils/syncCanvasNodesOnLoad.ts
@@ -3,10 +3,10 @@ import type { OnloadArgs } from "roamjs-components/types";
 import type { DiscourseNodeShape } from "~/components/canvas/DiscourseNodeUtil";
 import { DEFAULT_STYLE_PROPS } from "~/components/canvas/DiscourseNodeUtil";
 import { MAX_WIDTH } from "~/components/canvas/Tldraw";
-import {
-  extractFirstImageUrl,
+import type { TreeNode } from "roamjs-components/types";
+import calcCanvasNodeSizeAndImg, {
+  findFirstImage,
   getFirstImageByUid,
-  calcCanvasNodeDimensionsWithUrl,
 } from "./calcCanvasNodeSizeAndImg";
 import getDiscourseNodes from "./getDiscourseNodes";
 import resolveQueryBuilderRef from "./resolveQueryBuilderRef";
@@ -82,11 +82,11 @@ export const syncCanvasNodesOnLoad = async ({
   relationShapeTypeIds: string[];
   extensionAPI: OnloadArgs["extensionAPI"];
 }): Promise<void> => {
-  const { discourseNodeShapes, uidToTitle } = await syncCanvasNodeTitlesOnLoad(
+  const { discourseNodeShapes, uidToTitle } = await syncCanvasNodeTitlesOnLoad({
     editor,
     nodeTypeIds,
     relationShapeTypeIds,
-  );
+  });
   await syncCanvasKeyImagesOnLoad({
     editor,
     discourseNodeShapes,
@@ -94,11 +94,15 @@ export const syncCanvasNodesOnLoad = async ({
     extensionAPI,
   });
 };
-export const syncCanvasNodeTitlesOnLoad = async (
-  editor: Editor,
-  nodeTypeIds: string[],
-  relationShapeTypeIds: string[],
-): Promise<{
+export const syncCanvasNodeTitlesOnLoad = async ({
+  editor,
+  nodeTypeIds,
+  relationShapeTypeIds,
+}: {
+  editor: Editor;
+  nodeTypeIds: string[];
+  relationShapeTypeIds: string[];
+}): Promise<{
   discourseNodeShapes: DiscourseNodeShape[];
   uidToTitle: Map<string, string>;
 }> => {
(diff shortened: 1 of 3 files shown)


apps/roam/src/utils/syncCanvasNodesOnLoad.ts
```diff
@@ -104,4 +133,50 @@ export const syncCanvasNodeTitlesOnLoad = async (
...
+  const imageUpdates: {
+    id: DiscourseNodeShape["id"];
+    type: string;
+    props: { imageUrl: string; w: number; h: number };
+  }[] = [];
+
+  await Promise.all(
+    survivingShapes.map(async (shape) => {
+      const title = uidToTitle.get(shape.props.uid) ?? shape.props.title ?? "";
+      const { w, h, imageUrl } = await calcCanvasNodeSizeAndImg({
```
mdroidian: We probably don't want to run `calcCanvasNodeSizeAndImg` here if we're not using the `w` or `h`.  This will load each image to get the measurements for each node, which could take significant time based on the canvas.

mdroidian (changes requested):

apps/roam/src/utils/syncCanvasNodesOnLoad.ts
mdroidian: Need to update the comment to include images

apps/roam/src/utils/syncCanvasNodesOnLoad.ts
```diff
@@ -104,4 +135,66 @@ export const syncCanvasNodeTitlesOnLoad = async (
...
+    id: DiscourseNodeShape["id"];
+    type: string;
+    props: { imageUrl: string; w: number; h: number };
+  }[] = [];
+
+  // First pass: cheaply fetch imageUrls (no image loading) to find which shapes changed.
+  const urlResults = await Promise.all(
+    survivingShapes.map(async (shape) => {
+      const title = uidToTitle.get(shape.props.uid) ?? shape.props.title ?? "";
+      const imageUrl = await getCanvasNodeKeyImageUrl({
```
mdroidian: For each shape this is running `getNodeCanvasSettings` > `getDiscourseNodes()` > `getDiscourseRelations()`

Let's run `getDiscourseNodes()` once and get the canvas setting directly.

apps/roam/src/utils/syncCanvasNodesOnLoad.ts
```diff
@@ -104,4 +135,66 @@ export const syncCanvasNodeTitlesOnLoad = async (
...
+        nodeText: title,
+        uid: shape.props.uid,
+        nodeType: shape.type,
+        extensionAPI,
+      });
```
mdroidian, line 318: This is also running an individual query for each shape.  This could be hundreds of queries.

We'll have to write a custom query for any nodes that have the `key-image` option of `First image on page`.  So we can run one single query rather than a query for each node.  Those with `Query builder reference` will probably still be a `Promise.all`, unfortunately.

apps/roam/src/utils/syncCanvasNodesOnLoad.ts
```diff
@@ -104,4 +135,66 @@ export const syncCanvasNodeTitlesOnLoad = async (
...
+  );
+
+  const changedShapes = urlResults.filter(
+    ({ shape, imageUrl }) => (shape.props.imageUrl ?? "") !== imageUrl,
+  );
+
+  // Second pass: load images only for shapes whose URL changed, to compute new dimensions.
+  await Promise.all(
+    changedShapes.map(async ({ shape, title }) => {
+      const { w, h, imageUrl } = await calcCanvasNodeSizeAndImg({
```
mdroidian: Ideally we only run this when
- `imageUrl` was "" and is now not empty
- `imageUrl` was not empty and is now ""

This way we only need to load the images that we need to measure.

We should leave the `w`/`h` if just the image changes.  If the image dimensions change, display should have mostly already been taken care of with [ENG-1334: port over the key image resizing behavior from Obsidian](https://linear.app/discourse-graphs/issue/ENG-1334/port-over-the-key-image-resizing-behavior-from-obsidian)

apps/roam/src/utils/syncCanvasNodesOnLoad.ts
```diff
@@ -104,4 +135,66 @@ export const syncCanvasNodeTitlesOnLoad = async (
...
+        nodeText: title,
```
mdroidian: `calcCanvasNodeSizeAndImg` is running the query for `imageUrl` a second time here.  So for 100 shapes it would be 200 queries (if each shape changed).

As stated in the comment above, let's only  run `calcCanvasNodeSizeAndImage` when we need to set a new `w`, `h`.
And let's make sure we lass the `imageUrl` so we don't need to requery it.

mdroidian (changes requested):
### Summary

Functionally this direction makes sense, but **the main risk is how much Roam and the browser do on canvas open**. With **20+ discourse nodes that use key images**, the current flow can trigger **a large number of independent resolutions** (tree walks, query-builder runs, pulls, and image loads), not a small fixed cost. **The highest-impact follow-ups here should be performance-focused** before we lean on this in real graphs.

### Why this matters

- **`syncCanvasKeyImagesOnLoad`** runs **`getCanvasNodeKeyImageUrl` for every surviving node** in parallel (`Promise.all` over all shapes). Each call can be quite heavy.
- For nodes whose stored URL differs from the resolved URL, **`calcCanvasNodeSizeAndImg` calls `getCanvasNodeKeyImageUrl` again**, so **changed nodes pay for URL resolution twice** before dimensions are computed. That’s an easy **2×** on the hottest path when many URLs update at once (e.g. after settings or data changes).
- The second pass also **`loadImage`s in parallel** for every changed shape. That can be fine for a few images but **rough on large canvases** (many concurrent network decodes, main-thread work).

So the concern isn’t only “many shapes” — it’s **many shapes × (possibly expensive per-node Roam work + duplicate resolution + parallel image loads)**.

### Testing / acceptance (please block merge on this)

After the performance work lands, **verify on a large canvas**, not a 3-node fixture:

1. **Prefer an existing heavy canvas** in graph **`plugin-testing-akamatsulab2`**: open a page known to host a big tldraw canvas. Aim for **≥20 discourse node shapes with distinct key images** if possible.
2. **If no suitable page exists**, **create** a test page in that graph with **25+ nodes**, each with a **different** key image (exercises URL resolution + image load paths).
3. **Report concrete signals** in the PR : time-to-interactive or wall time until sync finishes, any UI jank, and whether Roam feels responsive. A before/after comparison on the same canvas is ideal.

mdroidian:
@trangdoan982 We should spend some time trying to optimize this query.  As we concluded in our meeting, this returning the entire tree for all nodes will not scale.<br><br>Have we tried only returning children with markdown images + block references?

PR #939, opened 2026-04-03
ENG-1580 Fix command palette availability

trangdoan982:
## Summary

- **Fix stale metadata cache**: `applyTemplate` was reading target file's frontmatter from `metadataCache` before computing the merge, but newly created files may not be indexed yet — causing template frontmatter to overwrite `nodeTypeId` and other fields. Fixed by computing the merge inside `processFrontMatter` where the live `fm` object is available.
- **Fix command palette availability**: `create-discourse-node` used `editorCallback`, which is only available when a markdown editor is active. Switched to `callback` so it's accessible from any view (canvas, graph, file explorer, etc.).
- **Simplify handler**: `createModifyNodeModalSubmitHandler` now accepts `editor?: Editor`, eliminating the duplicate inline fallback handler.

## Test plan

- [ ] Open command palette from a canvas view — "Create discourse node" should appear and work
- [ ] Open command palette from a markdown note — "Create discourse node" should work and insert `[[link]]` if creating from selected text
- [ ] Create a new discourse node with a template assigned — template frontmatter and body should be applied correctly without overwriting `nodeTypeId`
- [ ] Create a node via the `open-node-type-menu` command with a selection — template should still apply

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/939" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/obsidian/src/utils/registerCommands.ts
+++ b/apps/obsidian/src/utils/registerCommands.ts
@@ -26,7 +26,7 @@ type ModifyNodeSubmitParams = {
 
 const createModifyNodeModalSubmitHandler = (
   plugin: DiscourseGraphPlugin,
-  editor: Editor,
+  editor?: Editor,
 ): ((params: ModifyNodeSubmitParams) => Promise<void>) => {
   return async ({
     nodeType,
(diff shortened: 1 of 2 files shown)


trangdoan982:
some work done here for consistent command behavior but I think we can table this until we finish consolidating all the create node flows from Obsidian

PR #946, opened 2026-04-08
ENG-1617: se existing 'getConfigTree equivalent functions' for specific setting groups

sid597:
https://www.loom.com/share/3bed40fc5de943dcb4e7181888b42442

`getBlockPropBasedSettings` now does one Roam `pull` that returns the settings page's children with their string + uid + props in one shot, replacing the `q`-based `getBlockUidByTextOnPage` (~290ms per call) plus a second `pull` for props. `setBlockPropBasedSettings` reuses the same helper for the uid lookup so we have a single pull-and-walk pattern.

`SettingsDialog` captures a full settings snapshot once at mount via `useState(() => bulkReadSettings())` and threads `featureFlags` and `personalSettings` down to `HomePersonalSettings`. Each child component (`PersonalFlagPanel`, `NodeMenuTriggerComponent`,
`NodeSearchMenuTriggerSetting`, `KeyboardShortcutInput`) accepts an `initialValue` prop and seeds its local state from the snapshot instead of calling `getPersonalSetting` on mount. `PersonalFlagPanel`'s `initialValue` precedence flips so the prop wins when provided; `QuerySettings` callers without a prop still hit the existing fallback.

`getDiscourseNodes`, `getDiscourseRelations`, and `getAllRelations` narrow their snapshot parameter to `Pick<SettingsSnapshot, ...>` to declare which fields each function actually reads.

Adds a one-line `console.log` in `SettingsDialog` reporting the dialog open time, kept as an ongoing perf monitor.

## Release Notes

* **Refactor**
  * Restructured settings system architecture to use centralized settings snapshot instead of scattered getter functions, improving consistency and maintainability across settings components.

### DIFF
--- a/apps/roam/src/components/settings/Settings.tsx
+++ b/apps/roam/src/components/settings/Settings.tsx
@@ -28,6 +28,7 @@ import { FeedbackWidget } from "~/components/BirdEatsBugs";
 import { getVersionWithDate } from "~/utils/getVersion";
 import { LeftSidebarPersonalSections } from "./LeftSidebarPersonalSettings";
 import { LeftSidebarGlobalSections } from "./LeftSidebarGlobalSettings";
+import { bulkReadSettings } from "./utils/accessors";
 import posthog from "posthog-js";
 
 type SectionHeaderProps = {
@@ -72,6 +73,7 @@ export const SettingsDialog = ({
   onClose?: () => void;
   selectedTabId?: TabId;
 }) => {
+  const [mountStart] = useState(() => performance.now());
   const extensionAPI = onloadArgs.extensionAPI;
   const grammarNode = discourseConfigRef.tree.find(
     (node) => node.text === "grammar",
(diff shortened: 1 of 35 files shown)


sid597:
@coderabbitai full review

sid597:
@coderabbitai full review

apps/roam/src/components/settings/Settings.tsx
```diff
@@ -84,6 +85,8 @@ export const SettingsDialog = ({
   const [activeTabId, setActiveTabId] = useState<TabId>(
     selectedTabId ?? "discourse-graph-home-personal",
   );
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const snapshot = useMemo(() => bulkReadSettings(), [activeTabId]);
```
sid597: Rename this to settings instead of snapshot.

mdroidian (approved):

PR #967, opened 2026-04-17
ENG-906 Add node tag preview below tag input in settings

trangdoan982:
https://www.loom.com/share/7eaa1ccd091d4103885d2c2524b17154

## Summary
- Adds a live tag preview directly below the tag input field in node settings (General tab)
- Extracts tag styling logic into a shared `getNodeTagStyles` util to avoid duplication with the live Roam tag observer
- Refactors `initializeObserversAndListeners.ts` to use the shared util

## Test plan
- [x] Open node settings → General tab for a node with a tag set
- [x] Verify the tag preview appears below the input, styled as a colored pill matching the node's canvas color
- [x] Verify the preview updates live as you type
- [x] Verify with no canvas color set: preview shows unstyled (no color)
- [x] Verify live Roam tags (in blocks) still render with the same colored pill style
Screenshot 2026-04-17 at 13 11 13

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/967" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/apps/roam/src/utils/initializeObserversAndListeners.ts
+++ b/apps/roam/src/utils/initializeObserversAndListeners.ts
@@ -143,9 +143,8 @@ export const initObservers = async ({
           const normalizedNodeTag = node.tag ? getCleanTagText(node.tag) : "";
           if (normalizedTag === normalizedNodeTag) {
             renderNodeTagPopupButton(s, node, onloadArgs.extensionAPI);
-            const tagStyles = getNodeTagStyles(
-              node.canvasSettings?.color ?? "",
-            );
+            const color = node.canvasSettings?.color ?? "";
+            const tagStyles = color ? getNodeTagStyles(color) : {};
             if (tagStyles) {
               Object.assign(s.style, tagStyles);
             }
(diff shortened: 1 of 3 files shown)


apps/roam/src/utils/initializeObserversAndListeners.ts
```diff
@@ -145,29 +143,11 @@ export const initObservers = async ({
...
-                borderRadius: "12px",
-                margin: "0 2px",
-                fontSize: "0.9em",
-                whiteSpace: "nowrap",
-                boxShadow: "0 1px 2px rgba(0, 0, 0, 0.05)",
-                display: "inline-block",
-                cursor: "pointer",
-              });
+            const tagStyles = getNodeTagStyles(
+              node.canvasSettings?.color ?? "",
```
mdroidian: Minor regression: this now calls `getNodeTagStyles` even when there is no color.

It happens to be safe today because `formatHexColor` returns an empty string for falsy input,
so `getNodeTagStyles` ends up returning early indirectly. But that makes the call site rely on
helper internals rather than expressing the guard explicitly.

Prefer guarding on `color` before calling `getNodeTagStyles`.

You can either use an `if`, as the previous code did.  Or inline it
```typescript
const color = node.canvasSettings?.color ?? "";
const tagStyles = color ? getNodeTagStyles(color) : "";
```

mdroidian (approved):

PR #971, opened 2026-04-20
ENG-1640 Add custom sidebar commands, starting with Create Node

maparent:
https://linear.app/discourse-graphs/issue/ENG-1640/add-custom-sidebar-commands-starting-with-create-node

I responded to all comments; here is the updated loom.

https://www.loom.com/share/145b85a3792047fe9a7023eedd27e715

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/971" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/LeftSidebarCommands.tsx
+++ b/apps/roam/src/components/LeftSidebarCommands.tsx
@@ -4,9 +4,13 @@ import { OnloadArgs } from "roamjs-components/types";
 import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCommands";
 
 export const commands: Record<string, (ola: OnloadArgs) => Promise<void>> = {
+  /* eslint-disable @typescript-eslint/require-await */
+  // eslint-disable-next-line @typescript-eslint/naming-convention
   "Create Node": async (ola: OnloadArgs) => {
     createDiscourseNodeFromCommand(ola.extensionAPI);
+    // typescript-eslint/naming-convention
   },
+  /* eslint-enable @typescript-eslint/require-await */
 };
 
 export const sidebarCommandPopover = (onSelect: (value: string) => void) => {
--- a/apps/roam/src/components/LeftSidebarCommands.tsx
+++ b/apps/roam/src/components/LeftSidebarCommands.tsx
@@ -1,4 +1,4 @@
-import React, { MouseEvent } from "react";
+import React from "react";
 import { Popover, Position, Button, Menu, MenuItem } from "@blueprintjs/core";
 import { OnloadArgs } from "roamjs-components/types";
 import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCommands";
--- a/apps/roam/src/components/LeftSidebarCommands.tsx
+++ b/apps/roam/src/components/LeftSidebarCommands.tsx
@@ -25,7 +25,11 @@ export const commands: Record<
   /* eslint-enable @typescript-eslint/require-await */
 };
 
-export const sidebarCommandPopover = (onSelect: (value: string) => void) => {
+export const SidebarCommandPopover = ({
+  onSelect,
+}: {
+  onSelect: (value: string) => void;
+}) => {
   return (
     <Popover
       content={
--- a/apps/roam/src/components/LeftSidebarCommands.tsx
+++ b/apps/roam/src/components/LeftSidebarCommands.tsx
@@ -6,7 +6,9 @@ import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCo
 export const cleanCommandName = (name: string): string => {
   if (name.startsWith("{") && name.endsWith("}"))
     name = name.substring(1, name.length - 1);
-  // Should we make it title case as well?
+  name = name.trim();
+  // sentence case
+  name = name.charAt(0).toUpperCase() + name.slice(1);
   return name;
 };
 
--- a/apps/roam/src/components/LeftSidebarCommands.tsx
+++ b/apps/roam/src/components/LeftSidebarCommands.tsx
@@ -3,6 +3,13 @@ import { Popover, Position, Button, Menu, MenuItem } from "@blueprintjs/core";
 import { OnloadArgs } from "roamjs-components/types";
 import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCommands";
 
+export const cleanCommandName = (name: string): string => {
+  if (name.startsWith("{") && name.endsWith("}"))
+    name = name.substring(1, name.length - 1);
+  // Should we make it title case as well?
+  return name;
+};
+
 export const commands: Record<
   string,
   (onloadArgs: OnloadArgs) => Promise<void>
--- a/apps/roam/src/components/LeftSidebarCommands.tsx
+++ b/apps/roam/src/components/LeftSidebarCommands.tsx
@@ -3,11 +3,14 @@ import { Popover, Position, Button, Menu, MenuItem } from "@blueprintjs/core";
 import { OnloadArgs } from "roamjs-components/types";
 import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCommands";
 
-export const commands: Record<string, (ola: OnloadArgs) => Promise<void>> = {
+export const commands: Record<
+  string,
+  (onloadArgs: OnloadArgs) => Promise<void>
+> = {
   /* eslint-disable @typescript-eslint/require-await */
   // eslint-disable-next-line @typescript-eslint/naming-convention
-  "Create Node": async (ola: OnloadArgs) => {
-    createDiscourseNodeFromCommand(ola.extensionAPI);
+  "{create node}": async (onloadArgs: OnloadArgs) => {
+    createDiscourseNodeFromCommand(onloadArgs.extensionAPI);
     // typescript-eslint/naming-convention
   },
   /* eslint-enable @typescript-eslint/require-await */
--- a/apps/roam/src/components/LeftSidebarCommands.tsx
+++ b/apps/roam/src/components/LeftSidebarCommands.tsx
@@ -14,10 +14,6 @@ export const commands: Record<string, (ola: OnloadArgs) => Promise<void>> = {
 };
 
 export const sidebarCommandPopover = (onSelect: (value: string) => void) => {
-  const handleClick = (event: MouseEvent) => {
-    onSelect((event.target as Node).textContent!);
-  };
-
   return (
     <Popover
       content={
@@ -26,7 +22,7 @@ export const sidebarCommandPopover = (onSelect: (value: string) => void) => {
             <MenuItem
               key={commandName}
               text={commandName}
-              onClick={handleClick}
+              onClick={() => onSelect(commandName)}
             />
           ))}
         </Menu>
--- a/apps/roam/src/components/LeftSidebarCommands.tsx
+++ b/apps/roam/src/components/LeftSidebarCommands.tsx
@@ -1,3 +1,5 @@
+import React, { MouseEvent } from "react";
+import { Popover, Position, Button, Menu, MenuItem } from "@blueprintjs/core";
 import { OnloadArgs } from "roamjs-components/types";
 import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCommands";
 
@@ -5,4 +7,29 @@ export const commands: Record<string, (ola: OnloadArgs) => Promise<void>> = {
   "Create Node": async (ola: OnloadArgs) => {
     createDiscourseNodeFromCommand(ola.extensionAPI);
   },
+};
+
+export const sidebarCommandPopover = (onSelect: (value: string) => void) => {
+  const handleClick = (event: MouseEvent) => {
+    onSelect((event.target as Node).textContent!);
+  };
+
+  return (
+    <Popover
+      content={
+        <Menu>
+          {Object.keys(commands).map((commandName) => (
+            <MenuItem
+              key={commandName}
+              text={commandName}
+              onClick={handleClick}
+            />
+          ))}
+        </Menu>
+      }
+      position={Position.BOTTOM_LEFT}
+    >
+      <Button icon="cog" small minimal title="Commands" />
+    </Popover>
+  );
 };
--- a/apps/roam/src/components/LeftSidebarCommands.tsx
+++ b/apps/roam/src/components/LeftSidebarCommands.tsx
@@ -1,0 +1,8 @@
+import { OnloadArgs } from "roamjs-components/types";
+import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCommands";
+
+export const commands: Record<string, (ola: OnloadArgs) => Promise<void>> = {
+  "Create Node": async (ola: OnloadArgs) => {
+    createDiscourseNodeFromCommand(ola.extensionAPI);
+  },
+};
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -158,9 +158,9 @@ const SectionChildren = ({
           return void openTarget(e, child.text, onloadArgs);
         };
         return (
-          <div key={child.uid} className="pl-8 pr-2.5">
+          <div key={child.uid} className="bp3-dark pl-8 pr-2.5">
             {ref.type === "command" ? (
-              <Button onClick={onClick} minimal>
+              <Button onClick={onClick} minimal className="m-px">
                 {label}
               </Button>
             ) : (
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -8,6 +8,7 @@ import React, {
 } from "react";
 import ReactDOM from "react-dom";
 import {
+  Button,
   Collapse,
   Icon,
   Popover,
@@ -43,10 +44,13 @@ import { DISCOURSE_CONFIG_PAGE_TITLE } from "~/utils/renderNodeConfigPage";
 import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageUid";
 import { migrateLeftSidebarSettings } from "~/utils/migrateLeftSidebarSettings";
 import posthog from "posthog-js";
+import { commands } from "~/components/LeftSidebarCommands";
 
 const parseReference = (text: string) => {
   const extracted = extractRef(text);
-  if (text.startsWith("((") && text.endsWith("))")) {
+  if (commands[text]) {
+    return { type: "command" as const, uid: text, display: text };
+  } else if (text.startsWith("((") && text.endsWith("))")) {
     return { type: "block" as const, uid: extracted, display: text };
   } else {
     return { type: "page" as const, display: text };
@@ -58,7 +62,11 @@ const truncate = (s: string, max: number | undefined): string => {
   return s.length > max ? `${s.slice(0, max)}...` : s;
 };
 
-const openTarget = async (e: React.MouseEvent, targetUid: string) => {
+const openTarget = async (
+  e: React.MouseEvent,
+  targetUid: string,
+  onloadArgs: OnloadArgs,
+) => {
   e.preventDefault();
   e.stopPropagation();
   const target = parseReference(targetUid);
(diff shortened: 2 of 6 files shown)


apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx
```diff
@@ -256,13 +267,26 @@ const LeftSidebarGlobalSectionsContent = ({
     setNewPageInput(value);
   }, []);
 
+  const setPageValue = useCallback(
+    (value: string) => {
+      const input = document.getElementById(addPageInputId) as HTMLInputElement;
```
mdroidian: Why are we doing this imperatively instead of declaratively?  This is likely part of the problem.

apps/roam/src/components/settings/LeftSidebarPersonalSettings.tsx
```diff
@@ -338,6 +342,19 @@ const SectionItem = memo(
       }
     }, [childInput, section, addChildToSection]);
 
+    const setPageValue = useCallback(
+      (value: string) => {
+        const input = document.getElementById(
```
mdroidian: Same with this.

apps/roam/src/components/LeftSidebarCommands.tsx
```diff
@@ -0,0 +1,39 @@
...
+  "Create Node": async (ola: OnloadArgs) => {
+    createDiscourseNodeFromCommand(ola.extensionAPI);
+    // typescript-eslint/naming-convention
+  },
+  /* eslint-enable @typescript-eslint/require-await */
+};
+
+export const sidebarCommandPopover = (onSelect: (value: string) => void) => {
+  const handleClick = (event: MouseEvent) => {
+    onSelect((event.target as Node).textContent!);
```
mdroidian: Same with this, as graphite mentioned.

apps/roam/src/styles/styles.css
```diff
@@ -168,3 +168,17 @@
 .fuzzy-select-input-popover .bp3-popover-target {
   width: 100%;
 }
+
+.roam-sidebar-container .bp3-button:not([class*="bp3-intent-"]) {
```
mdroidian: FYI, one of the first things I will do if changes are submitted in `styles.css`, is to see if we can achieve this results without such changes.  Instead, we should try to achieve this result within the blueprint component itself, using tailwind wherever possible.

Take a look at https://blueprintjs.com/docs/versions/3/#core/components/button using dark theme

Image

Secondarily, we want to match existing Roam native components as much as possible.
Take note of Roam's settings itself, which is the same background color we are working with.

Image

apps/roam/src/utils/registerCommandPaletteCommands.ts
```diff
@@ -30,6 +30,76 @@ import { refreshAndNotify } from "~/components/LeftSidebarView";
 import { setPersonalSetting } from "~/components/settings/utils/accessors";
 import { sectionsToBlockProps } from "~/components/settings/LeftSidebarPersonalSettings";
 
+const getSelectionStartForBlock = (uid: string): number => {
```
mdroidian: Note: only move code when absolutely necessary.  It makes reading the diff much more difficult.
If moving code, please note `why` the code has been moved, and highlight any specific changes.

apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx
```diff
@@ -91,8 +95,14 @@ const LeftSidebarGlobalSectionsContent = ({
   const [autocompleteKey, setAutocompleteKey] = useState(0);
   const [isInitializing, setIsInitializing] = useState(true);
   const [isExpanded, setIsExpanded] = useState(true);
+  const addPageInputId = "globalAddPageId";
```
mdroidian: `const ADD_PAGE_INPUT_ID`

mdroidian (changes requested):
My initial impression is this might be an imperative-vs-declarative state sync issue in the input flow. I’ve left an inline note where I think it’s happening; could you take a pass at making that path consistently React-state-driven and confirm whether that resolves the stale/blank display behavior?

And as a quick test, I downloaded the transcript from the loom video and pasted it directly into cursor.  Cursor's response confirmed my intuition and even suggested changes for me. I think this would be a great opportunity for you to test using LLM tools.  Could you try explaining the problem to Cursor, Codex, or Claude to see if you can resolve it?

apps/roam/src/components/LeftSidebarCommands.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { MouseEvent } from "react";
+import { Popover, Position, Button, Menu, MenuItem } from "@blueprintjs/core";
+import { OnloadArgs } from "roamjs-components/types";
+import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCommands";
+
+export const commands: Record<string, (ola: OnloadArgs) => Promise<void>> = {
+  /* eslint-disable @typescript-eslint/require-await */
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  "Create Node": async (ola: OnloadArgs) => {
```
mdroidian: change to `{create node}`, as it is stated in the task.

apps/roam/src/components/LeftSidebarCommands.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { MouseEvent } from "react";
+import { Popover, Position, Button, Menu, MenuItem } from "@blueprintjs/core";
+import { OnloadArgs } from "roamjs-components/types";
+import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCommands";
+
+export const commands: Record<string, (ola: OnloadArgs) => Promise<void>> = {
```
mdroidian: lets stick to `onloadArgs` instead of `ola`, as it is used like this dozens of times in the repo already.

apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx
maparent: Because I tried declaratively and it did not work either; I did not know about the autocompleteKey workaround for AutocompleteInput. And yes, Claude code found that out for me, that was good advice. This is what it said:
```⏺ Found the root cause. AutocompleteInput has its own internal query state initialized once from value prop — it never syncs with external value prop changes after mount (line 12: `useState(() =>  itemToQuery(value)))`. It renders from query, not from value (line 106). So input.value = value gets overwritten by React reconciliation, and updating parent state with
  setNewPageInput/setChildInput has no effect on the display.```
I wonder if that could be corrected?

Comment (mdroidian) [reply apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx]: Can we not pass in the initial value?

Comment (maparent) [reply apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx]: Not sure what you mean here? Let me clarify on my side: The workaround which you use, and which I adopted in latest commit and solves the problem for now, is to increment the key of the AutocompleteInput whenever you set the value. My question was whether that workaround was needed. Maybe it is, I can vaguely see why, but it would be nice to just be able to just set the state and have the component adjust.

Comment (mdroidian) [reply apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx]: > Not sure what you mean here?

Most react components (especially from 3rd party libraries) have a way to set the initial value.
I haven't looked myself recently, but I'd assume it would be the case for `<AutocompleteInput>` as well.
So if the problem is that "AutocompleteInput has its own internal query state initialized once from value prop", we should be able to set that `value` prop directly.

Can we not pass the initial value to `<AutocompleteInput>`?

Comment (maparent) [reply apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx]: We can pass the initial value from a state variable, but when we update the state, the `AutocompleteInput` does not update accordingly, because the initial value of the state is cached in the internal state of the component. Changing the key is a workaround so we can both update the state and obtain essentially a new input instance with the updated state as a new initial value.

File: apps/roam/src/styles/styles.css (hunk 4 above)
Comment (maparent) [reply apps/roam/src/styles/styles.css]: Ok. I added the dark style, which was totally unlike the roam style; I put the roam style by hand in the css.

File: apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx (hunk 6 above)
Comment (maparent) [reply apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx]: obsolete.

File: apps/roam/src/utils/registerCommandPaletteCommands.ts (hunk 5 above)
Comment (maparent) [reply apps/roam/src/utils/registerCommandPaletteCommands.ts]: This was to make this function exportable. I think that was necessary.

File: apps/roam/src/components/LeftSidebarCommands.tsx (hunk 7 above)
Comment (maparent) [reply apps/roam/src/components/LeftSidebarCommands.tsx]: Should I shave off the `{}` while creating the button then? I think it's vaguely acceptable in the menu, horrible in the button.

Comment (mdroidian) [reply apps/roam/src/components/LeftSidebarCommands.tsx]: Sure, the button label can be "Create Node", but the menu item should be `{create node}`

File: apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx (hunk 1 above)
Comment (mdroidian) [reply apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx]: This shouldn't be the case, whenever a prop changes the state is refreshed and component is re-rendered.  Changing the `value` prop passed into `<Autocompleteinput>` will cause it to re-render with the new value.<br><br>[https://react.dev/reference/react/useState](<https://react.dev/reference/react/useState>)<br>"React will store the next state, render your component again with the new values, and update the UI."<br><br>Something else is play here.

**Comment by maparent** [comment]
Done. Right now I'm bothered by the fact that personal settings, unlike global settings, allow duplicates; but that should be a separate PR.

File: apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx (hunk 1 above)
Comment (maparent) [reply apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx]: It should indeed not be the case; I say that's a known issue with AutocompleteInput (known since the key workaround is used extensively in the code.)
The question is: is it something we want to solve?

File: apps/roam/src/components/LeftSidebarCommands.tsx (hunk 7 above)
Comment (maparent) [reply apps/roam/src/components/LeftSidebarCommands.tsx]: I removed the `{}` but did not use title case. Could be done in CSS or in JS, lmk what you think. There exists a title-case package that will handle prepositions correctly.

File: apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx (hunk 1 above)
Comment (maparent) [reply apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx]: https://www.loom.com/share/1d4bd6b4f2704f88ac792d76ca999275

File: apps/roam/src/components/LeftSidebarCommands.tsx [hunk 9]
```diff
@@ -0,0 +1,45 @@
...
+  /* eslint-disable @typescript-eslint/require-await */
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  "{create node}": async (onloadArgs: OnloadArgs) => {
+    createDiscourseNodeFromCommand(onloadArgs.extensionAPI);
+    // typescript-eslint/naming-convention
+  },
+  /* eslint-enable @typescript-eslint/require-await */
+};
+
+export const sidebarCommandPopover = (onSelect: (value: string) => void) => {
```
mdroidian: Prefer ({ onSelect }: { onSelect: … }) + PascalCase: React passes a props object, not a bare callback, and JSX expects PascalCase for custom components.

```suggestion
export const SidebarCommandPopover = ({
  onSelect,
}: {
  onSelect: (value: string) => void;
}) => {
```

apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx
mdroidian: So I reproed the issue and dug into the code.  The concise problem is that `AutocompleteInput` is rendering the `<Input>` from its internal `query` value state, not directly from the external `value` prop.  Claude was basically correct, but the initial imperative `input.value` was a very strong red herring, which masked the true problem.

Given that, resetting the `key` is acceptable as a workaround here, but we should make it explicit and document it in the code. I’d suggest wrapping it in something like `resetAutocomplete(value)` so it’s clear that we are intentionally remounting `AutocompleteInput` to refresh its internal query state.

```typescript
const resetAutocomplete = useCallback((nextValue = "") => {
  setNewPageInput(nextValue);

  // AutocompleteInput renders from its internal `query` state, which is only
  // initialized from the external `value` prop on mount. Bump the key to remount
  // it so the displayed input reflects the new parent state.
  setAutocompleteKey((prev) => prev + 1);
}, []);
```

mdroidian: > The question is: is it something we want to solve?

`<AutocompleteInput>` has had a couple behaviors we've needed to adjust, and Trang already created a slightly different version.  I do think it is now worth looking to creating a dg specific version.  I'll create a task for this.

apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx
```diff
@@ -256,13 +266,19 @@ const LeftSidebarGlobalSectionsContent = ({
     setNewPageInput(value);
   }, []);
 
+  const setPageValue = useCallback((value: string) => {
```
mdroidian: Use suggested `resetAutocomplete`

apps/roam/src/styles/styles.css
```diff
@@ -168,3 +168,14 @@
 .fuzzy-select-input-popover .bp3-popover-target {
   width: 100%;
 }
+
+.roam-sidebar-container .bp3-dark .bp3-button:not([class*="bp3-intent-"]) {
```
mdroidian: What problem is this css solving?

apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx
```diff
@@ -329,13 +345,13 @@ const LeftSidebarGlobalSectionsContent = ({
             <div className="mb-2 text-sm text-gray-600">
               Add pages that will appear for all users
             </div>
-            <div className="mb-3 flex items-center gap-2">
+            <div className="sidebarCommandPopoverContainer mb-3 flex items-center gap-2">
```
mdroidian: `sidebarCommandPopoverContainer` is unused.  Remove.

apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx
```diff
@@ -348,6 +364,7 @@ const LeftSidebarGlobalSectionsContent = ({
                 onClick={() => void addPage(newPageInput)}
                 title="Add page"
               />
+              {sidebarCommandPopover(setPageValue)}
```
mdroidian: `<SidebarCommandPopover onSelect={setPageValue}>`

apps/roam/src/components/settings/LeftSidebarPersonalSettings.tsx
```diff
@@ -432,6 +441,7 @@ const SectionItem = memo(
                   onClick={() => void handleAddChild()}
                   title="Add child"
                 />
+                {sidebarCommandPopover(setPageValue)}
```
mdroidian: `<SidebarCommandPopover onSelect={setPageValue}>`

apps/roam/src/components/settings/LeftSidebarPersonalSettings.tsx
```diff
@@ -338,6 +342,11 @@ const SectionItem = memo(
       }
     }, [childInput, section, addChildToSection]);
 
+    const setPageValue = useCallback((value: string) => {
```
mdroidian: Use suggested `resetAutocomplete`

apps/roam/src/components/LeftSidebarCommands.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Popover, Position, Button, Menu, MenuItem } from "@blueprintjs/core";
+import { OnloadArgs } from "roamjs-components/types";
+import { createDiscourseNodeFromCommand } from "~/utils/registerCommandPaletteCommands";
+
+export const cleanCommandName = (name: string): string => {
+  if (name.startsWith("{") && name.endsWith("}"))
+    name = name.substring(1, name.length - 1);
+  // Should we make it title case as well?
+  return name;
+};
```
mdroidian, line 13: Sentence case.  We could either store another `label` in `commands` or just capitalize the first letter.

apps/roam/src/components/LeftSidebarView.tsx
```diff
@@ -134,23 +148,31 @@ const SectionChildren = ({
...
-          <div key={child.uid} className="pl-8 pr-2.5">
-            <div
-              className={
-                "section-child-item page cursor-pointer rounded-sm leading-normal text-gray-600"
-              }
-              onClick={onClick}
-            >
-              {label}
-            </div>
+          <div key={child.uid} className="bp3-dark pl-8 pr-2.5">
```
mdroidian: We want the button to use the dark theme, not the surrounding div.  This affects all entries quite negatively.

Before:

Image

After:

Image

Doing a loom video before submitting for review is a good way to catch these regressions.

apps/roam/src/utils/registerCommandPaletteCommands.ts
```diff
@@ -30,6 +30,97 @@ import { refreshAndNotify } from "~/components/LeftSidebarView";
 import { setPersonalSetting } from "~/components/settings/utils/accessors";
 import { sectionsToBlockProps } from "~/components/settings/LeftSidebarPersonalSettings";
 
+type BlockSelection = {
```
mdroidian, line 33: As this is a large chunk of moved code, please highlight the specific changes with inline github comments.

mdroidian (changes requested):
Could you please update the body of the PR to reflect it's current state?
Could you also please include a loom video demonstrating the changes?

Thank you.

apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx
mdroidian: https://github.com/DiscourseGraphs/discourse-graph/pull/971#discussion_r3134335517

apps/roam/src/components/settings/LeftSidebarPersonalSettings.tsx
mdroidian: https://github.com/DiscourseGraphs/discourse-graph/pull/971#discussion_r3134335517

apps/roam/src/components/LeftSidebarCommands.tsx
mdroidian: Let's do sentence case, no package should be required.

apps/roam/src/utils/registerCommandPaletteCommands.ts
```diff
@@ -30,6 +30,97 @@ import { refreshAndNotify } from "~/components/LeftSidebarView";
...
+  }
+  const textLength = (getTextByBlockUid(uid) || "").length;
+  return {
+    selectionStart: textLength,
+    selectionEnd: textLength,
+    selectedText: "",
+  };
+};
+
+export const createDiscourseNodeFromCommand = (
```
mdroidian, line 77: When moving large chunks of code, please add inline github comments to highlight which piece(s) of the code were changed.  This was asked twice but still not done.

So I went through and compared this code line by line.

Here is an example of what is expected next time:

> This is the only line changed.  All other code in this change block is from moving the code so that it could be exported.  No other substantive changes.

apps/roam/src/components/settings/LeftSidebarGlobalSettings.tsx
```diff
@@ -92,7 +96,11 @@ const LeftSidebarGlobalSectionsContent = ({
   const [isInitializing, setIsInitializing] = useState(true);
   const [isExpanded, setIsExpanded] = useState(true);
 
-  const pageNames = useMemo(() => getAllPageNames(), []);
+  const commandNames = useMemo(() => Object.keys(commands), []);
+  const pageAndCommandNames = useMemo(
+    () => [...getAllPageNames(), ...commandNames],
```
mdroidian: Add `commandNames` first, as per ticket specs:

> {create node} appears in the supported custom-command autocomplete/options (**as the first option**).

apps/roam/src/components/settings/LeftSidebarPersonalSettings.tsx
```diff
@@ -693,6 +706,10 @@ const LeftSidebarPersonalSectionsContent = ({
   }, [sections, settingsDialogSectionUid]);
 
   const pageNames = useMemo(() => getAllPageNames(), []);
+  const pageAndCommandNames = useMemo(
+    () => [...pageNames, ...Object.keys(commands)],
```
mdroidian: Add `commands` first as per ticket specs:

> {create node} appears in the supported custom-command autocomplete/options (**as the first option**).

mdroidian (approved):
Let's make sure the `{create node}` command appears as the first command before merging.

PR #986, opened 2026-04-28
ENG-1615: Only add backlink when text is selected in "Create discourse node" command (Obsidian)

trangdoan982:
https://www.loom.com/share/a3267f6e354d4b3e9c61ab32758d7bd9

## Summary

- `createModifyNodeModalSubmitHandler` now takes `editor` as optional — only passed when an active editor is available
- Backlink (`editor.replaceSelection`) for linking to an existing node is now gated on `editor.somethingSelected()`
- `create-discourse-node` command switched from `editorCallback` to `callback`, resolving the case where the command is invoked from outside an active editor (editor becomes `undefined`)
- New node creation path already conditionally used `editor` via `createDiscourseNode` — no change needed there

Fixes: https://linear.app/discourse-graphs/issue/ENG-1615

## Test plan

- [x] With no text selected, run "Create discourse node" → node is created, no backlink inserted at cursor
- [x] With no text selected, run "Create discourse node" and link to an existing node → node linked, no backlink inserted
- [x] With text selected, run "Create discourse node" → node is created and backlink replaces the selected text
- [x] Run "Create discourse node" from outside any editor (e.g. file explorer focused) → modal opens, node is created, no crash

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/986" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/obsidian/src/utils/createNode.ts
+++ b/apps/obsidian/src/utils/createNode.ts
@@ -134,7 +134,7 @@ export const createDiscourseNode = async ({
     nodeType,
   });
 
-  if (newFile && editor) {
+  if (newFile && editor && editor.somethingSelected()) {
     editor.replaceSelection(`[[${formattedNodeName}]]`);
   }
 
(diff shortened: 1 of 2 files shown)


maparent (changes requested):
This looks good in principle, but I thought of an edge case:
What if the existing selection already has [[square brackets]], because the user added them preemptively?
I think we should test for that. Feel free to push back and decide it's scope creep, but I'd like it on record as a decision either way.

ETA: I tested, and this gets rejected as having invalid characters. Good enough.

maparent (approved):
Approved on second thought.

PR #989, opened 2026-04-30
ENG-1642 Discourse context overlay gets re-triggered every hover in a query block when view is render

maparent:
https://linear.app/discourse-graphs/issue/ENG-1642/discourse-context-overlay-gets-re-triggered-every-hover-in-a-query

https://www.loom.com/share/b8b7fcfe79334491946a95b2f0530103

Diagnostic: The RenderRoamBlockString inside CellRender seems to use useEffects to clear and re-render its contents. So when React asks CellRender to recreate the virtual DOM, in particular on every mouseEnter/mouseLeave to the resultsTable, the effects re-run, and the content is re-rendered from scratch. The solution provided by Claude was to apply React.memo on the CellRender,ResultRow,ResultTable stack, so these are not re-rendered when the props are the same. We had to make sure each prop was wrapped in either useMemo or useCallback.

Note: This induced a lot of whitespace changes. The only effective changes are introducing React.memo on those three components.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/989" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Refactor**
  * Optimized results table component rendering for improved performance and responsiveness.

### DIFF
--- a/apps/roam/src/components/results-view/ResultsTable.tsx
+++ b/apps/roam/src/components/results-view/ResultsTable.tsx
@@ -4,7 +4,10 @@ import React, {
   useEffect,
   useState,
   useCallback,
+  memo,
 } from "react";
+import ReactDOM from "react-dom";
+import nanoid from "nanoid";
 import { Button, HTMLTable, Icon, IconName } from "@blueprintjs/core";
 import { IconNames } from "@blueprintjs/icons";
 import { render as renderToast } from "roamjs-components/components/Toast";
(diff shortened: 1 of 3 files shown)


mdroidian:
@CodeRabbitAI full review

mdroidian:
@CodeRabbitai full review

maparent:
@mdroidian Just checking: I think you wanted me to look into why embeddings were not affected, correct? Or did you intend to do it as part of review?

mdroidian:
> @mdroidian Just checking: I think you wanted me to look into why embeddings were not affected, correct? Or did you intend to do it as part of review?

I can take a look.  Memoizing `ResultRow` and especially all of `ResultsTable` seems like overkill when `View Type` > `embed` doesn't have this same problem.  I think it's worth investigating to see if we can identify a more focused fix.

mdroidian (changes requested):
I think the memoization fix is treating the symptom rather than the root cause.

The overlay is being retriggered because hovering the query results currently updates React state in `ResultsView` (`revealMenuIcons`), which rerenders the results table. In `render` view, that means Roam’s `BlockString` output can be recreated, which causes the page-ref observer / discourse  overlay path to run again.

Memoizing `ResultsTable`/`ResultRow` can suppress that rerender in this case, but it is a broad workaround around a hover-only UI concern. It also risks hiding future legitimate updates because the whole table/row rendering path becomes dependent on memo boundaries.

A narrower fix is to remove the hover state entirely and let CSS reveal the query controls on hover. That keeps the same UI behavior without causing React to rerender the query results table every time the mouse enters or leaves the query block.

I have shown that here:
https://github.com/DiscourseGraphs/discourse-graph/pull/1006
https://www.loom.com/share/764c12c0caa948d2816c5f9c1e5ea74d

maparent:
That makes sense. Since you did write the fix, do you want to take this over?

mdroidian:
@coderabbitai full review

PR #992, opened 2026-05-01
ENG-1687 Memory leak: Unmounting the DiscourseContext Overlay leaves it dangling in Roam components

maparent:
https://linear.app/discourse-graphs/issue/ENG-1687/memory-leak-unmounting-the-discoursecontext-overlay-leaves-it-dangling

Problem: https://www.loom.com/share/b7207d93abd942839133327de691aa86

Demonstrating the solution: https://www.loom.com/share/5d6a22757ca24905b44f921b87faa9ce

Again the diagnosis is mine but the solution is Claude's.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory management for overlay components to ensure proper cleanup of disconnected DOM elements, enhancing stability and performance.

### DIFF
--- a/apps/roam/src/components/DiscourseContextOverlay.tsx
+++ b/apps/roam/src/components/DiscourseContextOverlay.tsx
@@ -366,6 +366,26 @@ const Wrapper = ({ parent, tag }: { parent: HTMLElement; tag: string }) => {
   );
 };
 
+const trackedContainers = new Set<HTMLElement>();
+let cleanupObserver: MutationObserver | null = null;
+
+const ensureObserver = () => {
+  if (cleanupObserver) return;
+  cleanupObserver = new MutationObserver(() => {
+    for (const el of trackedContainers) {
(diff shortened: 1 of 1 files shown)


mdroidian:
@CodeRabbitAI full review

mdroidian (approved):
Approving this direction. Tracking the overlay containers and unmounting when Roam removes them is the right fix for the leak.

One preference: instead of adding a local `trackedContainers` + `MutationObserver`, I’d lean on `roamjs-components/util/renderWithUnmount` here. We  already use it elsewhere in the extension, and it gives us the same DOM-removal cleanup while also registering/unregistering the React root and observer with the RoamJS extension registry. It also returns an explicit unmount function, which is useful when we intentionally remove overlays, e.g. when the setting is toggled off.

So I’m good with the lifecycle fix, but I’d prefer implementing it via `renderWithUnmount` to avoid maintaining a second cleanup mechanism.

PR #993, opened 2026-05-01
ENG-1377: Sync templates when syncing a node schema

trangdoan982:
https://www.loom.com/share/4415ec83680545b180fa2c1e8dc7cabc

## Summary

- When syncing node schemas to Supabase, include the template file content (`template_content`) so it can be shared across vaults
- When importing a node schema that carries `template_content`, automatically create the template file in the local Templates plugin folder
- Respects Templates plugin state: skips template creation (and clears the `template` reference) if the plugin is disabled or the folder path is not configured
- Does not overwrite existing local templates — the local file takes precedence

## Test plan

- [x] Sync a node type that has a template → verify `template_content` is sent to Supabase
- [x] Import a node schema from another vault with `template_content` → verify the template file is created locally
- [x] Import with Templates plugin disabled → verify notice is shown and `template` field is not set
- [x] Import with Templates plugin enabled but folder not configured → same behavior as above
- [x] Import when template file already exists → verify local file is kept, no overwrite

### DIFF
--- a/apps/obsidian/src/utils/conceptConversion.ts
+++ b/apps/obsidian/src/utils/conceptConversion.ts
@@ -35,10 +35,12 @@ export const discourseNodeSchemaToLocalConcept = ({
   context,
   node,
   accountLocalId,
+  templateContent,
 }: {
   context: SupabaseContext;
   node: DiscourseNode;
   accountLocalId: string;
+  templateContent?: string;
 }): LocalConceptDataInput => {
   const {
     description,
@@ -56,6 +58,7 @@ export const discourseNodeSchemaToLocalConcept = ({
     source_data: otherData,
   };
   if (template) literal_content.template = template;
+  if (templateContent) literal_content.template_content = templateContent;
   if (importedFromRid) literal_content.importedFromRid = importedFromRid;
   return {
     space_id: context.spaceId,
--- a/apps/obsidian/src/utils/conceptConversion.ts
+++ b/apps/obsidian/src/utils/conceptConversion.ts
@@ -34,12 +34,10 @@ const getNodeExtraData = (
 export const discourseNodeSchemaToLocalConcept = ({
   context,
   node,
-  accountLocalId,
   templateContent,
 }: {
   context: SupabaseContext;
   node: DiscourseNode;
-  accountLocalId: string;
   templateContent?: string;
 }): LocalConceptDataInput => {
   const {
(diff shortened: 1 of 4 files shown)


maparent (changes requested):
All that is there is good, but we miss a migration aspect. If the Concept has already been put in the database without templates, you need to modify the node schema somehow to ensure the template is uploaded. But there's no clear way to do that, or to know that the problem exists. I see three options:
1. Create a force sync command that wipes existing data for the space. 
2. Add a version number in metadata, and wipe any data that is older than the current version. 
3. If we know some node schema have templates, always query those Concept's metadata at sync to see if they need updating.

1 and 2 would be separate tasks. 3. is a bit costly, but probably what we'll end up doing short-term. But would be happy to discuss options.

Btw, I would have made the template into content... but that may not even be the best path. I think putting it in the metadata was a good call.

trangdoan982:
@maparent can you elaborate more on how option #2 might be useful to know that template sync problem exist?

I'm thinking that there's also a problem that template file is modified, but it won't show up in the `modifed` field of nodeType in data.json -> create a discrepancy regardless. so maybe we'll have to go with option 3? but this requires the source repo to be online at the time of query as well to enable sync.

I also don't feel comfortable with force-sync command because from UX perspective you would only want 1 command to sync everything.

Maybe we can just sync all schemas with force (lower cost), and keeping the same mechanism of sync based on modifedAt for other concepts (eg: node instances, relation instances)?

maparent:
Option 2: Any concept that does not have a metadata version field in its literal_content is considered at version 0. From now on, every concept's literal_content will have a "mv" (for "metadata_version") field, with an integer. (We could even add a small int column for this, make 0 the default.) We keep the current version in the code. Any Concept that has a version number lower than the current needs a migration.

Good catch re the template file, but we can easily add that to the computation of last_modified of the field at sync time. Not sure it makes a difference between options 1,2,3? <br>Agree 1 is the least preferred path.<br>Happy to do the table change if you prefer 2.

maparent:
(Could have a chat later today or tomorrow morning if you want)

maparent:
Thinking… can we somehow close this task and make the migration distinct? I think if we add the column or use a json field, you can set it to 1 within this task, and we can separate out the migration task for old values that have the implicit no-version.

trangdoan982:
yes that's what i was proposing =)) I'm down for this solution

maparent:
Good. So let's add the json field, and still expand the last_modified computation to include the template.

maparent:
Decision in meeting: Use "mv" json field now; and at every sync, look at last_modified of all templates, but the latter is a separate task.

maparent (approved):
LGTM

PR #995, opened 2026-05-01
ENG-1550: Fix # fragment encoding in markdown links and wikilinks during import

trangdoan982:
https://www.loom.com/share/a00c76ac173b44be803a0917fc1d02c9

## Summary

- **`encodePathForMarkdownLink`**: Split on the first `#` before percent-encoding path segments, so heading/block fragments (e.g. `#section`, `#^block-id`) are preserved as `#section` instead of being encoded as `%23section`
- **`processLink`**: Strip the `#fragment` before all file-resolution logic, then re-append it to the returned path — previously, successfully resolved wikilinks (e.g. `[[Note#Section]]`) would silently drop the heading reference
- **Wikilink `.md` stripping**: The `.endsWith(".md")` check now operates only on the path portion before `#`, so paths like `import/space/Note.md#Section` correctly strip the `.md` extension

## Root cause

`encodeURIComponent("#") === "%23"`, so the old `encodePathForMarkdownLink` turned `[text](Note.md#section)` into `[text](Note.md%23section)`. Obsidian interprets the latter as pointing to a file literally named `Note.md#section` rather than the `Section` heading inside `Note.md`.

The secondary bug in `processLink` passed the full `Note.md#section` string to `app.metadataCache.getFirstLinkpathDest`, which can't resolve a fragment-appended path, and returned a bare file path — discarding the fragment entirely.

## Test plan

- [x] Import a note with `[text](Note.md#section)` — verify the link renders as `[text](Note.md#section)` (relative equivalent), not `[text](Note.md%23section)`
- [x] Import a note with `[text](Note.md#^block-id)` — verify block reference is preserved
- [x] Import a note with `[[Note#Section]]` — verify section reference is preserved in the output wikilink
- [x] Normal links without `#` still work: `[text](Note.md)`, `[text](folder/Note.md)`
- [x] Spaces still encoded: `[text](Note With Spaces.md)` → `[text](Note%20With%20Spaces.md)`
- [x] External URLs pass through unchanged: `[text](https://example.com)`

### DIFF
--- a/apps/obsidian/src/utils/importNodes.ts
+++ b/apps/obsidian/src/utils/importNodes.ts
@@ -604,51 +604,59 @@ const updateMarkdownAssetLinks = ({
       return linkPath;
     }
 
-    // First, try to find if this link resolves to one of our imported assets
-    const importedAssetFile = findImportedAssetFile(linkPath);
-    if (importedAssetFile) {
-      return getRelativeLinkPath(importedAssetFile.path);
-    }
+    // Separate file path from heading/block fragment (e.g. "Note.md#section" → filePath="Note.md", fragment="#section")
+    // so that file resolution operates only on the file path portion.
+    const hashIndex = linkPath.indexOf("#");
+    const filePath = hashIndex !== -1 ? linkPath.slice(0, hashIndex) : linkPath;
+    const fragment = hashIndex !== -1 ? linkPath.slice(hashIndex) : "";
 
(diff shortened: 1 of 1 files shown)


maparent (changes requested):
Edit: Sorry this was typed in wrong window.

maparent (approved):
Sorry for the review mismatch. This is good.

PR #998, opened 2026-05-04
ENG-1688 Monthly upgrades of supabase, vercel, posthog for May 2026

maparent:
https://linear.app/discourse-graphs/issue/ENG-1688/monthly-upgrades-of-supabase-vercel-posthog-for-may-2026

This started as a minor upgrade; but the new supabase was more rigid about type checking, notably on upserts.  I did a lot of changes on typing to try to get this right, but they did not succeed in solving the upsets, so I just gave up on that. I believe those changes have value but I will propose them elsewhere as unrelated; this is closer to the minimum that is needed to make this compile with the new supabase version.

https://www.loom.com/share/744d191eab8749d18d0d0a5ef19aa319

### DIFF
--- a/apps/website/app/utils/supabase/dbUtils.ts
+++ b/apps/website/app/utils/supabase/dbUtils.ts
@@ -3,6 +3,7 @@ import type {
   PostgrestResponse,
   PostgrestSingleResponse,
 } from "@supabase/supabase-js";
+import { PostgrestError } from "@supabase/supabase-js";
 import type { Database, Tables, TablesInsert } from "@repo/database/dbTypes";
 
 export const KNOWN_EMBEDDING_TABLES: {
@@ -95,7 +96,9 @@ export const getOrCreateEntity = async <
 }): Promise<PostgrestSingleResponse<Tables<TableName>>> => {
   const result: PostgrestSingleResponse<Tables<TableName>> = await supabase
     .from(tableName)
-    .upsert(insertData, {
+    // Typescript gets confused with latest supabase
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .upsert(insertData as any, {
       onConflict: uniqueOn === undefined ? undefined : uniqueOn.join(","),
       ignoreDuplicates: false,
       count: "estimated",
@@ -138,14 +141,15 @@ export const getOrCreateEntity = async <
           console.log(`Found ${tableName} on re-fetch:`, reFetchedEntity);
           // Note: Using a PostgrestResult means I cannot have both data and error non-null...
           return {
-            error: {
+            error: new PostgrestError({
               ...result.error,
               message: `Upsert failed because of conflict with this entity: ${reFetchedEntity}"`,
-            },
+            }),
             statusText: result.statusText,
             data: null,
             count: null,
             status: 400,
+            success: false,
           };
         }
       }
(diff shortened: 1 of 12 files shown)


apps/website/app/utils/supabase/dbUtils.ts
```diff
@@ -95,7 +96,9 @@ export const getOrCreateEntity = async <
 }): Promise<PostgrestSingleResponse<Tables<TableName>>> => {
   const result: PostgrestSingleResponse<Tables<TableName>> = await supabase
     .from(tableName)
-    .upsert(insertData, {
+    // Typescript gets confused with latest supabase
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .upsert(insertData as any, {
```
mdroidian: @maparent  Have you tried an llm CLI?  Seems like codex 5.5 xhigh one-shotted this for me.
```
74 lines added, 63 lines removed. Net: +11 lines.
```

(it also fixes the `any` on ln 129 in the process)

```typescript
 insertData[key] as any, // TS gets confused here?
```

Give it a try and let me know how it turns out, and what model/setting/prompt you used if it fails.

mdroidian (changes requested):

apps/website/app/utils/supabase/dbUtils.ts
maparent: Ok, I tried Claude and it tells me this is a fundamental typescript limitation. Note that in this conversation, the "latest commit" refers to the typing-changes branch.

[2026-05-10-145100-the-update-to-supabase-js-introduced-a-typescript.txt](https://uploads.linear.app/a22a96b9-57e1-4add-a306-7b596f0b7ee3/62aa7662-736f-4353-8560-558b510f16f9/ff84bdd1-e69b-499c-ac9c-2a8beeae7497?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2EyMmE5NmI5LTU3ZTEtNGFkZC1hMzA2LTdiNTk2ZjBiN2VlMy82MmFhNzY2Mi03MzZmLTQzNTMtODU2MC01NThiNTEwZjE2ZjkvZmY4NGJkZDEtZTY5Yi00OTljLWFjOWMtMmE4YmVlYWU3NDk3IiwiaWF0IjoxNzc4NDM5MjcyLCJleHAiOjE4MTAwMDk4MzJ9.O1H7cGTOO4xVGHhnub1byRzo6YprAt8BOKJK70u0CMw)

mdroidian: I'm curious is Opus would at least suggest a fix.  Could you try Opus with prompt:

> This PR is using `any` here, which we're trying to avoid. Can you look and see why this is the case, why `any` might have been used? Is there any way we could actually define the general type structure that is expected? Or can you identify what the PR commenter means by "TypeScript gets confused" and how can we make it not confused?<br>This is from line 101 of apps\\website\\app\\utils\\supabase\\dbUtils.ts

For reference, [here's what codex provided](<https://github.com/DiscourseGraphs/discourse-graph/pull/1012>).  It is not a minimal fix, and has pros/cons.<br>At this point I'll defer to you on which direction to proceed.  But I am curious how Opus would respond.

maparent: Opus was taking forever and blocking me from other work. And the answers are not so good either. So I started with the Codex answer, made a few name changes and generalizations, and I think it's good to go.

mdroidian (approved):

PR #1005, opened 2026-05-06
ENG-1703 Allow to access a Web UI with a space's credentials

maparent:
https://linear.app/discourse-graphs/issue/ENG-1703/allow-to-access-a-web-ui-with-a-spaces-credentials

https://www.loom.com/share/282145e9dcd0490981fbd8d45030b1a6

### DIFF
--- a/apps/obsidian/src/components/AdminPanelSettings.tsx
+++ b/apps/obsidian/src/components/AdminPanelSettings.tsx
@@ -1,8 +1,10 @@
-import { useState, useCallback } from "react";
+import { useState, useCallback, useEffect } from "react";
 import { usePlugin } from "./PluginContext";
 import { Notice } from "obsidian";
 import { updateUsername } from "~/utils/supabaseContext";
 import { initializeSupabaseSync } from "~/utils/syncDgNodesToSupabase";
+import { nextRoot } from "@repo/utils/execContext";
+import { getLoggedInClient } from "~/utils/supabaseContext";
 
 export const AdminPanelSettings = () => {
   const plugin = usePlugin();
@@ -12,6 +14,26 @@ export const AdminPanelSettings = () => {
   const [username, setUsername] = useState<string>(
     plugin.settings.username || "",
   );
+  const [accessToken, setAccessToken] = useState<string | null>(null);
+  const [refreshToken, setRefreshToken] = useState<string | null>(null);
+  useEffect(() => {
+    if (syncModeEnabled) {
+      const fetchTokens = async () => {
+        const client = await getLoggedInClient(plugin);
+        if (client) {
+          const session = await client.auth.getSession();
+          if (session.data.session) {
+            setAccessToken(session.data.session.access_token);
+            setRefreshToken(session.data.session.refresh_token);
+          }
+        }
+      };
+      void fetchTokens();
+    } else {
+      setAccessToken(null);
+      setRefreshToken(null);
+    }
+  }, [syncModeEnabled]);
 
   const handleSyncModeToggle = useCallback(
     async (newValue: boolean) => {
@@ -78,6 +100,32 @@ export const AdminPanelSettings = () => {
             onChange={(e) => setUsername(e.target.value)}
             onBlur={(e) => void handleSetUsername(e.target.value)}
           />
+        </div>
+      </div>
+      <div
+        className={
+          "setting-item " + (accessToken && refreshToken ? "" : "hidden")
+        }
+      >
+        <div className="setting-item-info">
+          <div className="setting-item-name">Group management</div>
+          <div className="setting-item-description">
+            This will allow you to view and manage your sharing groups
+          </div>
+        </div>
+        <div className="setting-item-control">
+          {accessToken && refreshToken && (
+            <button
+              onClick={() => {
+                window.open(
+                  `${nextRoot()}/auth/token?t=${accessToken}&r=${refreshToken}&url=/`,
+                  "_blank",
+                );
+              }}
+            >
+              Manage groups
+            </button>
+          )}
         </div>
       </div>
     </div>
--- a/apps/obsidian/src/components/AdminPanelSettings.tsx
+++ b/apps/obsidian/src/components/AdminPanelSettings.tsx
@@ -33,7 +33,7 @@ export const AdminPanelSettings = () => {
       setAccessToken(null);
       setRefreshToken(null);
     }
-  }, [syncModeEnabled]);
+  }, [syncModeEnabled, plugin]);
 
   const handleSyncModeToggle = useCallback(
     async (newValue: boolean) => {
--- a/apps/obsidian/src/components/AdminPanelSettings.tsx
+++ b/apps/obsidian/src/components/AdminPanelSettings.tsx
@@ -15,7 +15,6 @@ export const AdminPanelSettings = () => {
     plugin.settings.username || "",
   );
   const [accessToken, setAccessToken] = useState<string | null>(null);
-  const [refreshToken, setRefreshToken] = useState<string | null>(null);
   useEffect(() => {
     if (syncModeEnabled) {
       const fetchTokens = async () => {
@@ -24,14 +23,12 @@ export const AdminPanelSettings = () => {
           const session = await client.auth.getSession();
           if (session.data.session) {
             setAccessToken(session.data.session.access_token);
-            setRefreshToken(session.data.session.refresh_token);
           }
         }
       };
       void fetchTokens();
     } else {
       setAccessToken(null);
-      setRefreshToken(null);
     }
   }, [syncModeEnabled, plugin]);
 
@@ -102,11 +99,7 @@ export const AdminPanelSettings = () => {
           />
         </div>
       </div>
-      <div
-        className={
-          "setting-item " + (accessToken && refreshToken ? "" : "hidden")
-        }
-      >
+      <div className={"setting-item " + (accessToken ? "" : "hidden")}>
         <div className="setting-item-info">
           <div className="setting-item-name">Group management</div>
           <div className="setting-item-description">
@@ -114,11 +107,11 @@ export const AdminPanelSettings = () => {
           </div>
         </div>
         <div className="setting-item-control">
-          {accessToken && refreshToken && (
+          {accessToken && (
             <button
               onClick={() => {
                 window.open(
-                  `${nextRoot()}/auth/token?t=${accessToken}&r=${refreshToken}&url=/`,
+                  `${nextRoot()}/auth/token?t=${accessToken}&url=/`,
                   "_blank",
                 );
               }}
--- a/apps/obsidian/src/components/AdminPanelSettings.tsx
+++ b/apps/obsidian/src/components/AdminPanelSettings.tsx
@@ -1,6 +1,6 @@
-import { useState, useCallback, useEffect } from "react";
+import { useState, useCallback } from "react";
 import { usePlugin } from "./PluginContext";
-import { Notice } from "obsidian";
+import { Notice, setIcon } from "obsidian";
 import { updateUsername } from "~/utils/supabaseContext";
 import { initializeSupabaseSync } from "~/utils/syncDgNodesToSupabase";
 import { nextRoot } from "@repo/utils/execContext";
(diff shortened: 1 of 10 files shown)


apps/obsidian/src/components/AdminPanelSettings.tsx
```diff
@@ -80,6 +102,32 @@ export const AdminPanelSettings = () => {
...
+          </div>
+        </div>
+        <div className="setting-item-control">
+          {accessToken && refreshToken && (
+            <button
+              onClick={() => {
+                window.open(
+                  `${nextRoot()}/auth/token?t=${accessToken}&r=${refreshToken}&url=/`,
+                  "_blank",
+                );
```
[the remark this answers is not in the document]
maparent: This is not possible in this case, as this will usually be executed in an electron environment, while the open will open in the default browser. So the only thing that is shared is a URL. What we could do is ensure that the token is not valid for very long.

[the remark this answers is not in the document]
maparent: Took another approach with the secret-token table.

apps/obsidian/src/components/AdminPanelSettings.tsx
```diff
@@ -41,6 +55,22 @@ export const AdminPanelSettings = () => {
...
+    const client = await getLoggedInClient(plugin);
+    if (!client) return;
+    const sessionData = await client.auth.getSession();
+    if (!sessionData.data.session) return;
+    /* eslint-disable @typescript-eslint/naming-convention */
+    const { access_token, refresh_token } = sessionData.data.session;
+    const { data, error } = await client.rpc("create_secret_token", {
+      v_payload: JSON.stringify({ access_token, refresh_token }),
+      expiry_interval: "10s",
+    });
```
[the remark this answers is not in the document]
mdroidian: @maparent

apps/obsidian/src/components/AdminPanelSettings.tsx
```diff
@@ -80,6 +110,23 @@ export const AdminPanelSettings = () => {
           />
         </div>
       </div>
+      <div className={"setting-item " + (isLoggedIn ? "" : "hidden")}>
+        <div className="setting-item-info">
+          <div className="setting-item-name">Group management</div>
```
mdroidian, line 116: What if I'm not an admin?  **In the future** we should name this something more generic "Sync management" or "Discourse Graph sync"

cc @jsmorabito

apps/obsidian/src/components/AdminPanelSettings.tsx
```diff
@@ -80,6 +110,23 @@ export const AdminPanelSettings = () => {
...
+          <div className="setting-item-description">
+            This will allow you to view and manage your sharing groups
```
mdroidian, line 118: It should be clear that this will open the browser.

Either via text or an icon.  We use this arrow icon in `General` settings to indicate external opening.

Image

apps/roam/src/components/settings/AdminPanel.tsx
```diff
@@ -266,13 +267,28 @@ const FeatureFlagsTab = (): React.ReactElement => {
   const [suggestiveOverlayValue, setSuggestiveOverlayValue] = useState(
     getFeatureFlag("Suggestive mode overlay enabled"),
   );
-
-  const syncAlreadyEnabled = duplicateNodeAlertValue || suggestiveOverlayValue;
```
mdroidian, line 270: keep this const, instead of repeating yourself

apps/obsidian/src/components/AdminPanelSettings.tsx
```diff
@@ -80,6 +110,23 @@ export const AdminPanelSettings = () => {
           />
         </div>
       </div>
+      <div className={"setting-item " + (isLoggedIn ? "" : "hidden")}>
```
mdroidian: Show/hide on `syncModeEnabled`, not `isLoggedIn`.  Handle login `onClick`

apps/obsidian/src/components/AdminPanelSettings.tsx
```diff
@@ -12,6 +14,18 @@ export const AdminPanelSettings = () => {
   const [username, setUsername] = useState<string>(
     plugin.settings.username || "",
   );
+  const [isLoggedIn, setIsLoggedIn] = useState(false);
+  useEffect(() => {
```
mdroidian: 1) Why are we doing this via `useEffect` instead of at the `<button>` itself?
2) We aren't handling failed logins at all

If a login fails, a user doesn't see anything.
So, as a user, if I want to go to group management, I load the admin panel.  If the login fails, I don't see a button, error message, nor a way to re-try.

Seeing as we already call `getLoggedInClient()` again in `handleLoginHandoff`, there is no need for this `useEffect` and it is actually counter productive.  Let's move everything to the `onClick` call.

There could be a case made for prefetching if a) it's a heavy call AND b) it's an often used action.  I believe neither is true in this case.

apps/roam/src/components/settings/AdminPanel.tsx
```diff
@@ -266,13 +267,28 @@ const FeatureFlagsTab = (): React.ReactElement => {
...
+  const [accessToken, setAccessToken] = useState<string | null>(null);
+  useEffect(() => {
```
mdroidian: Move this to the `onClick`, same comment as above in obsidian

apps/roam/src/components/settings/AdminPanel.tsx
```diff
@@ -393,6 +425,17 @@ const FeatureFlagsTab = (): React.ReactElement => {
       >
         Send Error Email
       </Button>
+      {accessToken && (
```
mdroidian: This shouldn't be gated on `accessToken`, use `syncAlreadyEnabled` instead and handle `accessToken` in `onClick`

apps/obsidian/src/components/AdminPanelSettings.tsx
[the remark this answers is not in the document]
maparent: Basically, this time interval is an interval during which the token has leaked to the logs, and a resident malware could pick it up; so I was going for a tight interval.
I think 60s to open a web page is overkill, even with slow computers. 
The one case where I see significant slowdown is if the browser needs to start up.
How do you feel about 30s?

apps/website/app/components/auth/LoginWithToken.tsx
```diff
@@ -0,0 +1,86 @@
...
+        setError(result.error.message);
+        return;
+      }
+      if (result.data == null) {
+        setError("This token does not exist");
+        return;
+      }
+      if (typeof result.data !== "string") {
+        setError("Payload is not a string");
+        return;
```
mdroidian, line 51: Error messages that we show to the user should be helpful to them. How do the following error messages help the user? One alternative could be a generic error message saying "The team has been notified", or "There has been an error. Please notify the team," but something like "Payload is not a string" doesn't really help the end user. 

We should log these in types of message in `postHog`, not show them to the user.

apps/website/app/utils/supabase/client.ts
```diff
@@ -0,0 +1,9 @@
+import { createBrowserClient } from "@supabase/ssr";
+import type { Database } from "@repo/database/dbTypes";
+
+export const createClient = () => {
+  return createBrowserClient<Database, "public">(
+    process.env.NEXT_PUBLIC_SUPABASE_URL!,
```
mdroidian: Avoid `!`

mdroidian (changes requested):
The biggest change requests are:
- moving the handling of the token fetch/login from a side effect to event handler
- gracefully handling token/login failure with error message, ability to retry

apps/obsidian/src/components/AdminPanelSettings.tsx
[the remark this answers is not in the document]
mdroidian: It's not just opening a web page.  Let's do 45 seconds.

apps/obsidian/src/components/AdminPanelSettings.tsx
maparent: Whoever you are, you can create a group, and that makes you admin of that specific group. So that's a group management operation you can always do.

apps/obsidian/src/components/AdminPanelSettings.tsx
maparent: Hmm... if I'm not logged in (eg offline), I feel I should not see a button that will fail. Maybe I could show it in all cases and make it disabled unless we have a login. But this is why I want to make the login happen before the button click.

apps/obsidian/src/components/AdminPanelSettings.tsx
maparent: I want to prefetch so the user will not get to attempt an action that is known to fail.
(I agree it's not a heavy call, but that's a reason why I think prefetch is cheap.)
How do you feel about disabling the button instead of hiding it?

mdroidian: It's only "known" to fail if we known the state we are testing for won't change.  <br><br>This would be true if we are in a roam graph that is of type `offline` but not true if the user is temporarily offline when they open the admin panel (ISP temporarily fail, wifi temporarily fail, network adaptor off, etc).<br><br>So `yes` to disable, but we'd still need a `refresh` / `try again` for the second case, because currently we aren't (and shouldn't) be continually re-checking if they have internet access.  In which case the question would be "why are we checking at admin panel load" anyway?

apps/obsidian/src/components/AdminPanelSettings.tsx
mdroidian: Addressed here: https://github.com/DiscourseGraphs/discourse-graph/pull/1005#discussion_r3215451632

apps/obsidian/src/components/AdminPanelSettings.tsx
maparent: Hmm… ok, point taken. It's easier to retry on button push. Will do.

apps/roam/src/components/settings/AdminPanel.tsx
maparent: I removed it because if I use it I have to use a Memo, since it's used inside an Effect. A single repetition felt better than such a simple Memo, but adding it for now.

apps/website/app/utils/supabase/client.ts
maparent: I will throw an exception if the variables are missing, though, otherwise the contract changes too much for callers.

maparent:
@mdroidian tagged you earlier and untagged you: there was a login issue I thought I saw, but I'm quite sure it was an artifact. So should be ready for review.

maparent:
Oh, unless you requested changes again? In which case not sure which ones you requested.

apps/roam/src/components/settings/AdminPanel.tsx
```diff
@@ -266,8 +268,10 @@ const FeatureFlagsTab = (): React.ReactElement => {
   const [suggestiveOverlayValue, setSuggestiveOverlayValue] = useState(
     getFeatureFlag("Suggestive mode overlay enabled"),
   );
-
-  const syncAlreadyEnabled = duplicateNodeAlertValue || suggestiveOverlayValue;
+  const syncAlreadyEnabled = useMemo(
```
mdroidian: **nit: `useMemo` isn't needed here**

This computation is a boolean OR of two primitives — cheaper than the bookkeeping `useMemo` does to track dependencies and cache the result. It also doesn't help with referential equality since the result is a primitive (compared by value), and it's only consumed locally within this component (not passed to memoized children or used as a hook dependency).

From the React docs on [`useMemo` — Should you add useMemo everywhere?](https://react.dev/reference/react/useMemo#should-you-add-usememo-everywhere):

> Optimizing with `useMemo` is only valuable in a few cases [...] In practice, you can make a lot of memoization unnecessary by following a few principles [...] **There is no benefit to wrapping a calculation in `useMemo` in other cases.**

See also:
- [You Might Not Need `useMemo`](https://react.dev/learn/you-might-not-need-an-effect) (and the broader [React rendering behavior](https://react.dev/learn/render-and-commit) guide)
- Dan Abramov, [Before You memo()](https://overreacted.io/before-you-memo/) — covers why reflexive memoization often adds cost without measurable benefit
- Kent C. Dodds, [When to useMemo and useCallback](https://kentcdodds.com/blog/usememo-and-usecallback)

mdroidian (approved):

PR #1011, opened 2026-05-10
ENG-1719 List groups I'm a member of

maparent:
https://linear.app/discourse-graphs/issue/ENG-1719/list-groups-im-a-member-of

This creates a vercel page with a group list, which uses the database token set up in ENG-1703.

https://www.loom.com/share/1897670dcaae4908a3fde4ce4cf3c22e


apps/website/app/components/auth/ListGroups.tsx
```diff
@@ -0,0 +1,106 @@
...
+          });
+          return;
+        }
+        setGroupData(groupResponse.data);
+        const membershipReq = await client
+          .from("group_membership")
+          .select("group_id,admin")
+          .eq("member_id", id);
+        if (membershipReq.error) {
+          const userMessage = "Could not access DiscourseGraphs";
```
mdroidian: *Discourse Graphs

apps/website/app/components/auth/ListGroups.tsx
```diff
@@ -0,0 +1,106 @@
...
+        internalError({
+          error,
+        });
+      }
+    };
+    void getGroups();
+  }, [internalError]);
+
+  return (
+    <div>
```
mdroidian: If the actual `<div>` isn't doing anything, use `<>` instead

mdroidian (changes requested):
## Codex PR Review

  ### Findings

  **High:** `apps/website/app/components/auth/LoginWithToken.tsx` should not perform the token exchange in a Client Component. The token is removed from the URL only after hydration via `useEffect`, but `PostHogPageView` reads `searchParams` and can capture `$current_url` before that happens. That risks leaking the one-time token to analytics.

  **High:** `router.replace(url)` uses a query-derived `url` without validation. Next.js warns against passing unsanitized URLs to `router.push`/`router.replace`. This should be restricted to same-origin relative paths, or default to `/auth/group`.

  **Medium:** `ListGroups.tsx` does not appear to need `"use client"`. It is session-aware data fetching plus rendering, and the app already has a cookie-backed Supabase server client. This should be an async Server Component instead of fetching in `useEffect`.

  ### Recommendation

  I would keep client code only where browser APIs are genuinely required. In this PR, `/auth/group` should be server-rendered:

  - Use `~/utils/supabase/server` in an async Server Component.
  - Fetch the current user, groups, and memberships on the server.
  - Render empty/not-logged-in states directly.
  - Use a colocated `error.tsx` for unexpected failures. The error boundary itself must be client-side, but that keeps `"use client"` isolated to error UI.

  For `/auth/token`, I would avoid a Client Component entirely if possible. A Route Handler is a better fit:

  - Read `t` and `url` from the request.
  - Validate the token payload.
  - Call `auth.setSession` using the server Supabase client so cookies are set in the response.
  - Redirect to a sanitized internal path.
  - Avoid exposing the token to hydration, analytics, or browser history.

  So: we do need `"use client"` for a Next error boundary, and we would need it if we keep `useSearchParams`, `useRouter`, `useEffect`, or `window.history`. But for this flow, the better Next.js-specific approach is Server Components plus a Route Handler, not expanding the client boundary.

mdroidian (changes requested):
Sorry, it doesn't look like the issues in the [review comment](https://github.com/DiscourseGraphs/discourse-graph/pull/1011#pullrequestreview-4291331403) were addressed.  Could you address those please?

maparent:
Ok, I did not see it. After research, I see it is possible to rewrite all that code as server components, but setting that up is a totally distinct task. I'll create it.

mdroidian:
Could you elaborate on why it is a distinct task?<br><br>I see it as the recommended way to achieve the current task ([using the NextJS recommended](<https://nextjs.org/docs/app/getting-started/server-and-client-components>) way to fetch and render data)

maparent:
I had to rewrite the token route to work server-side. It's a big PR, I'll try to think of ways to break it up tomorrow.

maparent:
We are now using the backend client. Added a test.

maparent:
Here I'm more embarrassed. Again, the PR is over size; a lot of that is due to testing. I do think it would be counter-productive to separate tests in another PR. I do not think it would be very helpful to split this PR in another way, though I could separate account.ts+tests, then all UI. Would that actually be helpful? But I think a generic argument should be made that tests need not the same level of review than the rest of the code, and should actually reduce the review burden; and that their LoC should be counted separately. What do you think?

mdroidian:
> Here I'm more embarrassed. Again, the PR is over size; a lot of that is due to testing. I do think it would be counter-productive to separate tests in another PR. I do not think it would be very helpful to split this PR in another way, though I could separate account.ts+tests, then all UI. Would that actually be helpful? But I think a generic argument should be made that tests need not the same level of review than the rest of the code, and should actually reduce the review burden; and that their LoC should be counted separately. What do you think?

The documented limit is:

> Recommended: ~200 lines changed
> Acceptable limit: up to 400 lines when well-scoped/self-contained

This PR is +281 -3
Lines changed: 281 additions & 3 deletions

I see that as falling within 

> Acceptable limit: up to 400 lines when well-scoped/self-contained

maparent:
Fair. But I think the question of test size needs to be addressed separately.

mdroidian:
@codex review

apps/website/app/(home)/auth/group/page.tsx
```diff
@@ -0,0 +1,11 @@
+import { ListGroups } from "~/components/auth/ListGroups";
+
+const Page = () => (
+  <main>
```
mdroidian, line 4: Do we need `<main>` and `<div>`?  Or can we just use `<main>` and add the classes there?

mdroidian (approved):
Also see the codex comment

apps/website/test/integration/listMyGroups.test.ts
```diff
@@ -0,0 +1,126 @@
+import assert from "assert";
```
mdroidian, line 1: @maparent 
repo/typescript filename convention: `listMyGroups.test.ts`

apps/website/app/(home)/auth/group/page.tsx
maparent: I tried and the look is completely different.

mdroidian: How do you mean?  Why do you think that is the case?

PR #1017, opened 2026-05-11
ENG-1690 Disable search dropdown in file explorer "Convert into" flow

trangdoan982:
https://www.loom.com/share/45627180e19b47e78e2fef7d055f88df

## Summary

- Adds a `disableExistingNodeSearch` prop to `ModifyNodeModal` / `ModifyNodeForm` that suppresses the search-as-you-type dropdown for existing nodes
- Passes `disableExistingNodeSearch: true` from both file-explorer "Convert into" call sites (markdown files and image files) in `index.ts` and `editorMenuUtils.ts`
- Updates the textarea placeholder to remove "Search for existing…" language when search is disabled

## Why

When converting a file from the file explorer context menu, the intent is unambiguously to **create** a new discourse node (not reuse an existing one). Showing the search dropdown adds noise and could confuse users into selecting an existing node when they meant to convert their file.

## What is not affected

- Editor "Turn into discourse node" flow (`currentFile` passed) — search remains enabled
- Canvas double-click edit flow (`initialFile` passed, `isEditMode = true`) — already search-free
- Hover-icon image conversion — `openConvertImageToNodeModal` is called without the flag, so search stays enabled there
- All other create-mode form semantics: heading "Create discourse node", editable node type selector, "Creating…" button label, relationship block

## Test plan

- [x] Right-click a non-discourse-node file in the file explorer → "Convert into" → pick a node type → confirm the title input does **not** show a search dropdown when typing
- [x] Right-click an image file in the file explorer → "Convert into" → same check
- [x] Use editor context menu "Turn into discourse node" → confirm search dropdown **still works**
- [x] Use hover icon on an embedded image → confirm search dropdown **still works**
- [x] Confirm file conversion still completes successfully (node is created, file is updated)

### DIFF
--- a/apps/obsidian/src/components/ModifyNodeModal.tsx
+++ b/apps/obsidian/src/components/ModifyNodeModal.tsx
@@ -53,8 +53,13 @@ type ModifyNodeFormProps = {
   onCancel: () => void;
   initialTitle?: string;
   initialNodeType?: DiscourseNode;
-  initialFile?: TFile; // for edit mode
-  currentFile?: TFile; // the file where the node is being created from
+  /** Present in edit mode — the file being modified. Also drives `isEditMode`. */
+  initialFile?: TFile;
+  /** Context file for the "Relationship with …" block (editor "Turn into" flow). */
+  currentFile?: TFile;
+  /** Hides the search-existing-nodes dropdown. Used when intent is always to create
+   *  (e.g. file explorer "Convert into"), not to reuse an existing node. */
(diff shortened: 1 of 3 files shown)


mdroidian (approved):
I won't block on this, but part of the problem the search results solve is looking for similar nodes so that you don't duplicate a node on creation.  This was previously a requested feature in similar contexts, so without a replacement, this might be an unwanted change.

PR #1018, opened 2026-05-11
ENG-1693 Add insert backlink checkbox to ModifyNodeModal

trangdoan982:
https://www.loom.com/share/2c08071a85d849f0b13e828541a0a429

## Summary

- Adds an **Insert backlink** checkbox at the bottom of `ModifyNodeModal` (create mode only)
- Smart defaults based on context — no new props needed:
  - `true` when the modal is opened with pre-filled text (user had text selected before invoking)
  - `true` when the user selects an existing node from the search
  - `false` when opened with no selected text (blank create)
  - User can always toggle the checkbox manually
- Gates backlink insertion in both the canvas flow (`nodeCreationFlow.ts`) and the editor command flow (`registerCommands.ts`) on the checkbox value

## Test plan

- [x] Open "Create discourse node" with no selected text → checkbox unchecked; submitting does not insert backlink
- [x] Select text in editor, open via right-click or command → checkbox checked; submitting inserts backlink at cursor
- [x] Open modal, pick an existing node from search → checkbox auto-checks; submitting inserts backlink
- [x] Open modal, pick existing node, clear selection → checkbox resets to initial default
- [x] User manually toggles checkbox in any case → submission respects the toggled value
- [x] Canvas: Don't show Insert backlink

Closes ENG-1693

### DIFF
--- a/apps/obsidian/src/components/canvas/utils/nodeCreationFlow.ts
+++ b/apps/obsidian/src/components/canvas/utils/nodeCreationFlow.ts
@@ -32,6 +32,7 @@ export const openCreateDiscourseNodeAt = (args: CreateNodeAtArgs): void => {
       selectedExistingNode,
       relationshipId,
       relationshipTargetFile,
+      insertBacklink,
     }) => {
       try {
         // If user selected an existing node, use it instead of creating a new one
@@ -54,11 +55,13 @@ export const openCreateDiscourseNodeAt = (args: CreateNodeAtArgs): void => {
           });
         }
 
-        const src = await addWikilinkBlockrefForFile({
-          app: plugin.app,
-          canvasFile,
-          linkedFile: fileToUse,
-        });
+        const src = insertBacklink
+          ? await addWikilinkBlockrefForFile({
+              app: plugin.app,
+              canvasFile,
+              linkedFile: fileToUse,
+            })
+          : undefined;
 
         let preloadedImageSrc: string | undefined = undefined;
         if (selectedNodeType.keyImage) {
--- a/apps/obsidian/src/components/canvas/utils/nodeCreationFlow.ts
+++ b/apps/obsidian/src/components/canvas/utils/nodeCreationFlow.ts
@@ -32,7 +32,6 @@ export const openCreateDiscourseNodeAt = (args: CreateNodeAtArgs): void => {
       selectedExistingNode,
       relationshipId,
       relationshipTargetFile,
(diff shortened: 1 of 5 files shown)


apps/obsidian/src/components/canvas/utils/nodeCreationFlow.ts
```diff
@@ -54,11 +55,13 @@ export const openCreateDiscourseNodeAt = (args: CreateNodeAtArgs): void => {
           });
         }
 
-        const src = await addWikilinkBlockrefForFile({
-          app: plugin.app,
-          canvasFile,
-          linkedFile: fileToUse,
-        });
+        const src = insertBacklink
```
mdroidian: I believe this breaks canvas.

https://www.loom.com/share/bdb2e4e379ee4d9a894e7a2072e77816

mdroidian (changes requested):
Requesting changes because this introduces a bug in the canvas create-node flow.

**Broader design concern**

Separate from the immediate bug, I’m worried about `ModifyNodeModal` continuing to accumulate workflow-specific props and submit fields.

The modal is now responsible for:

- create mode
- edit mode
- existing-node search
- relationship creation
- editor insertion
- canvas creation behavior
- backlink insertion

That makes the function hard to track because the meaning of a submit field depends heavily on which caller opened the modal.

It also makes bugs hard to debug. `insertBacklink` is a good example: in an editor flow, it means “insert a wikilink into the editor.” In the canvas flow, it ended up controlling whether the shape gets the blockref it needs to resolve to a file. Those are very different side effects hidden behind the same boolean.

I’d prefer we make caller capabilities explicit instead of inferring them inside the modal from global workspace state, and avoid using one generic submit payload for unrelated workflows. But we can continue that exploration in [ENG-1750: Refactor Obsidian `ModifyNodeModal` contract for clearer workflow ownership](https://linear.app/discourse-graphs/issue/ENG-1750/refactor-obsidian-modifynodemodal-contract-for-clearer-workflow)

mdroidian:
@codex review

mdroidian (approved):
Could you submit a video showing that the bug as been fixed?

Also, please take a look at https://github.com/DiscourseGraphs/discourse-graph/pull/1018#discussion_r3295494015

trangdoan982:
New test on editor and canvas:

https://www.loom.com/share/7729158024fc4cc28bd82bc268622eb9

PR #1020, opened 2026-05-11
ENG-1741 Prevent duplicate keyboard shortcuts for node types

trangdoan982:
https://www.loom.com/share/10168beedb5548b38b716ba59b10138f

## Summary

- On node creation, skip auto-assigning the first-letter shortcut if it is already taken by any existing node type; leave empty instead so the user is prompted to set a unique one manually
- In node type settings, validate the shortcut field against all other node types and surface an inline error when a conflict is detected (same warning pattern as the existing tag/format validation)

Fixes: [ENG-1741](https://linear.app/discourse-graphs/issue/ENG-1741/modify-keyboard-shortcut-behavior-for-node-type-dropdown-menu)

## Changes

- `DiscourseNodeConfigPanel.tsx` — collision-aware shortcut auto-assignment on "Add node"
- `NodeConfig.tsx` — `validateShortcut` callback + `shortcutError` state wired to the Shortcut text panel

## Test plan

- [x] Create a new node type whose label starts with the same letter as an existing node (e.g. add "Concept" when "Claim" with shortcut `C` exists) — shortcut should be empty, not `C`
- [x] In node settings, manually set a shortcut that matches another node's shortcut — inline error should appear immediately
- [x] Setting a unique shortcut clears the error
- [x] Existing nodes with no conflicts are unaffected

### DIFF
--- a/apps/roam/src/components/settings/NodeConfig.tsx
+++ b/apps/roam/src/components/settings/NodeConfig.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useCallback, useEffect } from "react";
-import { DiscourseNode } from "~/utils/getDiscourseNodes";
+import getDiscourseNodes, { DiscourseNode } from "~/utils/getDiscourseNodes";
 import DualWriteBlocksPanel from "./components/EphemeralBlocksPanel";
 import { getSubTree } from "roamjs-components/util";
 import Description from "roamjs-components/components/Description";
@@ -86,9 +86,11 @@ const NodeConfig = ({
   const [selectedTabId, setSelectedTabId] = useState<TabId>("general");
   const [tagError, setTagError] = useState("");
   const [formatError, setFormatError] = useState("");
+  const [shortcutError, setShortcutError] = useState("");
 
   const [tagValue, setTagValue] = useState(node.tag || "");
   const [formatValue, setFormatValue] = useState(node.format || "");
+  const [shortcutValue, setShortcutValue] = useState(node.shortcut || "");
   const validate = useCallback(
     ({
       tag,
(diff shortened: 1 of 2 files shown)


apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -148,6 +150,25 @@ const NodeConfig = ({
...
 
+  const validateShortcut = useCallback(
+    (value: string) => {
+      if (!value) return setShortcutError("");
+      const taken = getDiscourseNodes()
+        .filter((n) => n.type !== node.type && n.shortcut)
+        .map((n) => n.shortcut.toUpperCase());
+      setShortcutError(
+        taken.includes(value.toUpperCase())
+          ? `Shortcut "${value.toUpperCase()}" is already used by another node type.`
```
mdroidian: We have access to the nodes here, let's adjust `taken` to include the node label, and include the list (could be more than one) in the error message.  This way a user doesn't have to go through each one to find the duplicate.

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -148,6 +150,25 @@ const NodeConfig = ({
...
+          : "",
+      );
+    },
+    [node.type],
+  );
+
+  useEffect(() => {
```
mdroidian: Could we do this in the input instead?  eg: `onChange` or `onBlur`?

mdroidian (changes requested):

apps/roam/src/components/settings/NodeConfig.tsx
mdroidian: [You Might Not Need an Effect – Adjusting some state when a prop changes 
](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)

mdroidian (approved):

PR #1027, opened 2026-05-13
ENG-1702 Invite new members in the group

maparent:
https://linear.app/discourse-graphs/issue/ENG-1702/invite-new-members-in-the-group

One loom covers eng-1720, eng-1702 and vitest-web
https://www.loom.com/share/4dc27a35953340058f9227638e63ff62


maparent:
Note: I could make it smaller by separating the cucumber test if desired.

apps/website/app/utils/supabase/account.ts
```diff
@@ -34,16 +34,79 @@ export const getSessionUserData = async (
...
+    if (accountReq.error || !accountReq.data?.name) {
+      return null;
+    }
+    return { ...data, name: accountReq.data.name };
   }
-  return { id, name: accountReq.data.name, type: "person", email };
+  if (data.name === undefined) return null;
+  return data as {
+    id: string;
+    name: string; // typescript is being dumb
```
mdroidian: Could you elaborate please?

mdroidian (changes requested):
- Guard dg_account before deleting users - `packages/database/features/step-definitions/stepdefs.ts:83-83`  Because Supabase's .not(...) call does not narrow the generated row type, dg_account is still `string` | `null` here; client.auth.admin.deleteUser requires a string, so pnpm --filter @repo/database check-types now reports TS2345 on this line.

- Use the existing secret token schema path - `packages/database/supabase/config.toml:55-55`.  There is no packages/database/supabase/schemas/secret_token.sql; the schema file in  is secret_tokens.sql.

mdroidian (changes requested):
Re-triggering this "request changes" as GitHub/linear didn't catch it.

apps/website/app/utils/supabase/account.ts
maparent: elaborated in code.

maparent:
Ready for re-review (the git state does not show the option?)

mdroidian:
@codex review

apps/website/test/integration/groupInvitation.test.ts
```diff
@@ -0,0 +1,130 @@
+import { describe, it, expect, beforeAll, afterAll } from "vitest";
```
mdroidian, line 1: repo/ts naming convention: groupInvitation.test.ts

apps/website/app/utils/supabase/account.ts
```diff
@@ -36,16 +36,81 @@ export const getSessionUserData = async (
...
   }
-  return { id, name: accountReq.data.name, type: "person", email };
+  if (data.name === undefined) return null;
+  return data as {
+    id: string;
+    name: string; // typescript fails to infer that name is defined from above
+    type: AgentType;
+    email?: string;
+  };
+};
```
mdroidian, line 68:
```suggestion
  const { name } = data;
  if (name === undefined) return null;
  return { ...data, name };
```

mdroidian (approved):
Also see llm review comments

apps/website/app/utils/supabase/account.ts
maparent: Ah, nice!

PR #1040, opened 2026-05-15
ENG-1729: Create initial search with preview

trangdoan982:
https://www.loom.com/share/8c4c9052e8cc440584079b09d8468998

## Summary

Adds the first milestone of Roam advanced node search: a command-palette dialog with debounced DG node search, a capped results list, and a live preview panel.

- New `AdvancedNodeSearchDialog` using Blueprint `Dialog`, `InputGroup`, `Spinner`, `Tag`, and `NonIdealState`
- Queries user-backed DG node types via `roamAlphaAPI.data.async.fast.q`, then ranks matches with MiniSearch (50-result cap, 250ms debounce)
- Result rows show colored type badges, highlighted title matches, and ~200-char excerpts from `roamAlphaAPI.pull`
- Preview panel shows full content plus last modified, created date, and author metadata
- Registers `DG: Open Node Search Menu` in the Roam command palette

This PR targets the `roam-advanced-search` integration branch so advanced search milestones can land incrementally before merging to `main`.

## Test plan

- [x] Reload the Roam developer extension from this branch
- [x] Open command palette and run `DG: Open Node Search Menu`
- [x] Type a keyword and confirm loading, results, and empty states
- [x] Verify type badge colors, title highlighting, and excerpt text in the results list
- [x] Select results with mouse and arrow keys; confirm preview updates and scrolls active row into view
- [x] Confirm preview shows last modified, created date, author, and node content
- [x] Press `Escape` to close the dialog


apps/roam/src/components/AdvancedNodeSearchDialog/index.tsx
```diff
@@ -0,0 +1,374 @@
...
+      } else if (event.key === "ArrowUp" && results.length) {
+        event.preventDefault();
+        setActiveIndex((index) => Math.max(index - 1, 0));
+      } else if (event.key === "Escape") {
+        event.preventDefault();
+        onClose();
+      }
+    },
+    [onClose, results.length],
+  );
```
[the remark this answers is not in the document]
trangdoan982: not included in this ticket yet

mdroidian:
@codex review

apps/roam/src/components/AdvancedNodeSearchDialog/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+  result,
+}: {
+  active: boolean;
+  keywords: string[];
+  nodeConfig: DiscourseNode | undefined;
+  onClick: () => void;
+  onMouseEnter: () => void;
+  result: SearchResult;
+}) => (
+  <button
```
mdroidian: Blueprint `<Button>`

apps/roam/src/components/AdvancedNodeSearchDialog/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+    if (!isOpen) {
+      setSearchTerm("");
+      setDebouncedSearchTerm("");
+      setResults([]);
+      setActiveIndex(0);
+      setIndexError(false);
+      miniSearchRef.current = null;
+      allResultsRef.current = [];
+      return;
+    }
```
mdroidian: Why do we have set all of these states with `useEffect`?

apps/roam/src/components/AdvancedNodeSearchDialog/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+    let cancelled = false;
+    setIsIndexLoading(true);
+    setIndexError(false);
```
mdroidian, line 197: DRY, this is being set on ln 184 as well

apps/roam/src/components/AdvancedNodeSearchDialog/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+    setActiveIndex(0);
+  }, [debouncedSearchTerm, indexError, isIndexLoading, isOpen]);
+
+  useEffect(() => {
+    const panel = resultsPanelRef.current;
+    if (!panel) return;
+
+    const activeRow = panel.querySelector('[aria-selected="true"]');
+    activeRow?.scrollIntoView({ block: "nearest" });
+  }, [activeIndex]);
```
mdroidian: The scroll synchronization Effect misses result changes.

 If the user is already at index 0, scrolls the results pane, then searches again, `setActiveIndex(0)` may not trigger a render/effect, so the newly selected first result can remain out of view. Include a result identity dependency such as `activeResult?.uid` or results alongside `activeIndex`.

apps/roam/src/utils/registerCommandPaletteCommands.ts
```diff
@@ -332,6 +333,10 @@ export const registerCommandPaletteCommands = (onloadArgs: OnloadArgs) => {
   };
 
   // Roam organizes commands alphabetically
+  void addCommand("DG: Open Node Search Menu", () => {
```
mdroidian: Let's move this in alphabetically order with the rest. (after `DG: Open - Discourse settings`, before `DG: Open - Query drawer`

Also, let's follow the pattern and move the `posthogt.capture()` inside the callback function.

apps/roam/src/components/AdvancedNodeSearchDialog/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+    if (!matchedResults.length) {
+      setResults([]);
+      setActiveIndex(0);
+      return;
+    }
+
+    setResults(matchedResults);
+    setActiveIndex(0);
+  }, [debouncedSearchTerm, indexError, isIndexLoading, isOpen]);
```
mdroidian: Could we match the `ResultsView` pattern here and derive these displayed results with `useMemo` instead of storing them via an Effect?

`ResultsView` keeps the source `results` as data and derives `allProcessedResults` / `paginatedResults` from filters, sort, and pagination during render. This search result list is similar: once the index has loaded, `results` is derived from `debouncedSearchTerm`, index state, and loading/error state, rather than synchronized with an external system.

That would avoid an extra render with stale results after the query changes, and keep this aligned with React's "You Might Not Need an Effect" guidance.

apps/roam/src/components/AdvancedNodeSearchDialog/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+        setActiveIndex((index) => Math.max(index - 1, 0));
+      } else if (event.key === "Escape") {
+        event.preventDefault();
+        onClose();
+      }
+    },
+    [onClose, results.length],
+  );
+
+  const contentState = useMemo(() => {
```
mdroidian: This does not need a `useMemo`.  Memoization adds more machinery than value. React’s guidance is to calculate derived values during render unless the calculation is expensive.

apps/roam/src/components/AdvancedNodeSearchDialog/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+      autoFocus={false}
+      canEscapeKeyClose
+      canOutsideClickClose
+      className="flex-col overflow-hidden bg-white p-0"
+      enforceFocus={false}
+      isOpen={isOpen}
+      onClose={onClose}
+      style={{
+        height: "72vh",
+        maxWidth: "980px",
```
mdroidian: tailwind `max-w-4xl`
https://tailwindcss.com/docs/max-width

mdroidian (changes requested):
Quick first round of review.  Also see codex's comments.

mdroidian:
Also, we could put this behind an `admin` feature flag and merge to main instead of `roam-advanced-search` to make it easier for the team to test as we roll it out.

apps/roam/src/utils/registerCommandPaletteCommands.ts
```diff
@@ -339,10 +340,12 @@ export const registerCommandPaletteCommands = (onloadArgs: OnloadArgs) => {
   void addCommand("DG: Export - Current page", exportCurrentPage);
   void addCommand("DG: Export - Discourse graph", exportDiscourseGraph);
   void addCommand("DG: Open - Discourse settings", renderSettingsPopup);
-  void addCommand("DG: Open - Node search", () => {
-    posthog.capture("Node Search: Open Command Triggered");
-    renderAdvancedNodeSearchDialog();
-  });
+  if (getFeatureFlag("Advanced node search enabled")) {
```
mdroidian, line 415: This is probably fine for now, but will require a reload.

@sid597 would you have a suggested pattern to follow to build on top of the new reactive settings to add/remove a command pallet command on setting toggle?

mdroidian (approved):

apps/roam/src/components/AdvancedNodeSearchDialog/index.tsx
```diff
@@ -0,0 +1,361 @@
+import React, {
```
mdroidian, line 1: @trangdoan982 Let's try to follow the naming conventions already used in the repo (for both folder and file)

Look at 

`/canvas/Tldraw.tsx`
`/results-view/Kanban.tsx`

as examples.

because `index.tsx` is the main component, it should be named as such and follow React component naming conventions.

PR #1041, opened 2026-05-16
ENG-1763 Upgrade NextJS to version 16

maparent:
https://linear.app/discourse-graphs/issue/ENG-1763/upgrade-nextjs-to-version-16

This is needed for my WASM shortcut in MIRA; and while following supabase-nextjs instructions, they were referring to changes in latest next, so I thought to do the upgrade now to avoid having to port those changes backwards then forwards.

https://www.loom.com/share/4a96f62ce6ff4dffb55917ab5da24ed2

### DIFF
--- a/apps/website/next.config.ts
+++ b/apps/website/next.config.ts
@@ -24,6 +24,9 @@ const nextConfig: NextConfig = {
       "next-mdx-import-source-file": "./mdx-components.tsx",
     },
   },
+  images: {
+    qualities: [75, 85, 100],
+  },
   async headers() {
     return [
       {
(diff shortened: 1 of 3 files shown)


apps/website/next.config.ts
```diff
@@ -16,9 +16,6 @@ const withNextra = nextra({
 
 const nextConfig: NextConfig = {
   reactStrictMode: true,
-  serverRuntimeConfig: {
```
maparent, line 19: This is deprecated; according to the documentation we should set it in environment variables, which I'm pretty sure vercel does for us.

mdroidian (approved):

PR #1042, opened 2026-05-17
ENG-1762 Rewrite all supabase access as server-side components

maparent:
https://linear.app/discourse-graphs/issue/ENG-1762/rewrite-all-supabase-access-as-server-side-components

This is a rewrite of ENG-1703 so the server side has access to credentials; it will allow the rewrite of the subsequent PRs on the server side.

https://www.loom.com/share/a9122594679c4ff796e37c1829929167

Note: The PR looks big, but not if you realize that there were renames, as explained in the loom.


apps/website/app/(home)/auth/token/route.ts
```diff
@@ -0,0 +1,71 @@
...
+              ? error.message
+              : "Unkown error while logging you in.",
+          ),
+        request.nextUrl,
+      ),
+    );
+  }
+  return NextResponse.redirect(
+    new URL(url, url.startsWith("http") ? undefined : request.nextUrl),
+  );
```
mdroidian, line 75: codex
The route reads url from the query string, establishes the Supabase session, then redirects with `new URL(url, url.startsWith("http") ? undefined : request.nextUrl)`. Any absolute http:// or https:// value is accepted, so a valid token link can be used as an open redirect to an attacker-controlled destination immediately after login.

Recommendation
Keep the redirect target inside the application unless there is a product requirement for external destinations. Parse the URL inside the existing try block, reject non-HTTP(S) schemes, compare origin with request.nextUrl.origin, and redirect to a safe default on failure. If external redirects are required, use a narrow allowlist of trusted origins.

apps/website/app/utils/supabase/proxy.ts
```diff
@@ -3,25 +3,24 @@ import { NextResponse, type NextRequest } from "next/server";
...
-// Not usable yet, waiting for ENG-373
-// Inspired by https://supabase.com/ui/docs/nextjs/password-based-auth
+// as described here: https://supabase.com/docs/guides/auth/server-side/creating-a-client
 
 export const updateSession = async (request: NextRequest) => {
+  let supabaseResponse = NextResponse.next({
+    request,
+  });
   const dbEnv = envContents();
   const supabaseUrl = dbEnv.SUPABASE_URL;
```
mdroidian, line 13: codex
This PR adds a root `proxy.ts` whose matcher includes the public root and most website routes. Every matched request calls `updateSession`, which immediately throws when SUPABASE_URL or SUPABASE_PUBLISHABLE_KEY is absent. The database package documents `SUPABASE_USE_DB=none` as the default development mode, so the website now fails before rendering public pages in that supported setup.

Suggestion
Either limit the matcher to routes that genuinely need Supabase session refresh, or teach `updateSession` to return `NextResponse.next()` when the configured database mode is intentionally disabled. Keep the hard failure for deployment environments where Supabase is required.

mdroidian (changes requested):
Blocking concern: the token login route accepts an untrusted url query parameter and redirects to arbitrary absolute URLs after setting the session.
Blocking concern: the newly active proxy runs on public routes and throws when Supabase environment variables are absent, which breaks the documented database-less development mode.

apps/website/app/utils/supabase/proxy.ts
maparent: Narrowing to /auth/

apps/website/app/(home)/auth/token/route.ts
maparent: I am not aware of the need for external, disallowing totally for now.

mdroidian (approved):

PR #1053, opened 2026-05-18
ENG-1732: Add sorting to advanced node search

trangdoan982:
https://www.loom.com/share/3962e60744dd4477a24c1e993d8b9def

## Summary

- Add a sort icon control to the advanced node search dialog toolbar with a popover for relevance, alphabetical, date created, and author (each with asc/desc).
- Preserve MiniSearch scores and apply client-side sorting over the top 50 results after search.
- Fix toolbar layout so the search input, sort button, and close button do not overlap.

## Test plan

- [x] Open **DG: Open Node Search Menu** in Roam and run a keyword search
- [x] Confirm sort icon appears next to close without overlapping the search field
- [x] Sort by **Relevance** (desc): highest-scoring matches first
- [x] Sort by **Alphabetical** (asc/desc): titles reorder by stripped display title
- [x] Sort by **Date created** (asc/desc): order follows node creation time
- [x] Sort by **Author** (asc/desc): order follows author display name; empty authors sort last
- [x] Change sort while results are shown: list reorders immediately and selection resets to first row

### DIFF
--- a/apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchDialog.tsx
+++ b/apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchDialog.tsx
@@ -148,6 +148,7 @@ const AdvancedNodeSearchDialog = ({
   const [isIndexLoading, setIsIndexLoading] = useState(false);
   const [indexError, setIndexError] = useState(false);
   const [activeIndex, setActiveIndex] = useState(0);
+  const [results, setResults] = useState<SearchResult[]>([]);
   const [sort, setSort] = useState<SortConfig>(DEFAULT_SORT_CONFIG);
   const miniSearchRef = useRef<MiniSearch<
     SearchResult & { id: string }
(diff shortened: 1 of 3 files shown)


apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchDialog.tsx
```diff
@@ -191,6 +184,34 @@ const AdvancedNodeSearchDialog = ({
     };
   }, [isOpen]);
 
+  useEffect(() => {
+    if (!isOpen) {
```
mdroidian, line 188: The close/reset effect resets sort and results, but leaves searchTerm, debouncedSearchTerm, activeIndex, and indexError intact. That creates inconsistent reopen behavior: the input can reopen with the old query while the results were cleared, then results repopulate from stale search state.

Either preserve all search state intentionally, or reset all of it:

```
     if (!isOpen) {
       setSearchTerm("");
       setDebouncedSearchTerm("");
       setActiveIndex(0);
       setSort(DEFAULT_SORT_CONFIG);
       setResults([]);
       setIndexError(false);
```

apps/roam/src/components/DiscourseNodeSortControl.tsx
```diff
@@ -0,0 +1,182 @@
...
+      minimal
+      onClick={(event) => {
+        event.stopPropagation();
+        onDirectionChange("desc");
+      }}
+      onMouseDown={(event) => event.preventDefault()}
+      small
+    />
+  </ButtonGroup>
+);
```
mdroidian: I would strongly recommend we do not show separate ascending/descending icons for every sort option.

Only one sort field is active at a time, so repeating direction controls for every field adds visual clutter without adding much functionality. It also makes the menu feel heavier and more complicated than the underlying interaction.

I think the cleaner model is:

1. dropdown selects the sort field
2. a single adjacent direction button toggles ascending/descending for the active field

That gives us one direction icon instead of eight, keeps the dropdown focused on choosing the field, and makes the direction toggle a one-click action.


Here's an example of how Roam separates them in Linked References, if you want to do this in the dropdown.

Image

mdroidian:
@codex review

mdroidian (approved):

apps/roam/src/components/DiscourseNodeSortControl.tsx
trangdoan982: oh good point. thank you

PR #1056, opened 2026-05-20
ENG-949: Add query sections to left sidebar

sid597:
https://www.loom.com/share/bc4e7221aa9240ec9f58a9ba3f5657ab

## Summary by CodeRabbit

* **New Features**
  * Query blocks can now be used as dynamic sidebar sections with lazy-loading results.
  * Added customizable alias names for query sections.
  * Added configurable result limits for query section outputs.
  * Added refresh action and collapsible UI with loading/error/empty states.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DiscourseGraphs/discourse-graph/pull/1056?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -380,7 +380,10 @@ const QuerySectionItem = ({
   const queryLabel = useMemo(() => getTextByBlockUid(queryUid), [queryUid]);
   const displayName = alias || queryLabel || section.text;
   const truncateAt = section.settings?.truncateResult.value;
-  const resultLimit = section.settings?.resultLimit?.value ?? 0;
+  const resultLimit = Math.max(
+    0,
+    Math.trunc(section.settings?.resultLimit?.value ?? 10),
+  );
 
   const [isOpen, setIsOpen] = useState<boolean>(
     !!section.settings?.folded.value,
@@ -605,7 +608,7 @@ const PersonalSections = ({
       className="personal-left-sidebar-sections"
       renderItem={(section, handle) => {
         const sectionIndex = sections.findIndex((s) => s.uid === section.uid);
-        if (isQueryBlockRef(section.text)) {
+        if (isQueryBlockRef(section.text) && section.settings?.uid) {
           return (
             <QuerySectionItem
               section={section}
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -390,7 +390,7 @@ const QuerySectionItem = ({
   );
   const [results, setResults] = useState<ChildNode[]>([]);
   const [isLoading, setIsLoading] = useState(false);
-  const [hasLoaded, setHasLoaded] = useState(false);
+  const [hasCompletedInitialLoad, setHasCompletedInitialLoad] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const [isMenuOpen, setIsMenuOpen] = useState(false);
   const isTogglingRef = useRef(false);
@@ -416,15 +416,15 @@ const QuerySectionItem = ({
       setError("Query failed to run");
     } finally {
       setIsLoading(false);
-      setHasLoaded(true);
+      setHasCompletedInitialLoad(true);
     }
   }, [queryUid, onloadArgs.extensionAPI]);
 
   useEffect(() => {
-    if (isOpen && !hasLoaded) {
+    if (isOpen && !hasCompletedInitialLoad) {
       void loadResults();
     }
-  }, [isOpen, hasLoaded, loadResults]);
+  }, [isOpen, hasCompletedInitialLoad, loadResults]);
 
   const handleChevronClick = async () => {
     if (!section.settings) return;
@@ -460,7 +460,7 @@ const QuerySectionItem = ({
         onloadArgs={onloadArgs}
       />
     ));
-  } else if (hasLoaded) {
+  } else if (hasCompletedInitialLoad) {
     body = <div className="pl-8 pr-2.5 text-sm text-gray-500">No results</div>;
   }
 
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -30,11 +30,13 @@ import {
   settingKeys,
 } from "~/components/settings/utils/settingsEmitter";
 import {
+  isQueryBlockRef,
   type LeftSidebarConfig,
   type LeftSidebarPersonalSectionConfig,
   mergeGlobalSectionWithAccessor,
   mergePersonalSectionsWithAccessor,
 } from "~/utils/getLeftSidebarSettings";
+import runQuery from "~/utils/runQuery";
 import { sectionsToBlockProps } from "./settings/LeftSidebarPersonalSettings";
 import discourseConfigRef, { notify } from "~/utils/discourseConfigRef";
 import { getLeftSidebarSettings } from "~/utils/getLeftSidebarSettings";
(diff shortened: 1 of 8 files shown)


sid597:
@coderabbitai full review

sid597:
@codex review

apps/roam/src/components/settings/LeftSidebarPersonalSettings.tsx
```diff
@@ -337,6 +353,84 @@ const SectionItem = memo(
       setChildInputKey((prev) => prev + 1);
     }, []);
 
+    const handleAliasChange = useCallback(
```
mdroidian, line 356: Does this handle storing in block props as well?

apps/roam/src/components/settings/LeftSidebarPersonalSettings.tsx
```diff
@@ -349,6 +443,45 @@ const SectionItem = memo(
       (!section.settings && section.children?.length === 0) ||
       !section.children;
 
+    if (isQuery) {
+      return (
+        <div
+          className="personal-section rounded-md p-3 hover:bg-gray-50"
+          style={{ border: "1px solid rgba(51, 51, 51, 0.2)" }}
```
mdroidian: Any specific reason for using a custom color for the border?  If not, use tailwind.  If so, note where this style is coming from.

apps/roam/src/components/LeftSidebarView.tsx
```diff
@@ -362,6 +364,171 @@ const PersonalSectionItem = ({
...
+    0,
+    Math.trunc(section.settings?.resultLimit?.value ?? 10),
+  );
+
+  const [isOpen, setIsOpen] = useState<boolean>(
+    !!section.settings?.folded.value,
+  );
+  const [results, setResults] = useState<ChildNode[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [hasLoaded, setHasLoaded] = useState(false);
```
mdroidian: Why do we need variables tracking loaded state?

apps/roam/src/components/LeftSidebarView.tsx
```diff
@@ -362,6 +364,171 @@ const PersonalSectionItem = ({
...
+                  onClick={() => {
+                    void loadResults();
+                    setIsMenuOpen(false);
+                  }}
+                />
+                <MenuItem
+                  icon="document-open"
+                  text="Go to query block"
+                  onClick={() => {
+                    void window.roamAlphaAPI.ui.mainWindow.openBlock({
```
mdroidian: Should handle shift click open in sidebar as well.  Check if there is a function we have reused for this.

apps/roam/src/utils/getLeftSidebarSettings.ts
```diff
@@ -17,6 +22,19 @@ type LeftSidebarPersonalSectionSettings = {
   uid: string;
   truncateResult: IntSetting;
   folded: BooleanSetting;
+  alias?: StringSettingWithValueUid;
+  resultLimit?: IntSetting;
+};
+
+const BLOCK_REF_FULL_MATCH = new RegExp(`^${BLOCK_REF_REGEX.source}$`);
+const QUERY_BLOCK_MARKER = /\{\{query block(?::[^}]*)?\}\}/;
```
mdroidian: This should exist in a few places already.  Try to see if we can DRY this.

apps/roam/src/utils/getLeftSidebarSettings.ts
```diff
@@ -17,6 +22,19 @@ type LeftSidebarPersonalSectionSettings = {
...
+
+export const isQueryBlockRef = (text: string): boolean => {
```
mdroidian, line 48: How does the current query block observer handle this?  Can we reuse that logic instead of writing it again?

apps/roam/src/utils/getExportSettings.ts
```diff
@@ -81,6 +81,20 @@ export const getUidAndStringSetting = (props: Props): StringSetting => {
   };
 };
 
+export type StringSettingWithValueUid = StringSetting & { valueUid?: string };
+
+export const getUidAndStringSettingWithValueUid = (
```
mdroidian: Why is this being added and exported from `getExportSettings`?
It looks like it is only used in `getLeftSidebarSettings`.

mdroidian (approved):
A couple possible blockers:

- Could you point to me where the alias is being store in block props?  I see `createBlock` a few places, but no check/handle of `use new store settings`. 
- we should reuse the query builder observer regex/observer checking logic if possible for `{{query block}}`

apps/roam/src/components/settings/LeftSidebarPersonalSettings.tsx
sid597, line 63: @mdroidian For block props we don't do per-setting writes. We map the whole personal left-sidebar settings structure here because we have to update the data locally for the component, but also persist that same structure in block props.

Below, around line 81, we have the `syncAllSectionsToBlockProps` function  which writes this mapped structure to block props via `setPersonalSetting`. For this query alias path, we build `nextSections`, update local state with `setSections(nextSections)`, and then call `syncAllSectionsToBlockProps(nextSections)` around line 426.

apps/roam/src/components/settings/LeftSidebarPersonalSettings.tsx
sid597: yes on line 426 we call ` syncAllSectionsToBlockProps(nextSections);` which updates the block props

apps/roam/src/components/LeftSidebarView.tsx
sid597: `isLoading` shows “Loading…” while the query is running.

  `hasLoaded` tells us whether the query has run at least once. `results` starts as `[]`, but `[]` can also mean the query ran and returned no results. This lets us avoid showing “No results” before the query has run, and avoids rerunning the query every time the section is closed and opened again.

apps/roam/src/utils/getLeftSidebarSettings.ts
sid597: the existing spots:

  - `parseResultSettings.ts:41` has `/{{query block:(.*?)}}/`, but that only handles `{{query block:alias}}`.
  - `listActiveQueries.ts:8` checks `{{query block}}`, but not `{{query block:alias}}`.
  - `resolveQueryBuilderRef.ts:11` checks one known alias: `{{query block:${queryRef}}}`.
  - `initializeObserversAndListeners.ts:146-147` uses `createButtonObserver({ attribute: "query-block" })`, but that is for rendered DOM buttons, not raw block text.


So there is no existing exported helper that covers both raw block forms: `{{query block}}` and `{{query block:alias}}`.

apps/roam/src/components/LeftSidebarView.tsx
mdroidian: Can we rename one of the variables to be more clear then?  It sounds like you want to track "initial load" or something similar.

apps/roam/src/utils/getLeftSidebarSettings.ts
sid597: The current observer watches the button Roam displays for `{{query block}}` blocks (`initializeObserversAndListeners.ts:146-149`). In this sidebar path we only have the saved section text, usually a block ref like `((uid))`, and need to decide whether to render `QuerySectionItem` before rendering the sidebar row.


So we dereference the block ref with `getTextByBlockUid` and check the block text.

apps/roam/src/utils/getLeftSidebarSettings.ts
mdroidian: `listActiveQueries` should be checking for both.  We could reuse it there.

sid597: you mean to fix `listActiveQueries` and then use that here?

If we do that then there is circular dependency and need more refactoring to circumvent that do we want to do that here? I am fine either way.

mdroidian: What is the circular dependency?

sid597: -> Modify `listActiveQueries.ts`
-> Import it into `getLeftsidebarSettings.ts`

Now the import path is: 

```
  getLeftSidebarSettings.ts
    imports listActiveQueries.ts

  listActiveQueries.ts
    imports getQueryPages from QueryPagesPanel.tsx

  QueryPagesPanel.tsx
    imports getPersonalSetting from accessors.ts

  accessors.ts
    imports getLeftSidebarSettings.ts
```

mdroidian: Ah, no, just export and reuse the `QUERY_BLOCK_MARKER`

sid597: Should listActiveQueries.ts find both the normal query blocks and query blocks with aliases? currently it only searches for the `{{query block}}` ?

mdroidian: yes: DiscourseGraphs/discourse-graph#1056

PR #1057, opened 2026-05-20
ENG-1701: basic-group-management-ui

maparent:
https://linear.app/discourse-graphs/issue/ENG-1701/basic-group-management-ui

https://www.loom.com/share/2754a7979de04995baa775da5b23c898


maparent:
Interesting anomaly: I had lint errors from the GitHub task of the form
```Linting changed files in apps/website
Warning:   38:20  warning  Promise-returning function provided to attribute where a void return was expected  @typescript-eslint/no-misused-promises```
(Because I was using async functions directly as actions instead of wrapping them with a void.)
But those lint errors did not show up in a local `pnpm run lint`. Are the settings different in any way?

PR #1070, opened 2026-05-22
ENG-1791 Obsidian sync does not recover from Concept upsert failure

maparent:
https://linear.app/discourse-graphs/issue/ENG-1791/obsidian-sync-does-not-recover-from-concept-upsert-failure

https://www.loom.com/share/6c7c93bb2def43bd93722ab1cafd6938


apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -339,6 +343,29 @@ const buildChangedNodesFromNodes = async ({
...
+      supabaseClient
+        .from("my_concepts")
+        .select("source_local_id")
+        .eq("space_id", context.spaceId)
+        .eq("arity", 0)
+        .eq("is_schema", false)
+        .order("id"),
+      1000,
+    );
+    if (Array.isArray(existingConceptIds)) {
```
trangdoan982, line 358: sorry for the late review. i'll approve first to unblock for this particular bug.
but given our latest convos about wrapping upsert_concept and upsert_content in one transaction,  I wonder if we should do a more thorough missing check for both concepts and content -> upload whatever is missing. because otherwise we're doing a patch for a bug without being systematic about the root cause.
we can mark it as a clear TODO here for the next step.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -339,6 +343,29 @@ const buildChangedNodesFromNodes = async ({
     context.spaceId,
   );
   const changedNodes: ObsidianDiscourseNodeData[] = [];
+  let missing: Set<string> | undefined;
```
trangdoan982: nit: name to missing_concepts for clarify.
we should also add a comment explaining how this address a particular bug

trangdoan982 (approved):

PR #1071, opened 2026-05-22
ENG-1738: Render advanced search results as sidebar block

trangdoan982:
https://www.loom.com/share/08c356bb7cd841868b34bcea39cbab3b

## Summary
- wire the advanced search footer action and `Option + Enter` shortcut to open a dedicated sidebar results block instead of opening individual pages
- create a single summary block (`Advanced search results: \"{query}\"`) with child wikilink rows for each currently displayed result
- simplify the search dialog layout by removing preview-pane rendering and keeping results-focused list behavior

## Test plan
- [x] Open `DG: Open Node Search`, search for a query, press `Option + Enter`, and confirm a single right-sidebar block opens
- [x] Verify the new sidebar block title includes the query and child blocks are wikilinks for all displayed results
- [x] Verify `Open search sidebar` in footer triggers the same behavior as `Option + Enter`


mdroidian:
@codex review

mdroidian (changes requested):
This is quite different than the functionality I was expecting.

On Apr 23 and May 7th, we discussed replicating the same functionality as the Roam native search which is to render the search as a component in the sidebar.  But it looks like this did not make it into the scope doc/ticket.  Upon looking at the scope doc/ticket, I can see where the confusion could have arose: "The active result opens in the main panel or side bar accordingly".  This is quite ambiguous. 

I believe this will be quite confusing to users based on the wording and it will leave them with multiple backlinks to clean up.  I don't see the benefit of creating a page/blocks in this case as search is generally ephemeral.

That being said, there is a similar pattern in query builder called "Share To" where a user can insert search results as references to a given page, which may be something users request.  

But for this use case let's **change it to a rendered component** or **bring it up in the next Roam meeting** / All Hands for clarification.

trangdoan982:
new test: https://www.loom.com/share/815e68dbb06643b987ec786e75730122

mdroidian:
@codex review

apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchSidebarPanel.tsx
```diff
@@ -0,0 +1,410 @@
...
+  });
+  const selectedNodes = isTypeFilterActive
+    ? discourseNodes.filter((node) => selectedNodeTypeIds.includes(node.type))
+    : [];
+  const showSort = isNonDefaultSort(sort);
+
+  if (!isTypeFilterActive && !showSort) return null;
+
+  return (
+    <div className="dg-node-search-sidebar__filters mb-1 ml-[9px] mr-2 flex flex-col gap-1.5">
```
mdroidian: does roam accept ml-[9px]?

apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchSidebarPanel.tsx
```diff
@@ -0,0 +1,410 @@
...
+      {isTypeFilterActive && (
+        <div className="mt-1 flex flex-wrap items-center gap-1">
+          <span className="shrink-0 text-[0.8em] text-gray-500">
```
mdroidian: does roam accept text-[0.8em]?

apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchSidebarPanel.tsx
```diff
@@ -0,0 +1,410 @@
...
+                  style={{ backgroundColor: getNodeIndicatorColor(node) }}
+                />
+                {node.text}
+              </span>
+            </Tag>
+          ))}
+        </div>
+      )}
+      {showSort && (
+        <p className="text-[0.8em] text-gray-500">
```
mdroidian: does roam accept text-[0.8em]?

apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchSidebarPanel.tsx
```diff
@@ -0,0 +1,410 @@
...
+      <div className="rm-reference-main rm-search-query-content">
+        {indexError ? (
+          <NonIdealState
+            icon="error"
+            title="Search unavailable"
+            description="Reload the extension and try again."
+          />
+        ) : (
+          <>
+            <span className="ml-[9px] text-[0.9em]">
```
mdroidian: Does roam accept these classnames?

apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchSidebarPanel.tsx
```diff
@@ -0,0 +1,410 @@
...
+  const showSort = isNonDefaultSort(sort);
+
+  if (!isTypeFilterActive && !showSort) return null;
+
+  return (
+    <div className="dg-node-search-sidebar__filters mb-1 ml-[9px] mr-2 flex flex-col gap-1.5">
+      {isTypeFilterActive && (
+        <div className="mt-1 flex flex-wrap items-center gap-1">
+          <span className="shrink-0 text-[0.8em] text-gray-500">
+            Filtered to
```
mdroidian: Change to `Filtered by` or `Filter:`

apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchSidebarPanel.tsx
```diff
@@ -0,0 +1,410 @@
...
+
+    setResults(sortSearchResults({ hits: scoredHits, sort }));
+  }, [
+    debouncedSearchTerm,
+    indexError,
+    isDockedQuery,
+    isIndexLoading,
+    selectedNodeTypeIds,
+    sort,
+  ]);
```
mdroidian: **Suggestion: derive search results during render instead of syncing with `useEffect`**

Both `AdvancedSearchDialog` and `AdvancedSearchSidebarPanel` use an effect to run `searchIndexedNodes` → `sortSearchResults` → `setResults` whenever the debounced query, sort, filters, or index readiness change. That pattern causes an extra render on every search update (render → effect → `setState` → render).

Since search is synchronous, debounced, and capped at 50 results, this looks like derived data rather than a side effect. Consider replacing `results` state + the search `useEffect` with a `useMemo` (or inline derivation):

```tsx
const results = useMemo(() => {
  if (!debouncedSearchTerm) return [];
  if (isIndexLoading || indexError || !miniSearchRef.current) {
    // Sidebar only: keep docked placeholder while index loads
    return isDockedQuery ? dockedResults : [];
  }
  const scoredHits = searchIndexedNodes({ /* ... */ });
  return sortSearchResults({ hits: scoredHits, sort });
}, [debouncedSearchTerm, isIndexLoading, indexError, selectedNodeTypeIds, sort, /* sidebar: isDockedQuery, dockedResults */]);
```

**Benefits**
- One render per input change instead of two
- Search logic lives in one place; no effect/state drift
- Same behavior, simpler mental model (“results = f(query, index, filters, sort)”)

**Keep as `useEffect`**
- Async index build (`buildSearchIndex`) — that’s a real side effect
- Debounce timer — still belongs in an effect (or `useDebouncedValue` if you add one)

**Optional follow-up:** store the built index in state (`searchIndex`) instead of refs so the memo depends on real reactive data rather than `isIndexLoading` as a proxy for “refs are populated.” Could also extract a shared `useAdvancedNodeSearchResults` hook used by both dialog and sidebar.

apps/roam/src/utils/openDgSearchInSidebar.tsx
```diff
@@ -0,0 +1,160 @@
+import React from "react";
```
mdroidian: This file works as a bridge into Roam’s sidebar, but it’s doing a lot of imperative DOM/React bootstrapping in a `utils/` module, and the control flow is hard to follow. A few concerns:

**Architecture**
- This isn’t really a util — it renders React, keeps module-level mount state, and pokes Roam’s DOM. Consider moving it next to `AdvancedNodeSearchDialog/` (e.g. `mountAdvancedSearchInSidebar.tsx`) and keeping `utils/` for pure helpers.
- `DockedSearchState` lives here while `AdvancedSearchSidebarPanel` imports it from this file, and this file imports the panel — awkward coupling. The type should live in shared search types (`utils.ts` or `types.ts` in that folder).

**Behavior**
- When the sidebar is already open, we always hijack `getLatestOutlineWrapper()` and wipe it with `innerHTML = ""`. That destroys whatever outline pane the user had there. If intentional, worth a comment; otherwise consider always adding a dedicated “DG node search” window.
- Advanced search already has two sidebar paths: Alt+Enter docks via this hijack flow; Shift+Enter on a result uses plain `rightSidebar.addWindow`. Those should probably share one navigation helper.

**Fragility**
- Sidebar “open” is inferred from `#right-sidebar` width > 40px.
- We wait for new panes by counting `.rm-sidebar-outline-wrapper` nodes — could attach to the wrong pane if anything else adds a window concurrently.
- Title is set via `.window-headers span[style*='font-weight']` — brittle against Roam UI changes.
- When `rightSidebar.open()` exists, `openRightSidebar` returns `null` and execution falls through to the “find latest wrapper” path. Might work today, but it’s implicit and easy to break.

**Cleanup**
- `unmountSidebarSearch` is a module singleton with no public teardown if the user closes the pane in Roam. `SIDEBAR_ROOT_ID` is assigned but never queried — cleanup relies entirely on the closure, not the DOM id.

**Smaller nits**
- Mix of `../components/...` and `~/...` import styles.

mdroidian (changes requested):
The biggest blocking change is the issue we discussed in our 1:1 of the sidebar hijacking the last window instead of creating it's own window to use/persist.  We mentioned that bringing this (expected way to load custom windows in sidebar) up in Roam dev slack would be a good idea and suggesting a `custom` sidebar window type.

trangdoan982:
## New docked search results update:
- each time we dock search results, we add a new rightSidebar window, type: "search-query"
- id for each rightSidebar window is stored in localStorage
- then we use this id to render DG Node search result, hence allowing users to persist the results in the sidebar.

**Manual test checklist**
1. Dock search → confirm a **new** sidebar window (not hijacking the last one)
2. Filter/sort/type in sidebar → panel stays mounted
3. Close window → reload → no ghost restore
4. Leave window open → reload → panel remounts with saved query/filters
5. Collapse/expand right sidebar → panel still works
Test persistence:
https://www.loom.com/share/c3ed8df9a0a645539cc6f67d8e147478

mdroidian:
@codex review

mdroidian (approved):

PR #1072, opened 2026-05-23
ENG-1731: Add keyboard-only filtering with tag chips

trangdoan982:
https://www.loom.com/share/dd5b9491a87b436395a5fb2344b51c24

## Summary
- replace the advanced search text input with a keyboard-friendly chip input for node-type filters
- add ghost completion + `Tab` commit behavior (prefix-only) and chip keyboard navigation/removal
- keep chip state synchronized with the existing type filter dropdown and support chips-only result filtering

## Test plan
- [x] Open advanced search in Roam and verify ghost completion appears for unique node-type prefix
- [x] Press `Tab` to commit a ghost completion into a chip and confirm search term clears
- [x] Verify Backspace/Arrow Left-Right chip navigation/removal behavior without mouse
- [x] Toggle filters via dropdown and confirm chips/results stay in sync

### DIFF
--- a/apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchDialog.tsx
+++ b/apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchDialog.tsx
@@ -482,27 +482,25 @@ const AdvancedNodeSearchDialog = ({
               selectedTypeIds={selectedNodeTypeIds}
             />
           </div>
-          <div className="flex h-9 shrink-0 items-center gap-1">
-            <DiscourseNodeTypeFilter
-              layoutAnchorKey={selectedNodeTypeIds.length}
-              nodeTypes={discourseNodes}
-              onPopoverOpenChange={setIsTypeFilterPopoverOpen}
-              onSelectedTypeIdsChange={setSelectedNodeTypeIds}
-              selectedTypeIds={selectedNodeTypeIds}
-            />
-            <DiscourseNodeSortControl
-              disabled={isIndexLoading || indexError}
-              onSortChange={handleSortChange}
-              sort={sort}
-            />
-            <Button
-              className="shrink-0"
-              icon="cross"
-              minimal
-              onClick={onClose}
-              title="Close search"
-            />
-          </div>
+          <DiscourseNodeTypeFilter
+            layoutAnchorKey={selectedNodeTypeIds.length}
+            nodeTypes={discourseNodes}
+            onPopoverOpenChange={setIsTypeFilterPopoverOpen}
+            onSelectedTypeIdsChange={setSelectedNodeTypeIds}
+            selectedTypeIds={selectedNodeTypeIds}
+          />
+          <DiscourseNodeSortControl
+            disabled={isIndexLoading || indexError}
+            onSortChange={handleSortChange}
+            sort={sort}
+          />
+          <Button
+            className="shrink-0"
+            icon="cross"
+            minimal
+            onClick={onClose}
+            title="Close search"
+          />
         </div>
         <div className="flex min-h-0 w-full flex-1 overflow-hidden">
           {showSplitView ? (
--- a/apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchDialog.tsx
+++ b/apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchDialog.tsx
@@ -365,18 +365,21 @@ const AdvancedNodeSearchDialog = ({
     onClose();
   }, [activeResult, contentState, onClose]);
 
-  const onKeyDown = useCallback(
-    (event: React.KeyboardEvent<HTMLDivElement>) => {
-      if (event.defaultPrevented) return;
+  const handleSearchKeyDown = useCallback(
+    (event: React.KeyboardEvent): void => {
       if (event.key === "ArrowDown" && results.length) {
         event.preventDefault();
         setActiveIndex((index) =>
           Math.min(Math.max(index, 0) + 1, results.length - 1),
         );
-      } else if (event.key === "ArrowUp" && results.length) {
+        return;
+      }
+      if (event.key === "ArrowUp" && results.length) {
         event.preventDefault();
         setActiveIndex((index) => Math.max(index - 1, 0));
-      } else if (
+        return;
+      }
+      if (
         event.key === "Enter" &&
         event.altKey &&
         contentState === "results" &&
@@ -394,7 +397,9 @@ const AdvancedNodeSearchDialog = ({
         event.preventDefault();
         if (event.shiftKey) void onOpenInSidebar();
         else void onOpen();
-      } else if (
+        return;
+      }
+      if (
         event.key === "Enter" &&
         (event.metaKey || event.ctrlKey) &&
         contentState === "results" &&
@@ -403,7 +408,9 @@ const AdvancedNodeSearchDialog = ({
       ) {
         event.preventDefault();
         void onInsert();
-      } else if (event.key === "Escape") {
+        return;
+      }
+      if (event.key === "Escape") {
         if (isTypeFilterPopoverOpen) return;
         event.preventDefault();
         onClose();
@@ -421,6 +428,14 @@ const AdvancedNodeSearchDialog = ({
       onOpenInSidebar,
       results.length,
     ],
+  );
+
+  const onKeyDown = useCallback(
+    (event: React.KeyboardEvent<HTMLDivElement>) => {
+      if (event.defaultPrevented) return;
+      handleSearchKeyDown(event);
+    },
+    [handleSearchKeyDown],
   );
 
   const showSplitView = contentState === "results";
@@ -447,10 +462,7 @@ const AdvancedNodeSearchDialog = ({
         className="flex min-h-0 flex-1 flex-col overflow-hidden"
       >
         <div className="flex w-full flex-none items-start gap-2 border-b border-gray-200 px-3 py-2">
-          <div
-            className="flex min-h-9 w-full min-w-0 flex-1 items-center rounded border border-gray-300 bg-white px-2 py-1"
-            style={{ flex: "1 1 0", minWidth: 0 }}
-          >
+          <div className="flex min-h-9 min-w-0 flex-1 items-center rounded border border-gray-300 bg-white px-2 py-1">
             <Icon
               className="mr-2 shrink-0 self-center text-gray-500"
               icon="search"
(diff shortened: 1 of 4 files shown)


mdroidian:
@codex review

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+  const nodeTypeById = useMemo(
+    () =>
+      Object.fromEntries(
+        nodeTypes.map((nodeType) => [nodeType.type, nodeType]),
+      ),
+    [nodeTypes],
+  );
+
+  const selectedNodeTypes = useMemo(
```
mdroidian: Why does this need a memo?

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+  onArrowUp,
+  onEnter,
+  onShiftEnter,
+  onCmdEnter,
+  onEscape,
+}: NodeTypeChipsSearchInputProps): React.ReactElement => {
+  const [focusedChipIndex, setFocusedChipIndex] = useState(-1);
+  const chipRefs = useRef<(HTMLSpanElement | null)[]>([]);
+
+  const nodeTypeById = useMemo(
```
mdroidian: Why does this need a memo?

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+    () =>
+      getBestPrefixMatch({
+        nodeTypes,
+        query: searchTerm,
+        selectedTypeIds,
+      }),
+    [nodeTypes, searchTerm, selectedTypeIds],
+  );
+
+  const completionSuffix = useMemo(() => {
```
mdroidian: Why does this need a memo?

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+        {completionSuffix && (
+          <span className="pointer-events-none absolute inset-0 flex items-center overflow-hidden whitespace-nowrap text-sm">
+            <span className="invisible">{searchTerm}</span>
+            <span className="text-gray-400">{completionSuffix}</span>
+            <span className="ml-2 rounded bg-gray-100 px-1 text-[10px] uppercase tracking-wide text-gray-500">
+              Tab
+            </span>
+          </span>
+        )}
+        <input
```
mdroidian, line 276: Can we use blueprint TagInput or InputGroup for this?

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+              borderRadius: 3,
+            }}
+          >
+            <span
+              className="inline-flex items-center gap-1.5 rounded px-2 py-1 text-xs"
+              style={getNodeTagStyles(
+                nodeType.canvasSettings?.color ?? "#000000",
+              )}
+            >
+              <span className="max-w-[10rem] truncate leading-4">
```
mdroidian: Does Roam handle arbitrary width values?

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+                {nodeType.text}
+              </span>
+              <Button
+                className="!h-4 !min-h-0 !w-4 !min-w-0 !p-0"
```
mdroidian: Does Roam handle `!`?

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+      onEscape();
+    }
+  };
+
+  return (
+    <div className="flex min-w-0 flex-1 flex-wrap items-center gap-1 py-0.5">
+      {selectedNodeTypes.map((nodeType, index) => {
+        const isFocused = focusedChipIndex === index;
+        return (
+          <span
```
mdroidian, line 223: Can we use blueprint tags for this?

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+            </span>
+          </span>
+        );
+      })}
+      <span className="relative min-w-[12ch] flex-1">
+        {completionSuffix && (
+          <span className="pointer-events-none absolute inset-0 flex items-center overflow-hidden whitespace-nowrap text-sm">
+            <span className="invisible">{searchTerm}</span>
+            <span className="text-gray-400">{completionSuffix}</span>
+            <span className="ml-2 rounded bg-gray-100 px-1 text-[10px] uppercase tracking-wide text-gray-500">
```
mdroidian: Does Roam handle `text-[arbitrary value]`?

mdroidian (changes requested):
A few things I noticed with this PR that I think we should address:

**Input box alignment**
It looks like this PR introduces a misalignment of the search box with the icons.  I would consider this a regression that needs to be fixed:

Image


**Not using Blueprint Tags**
Did we try using Blueprint [Tag](https://blueprintjs.com/docs/versions/3/#core/components/tag), [TagInput](https://blueprintjs.com/docs/versions/3/#core/components/tag-input), or [InputGroup](https://blueprintjs.com/docs/versions/3/#core/components/text-inputs.input-group) for this?  Were there some issues stopping us from using the components that Roam style guide recommends?

**Tailwind**
As discussed in our last meeting, we noting how Roam doesn't handle all of the tailwind class syntax: specifically arbitrary values.  So let's do a quick run through to double check.

trangdoan982:
Addressed in 542ac305:

- **Search bar alignment:** Header row uses `items-center` again (matching eng-1730); removed `self-start` wrappers on filter/sort/close controls.
- **Tailwind / Roam:** Replaced arbitrary/`!` classes with inline styles (`CHIP_LABEL_MAX_WIDTH`, `INPUT_MIN_WIDTH`) and `text-xs` for the Tab hint.
- **useMemo:** Removed; lookups are plain derived values / small helpers.
- **ArrowDown guard:** `onArrowDown` / dialog `onKeyDown` no-op when `results.length === 0`, and index clamps with `Math.max(index, 0)`.

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
trangdoan982: TagInput won't be able to help us render the Ghost Tag component like this
image

It also doesn't have the flexible keyboard navigation to select previous tags and delete them like we built here.

But I did switch the tag to use BlueprintJS <Tag/> instead. now the tags look like this:
image

mdroidian: Could you explain what you mean re: flexible keyboard nav.  

> Tags can be removed by clicking their  buttons, or by pressing either backspace or delete repeatedly. Pressing delete mimics the behavior of deleting in a text editor, where trying to delete at the end of the line will do nothing. Arrow keys can also be used to focus on a particular tag before removing it. The cursor must be at the beginning of the text input for these interactions.

This is from the blueprint3 docs, and seems to suggest otherwise.

apps/roam/src/components/AdvancedNodeSearchDialog/AdvancedSearchDialog.tsx
```diff
@@ -416,19 +424,46 @@ const AdvancedNodeSearchDialog = ({
...
-            onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
-              setSearchTerm(event.target.value)
-            }
-            placeholder="Search discourse nodes..."
-            value={searchTerm}
-          />
+        <div className="flex w-full flex-none items-start gap-2 border-b border-gray-200 px-3 py-2">
+          <div
+            className="flex min-h-9 w-full min-w-0 flex-1 items-center rounded border border-gray-300 bg-white px-2 py-1"
+            style={{ flex: "1 1 0", minWidth: 0 }}
```
mdroidian: use class `flex-1`

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,323 @@
...
+  selectedTypeIds: string[];
+  inputRef: React.RefObject<HTMLInputElement>;
+  onSearchTermChange: (value: string) => void;
+  onSelectedTypeIdsChange: (ids: string[]) => void;
+  onArrowDown: () => void;
+  onArrowUp: () => void;
+  onEnter: () => void;
+  onShiftEnter: () => void;
+  onCmdEnter: () => void;
+  onEscape: () => void;
```
mdroidian: Unless we plan to reuse this component with different values, then including these as props is unnecessary.

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,323 @@
...
+    if (event.key === "Escape") {
+      event.preventDefault();
+      onEscape();
+    }
+  };
+
+  return (
+    <div
+      className="flex w-full min-w-0 flex-1 flex-wrap items-center gap-1 py-0.5"
+      style={{ flex: "1 1 0", minWidth: 0, width: "100%" }}
```
mdroidian: flex-1 w-full

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,323 @@
...
+              chipRefs.current[index] = element;
+            }}
+            role="button"
+            tabIndex={-1}
+            onClick={() => setFocusedChipIndex(index)}
+            onKeyDown={(event) => handleChipKeyDown(event, index)}
+            style={{
+              borderRadius: 3,
+              boxShadow: isFocused
+                ? "0 0 0 2px rgba(95, 87, 192, 0.2)"
```
mdroidian: Where is this styling coming from?  Does it align with other styling we use, Blueprint3 uses, or that Roam uses?

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
```diff
@@ -0,0 +1,323 @@
...
+        {completionSuffix && (
+          <span className="pointer-events-none absolute inset-0 flex items-center overflow-hidden whitespace-nowrap text-sm">
+            <span className="invisible">{searchTerm}</span>
+            <span className="text-gray-400">{completionSuffix}</span>
+            <span className="ml-2 rounded bg-gray-100 px-1 text-xs uppercase tracking-wide text-gray-500">
+              Tab
+            </span>
+          </span>
+        )}
+        <input
```
mdroidian, line 276: I would **strongly** recommend using `TagInput` or `InputGroup` here.  If it doesn't work for some reason, please document said reason.

mdroidian (approved):

apps/roam/src/components/AdvancedNodeSearchDialog/NodeTypeChipsSearchInput.tsx
trangdoan982: oh yes you're right. There was a prop I didn't know `interactive=true` to focus on different tags. let me use that. But I don't think we can do suggestive (the ghost "claim [tab]" when typing "cl") with either Tag or TagInput

mdroidian: Why not?

PR #1074, opened 2026-05-24
ENG-1767 Setup Linear release for Obsidian

trangdoan982:
https://www.loom.com/share/1c2c8bbce50a4d3d8cfe727c2d5edfb3

## Summary
- add a manual Obsidian release workflow that syncs releases to Linear, runs the existing `apps/obsidian/scripts/publish.ts` flow, and updates the Linear stage to `Sent to Obsidian for Review`
- set required CI environment variables for the publish flow, including `SUPABASE_USE_DB=production`, and keep monorepo `include_paths` filtering for Linear sync
- add a companion manual workflow to complete an Obsidian Linear release by version

## Test plan
- [x] run `Publish Obsidian Plugin` from GitHub Actions with a test version and verify `Sync Linear release` succeeds
- [x] verify the publish step runs successfully and creates/updates release artifacts in the Obsidian plugin repo
- [x] verify `Mark Linear release sent to Obsidian review` updates the target release stage in Linear
- [ ] run `Complete Obsidian Linear Release` with the same version and verify the release is marked complete in Linear

### DIFF
--- a/.github/workflows/obsidian-release-complete.yaml
+++ b/.github/workflows/obsidian-release-complete.yaml
@@ -16,19 +16,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Debug complete release inputs
-        env:
-          GITHUB_RUN_ID: ${{ github.run_id }}
-          GITHUB_EVENT_NAME: ${{ github.event_name }}
-          GITHUB_REF_NAME: ${{ github.ref_name }}
-          INPUT_VERSION: ${{ inputs.version }}
-        run: |
-          node -e '
-          // #region agent log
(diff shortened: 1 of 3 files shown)


mdroidian:
@codex review

mdroidian (changes requested):
I want to clarify the intended release flow here.

It sounds like we may be cutting a Linear release and `/discourse-graph-obsidian` release for every PR/feature. Is that correct? I’m assuming this is only for BRAT, not the Obsidian marketplace.

That would differ from the current Roam <> Linear flow, where each PR/feature is tagged with the current release, then when we cut a Roam Depot release, we update the Linear release status to indicate the collected work is being shipped.

Ideally, Roam and Obsidian should follow the same release pattern unless there’s a specific reason they need to differ.

Could you clarify:

* What is the expected process for sending an update to the Obsidian marketplace vs BRAT, especially given the new Obsidian community / AI review process?
* How do the Obsidian release YAMLs differ from the Roam release YAMLs, and why?
* If we cut a Linear release for every PR, how does the team know whether something is going to the Obsidian marketplace vs only to BRAT?
* In the video, you mentioned “release complete YAML here, which is called from Linear.” Where/how is that called from Linear?
* Where does the version number come from, and when/how is it updated?

A doc similar to `apps/roam/RELEASE.md` would probably help make this flow explicit.

trangdoan982:
note https://roamresearch.com/#/app/discourse-graphs/page/sZgJcfd0Q: now create every release on every relevant merge to Obsidian

mdroidian:
\*BRAT release, not linear release.

PR #1075, opened 2026-05-24
ENG-1793 Roam sync does not recover from Concept upsert failure

maparent:
https://linear.app/discourse-graphs/issue/ENG-1793/roam-sync-does-not-recover-from-concept-upsert-failure

https://www.loom.com/share/9cf9602e2c2e4ac1a449055b6e36917b

ETA after loom: undefined was never actually sent to the getAllMissingOrNewDiscourseNodes, which would affect first-ever upload. Corrected. Had some downstream impact, had to make sure getAllDiscourseNodesSince could handle undefined; refactored ISO string to ms number.


sid597:
@codex review

sid597 (approved):
LGTM!

PR #1076, opened 2026-05-24
ENG-1748: Add shared template import panel for node types

trangdoan982:
https://www.loom.com/share/6343ad0f31bf4e37a3cb62c1c53da546

## Summary
- add a dedicated "Import template from groups" flow in Node Type settings with candidate fetching, selection, and one-click import into the local templates folder
- Template Import UI is a split-panel modal with template list metadata (name, optional author, source space, publish time) and live markdown preview
- prevent duplicate imports by naming imported templates as `<template> (from <space>)` and filtering already imported candidates from the panel

## Test plan
- [x] Open Obsidian settings > Node Types > edit a node type > Template dropdown > Import template from groups
- [x] Confirm split layout: left candidate list + right markdown preview, with independent vertical scrolling
- [x] Import a template and verify local template name format is `<template> (from <space>).md`
- [x] Re-open import panel and verify the imported candidate no longer appears
- [x] Click "Import & use" and verify the selected node type template field is updated to the imported file

### DIFF
--- a/apps/obsidian/src/components/NodeTypeSettings.tsx
+++ b/apps/obsidian/src/components/NodeTypeSettings.tsx
@@ -1,11 +1,16 @@
-import { useState, useEffect, useRef } from "react";
+import { useState, useEffect, useRef, useCallback } from "react";
 import { validateNodeFormat, validateNodeName } from "~/utils/validateNodeType";
 import { usePlugin } from "./PluginContext";
-import { Notice, setIcon } from "obsidian";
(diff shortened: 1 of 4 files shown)


maparent (approved):
The code as written looks good to me, and I have tested it.
There are two issues that will make real-life usage more difficult in practice, but we decided that they are a distinct scope and should not block the PR. 
In order of urgency:

https://linear.app/discourse-graphs/issue/ENG-1810/initial-template-sync-migration

https://linear.app/discourse-graphs/issue/ENG-1809/sync-template-setting-if-user-rename-the-template-file

PR #1082, opened 2026-05-25
ENG-1477: Convert a tldraw arrow to a DG relation

sid597:
https://www.loom.com/share/2f7d93f683d74868b0a0193c1d1e7dac

This PR adds arrow-to-relation conversion without changing the existing relation storage decision. Both default relation creation and arrow conversion delegate persistence to the existing canvas relation util, which already chooses stored/reified relations vs legacy triples based on `getStoredRelationsEnabled()`.

  The split here is deliberate:
  - shared helpers for same-question logic: valid relation lookup, directional label, persistence plumbing
  - separate lifecycle functions for different behavior: fresh drag-handle relation creation vs converting an existing arrow

  That keeps the old drag-handle behavior default-centered while allowing conversion to preserve the user’s existing arrow geometry and only delete the original arrow after persistence succeeds.


apps/roam/src/components/canvas/overlays/DragHandleOverlay.tsx
```diff
@@ -256,103 +249,13 @@ export const DragHandleOverlay = () => {
...
-        return;
-      }
-
-      // Bind start and end
-      createOrUpdateArrowBinding(editor, newArrow, pending.sourceId, {
-        terminal: "start",
-        normalizedAnchor: { x: 0.5, y: 0.5 },
-        isPrecise: false,
-        isExact: false,
+      void createDefaultRelationBetweenNodes({
```
sid597, line 259: This replaces the old inline drag-handle relation creation block with the same default-create path in a named helper. The drag flow still creates a fresh relation arrow with default geometry and center bindings; conversion does not use this helper because it has to preserve an existing arrow's geometry.

apps/roam/src/components/canvas/overlays/relationCreation.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    const hexColor =
+      colorPalette[getRelationColor(relation.label)]?.solid ?? "#333";
+    validTypes.push({ id: relation.id, label, color: hexColor });
+  }
+
+  return validTypes;
+};
+
+export const createDefaultRelationBetweenNodes = async ({
```
sid597, line 121: This is intentionally the default/fresh relation creation path. It keeps the drag-handle behavior centered and default-styled; existing-arrow conversion lives separately because that flow has a different lifecycle and rollback contract.

apps/roam/src/components/canvas/overlays/relationCreation.ts
```diff
@@ -0,0 +1,187 @@
...
+import type { DiscourseRelation } from "~/utils/getDiscourseRelations";
+
+type RelationTypeOption = { id: string; label: string; color: string };
+
+type DirectionalRelation = Pick<
+  DiscourseRelation,
+  "label" | "complement" | "source" | "destination"
+>;
+
+export const getDirectionalRelationLabel = ({
```
sid597, line 29: The directional label/complement rule is shared because it is the same question in the dropdown, default creation, and arrow conversion paths. This keeps label selection single-source without merging the different shape-creation lifecycles.

apps/roam/src/components/canvas/uiOverrides.tsx
```diff
@@ -224,6 +237,140 @@ export const getOnSelectForShape = ({
...
+};
+
+const copyArrowBindingProps = (
+  binding: TLArrowBinding,
+): TLArrowBinding["props"] => ({
+  ...binding.props,
+  normalizedAnchor: { ...binding.props.normalizedAnchor },
+});
+
+const convertArrowToRelation = async ({
```
sid597, line 290: Conversion is intentionally separate from default relation creation. Here we transform an existing user-drawn arrow, so we preserve layout/bindings and only delete the original after relation persistence succeeds.

apps/roam/src/components/canvas/uiOverrides.tsx
```diff
@@ -224,6 +237,140 @@ export const getOnSelectForShape = ({
...
+    id: relationArrowId,
+    type: relationId,
+    parentId: arrow.parentId,
+    x: arrow.x,
+    y: arrow.y,
+    rotation: arrow.rotation,
+    opacity: arrow.opacity,
+    isLocked: arrow.isLocked,
+    meta: { ...arrow.meta },
+    props: {
```
sid597, line 327: Only geometry is inherited from the plain arrow (`bend`, terminals, label position). The DG relation visual contract is forced here so converted relations show the same direction/style as relations created through the drag-handle flow.

apps/roam/src/components/canvas/overlays/RelationTypeDropdown.tsx
```diff
@@ -25,49 +20,10 @@ export const RelationTypeDropdown = ({
...
-        const tldrawColor = getRelationColor(relation.label);
-        const hexColor = colorPalette[tldrawColor]?.solid ?? "#333";
-        validTypes.push({ id: relation.id, label, color: hexColor });
-      }
-    }
-
-    return validTypes;
-  }, [editor, sourceId, targetId]);
+  const validRelationTypes = useMemo(
+    () => getValidRelationTypesBetween(editor, sourceId, targetId),
```
sid597, line 24: This keeps the dropdown's relation filtering behavior the same, but reuses the same valid-relation lookup that the new arrow context menu needs.

mdroidian:
@codex review

mdroidian (approved):

apps/roam/src/components/canvas/uiOverrides.tsx
```diff
@@ -239,6 +386,22 @@ export const CustomContextMenu = ({
...
+    "arrowRelationOptions",
+    () => {
+      if (!selectedShape || selectedShape.type !== "arrow") return null;
+      const boundNodes = getArrowBoundNodeInfo(editor, selectedShape);
+      if (!boundNodes) return null;
+      const relationTypes = getValidRelationTypesBetween(
+        editor,
+        boundNodes.startId,
+        boundNodes.endId,
+      );
```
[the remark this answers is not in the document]
sid597: Fixed in 067af940. Suppressed self-loop conversion at the chokepoint — added `if (startBinding.toId === endBinding.toId) return null;` in `getArrowBoundNodeInfo`, which both the `arrowRelationOptions` submenu builder and `convertArrowToRelation` flow through. A self-loop arrow now offers no Relation submenu and cannot be converted, so the original arrow is never deleted and no ghost relation is created.

Went with suppression (your first suggestion) rather than deriving the source from the start terminal: that path would persist a node-relates-to-itself record through the shared `handleCreateRelationsInRoam`, which is not a meaningful relation. The guard keys off `toId` (shape instance), so same-type relations between two *distinct* nodes (e.g. Claim A → Claim B) are unaffected.

PR #1083, opened 2026-05-25
ENG-1373: Drag pages onto the canvas to create node shapes

sid597:
https://github.com/user-attachments/assets/85f08f70-a202-4df5-9269-dbad15a809a6


mdroidian:
@codex review

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -641,15 +643,63 @@ const TldrawCanvasShared = ({
 
   // Handle Roam block drag and drop
   useEffect(() => {
+    const temporaryDraggableAttribute = "data-roamjs-canvas-page-ref-draggable";
```
mdroidian: `TEMP_DRAG_ATTR`

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -1267,30 +1333,36 @@ const InsideEditorAndUiContext = ({
...
-            });
-            return;
+            return true;
+          };
+
+          // Check for page reference: [[pageName]]
+          const pageMatch = text.match(PAGE_REF_REGEX);
+          if (pageMatch?.[1]) {
+            if (await tryCreatePageNodeShape(pageMatch[1])) {
+              posthog.capture("Canvas: Node Added from External Content", {
```
mdroidian, line 1607: "External Content" here can be misleading.  Does this mean outside of Roam?

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -1303,6 +1375,14 @@ const InsideEditorAndUiContext = ({
           if (!blockText || !isLive)
             return await callDefaultTextHandler(content);
 
+          const refMatch = blockText.match(PAGE_REF_REGEX);
+          if (refMatch?.[1] && (await tryCreatePageNodeShape(refMatch[1]))) {
+            posthog.capture("Canvas: Node Added from External Content", {
```
mdroidian: "External Content" here can be misleading.  Does this mean outside of Roam?

mdroidian (changes requested):
https://www.loom.com/share/b7fa55afb36a48bd8fabe14391f8fe65

I think this needs one more pass before merging.

The goal is to drag a page reference from inside an outliner block onto the canvas. Right now, mixed-content blocks still open the page instead of allowing drag.

It also looks like this changes the behavior when a page reference is the only content in a block: dragging the block now creates a page node instead of a block node. I’m not sure we want to change that existing block-drag behavior.

Expected behavior: dragging the specific page reference should add that page to the canvas, similar to the discourse context behavior, without changing how the containing block itself drags.

Could you also test/show:

* link/render/embed query views
* multiple canvases, including sidebar canvases, to check `useEffect` behavior

mattakamatsu:
Ping! I wanted to use this feature recently, would love to put it back on @sid597's backlog, thankx

mdroidian (changes requested):
Just listing some discussion points from Friday

* Use **absolute positioning** for the drag handle so it overlays the text instead of causing text reflow on hover.
* Change the drag handle icon color from **red to gray** so it matches the existing icon palette.
* Fix the **post-drag bug** where the handle disappears from other blocks until the next click, likely due to the observer not re-firing.

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -174,6 +174,219 @@ export const MAX_WIDTH = "400px";
...
+  // SPAN-only page-ref observer never sees. Observed locally so anchors are
+  // not fed to its other handlers (overlays, previews).
+  const pageRefAnchorObserver = createHTMLObserver({
+    useBody: true,
+    tag: "A",
+    className: "rm-page-ref",
+    callback: observePageRef,
+  });
+  dragHandle.addEventListener("pointerdown", handleDragHandlePointerDown);
+  dragHandle.addEventListener("mousedown", handleDragHandlePointerDown);
```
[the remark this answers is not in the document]
sid597: the second listener stops mousedown from reaching Roam's block handlers, which stopPropagation on pointerdown can't do; the double call is idempotent

sid597:
@mdroidian Fixed all 3 points and updated the video to use latest code

mdroidian:
@codex review

mdroidian:
> @mdroidian Fixed all 3 points and updated the video to use latest code

Hey @sid597 , there is no mouse in the video so it's really hard to see that all three issues are solved.  I think I made out the second (absolute positioning) but it only displays for a very brief moment.

Did you test the post-drag bug fix yourself?

mdroidian:
@sid597 have you tried dragging an `[[EVD]]` or `[[RES]]`?  How do you handle the multiple referenced pages?

mdroidian (changes requested):
https://www.loom.com/share/4c2add16b45b49b1b2f3e464f56b8394

Change requests:

1. Match the background of the icon to the background of the window (white if main window, gray if sidebar)

2. Move the handle farther right so it sits outside the page reference’s closing `]]` brackets instead of overlapping them.

3. Stabilize handle placement for multiline references.

4. Do not render drag handles for nested references inside another reference, such as “EVD” or “Source” references embedded within a larger “Evidence” reference.

5. Change the mouse pointer to `grabbing` when over the drag icon

A few more issues found
https://www.loom.com/share/00f51beb434349b99d4b1a8dcf661c0d

- change: remove drag handles from titles
- bug: drag handle moves and is persistent when mouse is removed from the left

After making these changes, please ensure the video records a comprehensive manual regression pass.  Eg: demonstrating the drag behavior in the main window, sidebar, Kanban, and query link, embed, and render views. Include single-line, multiline, wrapping, and nested references in each applicable surface.

sid597:
https://www.loom.com/share/5c8ca4494d434736b42adc0397b18838

mdroidian:
@codex review

mdroidian (approved):

PR #1086, opened 2026-05-25
Eng 1721 basic group management UI

maparent:
[ENG-1721: Group membership UI](https://linear.app/discourse-graphs/issue/ENG-1721/group-membership-ui)

https://www.loom.com/share/2754a7979de04995baa775da5b23c898

### DIFF
--- a/apps/website/app/(home)/auth/group/[groupId]/page.tsx
+++ b/apps/website/app/(home)/auth/group/[groupId]/page.tsx
@@ -1,0 +1,78 @@
+import { redirect, notFound } from "next/navigation";
+import Link from "next/link";
+import { createClient } from "~/utils/supabase/server";
+import { getSessionBaseUserData } from "~/utils/supabase/account";
+import { GroupMemberList } from "~/components/auth/GroupMemberList";
+import { GroupInvite } from "~/components/auth/GroupInvite";
+import internalError from "~/utils/internalErrorSsr";
+
+const Page = async ({
+  params,
+  searchParams,
+}: {
+  params: Promise<{ groupId: string }>;
+  searchParams: Promise<{
+    token?: string;
+    tokenError?: string;
+    removeError?: string;
+  }>;
+}) => {
+  const { groupId } = await params;
+  const sp = await searchParams;
+
+  const client = await createClient();
+  const userData = await getSessionBaseUserData(client);
+  if (!userData)
+    redirect("/auth/error?error=" + encodeURIComponent("Not logged in"));
+
+  const membershipReq = await client
+    .from("group_membership")
+    .select("admin")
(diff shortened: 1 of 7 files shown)


maparent:
This is a UI task; there is a fair amount of boilerplate but not as much logic, as it was mostly handled in previous tasks.

mdroidian:
@codex review

apps/website/app/(home)/auth/group/[groupId]/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+    .eq("id", groupId)
+    .maybeSingle();
+  if (groupReq.error) {
+    internalError({ error: groupReq.error });
+    redirect("/auth/error?error=" + encodeURIComponent("Could not load group"));
+  }
+  const groupName = groupReq.data?.name ?? groupId;
+
+  return (
+    <main>
```
Comment (mdroidian) [inline apps/website/app/(home)/auth/group/[groupId]/page.tsx:53]: nit: shouldn't need `<main>` and `<div>`, just use `<main>`

apps/website/app/(home)/auth/group/page.tsx
```diff
@@ -1,11 +1,27 @@
...
+  searchParams: Promise<{
+    error?: string;
+    joined?: string;
+    createError?: string;
+    created?: string;
+  }>;
+}) => {
+  const params = await searchParams;
+  return (
+    <main>
```
mdroidian, line 17: nit: shouldn't need `<main>` and `<div>`, just use `<main>`

mdroidian (approved):

apps/website/app/(home)/auth/group/[groupId]/page.tsx
Comment (maparent) [reply apps/website/app/(home)/auth/group/[groupId]/page.tsx:53]: It does look different.
https://www.loom.com/share/bacbc70343964261877584171aa5ecb3

Comment (mdroidian) [reply apps/website/app/(home)/auth/group/[groupId]/page.tsx:53]: I don’t want to overstate the impact of one extra DOM element. That part is relatively minor.

The concern is that there are two downsides here:

1. The wrapper appears to be affecting layout, but we don’t yet understand why.
2. If the wrapper is not intentionally needed, then it is unnecessary structure, and we should generally avoid unnecessary DOM/layout complexity.

So the main issue is not “never use `<main><div>...`.” That structure can be valid when the wrapper has a clear purpose. The issue is preserving it as an unexplained workaround.

I would suggest checking the parent layout and comparing how `<main>` behaves as a direct child versus how the nested `<div>` behaves inside it. There is likely a specific CSS layout interaction causing the visual difference. Once that is understood, either keeping the wrapper or simplifying to a single `<main>` is fine, but the choice should be intentional.

PR #1087, opened 2026-05-26
eng-1802 update github actions

maparent:
https://linear.app/discourse-graphs/issue/ENG-1802/update-github-actions-was-github-actions-fail

The loom describes the original problem that led to this, though it ended up being spurious.
https://www.loom.com/share/6fff778b55584bd5ad8222ec2c3971e1

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,16 +14,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
-      - uses: pnpm/action-setup@v4
+      - uses: pnpm/action-setup@v6
         name: Install pnpm
         with:
           version: 10.15.1
           run_install: false
 
       - name: Setup Node.js environment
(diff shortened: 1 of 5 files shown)


mdroidian (changes requested):
Thanks for updating these actions.

Seeing as the issue was spurious, I'll treat this as just an update.

Since this is a major-version bump across several workflows, can you add a short compatibility note to the PR description and link to the release notes/changelog for each major bump?

The standard I’d use is: “Can someone reviewing this PR tell why the upgrade is safe without redoing all the research?” If the answer is no, then more context is required.

maparent:
Relevant breaking changes for setup-node:
6.0.0: Limit automatic caching to npm
5.0.0: Enhance caching in setup-node with automatic package manager detection
for pnpm/action-setup, no breaking change indicated. (But they are using node 24, fwiw)

The breaking changes in setup-node are not relevant to our scenario. We do follow the pattern described in the documentation for pnpm caching:
https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data

mdroidian (approved):

PR #1095, opened 2026-05-28
ENG-839: Guard incomplete relation types in canvas

trangdoan982:
https://www.loom.com/share/3977b131b059477595abae91916662f8

## Summary
- Prevent canvas relation creation from crashing when relation/node config is missing or stale.
- Block using incomplete relation types (missing label/complement/source/destination) and show actionable warnings.
- Disable saving incomplete relations in settings and filter incomplete relations from the relation dropdown.

## Test plan
- In Roam canvas, attempt to create a relation when:
  - no relations exist
  - the selected relation type has missing required fields
  - node types referenced by relations were deleted/recreated
- Confirm: no crash; warning toast appears; settings link works; normal creation still works.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1095" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationTool.tsx
+++ b/apps/roam/src/components/canvas/DiscourseRelationShape/DiscourseRelationTool.tsx
@@ -56,24 +56,6 @@ export const createAllReferencedNodeTools = (
         override onEnter = () => {
           this.didTimeout = false;
 
-          const selectedRelations = discourseContext.relations[action] || [];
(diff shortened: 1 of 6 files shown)


mdroidian (approved):
Let's either
- change where we set `"?"` to `""` and double check that doesn't break anything (then remove the `"?"` check

or

- Write this check in one place DRY, so that when we do remove the `"?"` in the future that we don't need to change it in multiple places

Also, I don't think `relation.complement` is ever set to `"?"`

PR #1099, opened 2026-06-01
ENG-1800 Create relation dialog closes when there is no match

maparent:
https://linear.app/discourse-graphs/issue/ENG-1800/create-relation-dialog-closes-when-there-is-no-match

Revised loom:
https://www.loom.com/share/e915bcae39f5473f8cf3bacf73de3ebd

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1099" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/CreateRelationDialog.tsx
+++ b/apps/roam/src/components/CreateRelationDialog.tsx
@@ -202,10 +202,12 @@ const CreateRelationDialog = ({
 
   const changeRelationType = (relName: string): void => {
     setSelectedRelationName(relName);
-    setPageOptions(getFilteredPageNames(relName));
+    const newPageOptions = getFilteredPageNames(relName);
+    setPageOptions(newPageOptions);
     if (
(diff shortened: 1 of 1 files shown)


mdroidian (commented):
Just quickly looked at the code/video and changing the type of setResult to accept string doesn't seem right.<br><br>We use a similar pattern `Result` with the AutocompleteInput component in multiple places in the repo, is this the same pattern we use in those places (eg Canvas, Share via, Query Builder, etc)?  Do those place also experience this same problem?

mdroidian (approved):
I would prefer if we stick to using `Result`, but this works as well I guess.

Also see graphite's comment.

maparent:
re using Result: Then we'd have to either modify AutoCompleteInput so it filters strings outside choices through a function that transforms strings into Results, or use my previous solution.

PR #1119, opened 2026-06-10
ENG-1842 Strip block uids from stored templates to fix template editing

sid597:
https://www.loom.com/share/1b3e8dc6a66c4528a3a7f6eba5f6c2b2

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1119" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/settings/utils/zodSchema.ts
+++ b/apps/roam/src/components/settings/utils/zodSchema.ts
@@ -57,7 +57,6 @@ export const IndexSchema = z.object({
 type RoamNode = {
   text: string;
   children?: RoamNode[];
-  uid?: string;
   heading?: 0 | 1 | 2 | 3;
   open?: boolean;
 };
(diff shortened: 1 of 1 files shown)


apps/roam/src/components/settings/utils/zodSchema.ts
```diff
@@ -57,7 +57,6 @@ export const IndexSchema = z.object({
 type RoamNode = {
   text: string;
   children?: RoamNode[];
-  uid?: string;
```
sid597, line 60: Why removing uid here fixes ENG-1842:

  - RoamNodeSchema is only used for DiscourseNodeSchema.template
  - The optional uid let migration store legacy template trees with their block uids in block props
  - #828's Template panel seeds an ephemeral buffer from that stored template → create-block replays the old uids → Block already exists (the originals still live under legacy Template) → buffer never created → panel dead. Shipped in v0.19.0, hence "suddenly broken"
  - Nothing reads template uids from the store: panel save (serializeBlockTree) and createBlocksFromTemplate (stripTemplateUids) both strip them already
  - Zod drops unknown keys on parse, so removal fixes both directions:
    - read: already-poisoned graphs get clean templates immediately, no data migration
    - write: future migrations store uid-free
  - Alternative was stripping in EphemeralBlocksPanel, but that patches one consumer and leaves the next one to hit the same landmine
  - Legacy mode (flag off) reads the live block tree and never touches this schema; fresh graphs were never poisoned

mdroidian (approved):

PR #1130, opened 2026-06-18
ENG-1656 Rename folder for imported vault

trangdoan982:
https://www.loom.com/share/39deddfaa529440d96484590a8c0d53f

> Note: the Loom predates review feedback — it shows the original `{owner}-{spaceName}` naming plus the auto-migration that has since been dropped.

## Summary

Imported vaults all landed in `import/{spaceName}`, which is just the source vault's name and does not say whose vault it was. This adds the owner to the folder name and moves folder identity off the folder name entirely.

- New import folders are named `{ownerName}-{spaceName}` under `import/`.
- The owner name is resolved from space membership — `LocalAccess` for the source `space_id`, joined to `my_accounts` filtered on `agent_type = "person"`. Node authors are deliberately not used: they tell us who wrote a given node, which can differ from the account the source space belongs to.
- The owner prefix is dropped when it adds nothing: Obsidian defaults the account name to the vault name, which is also the space name, so without this the folder would read `my-vault-my-vault` for anyone who has not set a username in settings. Same fallback when the owner lookup returns nothing, so a blocked query or a space with no person account degrades to the old `{spaceName}` naming rather than failing.
- Folder identity comes from `.dg.metadata` (`spaceUri`, `spaceName`, optional `userName`) rather than the folder name, so a user who renames an import folder keeps it — the next import finds it again by `spaceUri` and reuses it instead of creating a duplicate.
- Pre-existing folders with no `.dg.metadata` whose name matches the space name are adopted by writing metadata into them, rather than being abandoned.
- Duplicate metadata pointing at the same `spaceUri` is resolved by keeping the more recently modified one.

## Out of scope

Dropped or deferred after review discussion:

- **Auto-migration of existing import folders** — removed (72bf26dc). Existing folders keep their names; only new imports use the new scheme. This also removes the `migrated` metadata flag that came up in review.
- **Username changes after import** — flagged by @maparent; the folder is not renamed when the owner's username changes. Separate, low priority.
- **Renaming or moving the `import` root** — `IMPORT_ROOT` is a fixed path and only its immediate children are scanned, so moving the root or nesting import folders deeper hides existing imports and the next import recreates them. Recorded as a known limitation in the code (0f388d81).
- **Multi-user (Roam) source vaults** — @maparent's point that owner-in-name suits mono-user vaults and a platform prefix may be the better shape for Roam. Not addressed here since only Obsidian vaults are shared today.

## Test plan

Automated (run on the merge result, current `main` merged in):

- [x] `tsc --noEmit` clean for `apps/obsidian`
- [x] `eslint` clean on both changed files (0 errors, 0 new warnings)

Manual — needs a re-check since the naming rules changed after the Loom was recorded:

- [ ] Import from a space whose owner has a username set in settings → `import/{username}-{spaceName}/`
- [ ] Import from a space whose owner has no username set (name defaults to vault name) → `import/{spaceName}/`, not `import/{spaceName}-{spaceName}/`
- [ ] Rename an import folder, then import from that space again → the renamed folder is reused, no duplicate created
- [ ] Existing `import/{spaceName}/` with no `.dg.metadata` → adopted, metadata written, folder not renamed


apps/obsidian/src/types.ts
```diff
@@ -115,6 +115,8 @@ export type ImportFolderMetadata = {
   spaceUri: string;
   spaceName: string;
   userName?: string;
+  /** Set after auto-rename or when the user has a custom folder name. */
+  migrated?: boolean;
```
mdroidian: I'd suggest a more descriptive name.  When we come back to this in 6 months, will `migrated` be clear? Migrated what?

What if we need to do some other migration?

apps/obsidian/src/utils/importFolderMetadata.ts
```diff
@@ -1,19 +1,57 @@
...
+
 const generateShortId = (): string => Math.random().toString(36).slice(2, 8);
 
+const getImportFolderBasename = (folderPath: string): string =>
+  folderPath.split("/").pop() ?? "";
+
+const parseImportFolderMetadataRaw = (
+  raw: string,
+): ImportFolderMetadata | null => {
+  const tryParse = (content: string): unknown => JSON.parse(content);
```
mdroidian: nit: This wrapper doesn’t add behavior beyond JSON.parse. Consider inlining or replacing with a small loop over parse attempts so the control flow is easier to follow.

apps/obsidian/src/utils/importFolderMetadata.ts
```diff
@@ -1,19 +1,57 @@
...
+
+  let parsed: unknown;
+  try {
+    parsed = tryParse(raw);
+  } catch {
+    try {
+      parsed = tryParse(raw.replace(/,\s*([\]}])/g, "$1"));
```
mdroidian: suggestion: The trailing-comma retry is the only non-obvious part here, but there’s no comment explaining why. Was this for hand-edited .dg.metadata files? Legacy data? If we only ever write metadata via JSON.stringify, this path may be unnecessary.

apps/obsidian/src/utils/importFolderMetadata.ts
```diff
@@ -1,19 +1,57 @@
...
+    }
+  }
+
+  if (
+    parsed !== null &&
+    typeof parsed === "object" &&
+    "spaceUri" in parsed &&
+    typeof (parsed as Record<string, unknown>).spaceUri === "string"
+  ) {
+    return parsed as ImportFolderMetadata;
```
mdroidian: issue: Validation is weaker than the type. ImportFolderMetadata requires spaceName: string, but we only check spaceUri. A file like { "spaceUri": "abc" } would pass here and could cause issues downstream (e.g. reconcileImportFolderForSpace uses metadata.spaceName).

suggestion: Add a type guard that validates at least spaceUri and spaceName:

```
const isImportFolderMetadata = (value: unknown): value is ImportFolderMetadata =>
  typeof value === "object" &&
  value !== null &&
  "spaceUri" in value &&
  typeof value.spaceUri === "string" &&
  "spaceName" in value &&
  typeof value.spaceName === "string";
```

mdroidian: Prefer returning from a type guard instead of casting. The cast hides the gap between runtime checks and the declared type.

apps/obsidian/src/utils/importFolderMetadata.ts
```diff
@@ -80,112 +107,307 @@ const resolveMetadataDuplicate = async ({
...
+  folderPath: string,
+): string | undefined => {
+  const userNames = plugin.settings.userNames ?? {};
+  const files = plugin.app.vault
+    .getMarkdownFiles()
+    .filter((file) => file.path.startsWith(`${folderPath}/`));
+
+  for (const file of files) {
+    const authorId = plugin.app.metadataCache.getFileCache(file)?.frontmatter
+      ?.authorId as number | undefined;
```
mdroidian: always prefer typecheck over assertion

mdroidian (approved):
A few issues picked up, noted here and in comments.  I won't block on them, but combined probably worth another review and discussion in the dev meeting.  Especially the last question.

Preserve pre-existing custom import folder names
 - `discourse-graph\apps\obsidian\src\utils\importFolderMetadata.ts:252-253`
The task description states: "We should rename the folder when its original name is just the vault name.".  For vaults where a user manually renamed an import folder before this `migrated` flag existed, the metadata will not have `migrated: true`, so this branch treats the custom basename as stale and renames it to the generated `user-space` name as soon as a username is available. Since the previous resolver intentionally used `.dg.metadata` to find folders independent of their names, this discards existing custom organization instead of marking those folders as custom/migrated.

Migration picks first author not majority 
 - `resolveUserNameFromFolder` returns the username for the first markdown file (in vault iteration order) with a resolvable authorId, while imports use `resolveOwnerUserName` to pick the most frequent author in the space. Migration can rename folders with the wrong user prefix, and migrated: true prevents correction on later imports.

Question: why do imports derive the folder prefix from `resolveOwnerUserName()`?
- As written, imports still create one folder per `spaceUri`, not one folder per author. For a multi-author space, `resolveOwnerUserName()` picks the most frequent author in the current import batch, while `migrateImportFolderNames()` appears to use the first known `authorId` found in existing files. That means a space like `discourse-graphs` could be renamed to one contributor’s username, and future imports from that same space would continue landing in that single folder via `.dg.metadata.spaceUri`.
- Is that intended? If the desired schema is `{owner-username}-{spaceName}`, should we fetch the actual space owner and have one folder per author-spaceName instead of deriving it from imported node authors?

cc @maparent

mdroidian:
Actually, re "Migration picks first author not majority".  Why are we using top author/first author instead of querying for the actual synced author?  Why are we guessing instead of going to the source (username in supabase)?

mdroidian (changes requested):
Suggested PR scope change: https://linear.app/discourse-graphs/issue/ENG-1656/rename-folder-for-imported-vault#comment-c8adda2b

apps/obsidian/src/types.ts
trangdoan982: Resolved — `migrated` has been removed from the type entirely in the latest refactor. Import folders are no longer renamed automatically; the field is no longer needed.

apps/obsidian/src/utils/importFolderMetadata.ts
trangdoan982: Done — the `tryParse` wrapper has been inlined. The two `JSON.parse` calls are now directly in the try/catch blocks.

apps/obsidian/src/utils/importFolderMetadata.ts
trangdoan982: Added a comment explaining the reason: `// Tolerate trailing commas in case the file was hand-edited outside the plugin.` — this was triggered by a real bug where a hand-edited `.dg.metadata` with a trailing comma caused the folder to not be found and a duplicate to be created instead.

apps/obsidian/src/utils/importFolderMetadata.ts
trangdoan982: Done — replaced the cast with an `isImportFolderMetadata` type guard that validates both `spaceUri: string` and `spaceName: string`.

apps/obsidian/src/utils/importFolderMetadata.ts
trangdoan982: This comment was on `reconcileImportFolderForSpace` which has since been removed — the auto-rename migration logic is no longer part of this PR.

apps/obsidian/src/utils/importFolderNaming.ts
```diff
@@ -0,0 +1,60 @@
...
+  basename: string;
+  spaceName: string;
+  userName?: string;
+}): boolean => {
+  if (isLegacyFolderBasename(basename, spaceName)) return false;
+  if (userName && isExpectedMigratedBasename(basename, userName, spaceName)) {
+    return false;
+  }
+  return true;
+};
```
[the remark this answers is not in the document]
trangdoan982: `importFolderNaming.ts` was deleted and `migrateImportFolderNames` has been removed from this PR. The collision-suffix edge case no longer applies.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -281,6 +282,33 @@ export const fetchUserNames = async (
...
+  let topAuthorId = 0;
+  let topCount = 0;
+  for (const [authorId, count] of authorCounts) {
+    if (count > topCount) {
+      topCount = count;
+      topAuthorId = authorId;
+    }
+  }
+
+  return getUserNameById(plugin, topAuthorId);
```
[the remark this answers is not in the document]
trangdoan982: `resolveOwnerUserName` was updated in an earlier commit to return `undefined` instead of the synthetic `"user {id}"` fallback — addressed.

apps/obsidian/src/utils/importFolderMetadata.ts
trangdoan982: Done — replaced with an isImportFolderMetadata type guard (committed in the previous push).

maparent:
did not look in depth, but immediate reaction to reading this: `migrated: true` is an anti-pattern, it should be `version: 1`, so we can know to do subsequent migrations.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -281,6 +281,16 @@ export const fetchUserNames = async (
   await plugin.saveSettings();
 };
 
+const resolveOwnerUserName = (
+  nodes: ImportableNode[],
+  plugin: DiscourseGraphPlugin,
+): string | undefined => {
+  const authorId = nodes.find((n) => n.authorId !== undefined)?.authorId;
```
mdroidian: I don’t think deriving `ownerUserName` from the first imported node’s `authorId` matches the intent here. This gives us “an author of one selected node,” not the username/account associated with the source space. If the selected nodes are authored by multiple people, or the first selected node was authored by someone other than the vault account, the import folder could be named after the wrong person.

For Obsidian, the configured username is uploaded as `PlatformAccount.name` via `create_account_in_space`, and that account is linked to the vault’s space through `LocalAccess(space_id, account_id)`. Since this flow already has the source `spaceId`, could we derive the username from that space/account relationship instead? For example, query `LocalAccess` for this `spaceId`, join to `my_accounts`, prefer the relevant `agent_type === "person"` account, and fall back to `spaceName` if the lookup is unavailable or ambiguous.

mdroidian (approved):
I truly don’t think deriving `ownerUserName` from the first imported node’s is the right move here, but I won't block on it.

Also, @trangdoan982 don't forget to update the title/body to match the PR's new intent.

maparent:
Finally reading this seriously. I understand the motivation, and I think it's appropriate for Obsidian vaults which have a defined username. "multi-user" vaults currently refers to Roam vaults, where there is no name collision by design. This may not be true of future multi-user vaults. So… Intuition: I think that vault owner in the name is only appropriate for mono-user vaults. For Roam, maybe we should prefix with the platform, and leave it at that. And always prefix with (foreign) platform. Will we need another distinguisher for an eventual non-roam multi-user platform? We'll see when we get there, but I doubt that most common author will be the right one. (Admin authors? Org? I know we don't have those yet.)<br>All that said, I won't block on this, as we're only sharing obsidian vaults at this point in time and I'm happy with this heuristic for obsidian vaults. But I want to flag that the "most common author" may not be the right intuition to extend. (Is there a page where options are enumerated and discussed?)

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -281,6 +281,16 @@ export const fetchUserNames = async (
...
+  return authorId !== undefined
```
maparent: I have a concern, let me know if I'm missing something. We are looking at first author; but in obsidian first author is the default author, which is created in src/utils/supabaseContext.ts:101, and unless we've set a username in settings, the default name is the same as the vault name (L80). So we'd basically repeat the vault name in most cases. Can we detect that case, and consider that the author name is undefined if it's identical to the vault name?

maparent (changes requested):
Other than the author name comment, lgtm.
I also want to flag that we're not handling the case where the username changes. I also think that's a different issue, and of low priority.

apps/obsidian/src/utils/importFolderMetadata.ts
```diff
@@ -80,107 +105,139 @@ const resolveMetadataDuplicate = async ({
...
+const findImportFolderBySpaceUri = async ({
+  adapter,
+  spaceUri,
+}: {
+  adapter: DataAdapter;
+  spaceUri: string;
+}): Promise<{ folderPath: string; metadata: ImportFolderMetadata } | null> => {
   const importExists = await adapter.exists(IMPORT_ROOT);
-  if (!importExists) return map;
+  if (!importExists) return null;
```
maparent, line 136: I think that's the right call, but flagging that means we're not resilient to the user renaming the folder.

apps/obsidian/src/utils/importNodes.ts
trangdoan982: Agreed — fixed in 7f76128d. `resolveOwnerUserName` is gone; owner name now comes from `getSpaceOwnerNames`, which queries `LocalAccess` for the source `space_id`, joins to `my_accounts` filtered on `agent_type = "person"`, and returns the first person account per space. Node authors are no longer consulted at all.

Fallback behaviour: any space missing from that map yields `ownerUserName === undefined`, and `resolveFolderForSpaceUri` then falls back to the bare space name — so a failed/blocked query or a space with no person account degrades to the old naming rather than erroring.

A space with multiple person accounts is still ambiguous; first match wins. Documented that in the JSDoc rather than trying to pick a winner.

apps/obsidian/src/utils/importNodes.ts
trangdoan982: Good catch, you are not missing anything. Confirmed in `supabaseContext.ts`: the account is created with `name: plugin.settings.username || vaultName` and the space with `name: vaultName`, so the two are identical for anyone who has not set a username in settings.

Handled in 7f76128d, in `buildImportFolderBasename`:

```ts
const sanitizedUserName = sanitizeImportFolderName(userName);
const sanitizedSpaceName = sanitizeImportFolderName(spaceName);
if (!sanitizedUserName || sanitizedUserName === sanitizedSpaceName) {
  return sanitizedSpaceName;
}
return `${sanitizedUserName}-${sanitizedSpaceName}`;
```

So the default case gives `import/my-vault` rather than `import/my-vault-my-vault`, and the prefix only appears when it adds information. Comparison is on the sanitized forms so punctuation differences do not defeat it. The real owner name is still recorded in `.dg.metadata` as `userName` either way — only the folder basename drops it.

Note the first thread is also addressed: the name no longer comes from a node author at all, it comes from the space `LocalAccess` → `my_accounts` person account. That happens to be the same account you are describing, so this suppression is what keeps the common case clean.

apps/obsidian/src/utils/importFolderMetadata.ts
trangdoan982: Worth splitting into two cases, because the answer differs:

**Renaming an individual import folder is handled.** Folder identity is carried by `.dg.metadata` (`spaceUri`), not by the basename. `findImportFolderBySpaceUri` scans every child of `import/` and matches on `spaceUri`, so `import/alice-vault` → `import/whatever` is found again on the next import and reused rather than duplicated. That is the reason the auto-rename migration could be dropped in 72bf26dc.

**Renaming or moving the `import` root is not handled** — and I think that is the case you are pointing at, since the line is the `if (!importExists) return null;` guard. `IMPORT_ROOT` is the fixed string `"import"` and `adapter.list` is non-recursive, so renaming the root to `imported/`, moving it, or nesting import folders one level deeper makes existing imports invisible and the next import recreates them under a fresh `import/`. Same if `.dg.metadata` is deleted.

Not fixing that here — making the root configurable or the scan recursive is a bigger change than this ticket — but I recorded it as a known limitation on `IMPORT_ROOT` in 0f388d81 so it does not have to be rediscovered. Happy to file a follow-up if you think it is worth handling.

trangdoan982:
Addressed the outstanding review feedback — summary of what changed and what I deliberately did not change.

**Owner name no longer comes from node authors** (@mdroidian, @maparent). `resolveOwnerUserName` is gone. `getSpaceOwnerNames` now queries `LocalAccess` for the source `space_id` and joins to `my_accounts` filtered on `agent_type = "person"`. @maparent — this also answers your "most common author may not be the right intuition to extend" point: there is no author heuristic left to extend. The name comes from space membership, so it stays correct for a space whose nodes were written by several people.

**Owner prefix suppressed when redundant** (@maparent). Confirmed your reading of `supabaseContext.ts` — the account is created with `name: plugin.settings.username || vaultName` and the space with `name: vaultName`, so the two are identical unless a username is set. `buildImportFolderBasename` now compares the sanitized forms and returns the bare space name when they match, so the default case is `import/my-vault`, not `import/my-vault-my-vault`. The real owner name is still recorded in `.dg.metadata`; only the basename drops it.

**`migrated: true` is moot** (@maparent). The auto-migration was removed in 72bf26dc, and with it the flag — so the `version: 1` vs `migrated: true` question does not apply to the current diff. `ImportFolderMetadata` is now just `spaceUri`, `spaceName`, and optional `userName`. If versioned metadata is wanted later, agreed it should be a monotonic `version` rather than a boolean.

**Username changes after import** (@maparent) — not handled, agreed it is a separate low-priority issue. Noted in the PR body under out of scope rather than silently dropped.

**Multi-user / Roam vaults** (@maparent) — not addressed. Let me know if team has decided on a naming convention for Roam-Obsidian sync for folder name

**Known limitation recorded in code**: `IMPORT_ROOT` is a fixed path and only its immediate children are scanned, so renaming an *individual* import folder is safe (identity is `.dg.metadata`/`spaceUri`) but renaming or moving the `import` root is not. Documented on the constant in 0f388d81.

Also updated the title to the repo convention (`ENG-1656 Rename folder for imported vault`) and rewrote the body, which still described the removed migration flow — thanks for the nudge, @mdroidian.

Merged current `main` in (the branch was 92 commits behind) and re-verified against the merge result: `tsc --noEmit` clean, `eslint` clean on both changed files, all CI checks green.

@maparent this needs your re-review to clear the change request.

maparent (changes requested):
Ok, I had a look, and I approve all decisions, except one: I think that if the space has multiple owners, it's probably a Roam vault, and we should not give a name at all. That case should be easy to detect. <br>Rationale: Roam space names are inherently unique, and there is not (afaik) a way to determine the "owner", and naming a random person can even be politically awkward.

trangdoan982:
@maparent i was combing through all my opening PR and found this stale one. Given the current state of Obsidian-Roam sync, is this still helpful? or should I close this PR?

maparent:
Interesting. Some of my comments were addressed in [ENG-2298](https://linear.app/discourse-graphs/issue/ENG-2298/obsidian-import-panel-names-a-roam-space-after-one-of-its-authors). A lot of decisions in this thread were "good enough for now, we need to think about it more", but were not enacted. So… either closing and thinking about it again, or acting on the remaining decisions before next design round are viable solution. I am +0 for closing.

PR #1131, opened 2026-06-19
[ENG-1849] Persist Roam node type schema labels

sid597:
## Summary

Re-scopes ENG-1849 against the final merged ENG-1847 contract baseline.

The old fixture-based patch was built on a draft ENG-1847 artifact that no longer exists on `main`. Final ENG-1847 intentionally keeps the public surface to `@repo/database/crossAppNodeContract` and `@repo/database/lib/rid`, with typed examples only and no exported DB fixture API.

This PR now updates the actual Roam producer path instead: `discourseNodeSchemaToLocalConcept` always persists `literal_content.label` for node-type schema Concepts, including node types with no template. When a template exists, the existing `literal_content.template` behavior is preserved.

## Why this is the ENG-1849 fix

Roam already syncs node-type schema Concepts through the existing sync loop:

- schema Concepts use stable `source_local_id = node.type`
- schema Concepts are persisted with `is_schema: true`
- node instances still reference their schema via `schema_represented_by_local_id`
- dependency ordering already upserts schemas before instances

The remaining gap was that `literal_content.label` was only written when `node.template !== undefined`. Template-less node types, especially via the new settings store where empty templates become `undefined`, could sync without an explicit label field. ENG-1849 requires stable source identity plus label/name metadata for downstream mapping, so the label now becomes unconditional.

## Out of scope

- No shared contract expansion
- No recreated `packages/database/src/fixtures/crossAppNodes.ts`
- No destination-side schema mapping/materialization
- No relation schema or relation instance behavior

## Validation

- `pnpm --filter @repo/database check-types`
- `pnpm --filter roam exec eslint src/utils/conceptConversion.ts`
- `git diff --check origin/main...HEAD`

Full `pnpm --filter roam check-types` is currently blocked by unrelated merged-main Roam API typing errors in `src/utils/discourseNodeSearchProviders.ts` (`semanticSearchEnabled` / `semanticSearch`), not by this diff.


sid597:
@coderabbitai full review

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -78,17 +78,21 @@ export const discourseNodeSchemaToLocalConcept = (
   node: DiscourseNode,
 ): LocalConceptDataInput => {
   const titleParts = node.text.split("/");
+  const label = titleParts[titleParts.length - 1] ?? node.text;
```
sid597, line 81: Before a template-less node types could have name, but not the explicit `literal_content.label` field.

maparent (approved):

mdroidian (approved):

PR #1133, opened 2026-06-19
[ENG-1890] Add Publish tab to Roam query-results share dialog

sid597:
https://www.loom.com/share/d7cd9e76c02f46c6ad13fcda24088b90

Adds a "Publish" tab to the Roam Share Data / query-results dialog (Export.tsx), gated on isSyncEnabled(), that grants a sharing group access to the selected query-result discourse nodes via the existing SpaceAccess/ResourceAccess group-targeted model.

- Promote getMyGroups/getAvailableGroupIds/MyGroup from Obsidian's importNodes util to shared @repo/database/lib/groups; Obsidian's file becomes a behaviour-preserving re-export shim.
- New apps/roam/src/utils/publishNodesToGroups.ts: grants access only to nodes already synced (present in my_concepts) and reports the rest as not-yet-synced; upserts SpaceAccess (partial) + ResourceAccess for each node and its schema concept.
- PublishPanel with multi-group picker, non-discourse-result filter, and a success/skipped/failed count toast.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1133" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/Export.tsx
+++ b/apps/roam/src/components/Export.tsx
@@ -234,7 +234,9 @@ const ExportDialog: ExportDialogComponent = ({
         ? results
             .map((r) => {
               const node = findDiscourseNode({ uid: r.uid });
-              return node ? { uid: r.uid, type: node.type } : null;
+              return node && node.backedBy === "user"
+                ? { uid: r.uid, type: node.type }
+                : null;
             })
             .filter((n): n is PublishNode => n !== null)
         : [],
@@ -843,11 +845,16 @@ const ExportDialog: ExportDialogComponent = ({
         nonDiscourseCount,
         failedGroupCount: failedGroupIds.length,
       });
-      const messages = [
-        `Published ${publishedNodeUids.length} node${
-          publishedNodeUids.length === 1 ? "" : "s"
-        } to ${okGroupIds.length} group${okGroupIds.length === 1 ? "" : "s"}.`,
-      ];
+      const hasPublishedNodes = publishedNodeUids.length > 0;
+      const messages = hasPublishedNodes
+        ? [
+            `Published ${publishedNodeUids.length} node${
+              publishedNodeUids.length === 1 ? "" : "s"
+            } to ${okGroupIds.length} group${
+              okGroupIds.length === 1 ? "" : "s"
+            }.`,
+          ]
+        : ["No nodes were published."];
       if (skippedUnsyncedUids.length)
         messages.push(
           `${skippedUnsyncedUids.length} not synced yet — try again shortly.`,
@@ -862,10 +869,11 @@ const ExportDialog: ExportDialogComponent = ({
         );
       renderToast({
         content: messages.join(" "),
-        intent: failedGroupIds.length ? "warning" : "success",
+        intent:
+          failedGroupIds.length || !hasPublishedNodes ? "warning" : "success",
         id: "query-builder-publish-success",
       });
-      onClose();
+      if (hasPublishedNodes) onClose();
     } catch (e) {
       internalError({
         error: e as Error,
@@ -1160,7 +1168,7 @@ const ExportDialog: ExportDialogComponent = ({
   const PublishPanel = (
     <>
       <div className={Classes.DIALOG_BODY}>
-        {groupsLoading ? (
+        {groupsLoading || !groupsLoaded ? (
           <div className="my-2.5">Loading groups…</div>
         ) : groupsError ? (
           <div className="my-2.5">{groupsError}</div>
--- a/apps/roam/src/components/Export.tsx
+++ b/apps/roam/src/components/Export.tsx
@@ -11,7 +11,6 @@ import {
   Toast,
   Tooltip,
   Tab,
-  Tag,
   Tabs,
   RadioGroup,
   Radio,
@@ -87,7 +86,6 @@ import { AddReferencedNodeType } from "./canvas/DiscourseRelationShape/Discourse
 import posthog from "posthog-js";
 import { getMyGroups, type MyGroup } from "@repo/database/lib/groups";
 import {
-  getPublishedNodeCountsByGroup,
   publishNodesToGroups,
   type PublishNode,
 } from "~/utils/publishNodesToGroups";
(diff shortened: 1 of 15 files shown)


packages/database/src/lib/groups.ts
```diff
@@ -0,0 +1,49 @@
+import type { DGSupabaseClient } from "./client";
```
sid597, line 1: This is a shared-home move for existing Obsidian group helpers, not new group behavior. Roam now needs the same `getMyGroups` / `getAvailableGroupIds` logic for publishing, and importing from `apps/obsidian` would create an app-
  to-app dependency.

apps/roam/src/components/Export.tsx
```diff
@@ -139,6 +146,27 @@ const exportDestinationById = Object.fromEntries(
...
+    (window.roamAlphaAPI?.pull?.("[{:block/refs [:block/uid]}]", [
+      ":block/uid",
+      uid,
+    ]) as { [":block/refs"]?: { ":block/uid"?: string }[] } | null) || {};
+  return (result[":block/refs"] || []).flatMap((ref) =>
+    ref[":block/uid"] ? [ref[":block/uid"]] : [],
+  );
+};
+
+const getResultPublishNodes = (result: Result): PublishNode[] => {
```
sid597: Roam query results can return wrapper pages whose UID is not itself a discourse node. In that case, the visible row points at the real DG page through `:block/refs`, so publish resolves referenced user-backed DG nodes before granting access.

apps/roam/src/components/Export.tsx
```diff
@@ -139,6 +146,27 @@ const exportDestinationById = Object.fromEntries(
...
+  const directNode = findDiscourseNode({ uid: result.uid });
+  if (directNode && directNode.backedBy === "user")
```
sid597: We still require `backedBy === "user"` here because sync only uploads user-backed discourse node types; default Page/Block matches should not become publishable.

apps/roam/src/components/Export.tsx
```diff
@@ -213,6 +241,35 @@ const ExportDialog: ExportDialogComponent = ({
...
+  const [myGroups, setMyGroups] = useState<MyGroup[]>([]);
+  const [groupsLoading, setGroupsLoading] = useState(false);
+  const [groupsLoaded, setGroupsLoaded] = useState(false);
+  const [selectedGroupIds, setSelectedGroupIds] = useState<string[]>([]);
+  const [groupsError, setGroupsError] = useState("");
+  const [publishError, setPublishError] = useState("");
+
+  const { publishableNodes, nonDiscourseCount } = useMemo(() => {
+    if (!syncEnabled)
+      return { publishableNodes: [] as PublishNode[], nonDiscourseCount: 0 };
```
mdroidian: Avoid asserting when possible.  If not possible, make a comment explaining why we must resort to asserting.

apps/roam/src/components/Export.tsx
mdroidian: @sid597 Can you show an example of this where we wouldn't be expecting to export the `uid` of the queried row?

maparent (approved):
lgtm

maparent (changes requested):
Approved, but then saw missing functionality: When a node is already shared to a group, the button for that group should be already checked.

apps/roam/src/components/Export.tsx
sid597: Concrete repro from my local ENG-1890 Roam test graph: `test-1-graph`, page `06-23-2026`, query card `roamjs-query-page-yRauPWaLM`.

  For the visible row `[[[[QUE]] - cat]]`, `queryBuilder.runQuery("yRauPWaLM")` returned the row UID `DXRV7Vijg`, and `queryBuilder.isDiscourseNode("DXRV7Vijg")` was `false`.

  But the DG page referenced by that row was `[[QUE]] - cat` with UID `QgKNKlSx-`, and `queryBuilder.isDiscourseNode("QgKNKlSx-")` was `true`.

  So this is not changing normal export behavior for the queried row UID. It is publish-specific: when the query result row is a wrapper/ref page, publishing the row UID would skip it as non-DG/not-synced, while resolving `:block/refs` publishes the DG node the visible row points at.

mdroidian: If you are saying that the query is returning a block that references a single discourse node, then we should not try to "extract" said discourse node and publish it.  I'm not sure why we would try to include such behavior.<br><br>For example, if the query returned `[[[[QUE]] - cat]] some other text`, then `queryBuilder.runQuery("yRauPWaLM")` would return false and that would be correct behavior.

> So this is not changing normal export behavior for the queried row UID.

As described, this would be changing expected export behavior.<br><br>Could you elaborate on the specific use case?  Why would we want to "extract" a discourse node from a query row?  When would this be the desired behavior?

mdroidian: [Discussed in 1:1](https://roamresearch.com/#/app/discourse-graphs/page/YpkXItgN0), behavior to be removed.

sid597:
@codex review

apps/roam/src/components/Export.tsx
```diff
@@ -1042,6 +1174,68 @@ const ExportDialog: ExportDialogComponent = ({
...
+                ` (${nonDiscourseCount} non-discourse result${
+                  nonDiscourseCount === 1 ? "" : "s"
+                } will be skipped)`}
+            </div>
+          </>
+        )}
+      </div>
+      <div className={Classes.DIALOG_FOOTER}>
+        <div className={Classes.DIALOG_FOOTER_ACTIONS}>
+          <span style={{ color: "darkred" }}>{publishError}</span>
```
mdroidian: use tailwind `text-red-700`

I've added a task to consolidate this elsewhere in the code: [ENG-1963: Standardize Roam error text on text-red-700](https://linear.app/discourse-graphs/issue/ENG-1963/standardize-roam-error-text-on-text-red-700)

mdroidian (approved):

maparent (approved):
The functionality I requested will be handled separately.

sid597:
> Approved, but then saw missing functionality: When a node is already shared to a group, the button for that group should be already checked.


[ENG-1968: When a node is already shared to a group, the button for that group should be already checked](https://linear.app/discourse-graphs/issue/ENG-1968/when-a-node-is-already-shared-to-a-group-the-button-for-that-group)

PR #1146, opened 2026-06-22
ENG-1711 Reciprocal stored relations not updating after being added

trangdoan982:
Bug analysis:
https://www.loom.com/share/90f39d46cf4b49cf94c47ebb98645ba3

Bug fix test:
https://www.loom.com/share/1dbd7a00ab224f79903ee7e2af945e1a

## Summary
- Guard `matchDiscourseNode` against missing or malformed `specification` values to avoid discourse context query crashes.
- Add mutation-targeted discourse context refresh for stored relation create/delete by dispatching affected source/destination UIDs.
- Refresh only mounted contexts whose UID is affected, using `ignoreCache: true` for mutation-triggered refreshes while keeping passive/default loads cached.
- Ensure create dialog close/cancel paths do not trigger mutation refresh behavior.

## Why
Reciprocal stored relations could appear stale in open discourse context UI (especially across sidebar/main workflows) because mutation outcomes were not consistently propagating to mounted context views. This keeps default cached behavior intact and applies cache bypass only where mutation freshness is required.

## Test plan
- [x] Create a stored relation from an open discourse context and verify both affected source/destination contexts refresh if currently open.
- [x] Delete a stored relation from results table and verify affected open contexts refresh immediately.
- [x] Open discourse context passively (no mutation) and verify normal cached behavior remains unchanged.
- [x] Cancel/close the create relation dialog and verify no forced refresh occurs.
- [x] Verify no regression for `has title` specification matching and nodes with missing specifications.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1146" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/CreateRelationDialog.tsx
+++ b/apps/roam/src/components/CreateRelationDialog.tsx
@@ -24,9 +24,10 @@ import type { Result } from "~/utils/types";
 import internalError from "~/utils/internalError";
 import getDiscourseNodes from "~/utils/getDiscourseNodes";
 import posthog from "posthog-js";
+import { refreshDiscourseContextsForMutatedUids } from "~/utils/discourseContextMutationRefresh";
 
 export type CreateRelationDialogProps = {
-  onClose: () => void;
+  onClose: (created?: boolean) => void;
   sourceNodeUid: string;
 };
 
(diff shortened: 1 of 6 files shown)


mdroidian:
@trangdoan982 The ticket's title is "reciprocal stored relations not updating after being added" and the videos seem to reflect that as well, but the PR title and body says "Guard matchDiscourseNode when specification is missing".  Could you clarify how "guard matchDiscourseNode when specification is missing" is related to "reciprocal stored relations not updating after being added"?

trangdoan982:
@mdroidian yes I updated. My bad it was the title from the first commit that I forgot to update

mdroidian:
> @mdroidian yes I updated. My bad it was the title from the first commit that I forgot to update

@trangdoan982 I see the related commit is still there.  If it is no longer relevant to the ticket/PR, please remove it (and add it to a separate ticket if it is fixing a separate problem).  If it is relevant, please explain how it is relevant and what tests have been done to ensure this change is not causing any regression.

mdroidian:
@codex review

mdroidian (approved):
Potential issue: reference overlays may not refresh because the mutation-refresh listener is inside `ContextContent`.

The page title context works because `DiscourseContextCollapseOverlay` keeps `ContextContent` mounted. Reference overlays use the popover version, so the `ContextContent` listener likely only exists while the popover is open. The always-mounted overlay button state in `useDiscourseContext` does not currently listen for `roamjs:discourse-context:mutation-refresh`, which means the counters can stay stale after creating/deleting a relation.

I think the fix is small: move the listener into `useDiscourseContext` in `DiscourseContextOverlay.tsx`, and call `refresh(true)` when the mutated UID matches the overlay UID. Then either remove the listener from `ContextContent` or account for duplicate refreshes when the popover is open.

https://www.loom.com/share/123faca38f6648b7984746d15ab3fb56

PR #1147, opened 2026-06-22
[ENG-1852] Keep Roam shared node content fresh

sid597:
https://www.loom.com/share/2dcd282d7a0f4a878c8780cbca56c820

## Summary

- Extends the existing Roam Supabase sync loop to refresh already-shared source nodes.
- Detects shared source pages whose Roam page content changed via `:page/edit-time` or title-level `:edit/time`.
- Uploads refreshed `full` markdown content without running embeddings, while keeping direct/direct_and_description embedding updates on the existing changed-node path.
- Carries the matched node type through the shared freshness query so required schema metadata stays current.
- Adds the compact full-markdown validation artifact as a typed example.

## Manual demo validation:

  - Used local space `105833` and shared Roam page UID `tzmTaBmqT` (`[[CLM]] - cat -`).
  - Backfill case: confirmed the page already had `ResourceAccess` and `Content.variant = direct`, but no `Content.variant = full`; ran the existing Roam Supabase sync path; confirmed a new `full` row was created with Roam markdown
  content.
  - Freshness case: after `full` existed, edited the Roam page body; ran the existing Roam Supabase sync path again; confirmed the existing `full` row updated with the new page content.
  - No re-share/publish action was performed for either update.


apps/roam/src/utils/convertRoamNodeToFullContent.simple.example.ts
```diff
@@ -0,0 +1,47 @@
+import type { TreeNode } from "roamjs-components/types";
+import type { CrossAppNode } from "@repo/database/crossAppNodeContract";
+import { buildFullMarkdown } from "./convertRoamNodeToFullContent";
+import { contentTypes } from "@repo/content-model";
+
+/**
+ * Compact example of a Roam page tree rendered as shared `full` markdown
+ * content. Typechecking this file keeps the example aligned with the cross-app
```
maparent: Not sure about this statement, the CrossAppNode is only used marginally here.

apps/roam/src/utils/convertRoamNodeToFullContent.ts
```diff
@@ -1,10 +1,19 @@
 import { toMarkdown } from "./pageToMarkdown";
-import { type RoamDiscourseNodeData } from "./getAllDiscourseNodesSince";
 import { type DiscourseNode } from "./getDiscourseNodes";
 import getFullTreeByParentUid from "roamjs-components/queries/getFullTreeByParentUid";
 import getPageViewType from "roamjs-components/queries/getPageViewType";
 import type { TreeNode, ViewType } from "roamjs-components/types";
 import type { LocalContentDataInput } from "@repo/database/inputTypes";
+import { contentTypes } from "@repo/content-model";
+
+export type RoamFullContentNode = {
```
maparent, line 11: Is the intent to replace this with CrossAppContent eventually?

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -782,6 +826,163 @@ const getAllMissingOrNewDiscourseNodes = async ({
...
+
+  const sharedSourceLocalIds = new Set(
+    sharedResources.map((resource) => resource.source_local_id),
+  );
+  const syncedInstanceConcepts = await getAllPages(
+    supabaseClient
+      .from("my_concepts")
+      .select("source_local_id")
+      .eq("space_id", spaceId)
+      .eq("is_schema", false)
```
maparent, line 856: add arity = 0 to avoid relation instances

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -782,6 +826,163 @@ const getAllMissingOrNewDiscourseNodes = async ({
...
+      .order("source_local_id"),
+    1000,
+  );
+
+  if (!Array.isArray(syncedInstanceConcepts)) throw syncedInstanceConcepts;
+
+  return new Set(
```
maparent: This looks like just set.difference.
Unrelated: I added it to packages/utils/src/setOperations.ts#difference
I should delete that now that we use node 22.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -782,6 +826,163 @@ const getAllMissingOrNewDiscourseNodes = async ({
...
+  nodeTypes: DiscourseNode[];
+}): Promise<SharedFullContentUpdate[]> => {
+  const sharedSourceLocalIds = Array.from(sourceLocalIds);
+  if (sharedSourceLocalIds.length === 0 || nodeTypes.length === 0) {
+    return [];
+  }
+
+  const sinceMs = since ?? DEFAULT_SYNC_TIME;
+  const query = `[
+    :find ?node-title ?uid ?nodeCreateTime ?nodeEditTime ?pageEditTime ?author_local_id
```
maparent, line 937: I may be wrong, but I am a bit puzzled that this query uses only node, which seems to bind to the page and not to sub-nodes. I'd expect to find `?page`, `?node`, a `[?node :block/page ?page]` clause. Are we sure that nodeEditTime and pageEditTime are ever distinct in the query as written?

maparent (changes requested):

apps/roam/src/utils/syncDgNodesToSupabase.ts
sid597: pageEditTime gets updated when any block on the page changes whereas the nodeEditTime is tracking the title of the page. So when we combine both we get the info if this page as a whole entity has changed since last time.

apps/roam/src/utils/convertRoamNodeToFullContent.ts
```diff
@@ -38,26 +50,34 @@ export const buildFullMarkdown = ({
...
+        author: { localId: node.author_local_id },
+        localId: node.source_local_id,
+        createdAt: new Date(node.created || Date.now()),
+        modifiedAt: new Date(node.last_modified || Date.now()),
+        nodeType: { localId: node.node_type_id },
+        content: {
+          direct: {
+            localId: node.source_local_id,
+            value: title,
+          },
```
sid597, line 70: @maparent @mdroidian `crossAppNode` requires `direct` to be present but `crossAppNodeToDbContent` converter explicitly requires us to pass the variant type, so should we update the `crossAppNode.content` an `or` case?

maparent: It requires to pass the variant type to choose which of the two contents you're getting from the node. Will probably be updated with format when we have multiple format. I don't think we need to be an or case, or at least I don't see how it follows.

maparent: We could have a variant that returns all contents as an array, would that be more useful to you?

maparent (approved):
LGTM

apps/roam/src/utils/convertRoamNodeToFullContent.ts
sid597: I was saying that here I am passing `content.direct` and `content.full` but `direct` will not be used anywhere it is here because the `crossAppNode` type requires both `direct` and `full`.

maparent: Ah! Yes, full should be optional.

maparent: In a conference now, I'll make a task soon.

maparent: I do not see a case where the title is optional; do you?

sid597: In this case direct should be optional because it is not needed by the function that will consume this data

maparent: Hmm... Are you sure that sharing does not include the title at all? 
Possible underlying assumption: Do you expect the title to also be part of the full content?
In Obsidian we found it easier to separate the title from the content, so we assume that they are disjoint.

PR #1183, opened 2026-07-02
ENG-1983 Define intermediate schemas for node, relations and schemas

maparent:
https://linear.app/discourse-graphs/issue/ENG-1983/define-intermediate-schemas-for-node-relations-and-schemas

### DIFF
--- a/packages/database/src/crossAppContracts.ts
+++ b/packages/database/src/crossAppContracts.ts
@@ -14,6 +14,28 @@ type DbRef = {
 // Generalized reference
 export type Ref = LocalRef | DbRef;
 
+type SpaceRef = DbRef | { url: string; sourceApp: Enums<"Platform"> };
+
+// A potentially cross-space reference
+export type LocalOrRemoteRef =
+  | DbRef
+  | {
+      localId: string;
+      // Treat as LocalRef if space is absent
+      space?: SpaceRef;
+      // make the options mutually exclusive
+      dbId?: never;
+      rid?: never;
+    }
+  | {
+      // A string that contains combined space and localId
+      rid: string;
+      // make the options mutually exclusive
+      dbId?: never;
+      localId?: never;
+      space?: never;
+    };
+
 // Common attributes for most types
 export type CrossAppBase = LocalRef & {
   createdAt: Date;
@@ -84,4 +106,11 @@ export type CrossAppNode = CrossAppBase & {
     direct: InlineCrossAppContent;
     full: InlineCrossAppTypedContent;
   };
+};
+
+// A relation instance
+export type CrossAppRelation = CrossAppBase & {
+  relationType: Ref;
+  source: LocalOrRemoteRef;
+  destination: LocalOrRemoteRef;
 };
--- a/packages/database/src/crossAppContracts.ts
+++ b/packages/database/src/crossAppContracts.ts
@@ -85,18 +85,69 @@ export type CrossAppEmbedding = {
   embedding?: Enums<"EmbeddingName">;
 };
 
-// A Content object. It can be put inline inside a concept.
-// Missing CrossAppBase attributes are inferred from enclosing object.
-export type InlineCrossAppContent = Partial<CrossAppBase> & {
+// An asset reference
+export type CrossAppAsset = {
+  content: ArrayBuffer;
+  mimetype: string;
+  createdAt: Date;
+  modifiedAt?: Date;
+  filepath?: string;
+};
+
+// Document fields
+type CrossAppDocumentExtras = {
+  // MIME type
+  contentType: string;
+};
+
+// An inline document, to put inside Content or Concept.
+// Currently, we fully infer Documents (setting content_as_document=true)
+// since our nodes are pages; so this is not used yet.
+export type InlineCrossAppDocument = Partial<CrossAppBase> &
+  CrossAppDocumentExtras;
+
+// A standalone document, for `upsert_documents`. Not currently used.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+type StandaloneCrossAppDocument = LocalRef &
+  CrossAppBase &
+  CrossAppDocumentExtras;
+
+// Content fields
+type CrossAppContentExtras = {
   value: string;
   embedding?: CrossAppEmbedding;
   scale?: Enums<"Scale">;
+  partOf?: Ref;
+  assets?: CrossAppAsset[];
+  document?: InlineCrossAppDocument;
   contentType?: ContentType;
 };
+
+// A Content object. It can be put inline inside a concept.
+// Missing CrossAppBase attributes are inferred from enclosing object.
+export type InlineCrossAppContent = Partial<CrossAppBase> &
+  CrossAppContentExtras;
+
+// A standalone content object, for `upsert_content`
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+type StandaloneCrossAppContent = LocalRef &
+  CrossAppBase &
+  CrossAppContentExtras;
 
 // An inline Content with obligatory typing
 type InlineCrossAppTypedContent = InlineCrossAppContent & {
   contentType: ContentType;
+};
+
+// A platform account
+// either standalone for `upsert_account_in_space`,
+// or inline in Content or Concept
+export type CrossAppAccount = {
+  accountLocalId: string;
+  name?: string;
+  email?: string;
+  // agentType: Enums<"AgentType"> = 'person'  // inferred
+  // dgAccount?: string; // uuid
 };
 
 // A node instance
@@ -106,6 +157,8 @@ export type CrossAppNode = CrossAppBase & {
     direct: InlineCrossAppContent;
     full: InlineCrossAppTypedContent;
   };
+  // This is a way to define document globally for all contents
+  document?: InlineCrossAppDocument;
 };
 
 // A relation instance


packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,92 @@
+import type { ContentType } from "@repo/content-model";
```
mdroidian, line 1: @maparent 
Let's only export types that are being used elsewhere.  This adds positive signal and really helps navigate and grok the file.

When we export types that are not used, it adds a false positive and makes it harder to understand intent.

packages/database/src/crossAppNodeContract.example.ts
```diff
@@ -13,23 +13,20 @@ Multiple studies show that sleep after learning strengthens memory traces.
...
-    sourceNodeTypeId: "rCLM0schema",
-    label: "Claim",
+    local_id: "rCLM0schema",
   },
   content: {
     direct: { value: "Sleep improves memory consolidation" },
     full: { format: contentTypes.markdown, value: roamFullMarkdown },
   },
-  sourceModifiedAt: "2026-06-12T14:00:00.000Z",
+  createdAt: new Date("2026-06-12T14:00:00.000Z"),
```
sid597, line 30: We removed the `modifiedAt?` example from both cases. Also why is it optional? I think keeping it optional makes downstream code like `roamOriginNodeExmple.modifiedAt?` and then dependent on that. We can make the initial `modifiedAt` same as `createdAt` iff its not present.

packages/database/src/crossAppNodeContract.example.ts
```diff
@@ -13,23 +13,20 @@ Multiple studies show that sleep after learning strengthens memory traces.
...
-  sourceNodeId: ROAM_SOURCE_NODE_ID,
-  sourceNodeRid: spaceUriAndLocalIdToRid(
-    ROAM_SOURCE_SPACE_ID,
-    ROAM_SOURCE_NODE_ID,
-  ),
+  space: {
+    sourceApp: "Roam",
+    url: ROAM_SOURCE_SPACE_ID,
+  },
+  local_id: ROAM_SOURCE_NODE_ID,
```
sid597: snake case -> camelCase

What do you think about naming, like appending `source` vs direct e.g 
- `sourceLocalId`, vs `localId`
- `space.sourceApp` vs `space.app`

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,92 @@
...
+import { Enums } from "./dbTypes";
+
+export type DbRef = { dbId: number };
+export type SpaceRef = DbRef | { url: string; sourceApp: Enums<"Platform"> };
+
+export type LocalRef =
+  | {
+      localId: string;
+      // infer space from context if absent.
+      space?: SpaceRef;
```
sid597, line 25: Is this for future? Can `space` be absent in Obsidian <-> roam sync?

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export type CrossAppRelationTypeSchema = CrossAppBase & {
+  label: string;
+  reverseLabel: string;
+};
+
+export type CrossAppRelationTripleSchema = CrossAppBase & {
+  label: string;
+  reverseLabel: string;
+  relation?: Ref | CrossAppRelationTypeSchema;
```
sid597: Sorry this is confusing like we already have the `label`, and `reverseLabel` is that like source space's version? and this one is for the target's space (or vice-versa)? If so can we name them to reflect that?

sid597 (commented):
Some questions and suggestions, I read `CrossApp...` so many time that now I am confused what it even means lol

packages/database/src/crossAppContracts.ts
maparent: This is for the case where we have a reference to a node in another space. It currently happens in Obsidian; the source or destination of a relation may be an imported node, in which case we give the reference to the original node (imported nodes are not materialized again.)
But yes, it can be absent and inferred from context in most cases.

packages/database/src/crossAppContracts.ts
maparent: reverseLabel is the label of the reverse relation. So nothing to do with source space or import.

packages/database/src/crossAppNodeContract.example.ts
maparent: I have tended to remove source here, because everything was named sourceSomething, and it did not seem to add specific information (as you were remarking about CrossApp!) Maybe I missed something? Fully negotiable.

packages/database/src/crossAppNodeContract.example.ts
maparent: I could add it back in the example. It's optional precisely for the reason you gave: it uses createdAt if absent.

packages/database/src/crossAppContracts.ts
maparent: Ah, and I put it on the triple, which is redundant, because in Roam we don't have the relation as a separate object.

packages/database/src/crossAppNodeContract.example.ts
mdroidian: +1 to camelCase for TypeScript-facing contract fields

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,92 @@
...
+  createdAt: Date;
+  modifiedAt?: Date;
+  filepath?: string;
+};
+
+export type CrossAppContent = {
+  value: string;
+  localId?: string;
+  embedding?: CrossAppEmbedding;
+  author: CrossAppAccount | Ref;
```
mdroidian: When would `author` use `Ref` (specifically `LocalRef`)?

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,92 @@
...
+  partOf?: Ref;
+  assets?: CrossAppAsset[];
+};
+
+export type CrossAppTypedContent = CrossAppContent & {
+  format: ContentType;
+};
+
+export type CrossAppAccount = {
+  account: string;
```
mdroidian: Is this intended to be `accountLocalId`?

 `account` is ambiguous here: it could mean a display label, platform username, database account id, or the whole account object. Existing DB-facing shapes use `account_local_id`, and TypeScript-facing code generally uses `accountLocalId`.

packages/database/src/crossAppContracts.ts
mdroidian: Can we clarify the use cases for `LocalRef` and maybe narrow it to relation endpoints?

From the current Obsidian/database flow, the concrete cross-space case seems to be relation `source` / `destination`: Obsidian stores endpoint ids as either a local `nodeInstanceId` or an imported node’s `importedFromRid`, and the DB resolves those with `rid_or_local_id_to_concept_db_id`.

If that is the main case this type is supporting, I think it would be clearer to model it directly on relation endpoint refs (`source` / `destination`) rather than having all cross-app objects inherit a broad `LocalRef` union. If there are other planned use cases for `{ localId, space? } | { rid }`, can we add examples/docs for those?

packages/database/src/crossAppContracts.ts
mdroidian: Let's use complement here instead of reverseLabel

`reverseLabel` looks new in this PR, while `complement` is already the repo-wide term for the inverse/reverse relation label in both Roam and Obsidian. For example, Roam relation settings and query logic use `label` / `complement`, and Obsidian relation types also store `label` / `complement`.

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,92 @@
...
+    full: CrossAppTypedContent;
+  };
+  /**
+   * Source modified timestamp, or latest required `Content.modifiedAt` when
+   * deriving from persisted `direct` and `full` rows.
+   */
+  createdAt: Date;
+  modifiedAt?: Date;
+  author: CrossAppAccount | Ref;
+  contributors?: (CrossAppAccount | Ref)[];
```
mdroidian: I don't believe `contributors` is actively used anywhere today, correct? I see the DB tables for it, but I don't see app/converter paths producing or consuming it.

If that's right, I'd prefer we wait to add this to the cross-app contract until we have a concrete use case and are actually wiring it through. Otherwise it adds another field to reason about without a clear behavior

packages/database/src/crossAppNodeContract.example.ts
maparent: Agreed. Did the rename in the type and forgot to update the example.

packages/database/src/crossAppContracts.ts
maparent: Sure. I'm trying to abstract somewhat from the database, but it's always a judgement call. I agreee that `account` was unclear, maybe I should have gone with `accountId`; the `local` should be implied in the communication types. But it also adds cognitive load to think about the difference, so I'll use `accountLocalId`.

packages/database/src/crossAppContracts.ts
maparent: Hmmm... You're right it's not exercised now, and I'm fine with removing it; I'd leave a comment for it.

packages/database/src/crossAppNodeContract.example.ts
maparent: As for source: With the same reasoning as account, happy to reintroduce it where it fits database names. It was quite broad in the original file, but that argument does not hold anymore. Would you prefer that?

maparent:
A few residual questions inline in the code (eg: Is node color part of the contract?), but I think this is close to useable.

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,149 @@
...
+  createdAt: Date;
+  modifiedAt?: Date;
+  author: CrossAppAccount | Ref;
+};
+
+// Common base attributes for standalone types
+type CrossAppBase = LocalRef & BaseAttributes;
+
+// Partial version of above.
+export type InlineAbstractBase = WithOptionalLocalRef<Partial<BaseAttributes>>;
```
mdroidian: What is the purpose of this?  Where will it be used?

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,149 @@
...
+};
+
+// Document fields
+type CrossAppDocumentExtras = {
+  // Mime type
+  contentType: string;
+};
+
+// An inline document
+export type InlineCrossAppDocument = WithOptionalLocalRef<
```
mdroidian: What is "Inline" referring to?

mdroidian: And where is this intended to be used?

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,149 @@
...
+  Partial<BaseAttributes> & CrossAppDocumentExtras
+>;
+
+// future: A standalone document
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+type StandaloneCrossAppDocument = LocalRef &
```
mdroidian, line 111: What is a standalone document?  And where is it intended to be used?

If it's not used now, why is it included in this PR?

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,149 @@
...
+
+// An inline content object
+export type InlineCrossAppContent = WithOptionalLocalRef<
+  Partial<BaseAttributes> & CrossAppContentExtras
+>;
+
+// future: A standalone content object
+type StandaloneCrossAppContent = LocalRef &
+  BaseAttributes &
+  CrossAppContentExtras;
```
mdroidian: What is the difference between inline vs standalone?

Where is standalone going to be used?  What user need is it solving?

packages/database/src/crossAppContracts.ts
```diff
@@ -0,0 +1,149 @@
...
+
+// An inline Content with obligatory typing
+export type InlineCrossAppTypedContent = InlineCrossAppContent & {
+  contentType: ContentType;
+};
+
+// A standalone Content with obligatory typing
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+type StandaloneCrossAppTypedContent = StandaloneCrossAppContent & {
```
mdroidian: When would we use `StandaloneCrossAppContent` vs `StandaloneCrossAppTypedContent`

mdroidian (changes requested):
@maparent could you clarify the intended scope of this PR relative to ENG-1983 and its child tickets? This is fairly urgent because it is currently blocking @sid597.

The PR is attached to ENG-1983, but the changes seem to overlap with ENG-1984, ENG-1987, ENG-1863, and ENG-1864 as well. That may be the right approach, but as discussed [in the ticket](https://linear.app/discourse-graphs/issue/ENG-1983/define-intermediate-schemas-for-node-relations-and-schemas#comment-606a3139) and in our 1:1, it would help to make explicit which tickets this PR is meant to close and which remain follow-ups. I could not find that clarification.

Could you also add a short walkthrough, ideally a video, explaining the contract choices and intended usage? There are several decisions that are hard to evaluate from the patch alone, for example: `Inline` vs `Standalone`, multiple unused exports, and several types/fields marked as `future`.

Re: `future`

As part of our commitment to "Stop moving forward on technical proposals before the user need is clear", could you clarify the user need for any `future` types or fields, ideally tied to existing in-scope projects or tickets? If they are speculative, let's remove them and focus on the immediate need for this PR.

Right now I cannot tell which exported types are intended as stable API versus future scaffolding. My preference is to only export types that have a concrete producer/consumer or example in this PR, and keep future-facing shapes internal until the usage is clear.

Ideally, let’s focus this PR on the immediate problem that needs to be solved, and defer anything else that is not required for that scope.

packages/database/src/crossAppContracts.ts
maparent: This is a utility type used in eng-1985.

packages/database/src/crossAppContracts.ts
maparent: The db  upsert function has always allowed to put some objects inline inside related objects instead of passing a reference to the inline object; here, we can put documents in the content objects. The current usage is mostly to auto-generate the document, providing neither reference nor inline object, but this is only because our nodes are all pages. It will be false when we upload candidate nodes, for example.
When we provide an inline object, some shared attributes can be specified in the enclosing object, and hence there are less required attributes for the inline object than if the same object is upserted standalone. Hence the Inline/Standalone distinction.

packages/database/src/crossAppContracts.ts
mdroidian: @maparent

packages/database/src/crossAppContracts.ts
mdroidian: @maparent Unless we can tie it to an in scope project or ticket, let's remove it please

packages/database/src/crossAppContracts.ts
mdroidian: I'm assuming that localId/space can be used for more use cases? If not, then having it in a union this high up will just complicate things by making the downstream types just more complex for a single use case.

packages/database/src/crossAppContracts.ts
maparent: Standalone: See above. Here, I'm using "future" in a very restricted sense: Not yet used in ENG-1985. This is the type you would use if you were to call `upsert_document` on its own. It is not speculative in that way, and I think that having both the inline and standalone type clarifies the inline usage.

packages/database/src/crossAppContracts.ts
maparent: It solves the user need of understanding the full type.

packages/database/src/crossAppNodeContract.example.ts
mdroidian: @maparent Please add it back to the example.

packages/database/src/crossAppContracts.ts
maparent: If calling upsert_content, as we do now, we have to use StandaloneCrossAppContent. We will be using InlineCrossAppContent very very soon in ENG-1823/1824.

maparent:
Apologies I forgot to add a video here. I'm not working from home now, and it's not easy for me to do so until later today. I hope we can resolve the main issues asynchronously.

packages/database/src/crossAppContracts.ts
mdroidian: If the only current use cases auto-generate documents, then let's remove this and add it back when it is required for use.  Adding it now adds confusion for the dev who is trying to do the action.

When we have defined user need, then we add the scaffolding, but not before.

maparent: I will comment it out.

packages/database/src/crossAppContracts.ts
maparent: This allows a minimal reference to the author, i.e. enough for a reference, not enough for an upsert.  I did not check how we currently include authors in the sync, but I think it's closer to Ref usage than to an InlineAccount. Reusing the Ref type is a bit of overloading here, maybe we should define a separate AccountRef. But it would end up having exactly the same parameters.

maparent:
As for tickets: This currently covers all sub-tasks of ENG-1983.

packages/database/src/crossAppContracts.ts
mdroidian: That helps for `StandaloneCrossAppContent` and `InlineCrossAppContent`, thanks. I’m still unclear on `StandaloneCrossAppTypedContent` specifically: what path would use that instead of `StandaloneCrossAppContent`, and is there a concrete producer/consumer for it in this PR.

maparent: Fair enough; I was aiming for completeness here. We do need InlineCrossAppTypedContent so we require contentType for the full text; but StandaloneCrossAppTypedContent can go.

packages/database/src/crossAppContracts.ts
mdroidian: I’m still not following what “the full type” means here. Is that the full database-backed content shape, the full cross-app payload, or something else?

Assuming the user here is the developer, when do they need to understand or use this full type directly? For example, is this needed when calling `upsert_content`, writing converters, validating a received payload, or something else?

If the only current use case is internal type composition, I think we should avoid exposing it as part of the public contract until there is a concrete producer/consumer that needs it.

packages/database/src/crossAppContracts.ts
mdroidian: I’m still unclear on when `author` would actually use `Ref`, specifically `LocalRef`.  Could you give a real world example?

packages/database/src/crossAppContracts.ts
mdroidian: I'd prefer to remove it, as the example provided (candidate nodes) is not yet in our immediate timeline.

packages/database/src/crossAppContracts.ts
maparent: Sorry made that comment in the wrong place, I thought this was about document. 
I still think comments are useful for placeholders for future work.

packages/database/src/crossAppContracts.ts
maparent: On second thought: Document has a useful semantic, and should be in the contract. Removing it because we accidentally did not use it yet in the way we wrote the function so far actually makes some of the shared contract harder to understand. Actually, I should not comment it out at all, but add a comment "Not used in the current code paths yet" or something to that effect to direct attention.

packages/database/src/crossAppContracts.ts
mdroidian: I understand you feel that way, but without an in scope project or ticket, this would open us up to the codebase being riddled with possible future comments.  So for this PR, let's stick to the immediate need, and remove it please.

packages/database/src/crossAppContracts.ts
mdroidian: Great.  Then lets remove it from this PR and add to the PR for [ENG-1985](https://linear.app/discourse-graphs/issue/ENG-1985/define-converters-between-database-schemas-and-cross-app-schemas).  👍

packages/database/src/crossAppContracts.ts
maparent: It is used when calling upsert_content, yes. 
I think that having both inline and standalone type helps understand how to use the inline type.

packages/database/src/crossAppContracts.ts
mdroidian: I think this is the distinction I’m trying to keep clear for ENG-1983: this PR should define the client-side cross-app contract for nodes, relations, and schemas. It should describe what Roam/Obsidian produce and consume before conversion, not the shape the database upsert API happens to accept.

From that perspective, I’m not convinced the client contract should expose \`InlineCrossAppDocument\` or an inline/standalone distinction. Whether a document is passed inline to \`upsert_content\` or upserted separately feels like converter/persistence strategy, which doesn't belong in ENG-1983. A client producing a node/relation/schema payload should not need to know which database write shape the converter chooses.

If \`document\` itself is meaningful in the cross-app payload, let’s define the minimal client-facing document shape and show where it appears in a concrete node/content example. But I’d avoid exporting \`Inline...\` / \`Standalone...\` variants in this PR unless a current client-side producer/consumer needs to choose between them.

A comment saying “not used in current code paths yet” does not really solve the concern for me; it still makes future scaffolding part of the public contract.

packages/database/src/crossAppContracts.ts
mdroidian: I think we need to realign on what the goal of [ENG-1983](https://linear.app/discourse-graphs/issue/ENG-1983/define-intermediate-schemas-for-node-relations-and-schemas) was. And that was elaborating on the `crossAppNode` example that Sid provided that gave a shape for client side code to adhere to.  In that sense, the distinction between `Inline` and `Standalone` should not be something that the client side shape should need to be concerned about.  If `Standalone` provides some value at a future point, let's introduce it at that future point where the need is.  But debating it in the comments of a PR is not the place for this.<br><br>To help move forward, let's defer any code that is not solving a user need immediately related to what problem the ticket is meant to solve (in this case, namely: "We will propose specific platform-neutral schemas, following the example of CrossAppNode, but making it hew a bit closer to the database format.").<br><br>In this case I see that `InlineCrossAppContent` is used in `CrossAppNode` and `StandaloneCrossAppContent`is not used anywhere meaningful yet.<br><br>So let's **please remove** `StandaloneCrossAppContent` **so that we can move forward with this PR** with the minimal requirements, and plan to discuss `StandaloneCrossAppContent` at our next dev or 1:1 meeting.

maparent: The problem with that approach is that we're in the middle of a code transition, where we're currently using `upsert_content`, which would require StandaloneCrossAppContent after refactoring our sync conversions to use the CrossApp contract.
We want to transition to mostly using `upsert_concepts` with inline content. So both are needed across the transition. This is not a far future speculative change, it's on my plate now (ENG-1823/1824).
Even as I complete those two tasks, I expect it will make sense to still use `upsert_content` when making changes to content that do not affect the concepts, such as in the ENG-1852 PR I just reviewed for Sid. 
I promise if I find out I can eliminate all uses of `upsert_content`, I'll add a comment to this file that StandaloneCrossAppContent is not used. But that may change due to future optimizations, and I am -10 on removing it altogether.
(Edit: Grammar)

maparent: I'll put it another way: You want to minimize the cognitive surface. I understand that. But the question is: Is this something the user of these structures have to know about to use them? And I think in this case the answer is a resounding YES.

mdroidian: I understand we disagree.  To move forward, **please remove it so we push forward with this PR** and can continue the conversation elsewhere about where it should be added, knowing that this will be added shortly if need be.

packages/database/src/crossAppContracts.ts
maparent: Done. For the record: I think the absence of future comments is a flaw.

packages/database/src/crossAppContracts.ts
maparent: In convertRoamNodeToFullContent, we use author_local_id which would be the equivalent of a LocalRef.

maparent:
Decision: Will be split in sub-tasks.

PR #1184, opened 2026-07-02
ENG-1985 Define converters between database schemas and cross-app schemas

maparent:
https://linear.app/discourse-graphs/issue/ENG-1985/define-converters-between-database-schemas-and-cross-app-schemas


maparent:
I have a first pass of the converters written, next step would be adding tests.

PR #1197, opened 2026-07-06
eng-1863 Relation cross-app contract

maparent:
[ENG-1863: Define cross-app relation contract and fixtures](https://linear.app/discourse-graphs/issue/ENG-1863/define-cross-app-relation-contract-and-fixtures)

https://www.loom.com/share/3deac862860d491884d212fc0f8b178a

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1197" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/database/src/crossAppContracts.ts
+++ b/packages/database/src/crossAppContracts.ts
@@ -4,6 +4,9 @@ import { Enums, type Json } from "./dbTypes";
 // An identifier for objects in the platform. Expected to be unique within the platform.
 export type LocalId = string;
 
+// A composite identifier for objects in other spaces.
+export type Rid = string;
+
 // Common attributes for most types
 export type CrossAppBase = {
   localId: LocalId;
@@ -75,4 +78,13 @@ export type CrossAppNode = CrossAppBase & {
     direct: InlineCrossAppContent;
     full?: InlineCrossAppTypedContent;
   };
+};
+
+// A relation instance
+export type CrossAppRelation = CrossAppBase & {
+  relationType: LocalId;
+  /* eslint-disable @typescript-eslint/no-duplicate-type-constituents */
+  source: LocalId | Rid;
+  destination: LocalId | Rid;
+  /* eslint-enable @typescript-eslint/no-duplicate-type-constituents */
 };


packages/database/src/crossAppContracts.ts
```diff
@@ -14,6 +14,28 @@ type DbRef = {
...
+      rid?: never;
+    }
+  | {
+      // A string that contains combined space and localId
+      rid: string;
+      // make the options mutually exclusive
+      dbId?: never;
+      localId?: never;
+      space?: never;
+    };
```
mdroidian: Are all of these union variants required? If so, could we add examples showing where each is already used or needed by upcoming work? Otherwise, I’d prefer to keep the contract limited to the variants we currently need and add others when a concrete use case arises.

Including speculative variants increases the contract’s conceptual surface area and requires additional validation, tests, and converter paths, which may create unnecessary confusion and complexity.

mdroidian (changes requested):

packages/database/src/crossAppContracts.ts
maparent: We do use the Rid in https://github.com/DiscourseGraphs/discourse-graph/blob/7e058c152242359771f50ea134491a3c4add687c/apps/obsidian/src/utils/conceptConversion.ts#L224
And I expect this pattern to hold: We store a Rid and send it back.
We do not currently have a version that requires space directly at the upsert_concept call points, because the burden of computing the Rid from space+localId is currently on the platform (8 calls to `spaceUriAndLocalIdToRid` in the Obsidian code base, many upstream of upsert_concept.) 
Having the space+localId variant in the contract will  help centralize the Rid computation in the database package.

mdroidian: Thanks, that confirms the `{ rid }` variant is required.

I agree that centralizing the `spaceUri + localId → RID` normalization could be useful. What I’m trying to separate is centralizing the RID implementation—which already lives in `@repo/database/lib/rid`—from expanding the shared relation contract to accept another representation of the same identity.

The existing Obsidian call sites use RIDs for several platform-level purposes, including imported provenance, duplicate detection, previews, and endpoint lookup. Because of that, moving normalization into the upsert converter would not replace all eight call sites.

This rationale also seems to support, at most, a `{ localId, spaceUri }` input. I don’t see the same justification yet for the top-level `{ dbId }`, `{ localId, space: { dbId } }`, or the `sourceApp` property. The current converter does not use `sourceApp`, and its `space.dbId` branch is explicitly unresolved.

I’d prefer to keep the shared endpoint contract as `LocalRef | RidRef` for now:

```ts
type RelationEndpointRef =
  | { localId: string }
  | { rid: string };
```

This gives the shared boundary one canonical representation for remote identity: an RID. A platform that naturally has `spaceUri + localId` can normalize it before constructing the contract.

If an upcoming Roam or Obsidian producer genuinely benefits from emitting `{ localId, spaceUri }`, we can add that form alongside the producer and a test demonstrating its normalization to the expected RID.

maparent: Removed all the Reference types.

mdroidian (approved):

PR #1199, opened 2026-07-06
ENG-1990: Restore sidebar navigation for blocks containing a smartblock

sid597:
https://www.loom.com/share/f123ad86812345f1a21cb41205d852f6

Sidebar rows rendered via RoamRenderedBlock (blocks containing a SmartBlock button) had no click handler, so clicking the text did nothing. Per the approach agreed on the ticket: a click watcher on the rendered block now navigates via the existing openTarget unless the click originated on a data-roamjs-smartblock-button element, which passes through to the SmartBlocks extension.

Fixes ENG-1990

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1199" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -179,9 +179,6 @@ const toggleFoldedState = async ({
   }
 };
 
-const RENDERED_BLOCK_INTERACTIVE_SELECTOR =
-  "a, button, input, [data-link-title], [data-tag], .rm-block-ref";
-
 const RoamRenderedBlock = ({
   uid,
   onNavigate,
@@ -203,7 +200,7 @@ const RoamRenderedBlock = ({
 
   const handleClick = (e: React.MouseEvent) => {
     const target = e.target as HTMLElement;
-    if (target.closest(RENDERED_BLOCK_INTERACTIVE_SELECTOR)) return;
+    if (target.closest("button")) return;
     onNavigate(e);
   };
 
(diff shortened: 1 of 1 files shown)


sid597:
Review addressed in a466f514. The double-navigation finding was valid — page refs, tags, block refs, links and inputs inside the rendered block act on their own, so the outer handler double-fired. The exclusion is now a selector covering all interactive elements (a, button, input, [data-link-title], [data-tag], .rm-block-ref); button subsumes the SmartBlock trigger. Navigation only fires for clicks on inert content, which keeps the agreed behavior from the ticket.

On the overlay alternative: a clickable overlay above the rendered block would intercept the SmartBlock button itself, which is the one thing that must stay clickable, so the exclusion-selector approach was kept. e.defaultPrevented isn't reliable here since Roam's internal handlers don't consistently call preventDefault.

sid597:
Approach revised in 39520f0d after product clarification: the rendered block is now fully inert except buttons — pointer-events are disabled on everything inside .dg-sidebar-rendered-block and re-enabled only on button elements. Clicking text, refs, tags, or empty space navigates to the block (and can never enter Roam's edit mode or trigger ref navigation, since those clicks no longer reach Roam's elements); clicking a button (e.g. a SmartBlock trigger) only runs its action. This replaces the interactive-element exclusion selector, which is no longer needed beyond the button check on the bubbled click.

mdroidian (approved):

PR #1200, opened 2026-07-06
ENG-1819: Make global left sidebar fold toggle personal

sid597:
Fold state of the GLOBAL sidebar section moves from the shared global setting to a per-user personal setting (PERSONAL_KEYS.globalSectionFolded, legacy dual-write via a {userUid}/Global-Section-Folded marker block, following the {userUid}/Personal-Section naming). The admin Collapsable flag is removed: the section is now always foldable, per user.

Supersedes the earlier eng-1819 branch (d51649d9), reconciled onto current main; the old remote branch was replaced by this one.

Fixes ENG-1819

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1200" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/LeftSidebarView.tsx
+++ b/apps/roam/src/components/LeftSidebarView.tsx
@@ -638,23 +638,24 @@ const GlobalSection = ({
     try {
       const nextIsOpen = !isOpen;
       setIsOpen(nextIsOpen);
-      const markerText = getGlobalSectionFoldedMarkerText(
-        window.roamAlphaAPI.user.uid() || "",
-      );
-      if (nextIsOpen) {
-        const children = getBasicTreeByParentUid(leftSidebarUid);
-        await Promise.all(
-          children
-            .filter((c) => c.text === markerText)
-            .map((c) => deleteBlock(c.uid)),
-        );
-      } else {
-        await createBlock({
-          parentUid: leftSidebarUid,
-          node: { text: markerText },
-        });
+      const userUid = window.roamAlphaAPI.user.uid();
+      if (userUid) {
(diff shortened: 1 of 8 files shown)


sid597:
Null-uid finding addressed in efcb1d6d: roamAlphaAPI.user.uid() is typed string | null, so with no uid the toggle is now session-only (no marker read/write) instead of falling back to a shared "/Global-Section-Folded" marker. Note the pre-existing personal-sections code (getLeftSidebarSettings.ts, targetUserUid + "/Personal-Section") has the same degenerate case and isn't touched here.

On migrating the old Global-Section > Settings > Folded state: deliberately not done. Per Michael on ENG-1819 the fold feature hasn't shipped in a release, so the old state only exists in dev graphs; the cost is a one-time default-open until re-toggled. Adding a permanent legacy fallback read for unreleased shared state (which shouldn't seed per-user prefs anyway) didn't seem worth it — happy to add if you disagree.

sid597:
No video because we need more than one user to test this out, we have removed the collapsable setting from the global panel. Now user can close and open directly from in the left sidebar just as the other personal left sidebar panels they have

mdroidian:
> No video because we need more than one user to test this out, we have removed the collapsable setting from the global panel. Now user can close and open directly from in the left sidebar just as the other personal left sidebar panels they have

Tip: In chrome you can create multiple profiles and roam accounts are free to create.  So it is possible (and quite low effort) to create a video side by side in the same graph with different accounts. 👍

mdroidian (approved):
https://www.loom.com/share/4bfaf8e6f4744c888eec5cf5d8e11a00

PR #1202, opened 2026-07-06
ENG-1965: Rename Template-Block-props buffer block to Template

sid597:
Renames the ephemeral template buffer block users were seeing in the node settings Template tab (and on the raw node type page) from Template-Block-props to Template.

Safety note on the text-match lookups: getSubTree({parentUid: node.type, key: "Template"}) in NodeConfig.tsx / DiscourseNodeSuggestiveRules.tsx also matches this text, but getBasicTreeByParentUid returns children order-sorted and the legacy Template block is auto-created at order 0 before the buffer (order: last) can exist, so the lookup deterministically resolves to the legacy block. No read path matches the old Template-Block-props text, so no compatibility shim is needed.

Fixes ENG-1965

### DIFF
--- a/apps/roam/src/components/settings/components/EphemeralBlocksPanel.tsx
+++ b/apps/roam/src/components/settings/components/EphemeralBlocksPanel.tsx
@@ -14,7 +14,7 @@ import {
 import type { DiscourseNodeBaseProps } from "./BlockPropSettingPanels";
 
 const DEBOUNCE_MS = 250;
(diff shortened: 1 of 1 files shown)


sid597:
Revised in 18a55097: an interim commit used "Template (editing)" to sidestep the getSubTree text-match collision, but the collision turns out to be closed by ordering (the legacy Template block is always first-ordered; see updated PR description), so the buffer now uses plain "Template" per the ticket.

mdroidian (approved):

PR #1203, opened 2026-07-06
ENG-1792: Add edit block button to rendered dg-canvas

sid597:
https://www.loom.com/share/5a61fddb73824341865e254fb3431243

Adds a minimal Blueprint edit button overlaid top-right on the embedded canvas that focuses the underlying {{dg-canvas}} block via setBlockFocusAndSelection, using the same getUids/.roam-block idiom as the ResultsView Edit Block action so it also works for canvases in the right sidebar.

Fixes ENG-1792

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1203" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/canvas/CanvasEmbed.tsx
+++ b/apps/roam/src/components/canvas/CanvasEmbed.tsx
@@ -51,7 +51,7 @@ const CanvasEmbedChrome = ({
   <div className="relative h-full w-full">
     <TldrawCanvas title={title} />
     <Button
-      className="absolute right-2 top-2 z-20"
+      className="absolute bottom-2 right-8 z-20"
       icon="edit"
       minimal
       small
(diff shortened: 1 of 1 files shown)


sid597:
The crash claim is incorrect: getUids (roamjs-components/dom/getUids) accepts a null element and returns { blockUid: "", windowId: "" }, so a missing .roam-block ancestor makes the edit button a no-op, not a render crash — same failure mode as the identical idiom in ResultsView. The cast was misleading about nullability though, so d53a010d switches to the closest<HTMLDivElement>() generic, which types the lookup as nullable with no assertion.

On the button hit area over the canvas: acknowledged — it mirrors the existing bottom-right sync-icon overlay, and the top-right corner sits outside tldraw's own toolbar zones. Will verify interaction in the demo recording.

mdroidian (approved):

PR #1208, opened 2026-07-08
eng-1910-add-no-op-content-resolve-endpoint

maparent:
[ENG-1910: Add no-op Next.js content resolve endpoint](https://linear.app/discourse-graphs/issue/ENG-1910/add-no-op-nextjs-content-resolve-endpoint)

Content and concept upsert, based on cross-app types.

Note: This is ready for review in the sense it's ready for an iteration, we should discuss PR boundaries a bit more. Please focus the review on the PR boundaries, the detailed review can come after.
The PR has two routes, maybe they could be separated.
It introduces a StandaloneContent type for upserting contents outside of concepts, following existing sync functions.
In the loom, I mention an issue with inline authors; but I checked and we do not currently use this anywhere in the sync code.
That said, the upsert_concept route is only useable for nodes, we will expand it when we have checked in the converters for more types.

https://www.loom.com/share/0d11ac2c3dea41e3bfe52141701af5e7

### DIFF
--- a/apps/website/app/api/internal/space/[id]/concept/route.ts
+++ b/apps/website/app/api/internal/space/[id]/concept/route.ts
@@ -1,0 +1,57 @@
+import { NextResponse, NextRequest } from "next/server";
+
+import { createClient } from "~/utils/supabase/server";
+import {
+  createApiResponse,
+  handleRouteError,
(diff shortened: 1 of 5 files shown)


maparent:
This was meant for eng-1909

PR #1214, opened 2026-07-10
ENG-1855 Add Roam shared-node import discovery

sid597:
https://www.loom.com/share/7210cfc4b3984f2ab927e7092dfab9c1
## Summary

Adds a read-only Roam discovery surface for group-visible shared nodes.

- Starts from the current user's group grants in `ResourceAccess`, so discovery reflects shared persistence rather than direct source-app access or every resource the user can read.
- Resolves the source app, source space, direct title, optional source-local ID, and latest upstream modified time from `my_spaces`, `my_concepts`, and `my_contents`.
- Requires an instance concept with a node type plus both `direct` and typed `full` content before a resource is shown.
- Builds a stable source RID from the source-space URL and source-local ID, then marks matching `discourse-graph.importedFrom.sourceNodeRid` pages as imported.
- Adds `DG: Discover shared nodes` with Blueprint loading, retry, empty, search, reload, and status states.
- The dialog sets `autoFocus={false}`/`enforceFocus={false}` (same as `AdvancedSearchDialog`): Roam's command palette runs commands on Enter keyup, and without this the keyup clicks the autofocused close button and instantly dismisses the dialog.

## Imported status

Nothing writes `importedFrom` yet, so every node shows **Available** for now. This PR defines the prop path it reads (`discourse-graph.importedFrom.sourceNodeRid`); the import action (ENG-1859) writes that metadata when it materializes nodes.

## How to read this diff

This is 766 added lines because Roam had no discovery surface yet; it is four bounded pieces, not a broad rewrite.

| Area | Lines | Review focus |
| --- | ---: | --- |
| Discovery utility | 333 | Group access boundary, contract filtering, stable identity |
| Blueprint dialog | 263 | Read-only required fields and ordinary UI states |
| Unit tests | 163 | Contract and identity edge cases |
| Existing command registry | 7 | Sync-gated entry point |

The command registry is the only existing production file changed. There are no changes to shared packages, persistence schemas, migrations, sync behavior, or existing import code.

Suggested review order: start at `getGroupSharedResources`, follow the rows through `buildDiscoveredSharedNodes`, then scan the table and tests.

## Review guide

1. `discoverSharedNodes.ts`: group-backed candidate selection and contract filtering.
2. `DiscoverSharedNodesDialog.tsx`: read-only presentation of the ticket's source metadata.
3. `discoverSharedNodes.test.ts`: resource identity, contract completeness, imported RID matching, current-space exclusion, and sorting.
4. `registerCommandPaletteCommands.ts`: sync-gated command entry only.

## Scope

This PR intentionally stops at discovery. It does not materialize or refresh Roam nodes, import relations or assets, change sync behavior, or modify the database schema.

## Validation

- `pnpm --filter roam test` - 10 files / 53 tests passed
- `pnpm --filter roam check-types`
- `pnpm --filter roam build` - built with 0 errors
- `pnpm --filter roam lint` - 0 errors; existing warnings unchanged
- Prettier and `git diff --check`


apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,333 @@
...
+const getGroupSharedResources = async (
+  client: DGSupabaseClient,
+): Promise<ResourceAccess[]> => {
+  const groupIds = await getAvailableGroupIds(client);
+  if (groupIds.length === 0) return [];
+
+  const resources: ResourceAccess[] = [];
+  for (let from = 0; ; from += PAGE_SIZE) {
+    const response = await client
+      .from("ResourceAccess")
```
sid597, line 220: Review starts here: candidate identities come only from `ResourceAccess` rows granted to the user's groups. The later `my_*` reads only resolve those keys.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,333 @@
...
+        sourceLocalId: concept.source_local_id,
+        spaceId: concept.space_id,
+      });
+      if (!sharedResourceKeys.has(resourceKey)) return [];
+
+      const space = spacesById.get(concept.space_id);
+      const variants = contentByResource.get(resourceKey);
+      const direct = variants?.direct;
+      const full = variants?.full;
+      if (
```
sid597, line 164: This is the ENG-1847 contract gate: only typed instances with both `direct` and typed `full` content reach the dialog.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,333 @@
...
+          concept.source_local_id,
+          space.platform === "Obsidian" ? "note" : undefined,
+        );
+      } catch {
+        return [];
+      }
+
+      return [
+        {
+          alreadyImported: importedSourceRids.has(sourceNodeRid),
```
sid597, line 194: Imported status joins on the stable RID, not title or local ID, so identical local IDs in different source spaces remain distinct.

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -0,0 +1,263 @@
...
+              icon="search"
+              title={search ? "No matching shared nodes" : "No shared nodes"}
+            />
+          </div>
+        ) : (
+          <div style={{ minHeight: 0, overflow: "auto" }}>
+            <HTMLTable striped style={{ width: "100%" }}>
+              <thead>
+                <tr>
+                  <th>Source app</th>
```
sid597, line 197: This stays intentionally read-only: these columns match ENG-1855's pre-import metadata, while materialization remains in ENG-1859.

sid597 (commented):
Four short guideposts through the ENG-1855 data flow.

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -0,0 +1,229 @@
...
+        node.sourceSpaceId,
+        node.title,
+        node.sourceNodeId,
+      ].some((value) => value?.toLocaleLowerCase().includes(normalizedSearch)),
+    );
+  }, [nodes, search]);
+
+  return (
+    <Dialog
+      autoFocus={false}
```
sid597, line 137: Roam's command palette dispatches commands on Enter keyup. Without `autoFocus={false}` that keyup lands as a click on the freshly-focused close button and dismisses the dialog immediately. Same props as `AdvancedSearchDialog`.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,329 @@
...
+const getGroupSharedResources = async (
+  client: DGSupabaseClient,
+): Promise<ResourceAccess[]> => {
+  const groupIds = await getAvailableGroupIds(client);
+  if (groupIds.length === 0) return [];
+
+  const resources = await getAllPages(
+    client
+      .from("ResourceAccess")
+      .select("space_id, source_local_id")
```
mdroidian, line 221: from codex
> Make ResourceAccess pagination order unique
When a user belongs to multiple groups that grant the same resource and the result crosses the 1,000-row page boundary, multiple ResourceAccess rows tie on both current order fields. Offset pagination does not guarantee a stable order among those ties, so adjacent range requests can duplicate one row and omit another, causing an otherwise visible node to disappear.

Unlikely we will be above 1000 anytime soon, so probably ok to punt.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,329 @@
...
+  | "text"
+  | "variant"
+>;
+type SharedSpace = Pick<
+  Tables<"my_spaces">,
+  "id" | "name" | "platform" | "url"
+>;
+type ValidSharedSpace = {
+  name: string;
+  platform: "Roam" | "Obsidian";
```
mdroidian, line 35: Is there not already a type for platforms?

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,329 @@
...
+  url: string;
+};
+
+export type DiscoveredSharedNode = {
+  alreadyImported: boolean;
+  modifiedAt: string;
+  sourceApp: "Roam" | "Obsidian";
```
mdroidian, line 42: Is there not already a type for platforms?

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+          Imported
+        </Tag>
+      ) : (
+        <Tag minimal>Available</Tag>
+      )}
+    </td>
+  </tr>
+);
+
+const DiscoverSharedNodesDialog = ({ onClose }: { onClose: () => void }) => {
```
mdroidian, line 80: Probably worth modelling this off Obsidian or passing this to Johnny, but that can be deferred.

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState("");
+  const [search, setSearch] = useState("");
+
+  const loadNodes = useCallback(async (): Promise<void> => {
+    setLoading(true);
+    setError("");
+    try {
+      const context = await getSupabaseContext();
+      if (!context) throw new Error("Could not connect to shared persistence.");
```
mdroidian, line 91: Pass these errors to posthog

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+      const client = await getLoggedInClient();
+      if (!client) throw new Error("Could not connect to shared persistence.");
+      setNodes(
+        await discoverSharedNodes({
+          client,
+          currentSpaceId: context.spaceId,
+        }),
+      );
+    } catch (loadError) {
+      console.error("Failed to discover shared nodes:", loadError);
```
mdroidian: Prefer posthog instead of console.error.

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+      )}
+    </td>
+  </tr>
+);
+
+const DiscoverSharedNodesDialog = ({ onClose }: { onClose: () => void }) => {
+  const [nodes, setNodes] = useState<DiscoveredSharedNode[]>([]);
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState("");
+  const [search, setSearch] = useState("");
```
mdroidian: Let's be more specific.  Set search what?  EG: setSearchTerm, setSearchQuery, etc

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,329 @@
...
+  sourceNodeId?: string;
+  sourceNodeRid: string;
```
mdroidian, line 44: Can we derive some of the other properties from RID?  If so, probably worth dropping them from the definition.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,329 @@
...
+          .eq("is_schema", false)
+          .in("source_local_id", ids),
+        client
+          .from("my_contents")
+          .select(
+            "content_type, last_modified, source_local_id, space_id, text, variant",
+          )
+          .eq("space_id", spaceId)
+          .in("source_local_id", ids)
+          .in("variant", ["direct", "full"]),
```
mdroidian, line 291: Do we need to pull in `full` at this stage?  Or should we defer this until a user actually decides to import it?

If we need it as a gate, could we just check for the existance of `full` in the query, so that we don't need to pass the entire `content` to the client for every possible node?

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,329 @@
...
+
+const getResourceKey = ({
+  sourceLocalId,
+  spaceId,
+}: {
+  sourceLocalId: string;
+  spaceId: number;
+}): string => `${spaceId}:${sourceLocalId}`;
+
+const getLatestTimestamp = (timestamps: (string | null)[]): string | null => {
```
mdroidian, line 71: `Concept.last_modified` and `Content.last_modified` are PostgreSQL timestamp without time zone columns. The application writes UTC values, but PostgREST returns them without a Z or offset, so Date.parse(timestamp) interprets them in the browser’s local timezone. This definitely shifts the displayed time and can also affect latest-selection/sorting around DST transitions.

Could we normalize these values here before any comparison and have getLatestTimestamp return a canonical `toISOString()` value? That keeps the PR-local fix small and also makes the existing dialog formatter safe. The test fixtures should use the production shape, such as `2026-06-14T15:00:00`, rather than including Z.

Out of scope, but this is something that should be looked at repo wide: [ENG-2034: Standardize UTC timestamp handling across the repository](https://linear.app/discourse-graphs/issue/ENG-2034/standardize-utc-timestamp-handling-across-the-repository)

mdroidian (approved):
Sorry for the delay in review.

I'm going to approve this so we can unblock the stacked PRs, but there are 2 big concerns:

1. We shouldn't pass the entire `full` content to the client for all possible nodes.  That should be deferred to only when a user decides to import.  Let's make sure we either fix that here or create a follow up ticket to fix it. ([link](https://github.com/DiscourseGraphs/discourse-graph/pull/1214#discussion_r3583059835))
2. The timestamp isn't normalized to UTC so it will provide different values to different clients. ([link](https://github.com/DiscourseGraphs/discourse-graph/pull/1214#discussion_r3583162168))

The [RID comment](https://github.com/DiscourseGraphs/discourse-graph/pull/1214#discussion_r3582881655) is the next strongest to look at.

The rest are **strong** recommendations.

maparent:
I made full optional in [ENG-2016](https://linear.app/discourse-graphs/issue/ENG-2016/adjustment-to-cross-app-node-contract-make-full-content-optional), which should make this easier.<br>Sorry, I thought I could review tonight but it's late.<br>I agree with the RID comment. What we're converging on elsewhere: Use localId, which will sometimes contain a RID. It may be worth adding a `is_rid` function; it would basically look for either `https://` or `orn:` prefix at this point.

maparent:
And yes UTC!

maparent:
For me UTC is  a blocker, because it means inserting invalid data in the database.

apps/roam/src/utils/discoverSharedNodes.ts
sid597: Will address this in ENG-2019 (#1216), under the shared discovery contract

apps/roam/src/utils/discoverSharedNodes.ts
sid597: Will address this in ENG-2019 (https://github.com/DiscourseGraphs/discourse-graph/pull/1216), under the shared discovery contract

apps/roam/src/utils/discoverSharedNodes.ts
sid597: Deferring this contract simplification to ENG-2019 (#1216), where SharedNodeCandidate becomes the canonical model shared by Roam and Obsidian.

apps/roam/src/utils/discoverSharedNodes.ts
sid597: Deferred to ENG-2019 (#1216). The shared discovery implementation will separate direct rows from full summaries, treat full as optional, and avoid selecting full text during discovery.

apps/roam/src/utils/discoverSharedNodes.ts
sid597: Addressed: production-shaped timestamps get Z before parsing and getLatestTimestamp returns canonical toISOString() values.

sid597:
ENG-1855 — addressed in this PR
  ├── RID correctness: preserve source local IDs that are already https:// or orn: RIDs
  ├── UTC correctness: normalize PostgREST timestamps before comparison and return canonical ISO values
  ├── searchTerm / setSearchTerm rename
  └── PostHog/internalError reporting for discovery-load failures

  ENG-2019 — follow-up in #1216
  ├── shared discovery query and candidate architecture
  ├── no full-text transfer during discovery; defer fetching full content until import
  ├── optional full content, following ENG-2016
  ├── separate direct-content rows from full-content summaries
  ├── unique ResourceAccess pagination ordering
  ├── named shared Platform type backed by Enums<"Platform">
  ├── evaluate and remove candidate properties that can safely be derived from RID
  ├── preserve the ENG-1855 RID and UTC behavior during extraction
  └── comprehensive shared-module tests for Roam and Obsidian consumers

sid597:
@mdroidian @maparent addressed some comments others will be addressed in next pr

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,336 @@
...
+
+const getSharedNodeRows = async ({
+  client,
+  currentSpaceId,
+}: {
+  client: DGSupabaseClient;
+  currentSpaceId: number;
+}): Promise<SharedNodeRows> => {
+  const resources = (await getGroupSharedResources(client)).filter(
+    (resource) => resource.space_id !== currentSpaceId,
```
maparent, line 255: nit: this should be pushed to Db request. Not vital.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,336 @@
...
+  for (const spaceId of spaceIds) {
+    const sourceLocalIds = resources
+      .filter((resource) => resource.space_id === spaceId)
+      .map((resource) => resource.source_local_id);
+    for (const ids of chunk(sourceLocalIds, RESOURCE_ID_CHUNK_SIZE)) {
+      const [conceptsResponse, contentsResponse] = await Promise.all([
+        client
+          .from("my_concepts")
+          .select(
+            "is_schema, last_modified, schema_id, source_local_id, space_id",
```
maparent, line 278: I'd expect metadata and author_id here, but I guess they can both come later.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -0,0 +1,336 @@
...
+          )
+          .eq("space_id", spaceId)
+          .eq("is_schema", false)
```
maparent, line 281: Add `.eq("arity", 0)` to avoid relations.

maparent (changes requested):
One tiny needed change, two comments. Focused on the backend.

apps/roam/src/utils/discoverSharedNodes.ts
sid597: Resolved in #1216: the shared SharedNodeCandidate keeps rid + the raw fields deliberately — Obsidian's consumer needs raw sourceLocalId and the numeric spaceId, which aren't derivable from the RID, while computing rid centrally keeps the isRid preservation (https:// / orn: local IDs) in one place instead of both apps re-deriving it.

apps/roam/src/utils/discoverSharedNodes.ts
sid597: Deferring per your "can come later": in #1216 the shared module's my_contents (direct) select already carries author_id, created, metadata for the Obsidian consumer, but nothing consumes them from the concepts select yet — will add them there when a consumer appears.

apps/roam/src/utils/discoverSharedNodes.ts
sid597: Done in #1216: discovery now reads full as a summary (last_modified only) and full is optional per ENG-2016 — full text is only fetched at import time.

apps/roam/src/utils/discoverSharedNodes.ts
sid597: Done in #1216: the current-space filter is now .neq("space_id", currentSpaceId) on the ResourceAccess request (and pagination orders by the full PK for a unique total order).

maparent (approved):

PR #1216, opened 2026-07-10
ENG-2019: Extract shared cross-space node discovery into packages/database

sid597:
https://github.com/user-attachments/assets/f3ea6116-2fc7-429a-a3dd-a5fbb8c267af

> < />!NOTE< />  <br>Was stacked on DiscourseGraphs/discourse-graph#1214 ([ENG-1855](https://linear.app/discourse-graphs/issue/ENG-1855/add-roam-shared-node-import-discovery)), now merged — base is `main` and the diff shows only the extraction.

Closes [ENG-2019](https://linear.app/discourse-graphs/issue/ENG-2019/extract-shared-cross-space-node-discovery-into-packagesdatabase).

## What

One shared module — `@repo/database/lib/sharedNodes` — now lists group-visible shared nodes cross-space for both apps:

| Old | New |
| -- | -- |
| Roam `discoverSharedNodes.ts`: `getGroupSharedResources`, `getSharedNodeRows`, `buildDiscoveredSharedNodes`, `chunk`, `getLatestTimestamp` | moved to `packages/database/src/lib/sharedNodes.ts` (`listGroupSharedNodes` + pure `buildSharedNodeCandidates`) |
| Obsidian `getPublishedNodesForGroups`: inline `my_contents` query | thin wrapper over `listGroupSharedNodes`, maps to the existing `PublishedNode` shape |

Roam keeps its `DiscoveredSharedNode` type, `discoverSharedNodes` entry point, and app-side imported-RID detection (`roamAlphaAPI`); it maps candidates through a small exported `toDiscoveredSharedNodes`. Already-imported detection stays app-side in both apps by design.

## The one behavior change (Obsidian)

Obsidian's import listing previously returned **any** cross-space `my_contents` row (RLS-trust only — its `groupIds` parameter was dead code — no pagination, no contract check). It now uses the shared semantics extracted from Roam:

* **Contract gate**: only nodes with an instance concept (`schema_id` set) and a `direct` content are listed. `full` is optional ([ENG-2016](https://linear.app/discourse-graphs/issue/ENG-2016/adjustment-to-cross-app-node-contract-make-full-content-optional)) and only its `last_modified` is read at discovery — full text is never fetched until import.
* **Explicit group gating**: candidates come from `ResourceAccess` ∩ `getAvailableGroupIds` (RLS stays as defense-in-depth). Whether to move to RLS-trust instead is deliberately out of scope.
* **Pagination**: `getAllPages` + 100-id chunked reads (the old query silently capped at the PostgREST row limit).
* Minor: `modifiedAt` now takes the latest timestamp across the concept **and** both content variants (was: content rows only).

## Deviations from a pure code move

* The `my_contents` select gains `author_id, created, metadata` — fields the Obsidian consumer needs.
* `metadata !== null` guard in the Obsidian mapper: the old code crashed on a `direct` row with null metadata (`typeof null === "object"` passes, then `.filePath` throws); Roam-origin rows make that case likely now.
* The Obsidian wrapper re-wraps query failures in `Error` (the shared module throws raw `PostgrestError`, which the modal's Notice would render as `[object Object]`).
* `listGroupSharedNodes` takes optional `groupIds` so the Obsidian modal (which already fetches them for its empty-groups notice) doesn't trigger a second `getAvailableGroupIds` query. Roam omits it.

## Infra

* `packages/database` gets a minimal vitest setup mirroring `packages/content-model`; script is `test:unit` because cucumber owns `test`. Roam's pure-builder tests moved over (9 tests); Roam keeps 2 mapper tests.
* `turbo.json` + CI gain a `test:unit` step so the moved tests stay enforced (they previously ran via Roam's CI-filtered suite).
* `pnpm-lock.yaml` is a 3-line hand edit reusing the catalog's existing vitest resolution — a regenerated lockfile re-resolved unrelated packages and broke Obsidian's typecheck, so re-resolution was avoided entirely.

Out of scope, unchanged on purpose: the 4 other `chunk` copies, `importPreview.ts` (preview-time schema checks are a different concern), and any RLS-vs-explicit gating redesign.

## Review feedback from DiscourseGraphs/discourse-graph#1214, resolved here

* **No full-text transfer during discovery** (mdroidian's top concern): the `my_contents` read is split — `direct` rows carry the listing fields; `full` is a summary select (`last_modified, source_local_id, space_id` only).
* **Platform type**: inline `"Roam" | "Obsidian"` unions replaced with `Enums<"Platform">` (local alias, matching `supabaseContext.ts` usage).
* **Predicates pushed to the request** (maparent): `.neq("space_id", currentSpaceId)` on the `ResourceAccess` query. `buildSharedNodeCandidates` intentionally keeps its own current-space and variant checks — the exported pure builder is the semantic contract and shouldn't depend on callers pre-filtering; the query-side filters are transport optimizations.
* **Unique pagination order**: `ResourceAccess` pages order by `(space_id, source_local_id, account_uid)` — the full primary key, so offset pagination can't duplicate/drop tied rows.
* **1855 fixes preserved through extraction** (ported into the shared module during the base merge): `arity = 0` filter, UTC normalization (`getLatestTimestamp` returns canonical `toISOString()`; `created` normalized the same way, so the Obsidian mapper drops its `+ "Z"`), and RID-shaped source-local IDs preserved via `isRid`.

**Deferred, on purpose**: `metadata`/`author_id` on the concepts select (no consumer needs them yet — maparent agreed "can come later") and slimming the candidate to RID-derived fields (Obsidian needs raw `sourceLocalId` + numeric `spaceId`, which aren't derivable from the RID; computing `rid` centrally keeps the `isRid` preservation in one place — replied on DiscourseGraphs/discourse-graph#1214).

## Verification

* `tsc --noEmit` clean in `packages/database`, `apps/roam`, `apps/obsidian`; eslint/prettier clean on all touched files (importNodes.ts keeps its 21 pre-existing warnings, count unchanged).
* 10 shared-module tests + Roam mapper tests green; `turbo run test:unit` verified locally.
* Roam built with `SUPABASE_USE_DB=local`; dist staged at `dg-test-eng-2019` for manual verification.

---

Open in Devin Review


packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  return {
+    concepts,
+    contents,
+    resources,
+    spaces: spacesResponse.data,
+  };
+};
+
+export const listGroupSharedNodes = async ({
```
sid597, line 250: Entry point. Everything below this file's exports started as Roam's ENG-1855 code moved over; the deltas beyond a pure move are the optional `groupIds` (caller-supplied to avoid a duplicate `getAvailableGroupIds` query — Obsidian passes it, Roam doesn't) and the #1214 review items applied here: `full` fetched as a summary only, `.neq` current-space filter + full-PK ordering on ResourceAccess, `Enums<"Platform">`, and UTC/RID/arity preserved from the 1855 fixes.

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,319 @@
...
+    (timestamp): timestamp is string =>
+      typeof timestamp === "string" && !Number.isNaN(Date.parse(timestamp)),
+  );
+  if (validTimestamps.length === 0) return null;
+  return validTimestamps.reduce((latest, timestamp) =>
+    Date.parse(timestamp) > Date.parse(latest) ? timestamp : latest,
+  );
+};
+
+export const buildSharedNodeCandidates = ({
```
sid597: Pure builder = the shared semantics both apps now use: contract gate (instance concept with `schema_id` + `direct` title; `full` is optional per ENG-2016 and only contributes `last_modified`), exact ResourceAccess identity match, current-space exclusion, newest-first sort. This is the semantics reconciliation — Obsidian previously had none of these gates. The current-space and variant checks are intentionally duplicated here even though the queries also filter them: the exported pure builder is the contract and shouldn't depend on callers pre-filtering.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -47,255 +18,23 @@ export type DiscoveredSharedNode = {
...
-  return validTimestamps.reduce((latest, timestamp) =>
-    Date.parse(timestamp) > Date.parse(latest) ? timestamp : latest,
-  );
-};
-
-export const buildDiscoveredSharedNodes = ({
-  concepts,
-  contents,
-  currentSpaceId,
+export const toDiscoveredSharedNodes = ({
```
sid597, line 22: Roam swap: behavior-identical output. What stays app-side is the imported-RID lookup (`roamAlphaAPI` block props) and this mapper to the existing `DiscoveredSharedNode` shape (`sourceSpaceId` was always the space URI — the candidate now carries it directly).

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -40,84 +41,35 @@ export const getPublishedNodesForGroups = async ({
...
-  // the latest last_modified per node and prefer "direct" for text (title).
-  const { data, error } = await client
-    .from("my_contents")
-    .select(
-      "source_local_id, space_id, text, created, last_modified, variant, metadata, author_id",
-    )
-    .neq("space_id", currentSpaceId);
-
-  if (error) {
+  const candidates = await listGroupSharedNodes({
```
sid597: Obsidian swap — the PR's one behavior change lands here (see description). Same `PublishedNode` shape out. The old `+ "Z"` conversions are gone on purpose: the shared module now returns canonical `toISOString()` values for `created`/`lastModified`, so appending another Z would produce `NaN`. Two hardening deltas vs the old code: null-metadata guard (old code threw on `metadata: null`) and re-wrapping `PostgrestError` so the modal Notice shows a real message.

sid597 (commented):
Reviewer orientation — the four places to read, in order.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -47,255 +18,23 @@ export type DiscoveredSharedNode = {
...
 };
 
-type SharedNodeRows = {
-  concepts: SharedConcept[];
-  contents: SharedContent[];
-  resources: ResourceAccess[];
-  spaces: SharedSpace[];
-};
-
-const getResourceKey = ({
```
sid597, line 57: → sharedNodes.ts. `getResourceKey` and `chunk` are verbatim copies. `getLatestTimestamp` carries the 1855 UTC fix: it normalizes Z-less PostgREST timestamps and returns canonical `toISOString()` values (`created` goes through the same normalization). `SharedContent` gains `author_id, created, metadata` (Obsidian consumer needs them) and drops `content_type` (only the old full-gate used it).

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -47,255 +18,23 @@ export type DiscoveredSharedNode = {
...
-    (timestamp): timestamp is string =>
-      typeof timestamp === "string" && !Number.isNaN(Date.parse(timestamp)),
-  );
-  if (validTimestamps.length === 0) return null;
-  return validTimestamps.reduce((latest, timestamp) =>
-    Date.parse(timestamp) > Date.parse(latest) ? timestamp : latest,
-  );
-};
-
-export const buildDiscoveredSharedNodes = ({
```
sid597, line 81: → sharedNodes.ts as `buildSharedNodeCandidates`. Sort/identity logic unchanged; the gate changed per ENG-2016 + #1214 review: `full` is optional and only its `last_modified` is considered (no full-text requirement). Also modified: drops `importedSourceRids`/`alreadyImported` (now mapped app-side, L29 of this file) and outputs the neutral candidate shape (`rid`, `platform`, `lastModified`, + `spaceUri`/`created`/`authorId`/`directMetadata` pass-throughs).

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -47,255 +18,23 @@ export type DiscoveredSharedNode = {
...
-      ];
-    })
-    .sort(
-      (left, right) =>
-        Date.parse(right.modifiedAt) - Date.parse(left.modifiedAt) ||
-        left.title.localeCompare(right.title),
-    );
-};
-
-const getGroupSharedResources = async (
```
sid597, line 212: → sharedNodes.ts `getGroupSharedResources`. Deltas vs the 1855 code: optional `groupIds` param (callers that already fetched group ids — the Obsidian modal — skip the internal `getAvailableGroupIds` refetch), the current-space exclusion pushed into the request as `.neq("space_id", currentSpaceId)` (maparent's #1214 nit), and ordering extended to `(space_id, source_local_id, account_uid)` — the full ResourceAccess PK — so offset pagination has a unique total order.

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -47,255 +18,23 @@ export type DiscoveredSharedNode = {
...
-    ).values(),
-  ];
-};
-
-const chunk = <T>(values: T[], size: number): T[][] =>
-  Array.from({ length: Math.ceil(values.length / size) }, (_, index) =>
-    values.slice(index * size, (index + 1) * size),
-  );
-
-const getSharedNodeRows = async ({
```
sid597, line 247: → sharedNodes.ts `getSharedNodeRows`. The single `my_contents` read is split per mdroidian's #1214 concern: `direct` rows carry the listing fields (title + `author_id, created, metadata`), `full` is a summary select (`last_modified, source_local_id, space_id`) — no full text leaves the DB at discovery. Also threads `groupIds` through and keeps the 1855 `arity = 0` filter on concepts.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -40,84 +41,36 @@ export const getPublishedNodesForGroups = async ({
   groupIds: string[];
   currentSpaceId: number;
 }): Promise<Array<PublishedNode>> => {
-  if (groupIds.length === 0) {
-    return [];
-  }
-
-  // Query my_contents (RLS applied); exclude current space. Get both variants so we can use
-  // the latest last_modified per node and prefer "direct" for text (title).
-  const { data, error } = await client
```
sid597, line 49: Not moved — replaced. This query's semantics (any cross-space row, RLS-only, unpaginated) are superseded by the shared module's gated listing; the variant-grouping/direct-preference logic it did by hand is what `buildSharedNodeCandidates` does behind the gate.

sid597 (commented):
Moved-to map for the deleted code — each comment marks where it landed in `packages/database/src/lib/sharedNodes.ts` and whether it changed.

sid597:
ENG-1855 — addressed in this PR
  ├── RID correctness: preserve source local IDs that are already https:// or orn: RIDs
  ├── UTC correctness: normalize PostgREST timestamps before comparison and return canonical ISO values
  ├── searchTerm / setSearchTerm rename
  └── PostHog/internalError reporting for discovery-load failures

  ENG-2019 — follow-up in #1216
  ├── shared discovery query and candidate architecture
  ├── no full-text transfer during discovery; defer fetching full content until import
  ├── optional full content, following ENG-2016
  ├── separate direct-content rows from full-content summaries
  ├── unique ResourceAccess pagination ordering
  ├── named shared Platform type backed by Enums<"Platform">
  ├── evaluate and remove candidate properties that can safely be derived from RID
  ├── preserve the ENG-1855 RID and UTC behavior during extraction
  └── comprehensive shared-module tests for Roam and Obsidian consumers

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,360 @@
...
+    fullContentSummaries,
+    resources,
+    spaces: spacesResponse.data,
+  };
+};
+
+export const listGroupSharedNodes = async ({
+  client,
+  currentSpaceId,
+  groupIds,
```
mdroidian: Looks like this is only used in Obsidian, but has no use other than listGroupSharednodes. The same call is used downstream at getGroupSharedResources.

So remove it from Obsidian's listGroupSharedNodes call, then we can remove it from the subsequent listGroupSharedNodes and getGroupSharedResources.

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,360 @@
...
+>;
+type Platform = Enums<"Platform">;
+
+type ValidSharedSpace = {
+  name: string;
+  platform: Platform;
+  url: string;
+};
+
+export type SharedNodeCandidate = {
```
mdroidian: The term "candidate" means something specific in DG: Candidate Nodes.  It also highly likely we'll add these to the db sooner than later.

So let's refrain from using "candidate" in types/variables/functions unless it is specifically referring to "Candidate Nodes".

I believe obsidian used "publishedNodes", which makes sense to me.  But I am open to other suggestions.

apps/obsidian/src/utils/importNodes.ts
mdroidian: `getPublishedNodesForGroups` is now redundant.  It looks like the only thing it is doing is adding `filePath`, which doesn't warrant an entire wrapped function.

apps/obsidian/src/utils/importNodes.ts
mdroidian: It looks like a few specific functions weren't replaced 1:1.  Could you walk through those decisions?  For example:
- text
- modifiedAt

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,360 @@
...
+
+export const listGroupSharedNodes = async ({
+  client,
+  currentSpaceId,
+  groupIds,
+}: {
+  client: DGSupabaseClient;
+  currentSpaceId: number;
+  groupIds?: string[];
+}): Promise<SharedNodeCandidate[]> => {
```
mdroidian: Are we planning on re-using a CrossApp contract for the return type?

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,360 @@
...
+      const space = spacesById.get(concept.space_id);
+      const direct = directByResource.get(resourceKey);
+      if (!space || typeof direct?.text !== "string") return [];
+
+      const lastModified = getLatestTimestamp([
+        concept.last_modified,
+        direct.last_modified,
+        fullModifiedByResource.get(resourceKey) ?? null,
+      ]);
+      if (!lastModified) return [];
```
mdroidian: We could be more lenient and use createdAt if lastModified is empty.

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,360 @@
...
+
+      let rid: string;
+      try {
+        rid = isRid(concept.source_local_id)
```
mdroidian: When would source_local_id be rid?

mdroidian (changes requested):

apps/obsidian/src/utils/importNodes.ts
mdroidian: We can either make that change here or create/point to a linear ticket where that change will be made.  (the latter is preferred with the current size of this PR.)

packages/database/src/lib/sharedNodes.ts
sid597: I went with SharedNode rather than publishedNodes since this module is app-neutral (Roam calls it 'discovered', Obsidian 'published')

apps/obsidian/src/utils/importNodes.ts
sid597: [ENG-2045: Remove Obsidian getPublishedNodesForGroups wrapper, consume shared SharedNode directly](https://linear.app/discourse-graphs/issue/ENG-2045/remove-obsidian-getpublishednodesforgroups-wrapper-consume-shared)

apps/obsidian/src/utils/importNodes.ts
sid597: text (title)
- Old: const text = direct?.text ?? latest.text ?? ""; — preferred the direct row, but fell back to the latest-modified row's text (which can be the full variant, i.e. the entire markdown body), then to "".
- New: title: direct.text, and the node is dropped if there's no direct row with string text (sharedNodes.ts:183).
- Why: the contract gate. A node with no direct content isn't a valid shared node, so instead of silently putting the full-markdown body (or an empty string) into the title, it's filtered out. The old ?? latest.text could surface a node's whole body as its title.

modifiedAt
- Old: max last_modified across the content rows only, + "Z" → epoch, 0 if null.
- New: getLatestTimestamp([concept.last_modified, direct.last_modified, full.last_modified]), canonical toISOString(), mapped to epoch; falls back to created if all are null (612e6516).
- Why: (a) it now also folds in the concept timestamp — a node's type/schema metadata can change without a content edit, and discovery should reflect the newer of the two; (b) created fallback instead of 0, so a node with only a created date still sorts/displays sensibly.

packages/database/src/lib/sharedNodes.ts
sid597: tbh I am not sure this is being carried over from past and for our existing usecases for roam and obsidian we are not using rid, and then case when we will be using it is in future. So I think we should remove this altogether, what do you guys think? @mdroidian  @maparent 

The future usecase is  re-sharing (a node imported from one app, re-published under its origin RID via the converter)

packages/database/src/lib/sharedNodes.ts
sid597: CrossApp contract is for the setters and getters converters, we do have the setters converter for getters we will do it in ENG-1856 so, this will be handled in that pr. 

SharedNode is a lightweight listing of what a user sees for the nodes that are in shared space.

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,353 @@
...
+const getGroupSharedResources = async (
+  client: DGSupabaseClient,
+  currentSpaceId: number,
+): Promise<ResourceAccess[]> => {
+  const availableGroupIds = await getAvailableGroupIds(client);
+  if (availableGroupIds.length === 0) return [];
+
+  const resources = await getAllPages(
+    client
+      .from("ResourceAccess")
```
maparent: Trying to understand what's the point of looking at ResourceAccess instead of directly at Content/Concept. You have a good point that the groupId was not used in Obsidian, but do we ever filter on it? (Counterpoint: We might need to do so to isolate sharing to public at some point. But even then, I think we can find another way to do so and eliminate this call altogether.)
I think the Obsidian pattern of getting straight at the data and counting on RLS was actually more efficient.

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,353 @@
...
+  if (spacesResponse.error) throw spacesResponse.error;
+
+  const concepts: SharedConcept[] = [];
+  const directContents: SharedContent[] = [];
+  const fullContentSummaries: SharedContentSummary[] = [];
+  for (const spaceId of spaceIds) {
+    const sourceLocalIds = resources
+      .filter((resource) => resource.space_id === spaceId)
+      .map((resource) => resource.source_local_id);
+    for (const ids of chunk(sourceLocalIds, RESOURCE_ID_CHUNK_SIZE)) {
```
maparent: Ok I think this should be in the skills somewhere: Never make a database query inside a loop if you can avoid it, that's a performance sink. In this case it's particularly easy to NOT restrict by space_id, just get everything and redistribute if desired (which is not even needed here.)

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,353 @@
...
+      const [conceptsResponse, directResponse, fullResponse] =
+        await Promise.all([
+          client
+            .from("my_concepts")
+            .select(
+              "is_schema, last_modified, schema_id, source_local_id, space_id",
```
maparent: Much less obvious optimization: Can be done in a single query.
```
      const conceptsResponse = await client
        .from("my_concepts")
        .select(
          `is_schema, last_modified, schema_id, source_local_id, space_id,
              direct:content_of_concept!inner(author_id, created, last_modified, metadata, source_local_id, space_id, text, variant),
              full:content_of_concept!inner(last_modified, source_local_id, space_id)`,
        )
        .eq("space_id", spaceId)
        .eq("is_schema", false)
        .eq("arity", 0)
        .eq("direct.variant", "direct")
        .eq("full.variant", "full")
        .in("source_local_id", ids);
      if (conceptsResponse.error) throw conceptsResponse.error;
      fullContentSummaries.push(...conceptsResponse.data.map((c) => c.full));
      directContents.push(...conceptsResponse.data.map((c) => c.direct));
      concepts.push(
        ...conceptsResponse.data.map((c) => {
          const { direct, full, ...other } = c;
          return other;
        }),
      );
```
(Assuming full not optional in this context.)

packages/database/src/lib/sharedNodes.ts
maparent: The use case for this is when we have a local link to a node that was imported from another space. Then in the database storage we use the original node's dbId, and in source_local_id we use Rid.
So this should stay.

maparent (changes requested):
Many optimizations can be done to reduce Db access. We should discuss the issue of whether we will filter by group and how today.

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,353 @@
...
+    .map(normalizeUtcTimestamp)
+    .filter((timestamp): timestamp is string => typeof timestamp === "string");
+  if (validTimestamps.length === 0) return null;
+  return validTimestamps.reduce((latest, timestamp) =>
+    timestamp > latest ? timestamp : latest,
+  );
+};
+
+export const buildSharedNodes = ({
+  concepts,
```
mdroidian: if this is just taking in nodes, let's name it "nodes" or something similar, eventually should be CrossAppNode contract, but that can be defered to the next project

packages/database/src/lib/sharedNodes.ts
mdroidian: Discussed 2026-07-22, MAP said this Rid is only for relations while the call it is being used on is for nodes, so it is not required.

packages/database/src/lib/sharedNodes.ts
sid597: This comment is to be addressed we can do the other's later

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,345 @@
...
+        )
+        .eq("is_schema", false)
+        .eq("arity", 0)
+        .in("source_local_id", ids),
+      client
+        .from("my_contents")
+        .select(
+          "author_id, created, last_modified, metadata, source_local_id, space_id, text, variant",
+        )
+        .in("source_local_id", ids)
```
maparent: Sorry, one more small round. In theory, there is a small risk that a source_local_id in your own space collides with one found in ResourceAccess. It is easier to just ask for resources outside your space, `.neq(space_id, currentSpaceId)`, and to drop the `.in("source_local_id", ids)` clause altogether. (And in that case you don't need the call to `getGroupSharedResources` at all.)

packages/database/src/lib/sharedNodes.ts
```diff
@@ -0,0 +1,345 @@
...
+        .eq("variant", "direct"),
+      client
+        .from("my_contents")
+        .select("last_modified, source_local_id, space_id")
+        .in("source_local_id", ids)
```
maparent: Same as line 311.

maparent (changes requested):
One last small logic flaw.

maparent (approved):
lgtm, thanks!

mdroidian (approved):
Let's go ahead based on MAP's review

PR #1221, opened 2026-07-11
ENG-1865 Publish Roam stored relations for shared nodes

maparent:
https://linear.app/discourse-graphs/issue/ENG-1865/publish-roam-stored-relations-for-shared-nodes

PR walkthrough:
https://www.loom.com/share/c8a96f0cdbc74dcf8fada181ca2f7144

Note: In the loom, I mention a bit of code waiting for 1856. That got merged, and I did make the adjustments accordingly, so disregard that.

Demo:
https://www.loom.com/share/2d93fd0570784f2eb25fac7c514911a3

### DIFF
--- a/apps/roam/src/utils/publishNodesToGroups.ts
+++ b/apps/roam/src/utils/publishNodesToGroups.ts
@@ -156,12 +156,12 @@ export const gatherCorrespondingRelations = async ({
         relations
           .filter(
             (r) =>
-              (publishedIds.has(r.sourceUid) ||
-                (forNodeIds ? forNodeIds.has(r.sourceUid) : false) ||
-                groupSpaceIds.has(isImportedFrom(r.sourceUid) || 0)) &&
-              (publishedIds.has(r.destinationUid) ||
-                (forNodeIds ? forNodeIds.has(r.destinationUid) : false) ||
-                groupSpaceIds.has(isImportedFrom(r.destinationUid) || 0)),
+              (publishedIds.has(r.sourceUid) || // source already published
+                (forNodeIds ? forNodeIds.has(r.sourceUid) : false) || // source will be published
+                groupSpaceIds.has(isImportedFrom(r.sourceUid) || 0)) && // source imported from known space
+              (publishedIds.has(r.destinationUid) || // destination already published
+                (forNodeIds ? forNodeIds.has(r.destinationUid) : false) || // destination will be published
+                groupSpaceIds.has(isImportedFrom(r.destinationUid) || 0)), // destination imported from known space
           )
           .map((r) => r.relationId),
       ];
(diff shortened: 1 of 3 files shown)


packages/database/src/inputTypes.ts
```diff
@@ -6,7 +6,7 @@ export type LocalAccountDataInput = Partial<
 export type LocalDocumentDataInput = Partial<
   Omit<
     Database["public"]["CompositeTypes"]["document_local_input"],
-    "author_inline"
+    "author_inline" | "contents"
```
maparent, line 9: "content" is of type unknown, so we get typescript errors when using the LocalDocumentDataInput explicitly (even though indirectly through LocalConceptDataInput, LocalContentDataInput.) 
It is not used yet, so just removing from input type for now.
I guess this did not arise earlier because we were building objects, not using the type explicitly.

apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -13,6 +30,300 @@ type PublishNodesResult = {
...
+    await getSyncedRelationLocalIds(client, spaceId),
+  );
+  const isImportedFrom = (nodeLocalId: string): number => {
+    let cached = sourceIdOfNodes[nodeLocalId];
+    if (cached === undefined) {
+      cached = sourceIdOfNodes[nodeLocalId] =
+        getSpaceIdOf(nodeLocalId) || spaceId;
+    }
+    return cached === spaceId ? 0 : cached;
+  };
```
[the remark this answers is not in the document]
maparent: This is valid, but it's related to function that is expected to be integrated in ENG-1865 (forthcoming.) So the bug is really a placeholder.

mdroidian:
My understanding from the [last meeting](https://roamresearch.com/#/app/discourse-graphs/page/zGDM8LV_6) is that this PR is will be refactored.

I'll move it to draft for now to reflect the status.

mdroidian:
MAP would like to base this on the upsert route [ENG-2039](https://linear.app/discourse-graphs/issue/ENG-2039/create-a-generic-upsert-route)

mdroidian:
@codex review

apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -1,27 +1,211 @@
...
+        groupId,
+        relations
+          .filter(
+            (r) =>
+              (publishedIds.has(r.sourceUid) ||
+                (forNodeIds ? forNodeIds.has(r.sourceUid) : false) ||
+                groupSpaceIds.has(isImportedFrom(r.sourceUid) || 0)) &&
+              (publishedIds.has(r.destinationUid) ||
+                (forNodeIds ? forNodeIds.has(r.destinationUid) : false) ||
+                groupSpaceIds.has(isImportedFrom(r.destinationUid) || 0)),
```
mdroidian: Let's make an appropriately named const for each semantic chunk of this filter

mdroidian (approved):
See codex comments

apps/roam/src/utils/publishNodesToGroups.ts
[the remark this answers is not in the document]
maparent: Will do this downstream.

maparent: Adding comments instead. Adding consts means we lose the short-circuit evaluation.

mdroidian: How would adding const mean we lose the short-circuit evaluation?

PR #1222, opened 2026-07-12
ENG-2027: See individual Canvas frame in a block (dg-canvas frame embed)

mattakamatsu:
Closes ENG-2027. See the issue for the product framing; this is the implementation + self-review notes.

## What

One block syntax, `{{dg-canvas:}}`, now drives both the whole-canvas embed and a frame-anchored one:

- `{{dg-canvas: [[Canvas]]}}` → the original whole-canvas embed, **unchanged**.
- `{{dg-canvas: [[Canvas]] "Frame" shape:ID}}` → live embed zoomed to that frame; tracks it as it moves/resizes, ⌖ re-center button, tldraw chrome hidden by default (⌘. restores it).
- Frame referenced by **shape id** (stable across renames) with **name fallback**. Any broken / unknown / malformed frame reference **silently degrades to the whole canvas**.
- Insert via slash command → pick canvas → pick frame.

The `{{dg-frame}}` attribute from the first draft is gone — merged into `{{dg-canvas}}`.

## Shared-code bugs fixed along the way

Both pre-existing, surfaced by exercising the embed on real canvases:

- **Remote-sync `ValidationError` that blanked canvas pages.** The pull-watch handler applied the raw persisted store via `applyDiff`, which validates but never migrates — so any record predating `MigrateNodeTypeToDiscourseNode` (a legacy node-uid shape) threw inside `mergeRemoteChanges` and tore the page down, for plain `{{dg-canvas}}` too. Extracted to `canvasRemoteMerge.ts`: migrate the incoming snapshot first (same machinery as `loadSnapshot`), drop still-invalid records individually, report failures instead of throwing. Unit-tested against real tldraw stores.
- **CSS that hid all page blocks when an embed mounted.** The block-hiding `:has()` rule only excluded `.dg-canvas-embed`; any new embed wrapper blanked the host page.

## Self-review (10-angle pass, findings addressed in the fix commits)

- `fix`: a `}` in a frame name no longer parses to null and blanks the block (lazy tail + serializer sanitize).
- `fix`: cancel the zoom-on-mount `requestAnimationFrame` on unmount; re-center refreshes the not-found state.
- `perf`: resolve frames from the raw persisted store instead of building a throwaway migrated editor (also dedupes with the picker and stops a legacy-store throw).
- `chore`: remove the unused `instanceKey` session-key plumbing left dead by the merge.
- `style`: use Blueprint surfaces for the embed chrome instead of a hardcoded `rgba(...)` (AGENTS.md).
- `chore`: explicit return types on the new exports.

**For reviewers:** `refreshConfigTree.ts` gained a null-guard (a phantom null child was aborting extension load). It's unrelated to this feature and only hardens one of the ~16 callers of that helper — happy to split it into its own PR if preferred. Canvas concurrency remains last-write-wins (a remote edit in the same 350 ms window as your own save can be dropped until the next sync), unchanged in spirit from today.

## Verification

- `check-types`, `lint` (0 errors), and `test` (62) green.
- Live-verified in `plugin-testing-akamatsulab2`: all four routing cases (plain / valid-frame / unknown-frame / malformed-tail) route as expected with no errors, the frame embed zooms to its frame, and the ⌖ button + degradation paths work.


mattakamatsu:
it also picked up a couple of pre-existing bugs and linting issues; those can be ignored/rejected if desired

sid597:
@mdroidian is this still on your radar or should I pick this up?

apps/roam/src/components/canvas/CanvasFrameEmbed.tsx
```diff
@@ -0,0 +1,170 @@
...
+  // Manage the camera ourselves (zoom-to-frame on every mount, no session
+  // persistence) and start with the tldraw chrome hidden (cmd+. restores it).
+  const embedOptions: CanvasEmbedOptions = {
+    disableSessionPersistence: true,
+    onEditorMount: handleEditorMount,
+    defaultFocusMode: true,
+  };
+
+  return (
+    <div style={{ position: "relative", height: "100%", width: "100%" }}>
```
mdroidian, line 141: use tailwind styles instead of inline

apps/roam/src/components/canvas/CanvasFrameEmbed.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { Button, Card, Elevation } from "@blueprintjs/core";
+import type { Editor, TLFrameShape, TLShapeId } from "tldraw";
+import { TldrawCanvas, type CanvasEmbedOptions } from "./Tldraw";
+import { getCanvasFrameShapes } from "./useRoamStore";
+
+const FRAME_ZOOM_INSET = 16;
+
+export type FrameRef = { name?: string; shapeId?: string };
+
+// Decide whether a parsed frame argument actually maps to a frame on the canvas
```
mdroidian, line 11: Let's remind the agent to view the agents.md re: comments

`Add comments only when necessary; descriptive names should minimize the need for comments`

apps/roam/src/components/canvas/CanvasFrameEmbed.tsx
```diff
@@ -0,0 +1,170 @@
...
+      .getCurrentPageShapes()
+      .filter(
+        (shape): shape is TLFrameShape =>
+          shape.type === "frame" &&
+          ((shape as TLFrameShape).props.name ?? "").trim().toLowerCase() ===
+            target,
+      );
+    if (matches.length > 1) {
+      // eslint-disable-next-line no-console
+      console.warn(
```
mdroidian, line 69: Generally, only use `console` for debugging
Use a toast if we need to communicate to the user, posthog if want to record the error

apps/roam/src/components/canvas/CanvasFrameEmbedDialog.tsx
```diff
@@ -0,0 +1,198 @@
+import React, { useCallback, useEffect, useMemo, useState } from "react";
```
mdroidian, line 1: Let's tie this into the existing embed dialog rather than creating a new dialog/separate slash command

apps/roam/src/components/canvas/CanvasFrameEmbedDialog.tsx
```diff
@@ -0,0 +1,198 @@
...
+      });
+      onClose();
+    },
+    [selectedPage, onSelect, onClose],
+  );
+
+  const renderPageStep = () => {
+    if (canvasPages === null)
+      return (
+        <div className="text-sm text-[#5c7080]">Loading canvas pages...</div>
```
mdroidian, line 101: Does roam recognize `text-[#5c7080]`?
Regardless, why are we using a custom color here?

apps/roam/src/components/canvas/canvasRemoteMerge.ts
```diff
@@ -0,0 +1,228 @@
+// Merging persisted canvas state from Roam block props into a live tldraw
```
mdroidian, line 1: This should be a separate ticket

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -190,7 +190,29 @@ export const isPageUid = (uid: string) =>
     ":node/title"
   ];
 
-export const TldrawCanvas = ({ title }: { title: string }) => {
+// Options for mounting the canvas inside a block embed (`{{dg-canvas: ...}}`).
```
mdroidian, line 193: Claude, chill on the comments

apps/roam/src/components/canvas/tldrawStyles.ts
```diff
@@ -1,12 +1,14 @@
 // tldrawStyles.ts because some of these styles need to be inlined
 export default /* css */ `
-  /* Hide Roam Blocks only when a full-page canvas is present (not embedded) */
-  .roam-article:has(.roamjs-tldraw-canvas-container:not(.dg-canvas-embed *)) .rm-block-children  {
+  /* Hide Roam Blocks only when a full-page canvas is present (not embedded).
+     Every block-embed wrapper class must be excluded here, or mounting that
```
mdroidian, line 4: The original comment captured the intent just fine.

apps/roam/src/components/canvas/CanvasFrameEmbed.tsx
```diff
@@ -0,0 +1,170 @@
...
+export type FrameRef = { name?: string; shapeId?: string };
+
+// Decide whether a parsed frame argument actually maps to a frame on the canvas
+// by scanning the persisted frame shapes (no editor mounted). The classic embed
+// renderer uses this to route: no match -> the frame argument is ignored and the
+// whole canvas renders. Sync-mode canvases may have a slightly stale snapshot;
+// the worst case is falling back to the whole-canvas embed.
+export const findCanvasFrameRef = ({
+  pageUid,
+  frameName,
```
mdroidian, line 18: Let's not store/search for `frameName` (which is subject to change), only use `frameShapeId` which is stable.

`frameName` should be part of the UI when choosing, but not part of the saved data

apps/roam/src/components/canvas/useRoamStore.ts
```diff
@@ -182,6 +138,27 @@ export const getRoamCanvasSnapshot = ({
   };
 };
 
+// The raw persisted record map from block props, without constructing or
```
mdroidian, line 141: Claud, chill on the comments

apps/roam/src/components/canvas/useRoamStore.ts
```diff
@@ -492,18 +385,42 @@ export const useRoamStore = ({
           (after?.[":block/props"] || {}) as json,
         ) as Record<string, json>;
         const rjsqb = props["roamjs-query-builder"] as Record<string, unknown>;
+        // Any pending merge was computed from an older props state than this
```
mdroidian, line 388: What's the use case for a "pending merge"?  When is this likely to happen?

apps/roam/src/utils/refreshConfigTree.ts
```diff
@@ -1,13 +1,46 @@
-import getBasicTreeByParentUid from "roamjs-components/queries/getBasicTreeByParentUid";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
```
mdroidian, line 1: This change should be it's own ticket where the problem is clearly documented/repeated.  The fix should also be in roamjs-components, not here.

mdroidian (changes requested):
> it also picked up a couple of pre-existing bugs and linting issues; those can be ignored/rejected if desired

@mattakamatsu Yeah, generally speaking we want each PR/Ticket to solve a single problem and be less than 500 lines (before tests).  For the sake of a quicker review merge, let's drop the `cavasRemoteMerge`/`useRoamStore` and the `refreshConfigTree` fixes.

I also did a quick once over.

PR #1223, opened 2026-07-12
Render dg-canvas frame embeds as static snapshots by default (v1)

mattakamatsu:
Stacked on #1222 (v0 frame embeds) — **do not merge before it**; retarget to `main` after v0 lands.

## What

Frame-anchored `{{dg-canvas: [[Page]] "Frame" shape:ID}}` embeds now render a **read-only SVG image** of the frame exported from persisted canvas data, instead of mounting a full interactive tldraw editor per embed. Spec: `dg-prototypes/canvas-frame-embed/README.md` (v1 section, decisions resolved 2026-07-11).

- **Default = snapshot** for frame embeds; a trailing `live` token (`{{dg-canvas: [[Page]] "Frame" shape:ID live}}`) opts back into the v0 live editor. Frameless embeds unchanged (always live).
- **Sync-mode canvases** always fall back to the live embed with an info badge — their block-props snapshot is frozen at the local→sync switch, so a static render could be arbitrarily stale.
- **Hover chrome**: *Open canvas* (navigates to the canvas page and zooms to the frame once on arrival, via a one-shot sessionStorage hint) and *Edit here* (swaps the live embed in place until the block re-renders).
- **Refresh model**: recompute on mount — block re-render = fresh snapshot, no watchers.

## Why not `TldrawImage`

In tldraw 2.4.6, `TldrawImage` builds its internal store with `useTLStore({ snapshot, shapeUtils })` only — no `bindingUtils`, no `migrations` — so DG canvases (per-relation-id binding records, legacy node-uid shape types) throw during store creation. `CanvasFrameSnapshot` replicates its export flow (temporary headless `Editor` → `getSvgString` → blob URL → ``) on top of a store built with the exact live-mount recipe (`buildFrameExportStore`), so legacy records migrate on load the same way they do in the live embed. tldraw's single-frame export handles the rest: descendants included and clipped to the frame, neighbors excluded, no frame border/label, bounds = frame page bounds — no store filtering, no hand-rolled transforms.

## Tests

- `dgCanvasEmbed.test.ts`: `live` token grammar (all arg shapes, degradation on misplaced/miscased/junk tails, serialize round-trip) + `getFrameEmbedMode` matrix
- `canvasFrameExportStore.test.ts`: legacy node-uid-typed child of a frame migrates on load; nested-frame parentId chain survives; missing migrations throw (documents why the dep-builder matters)
- `canvasFrameZoomHint.test.ts`: one-shot semantics, per-page scoping, TTL expiry, corrupt-value cleanup, null-storage safety
- `buildAllAddReferencedNodeByAction.test.ts`: extraction from `Tldraw.tsx` pinned

`npm test` (82 passing), `check-types`, eslint, and `npm run build` all clean.

## Not yet done

- Live verification in `plugin-testing-akamatsulab2` (the verification matrix is in the spec README — snapshot vs live rendering, legacy canvas, placeholders, Open-canvas zoom, Edit-here swap). The on-disk `apps/roam/dist` is built from this branch and ready to load.
- Cosmetic: exported SVG text falls back to system fonts (`DiscourseNodeUtil.toSvg` doesn't embed font defs) — noted as an optional follow-up in the spec.

### DIFF
--- a/apps/roam/src/components/canvas/CanvasEmbed.tsx
+++ b/apps/roam/src/components/canvas/CanvasEmbed.tsx
@@ -9,7 +9,8 @@ import getUids from "roamjs-components/dom/getUids";
 import getTextByBlockUid from "roamjs-components/queries/getTextByBlockUid";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
 import { TldrawCanvas } from "./Tldraw";
-import { CanvasFrameEmbed, findCanvasFrameRef } from "./CanvasFrameEmbed";
+import { findCanvasFrameRef } from "./CanvasFrameEmbed";
+import { CanvasFrameEmbedRouter } from "./CanvasFrameSnapshot";
 import { parseDgCanvasEmbed } from "~/utils/dgCanvasEmbed";
 
 const CanvasEmbedPlaceholder = ({ message }: { message: string }) => (
@@ -101,7 +102,12 @@ export const renderCanvasEmbed = (
   renderWithUnmount(
     <ExtensionApiContextProvider {...onloadArgs}>
       {frame ? (
-        <CanvasFrameEmbed title={parsed.title} frame={frame} />
+        <CanvasFrameEmbedRouter
+          title={parsed.title}
+          pageUid={pageUid}
+          frame={frame}
+          live={parsed.live}
+        />
       ) : (
         <CanvasEmbedChrome title={parsed.title} location={location} />
       )}
--- a/apps/roam/src/components/canvas/CanvasFrameEmbed.tsx
+++ b/apps/roam/src/components/canvas/CanvasFrameEmbed.tsx
@@ -1,12 +1,11 @@
 import React, { useCallback, useEffect, useRef, useState } from "react";
-import { Button, Card, Elevation } from "@blueprintjs/core";
-import type { Editor, TLFrameShape, TLShapeId } from "tldraw";
+import { Button, Card, Elevation, Icon } from "@blueprintjs/core";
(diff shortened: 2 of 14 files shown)


mattakamatsu:
**Live-verified in `plugin-testing-akamatsulab2`** (against `Canvas/frame-testing`, frames `test1f`/`test2f`, which includes discourse-node shapes and a legacy-typed record):

- ✅ Frame snapshot renders the frame content only — velocity-plot image + red `[[RES]]` and green `[[CLM]]` discourse nodes with correct colors; neighbors excluded, no frame border/label. Renders headlessly on a page where the canvas was never opened.
- ✅ Refresh-on-render: fresh blob image on every mount, no watchers.
- ✅ `live` token → v0 live editor; frameless `{{dg-canvas}}` → live whole-canvas embed.
- ✅ Hover chrome: **Edit here** swaps the live editor in place (zoomed to the frame, ⌖ button present); **Open canvas** navigates to the canvas page and zooms to the frame on arrival.
- ✅ No blank-page regression; blocks stay visible.

**One bug found and fixed during testing** (commit 8c66e4df): Open-canvas zoom-on-arrival used a single `requestAnimationFrame` after `onMount`. That works on a warm canvas load but on a **cold** load the editor's viewport isn't measured yet one frame in, so `zoomToBounds` no-ops and the camera stays at the default origin. Reproduced deterministically (cleared session camera → cold load ended at `{0,0}` despite the hint being consumed and the frame resolving). Fixed by retrying across frames until `getViewportScreenBounds` is measured, then zooming once (bounded to ~90 frames). Re-verified live: cold full-page load now lands framed on `test1f`.

PR #1218, opened 2026-07-10
ENG-1856 Store source identity metadata for Roam imported nodes

sid597:
https://www.loom.com/share/92670b3f35cd4341ae3adb9c82334ba8

## Summary

- store Roam imported-node provenance under `discourse-graph.importedFrom`
- use `sourceNodeRid` as the canonical cross-app identity and keep `sourceModifiedAt` as refresh state
- add consistent read, write, bulk lookup, and single-node lookup helpers
- route Roam shared-node discovery through the shared bulk lookup helper
- preserve sibling block props when imported metadata is written or refreshed

## Why

Roam needs a stable way to recognize an imported shared node after its page title or content changes. Matching by display title would make duplicate prevention and refresh dependent on presentation and namespace decisions.

Obsidian already uses `importedFromRid` as the canonical imported identity and stores the imported source timestamp as `lastModified`. This implementation follows the same boundary in Roam:

- `sourceNodeRid` is the identity
- `sourceModifiedAt` is refresh state
- source app, source space, and source-local ID are not duplicated because they are encoded by the RID
- source title is not persisted as identity

This supersedes the closed v1 PR, #1161, and addresses its review feedback about the redundant six-field identity shape.

## Implementation

The metadata is stored in page/block props rather than page content:

```ts
{
  "discourse-graph": {
    importedFrom: {
      sourceNodeRid,
      sourceModifiedAt,
    },
  },
}
```

Writing merges the existing `discourse-graph` object so unrelated metadata, including relation migration state, survives. Content replacement during import or refresh does not touch these block props.

The helper exposes:

- parsing and reading imported source metadata
- writing updated source metadata
- collecting all imported source RIDs for discovery duplicate prevention
- finding an imported local node UID by canonical source RID

## Dependency

This PR is intentionally stacked on ENG-2019 / #1216, which is itself stacked on ENG-1855. ENG-1855 introduced the Roam discovery consumer and ENG-2019 extracted the cross-space listing logic while leaving already-imported detection app-side.

Once #1216 merges, this PR can be retargeted to `main` without changing the ENG-1856 implementation diff.

## Validation

- `pnpm --filter roam test -- src/utils/__tests__/importedSourceIdentity.test.ts src/utils/__tests__/discoverSharedNodes.test.ts`
  - 11 test files passed
  - 51 tests passed
- `pnpm --filter roam check-types`
- targeted ESLint for the changed files
- Prettier
- `git diff --check`

[...]

PR #1224, opened 2026-07-12
ENG-1858 Materialize Obsidian-origin markdown into Roam

sid597:
## What

A Roam materializer for one Obsidian-origin `CrossAppNode`: create a local page from the shared markdown, or update the page already imported from the same source RID instead of duplicating it, preserving imported source identity for later refresh.

Base is `main` (ENG-1856 landed in #1218).

## Behaviour

**First import** — validate the input, refuse if the title is already held by a page we did not import, create the page from the markdown body via Roam's native `data.page.fromMarkdown`, then store the source RID and source-modified time. If the identity write fails the new page is removed, so no untracked partial import is left behind.

**Re-import** — find the page by source RID, materialize replacement blocks and delete the previously captured children only after the replacement succeeds, then rename the title, then refresh source identity.

Failures return a `stage` (`validate-input`, `find-imported-node`, `title-collision`, `create-page`, `replace-page-content`, `update-page-title`, `write-source-identity`) and never drop the source RID or timestamp, so ENG-1859 can report per-node outcomes.

## Decisions a reviewer will want

**Obsidian frontmatter is stripped before Roam sees the markdown.** The `full` content variant is the raw note file — `apps/obsidian/src/utils/upsertNodesAsContentWithEmbeddings.ts:74-81` sets it to `vault.read(node.file)`, frontmatter included (`node.frontmatter` is stored additionally in `metadata`, not instead). Roam has no frontmatter concept, so passing it through materialized `---`/`nodeTypeId:`/… as literal blocks. The mirror-direction importer does the same thing with gray-matter (`apps/obsidian/src/utils/importNodes.ts:948`). `stripFrontmatter` lives in `@repo/content-model` next to `normalizeLineEndings` rather than as a Roam-local copy, since it is pure text logic both apps care about. **Open question for @maparent:** the contract says `direct` = title, `full` = body — should the Obsidian producer be stripping frontmatter at publish time instead? This strip is needed either way for rows already published.

**Origin comes from the `Platform` enum, not the RID string.** `discoverSharedNodes` already surfaces `sourceApp: Enums<"Platform">` from `my_spaces.platform`, so `materializeObsidianNode` takes it as input rather than testing the RID for an `obsidian:` prefix — that prefix test also fails for the web-URL RID form that `spaceUriAndLocalIdToRid` emits for `http`-shaped space URIs. RID shape is checked with `isRid`.

**Title collision is refused, not resolved.** Roam page titles are unique and `findImportedNodeUidBySourceRid` cannot see a page the user authored locally. Creating over one is the only genuinely destructive outcome available here, so the create path returns `title-collision` and writes nothing. Richer collision UX is out of scope per the issue; this is the documented MVP0 behaviour.

**`setBlockProps` stays the only block-props writer.** Making the identity write observable needs an awaitable write, and the reason it was not awaitable is a floating promise inside `setBlockProps` (which already carried a `no-floating-promises` warning). Rather than keep a second copy of the merge logic in `writeImportedSourceIdentity`, `setBlockProps` now returns the update promise. The 10 existing call sites are prefixed with `void`, which preserves their current fire-and-forget behaviour exactly — whether any of them *should* await is pre-existing and unchanged by this PR.

**`node.nodeType` is intentionally unused.** Obsidian types a note through frontmatter; Roam types a page through its title-format regex. Reconciling those is a cross-grammar decision, so MVP0 keeps the source title verbatim — an imported node is recognised as its type only when the title already matches a local format. Active type reconciliation/creation is follow-up work.

**Only `text/obsidian+markdown` is accepted for `full`.** No producer emits `text/markdown` for an Obsidian node, and `text/markdown` payloads embed the title as an H1 (`buildFullMarkdown` in `roamToCrossAppConverters.ts`), which would duplicate the title into the page body.

## Scope

Materializer API only — nothing calls it yet. ENG-1859 wires the discovery selection UI and multi-node reporting; it is blocked by this issue. Relations, assets and refresh-all are ENG-1872 / ENG-1860.

Over the size guideline at ~15 files. The coupling: 341 lines are the two test files, `setBlockProps` + its 10 `void` call sites are one mechanical change needed to make the identity write awaitable, and `stripFrontmatter` is 20 lines in `@repo/content-model`. Splitting the `setBlockProps` change into its own base PR is fine if preferred.

## Validation

- `pnpm --filter roam test` — 12 files / 63 tests pass
- `pnpm --filter @repo/content-model test` — 2 files / 9 tests pass
- `pnpm --filter roam check-types`, `pnpm --filter @repo/content-model check-types` — pass
- `pnpm --filter roam build` — 0 errors
- ESLint / Prettier clean on every changed file (the remaining warnings in `DiscourseNodeConfigPanel.tsx` and `supabaseContext.ts` are pre-existing, on lines this PR does not touch)

## Runtime proof still owed

`data.page.fromMarkdown` and `data.block.fromMarkdown` are not in `roamjs-components`' `roamAlphaAPI` type, so their signatures are declared locally and every test mocks that declaration. **Green tests here do not prove the real API behaves as declared.** Not yet driven against a live graph: the created/updated page shapes, whether `page.fromMarkdown` honours a caller-supplied uid, what `page.update` with `"merge-pages": false` does when the target title is taken, and how Obsidian wikilinks/embeds land in Roam blocks (markdown fidelity limits are ENG-1882). Do not merge on the unit tests alone.

## Known boundary

Roam mutations are not transactional. If replacement markdown is imported but deleting a previous child fails, the result reports `replace-page-content` and the page can briefly contain both new and leftover old blocks.

Also worth a follow-up: `findImportedNodeUidBySourceRid` (from #1218) takes the first row of an unordered datalog result, so if two pages ever carry the same source RID, which one a re-import overwrites is non-deterministic.

## Linear

[ENG-1858](https://linear.app/discourse-graphs/issue/ENG-1858/materialize-obsidian-origin-markdown-into-roam)

### DIFF
--- a/apps/roam/src/components/canvas/Clipboard.tsx
+++ b/apps/roam/src/components/canvas/Clipboard.tsx
@@ -202,7 +202,7 @@ export const ClipboardProvider = ({
     if (!isInitialized || !clipboardBlockUid) return;
 
     try {
-      setBlockProps(clipboardBlockUid, {
+      void setBlockProps(clipboardBlockUid, {
         [CLIPBOARD_PROP_KEY]: pages,
         [CLIPBOARD_SHOW_NODES_ON_CANVAS_PROP_KEY]: showNodesOnCanvas,
       });
--- a/apps/roam/src/components/canvas/canvasSyncMode.ts
+++ b/apps/roam/src/components/canvas/canvasSyncMode.ts
@@ -42,7 +42,7 @@ const setRoamJsQueryBuilderProps = ({
   pageUid: string;
   nextRjsqb: Record<string, json>;
 }): void => {
-  setBlockProps(pageUid, {
+  void setBlockProps(pageUid, {
     [QUERY_BUILDER_PROP_KEY]: nextRjsqb,
   });
 };
--- a/apps/roam/src/components/settings/DiscourseNodeConfigPanel.tsx
+++ b/apps/roam/src/components/settings/DiscourseNodeConfigPanel.tsx
@@ -110,7 +110,7 @@ const DiscourseNodeConfigPanel: React.FC<DiscourseNodeConfigPanelProps> = ({
                 },
               ],
             }).then((valueUid) => {
-              setBlockProps(
+              void setBlockProps(
                 valueUid,
                 DiscourseNodeSchema.parse({
                   text: label,
--- a/apps/roam/src/components/settings/utils/accessors.ts
+++ b/apps/roam/src/components/settings/utils/accessors.ts
@@ -657,7 +657,7 @@ const setBlockPropAtPath = (
     return currentContext[currentKey];
   }, updatedProps);
 
-  setBlockProps(blockUid, updatedProps, false);
+  void setBlockProps(blockUid, updatedProps, false);
 };
 
 const setBlockPropBasedSettings = ({
@@ -1193,7 +1193,7 @@ export const getAllDiscourseNodes = (): DiscourseNode[] => {
       );
       const retryResult = DiscourseNodeSchema.safeParse(migrated);
       if (retryResult.success) {
-        setBlockProps(pageUid, retryResult.data, false);
+        void setBlockProps(pageUid, retryResult.data, false);
         nodes.push(
           toDiscourseNode({
             ...retryResult.data,
--- a/apps/roam/src/components/settings/utils/init.ts
+++ b/apps/roam/src/components/settings/utils/init.ts
@@ -163,7 +163,7 @@ const initializeSettingsBlockProps = (
         Object.keys(existingProps).length === 0 ||
         !schema.safeParse(existingProps).success
       ) {
-        setBlockProps(uid, defaults, false);
+        void setBlockProps(uid, defaults, false);
       }
 
       // Reconcile placeholder relation keys with real block UIDs.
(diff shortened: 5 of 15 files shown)


maparent:
Ok. First Rid because it's an easy question: I think we could augment ridToSpaceUriAndLocalId so it also delivers the Platform in the structure. The information is present for the platforms we know, and I think it's a safe assumption that the Rids we design will always include that information. Would that help?

maparent:
Stripping Frontmatter is more complex. Right now we put the frontmatter in two places: the Concept's literal_content (under `source_data`), and we also keep it in the concept's body. In theory this is redundant; in practice, the literal_content is jsonb, does not preserve key order, and hence the transfer would be lossy if we did not keep it there. Possibly acceptably so? OTH, putting it in literal_content allows it to be indexed. So I would tend to keep this design, but also open to discussion. (Which does not mean it could not be improved. I think that a lot of the internal keys should be stripped from source_data.)<br>We said we woudl translate the frontmatter to the Roam properties block; though I guess that is another scope. There is a question whether we'd take the properties from the source_data or from re-parsing the full text. I prefer the former, as Roam should not have to care about Obsidian specifics. By that rule, it should not have to strip the frontmatter either; this is something that I hope we revisit with the transfer to ATJson. We should be able to mark that region as something that should be stripped in cross-platform transfer, in a way that each platform does not have to know about other platforms. But again I see that as future scope. I hope that's helpful.

PR #1229, opened 2026-07-13
ENG-2030 Discourse context under page title doesn't show up unless `overlay` is on

mdroidian:
## Summary

- Always render discourse context under the title of discourse-node pages.
- Keep the `Overlay` preference scoped to discourse context on page references.

## Root cause

The page-title renderer was gated by the same personal preference that enables page-reference overlays, so disabling `Overlay` unintentionally hid the context beneath page titles too.

## Impact

Discourse context now appears beneath page titles in both the main window and sidebar regardless of the page-reference overlay preference.

## Validation

- Roam unit suite: 43 tests passed
- ESLint: changed file passed

### DIFF
--- a/apps/roam/src/utils/initializeObserversAndListeners.ts
+++ b/apps/roam/src/utils/initializeObserversAndListeners.ts
@@ -133,9 +133,7 @@ export const initObservers = ({
             nodeType: node.type,
           });
         }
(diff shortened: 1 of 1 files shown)


mdroidian:
@codex review

PR #1237, opened 2026-07-18
ENG-2039 Create a generic upsert route

maparent:
[FEE-890](https://linear.app/discourse-graphs/issue/FEE-890/create-a-generic-upsert-route)

[https://www.loom.com/share/b391b33f651a4788b24e0e98dd4214d8](<https://www.loom.com/share/b391b33f651a4788b24e0e98dd4214d8>)

---

Open in Devin Review

### DIFF
--- a/apps/website/app/api/internal/space/[id]/route.ts
+++ b/apps/website/app/api/internal/space/[id]/route.ts
@@ -103,7 +103,7 @@ export const POST = async (
           nodeSchemasById[c.localId] = c;
         });
       schemaResult.data
-        .filter((d) => d.arity === 0)
+        .filter((d) => d.arity === 2)
         .map((d) => {
           const c = dbRelationTypeSchemaToCrossApp(d as Concept, authorMap);
           relationTypesById[c.localId] = c;
--- a/apps/website/app/api/internal/space/[id]/route.ts
+++ b/apps/website/app/api/internal/space/[id]/route.ts
@@ -1,0 +1,167 @@
+import { NextResponse, NextRequest } from "next/server";
+
+import { createClient } from "~/utils/supabase/server";
+import {
+  createApiResponse,
+  handleRouteError,
+  defaultOptionsHandler,
+} from "~/utils/supabase/apiUtils";
+import { asPostgrestFailure } from "@repo/database/lib/contextFunctions";
+import type { Json, Tables } from "@repo/database/dbTypes";
+import { CrossAppUpsertData } from "@repo/database/crossAppContracts";
+import { LocalConceptDataInput } from "@repo/database/inputTypes";
+import {
+  crossAppNodeSchemaToDbConcept,
+  crossAppNodeToDbConcept,
+  crossAppRelationToDbConcept,
+  crossAppRelationTripleSchemaToDbConcept,
+  crossAppRelationTypeSchemaToDbConcept,
+  dbNodeSchemaToCrossApp,
+  dbRelationTypeSchemaToCrossApp,
+} from "@repo/database/lib/crossAppConverters";
+import { getAccountId } from "~/utils/supabase/account";
+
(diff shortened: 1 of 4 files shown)


packages/database/src/lib/crossAppConverters.ts
```diff
@@ -103,6 +104,24 @@ export const crossAppNodeSchemaToDbConcept = (
   });
 };
 
+export const dbNodeSchemaToCrossApp = (
+  schema: Concept,
+  authorMap: Record<number, string>,
+): CrossAppNodeSchema => {
+  const { template, template_content, ...other } =
+    schema.literal_content as Record<string, Json>;
```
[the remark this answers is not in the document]
maparent: even without template , the literal_content defaults to  an empty object, not null. This is not needed.

apps/website/app/api/internal/space/[id]/route.ts
```diff
@@ -0,0 +1,167 @@
...
+      asPostgrestFailure("Cannot parse space id", "invalid", 403),
+    );
+
+  try {
+    const supabase = await createClient();
+    const userId = await getAccountId(supabase);
+    if (userId === undefined)
+      return createApiResponse(
+        request,
+        asPostgrestFailure("Please login", "invalid", 401),
```
Comment (mdroidian) [inline apps/website/app/api/internal/space/[id]/route.ts:47]: Will this be passed to the user?  If so, as a user, when I receive this message, what should I do with it?  Stated differently, what's the use case for when a user would hit this failure route and what action could they take to rectify it?

apps/website/app/api/internal/space/[id]/route.ts
```diff
@@ -0,0 +1,167 @@
+import { NextResponse, NextRequest } from "next/server";
```
Comment (mdroidian) [inline apps/website/app/api/internal/space/[id]/route.ts:1]: Let's name this route something more explicit to what it's doing.  Right now it only conveys something about the space.
How about `/api/internal/space/[id]/graph/upsert` or just `/graph` if we plan on adding the `GET` here too.

Something to convey this is `Concept, Content, and Document` but not `User, Space, etc`

apps/website/app/api/internal/space/[id]/route.ts
```diff
@@ -0,0 +1,167 @@
...
+      (id) => !(id in nodeSchemasById),
+    );
+    const missingRelationTypeSchemaIds = [
+      ...neededRelationTypeSchemaIds,
+    ].filter((id) => !(id in relationTypesById));
+    const missingSchemaIds = [
+      ...missingNodeSchemaIds,
+      ...missingRelationTypeSchemaIds,
+    ];
+    if (missingSchemaIds.length > 0) {
```
Comment (mdroidian) [inline apps/website/app/api/internal/space/[id]/route.ts:77]: @maparent My understanding from our discussion on the 17th (discussion [ref](https://roamresearch.com/#/app/discourse-graphs/page/lf-EOV7a6), specifically this [ref](https://roamresearch.com/#/app/discourse-graphs/page/SNYOjt_6j)) was that the action-specific functions would assemble the data required for the action before calling this generic route. This block instead fetches schemas that weren’t included in the request and converts them back to `CrossApp` types. 

For example, if `publishNodes` is sharing a Claim and Evidence connected by Supports, the action-specific function should send the nodes, their node schemas, the relation, the Supports relation type schema, and the Claim–Supports–Evidence triple schema. Given that callers are expected to go through these action-specific functions, when would this route receive a relation triple schema without its referenced node and relation type schemas? If that shouldn’t happen, could we validate and reject the incomplete payload instead of fetching the dependencies here? That would also remove the need for the DB-to-CrossApp converters in this PR.

mdroidian (commented):
Leaving this as a comment mid review, because the [action-specific function comment](https://github.com/DiscourseGraphs/discourse-graph/pull/1237#discussion_r3611525400) should be discussed.

PR #1238, opened 2026-07-19
ENG-2040 Test the generic upsert route

maparent:
https://linear.app/discourse-graphs/issue/ENG-2040/test-the-generic-upsert-route

https://www.loom.com/share/7a958930a0e641d0b5552f42bb2863f2

Added after the loom: I also now have to start the next server in the GitHub test database action.


packages/database/scripts/serveAndTest.ts
```diff
@@ -16,58 +17,96 @@ if (process.env.SUPABASE_PROJECT_ID !== "test") {
...
+      stdio: ["ignore", "pipe", "inherit"],
+    });
+    this.promise = new Promise((rsc, rjc) => {
+      this.resolveCallback = rsc;
+      this.rejectCallback = rjc;
 
-const servingReady = new Promise((rsc, rjc) => {
-  resolveCallback = rsc;
-  rejectCallback = rjc;
+      // Add timeout
```
mdroidian, line 51: .

packages/database/scripts/serveAndTest.ts
```diff
@@ -16,58 +17,96 @@ if (process.env.SUPABASE_PROJECT_ID !== "test") {
...
+      this.timeoutClear = setTimeout(() => {
+        rjc(new Error("Timeout waiting for functions to serve"));
+      }, 30000); // 30 second timeout
```
mdroidian, line 54: .

packages/database/scripts/serveAndTest.ts
```diff
@@ -16,58 +17,96 @@ if (process.env.SUPABASE_PROJECT_ID !== "test") {
...
   } finally {
-    if (serve.pid)
+    if (supabaseServer.process.pid)
+      try {
+        process.kill(-supabaseServer.process.pid);
+      } catch (e) {
+        console.error("Could not kill the process");
+        // maybe it just ended on its own.
+      }
+    if (nextServer.process.pid)
```
mdroidian, line 107: On Windows process.kill(-pid) returned ESRCH and left the detached child alive, leaking port 3000.

packages/database/features/addGeneric.feature
```diff
@@ -0,0 +1,134 @@
...
+                    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+                  ]
+                }
+              }
+            }
+          }
+        ],
+        "relations": [
+          {
+            "name": "a supports relation",
```
mdroidian, line 121: Should this match `CrossAppRelation`?

mdroidian (changes requested):
Do we want to start Next for the entire database Cucumber suite? Only `addGeneric.feature` needs it, so this adds another startup dependency and failure mode to every database test.

I think the route-level coverage is worthwhile, but I’d isolate it in a separate command/job that starts Next and runs only this feature. The regular database suite could remain Supabase-only. We should also poll an HTTP health check instead of relying on the `"Ready in "` console output.

Also, are we missing `relationTypeSchemas`?

PR #1247, opened 2026-07-24
FEE 871 Mapping claims and evidence

KarolaKirsanow:
This adds a short article to obsidian/fundamentals called "Mapping claims and evidence" as suggested by our SEO analysis. The article can be copied to roam/fundamentals; it has no platform-specific content


KarolaKirsanow:
request a read from @joelchan and/or @jsmorabito for content.

jsmorabito (approved):
Looks great!! Approved 👍

PR #1253, opened 2026-07-27
ENG-1767 Setup Linear release for Obsidian

trangdoan982:
## Summary

* Add a manual `stable_obsidian_release.yaml` workflow (`workflow_dispatch`) that publishes an official stable Obsidian version via `publish.ts`, then syncs and completes the matching Linear release
* Fix `publish.ts` so it writes the resolved version back into the local `apps/obsidian/package.json`/`manifest.json` (previously only the mirror-repo copy was updated, so the local version field was permanently stale at `0.1.0`)
* One-time correction: bump local `package.json`/`manifest.json` from stale `0.1.0` to `1.5.2`, matching the actual latest stable release in the `discourse-graph-obsidian` mirror repo
* Document the release process in `apps/obsidian/README.md`

Supersedes closed PR DiscourseGraphs/discourse-graph#1074, which used an older 3-workflow design without the merge-time tracking piece. That piece is now split out to [ENG-1994](https://linear.app/discourse-graphs/issue/ENG-1994/automatically-add-merged-obsidian-issues-to-in-progress-linear-release) (DiscourseGraphs/discourse-graph#1254, stacked on this PR).

## Test plan

To test this workflow, I need to merge them first. Rationale:<br>GitHub only allows manually triggering a workflow_dispatch workflow once its file exists on the default branch (main) — per their docs: "[This event will only trigger a workflow run if the workflow file exists on the default branch](<https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch>)."

- [X] `js-yaml` parse check on the new workflow YAML
- [X] `tsc --noEmit` and `eslint` clean on `publish.ts`
- [X] Dry-ran the version-guard bash logic against edge cases (backwards version, re-releasing an already-shipped version, normal beta→stable handoff, skipping straight to a new stable)
- [ ] Manually dispatch `stable_obsidian_release.yaml` with a real version matching an existing Linear release and confirm the GitHub release, mirror repo push, and Linear sync/complete all succeed


mdroidian:
@codex review

.github/workflows/stable_obsidian_release.yaml
```diff
@@ -0,0 +1,68 @@
...
+            git config user.email "github-actions[bot]@users.noreply.github.com"
+            git add apps/obsidian/package.json apps/obsidian/manifest.json
+            git commit -m "chore: release obsidian ${{ inputs.version }} [skip ci]"
+            git push
+          fi
+
+      - name: Sync Linear release
+        uses: linear/linear-release-action@v0
+        with:
+          access_key: ${{ secrets.LINEAR_RELEASES_ACCESS_KEY }}
```
mdroidian: This key is already used for the Roam pipeline.  I don't think we can reuse this same access key.

.github/workflows/obsidian-release.yaml
```diff
@@ -0,0 +1,68 @@
+name: Stable Obsidian Release
```
mdroidian, line 1: let's match the naming style of all the other files in this folder

.github/workflows/obsidian-release.yaml
```diff
@@ -0,0 +1,68 @@
...
+          node-version: "22"
+          cache: "pnpm"
+
+      - name: Install Dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Publish stable release
+        run: cd apps/obsidian && npx tsx scripts/publish.ts --version ${{ inputs.version }}
+
+      - name: Commit version bump
```
mdroidian, line 53: Is the intent that the PR changes the manifsest.json and package.json (as the README says), or that the GitHub action will do this?

.github/workflows/stable_obsidian_release.yaml
```diff
@@ -0,0 +1,68 @@
...
+        uses: actions/setup-node@v6
+        with:
+          node-version: "22"
+          cache: "pnpm"
+
+      - name: Install Dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Publish stable release
+        run: cd apps/obsidian && npx tsx scripts/publish.ts --version ${{ inputs.version }}
```
mdroidian: `${{ inputs.version }}` is inserted into the generated Bash script before it runs. A value like `1.6.0; <command>` ends the publisher command and executes the appended command with repository write access and the job’s mirror token and Supabase values.

`publish.ts` can’t prevent this because Bash parses the separator before Node receives the arguments. Its version regex is also only anchored at the start, so it accepts trailing content when that content reaches Node.

Pass the input through an environment variable, validate it against a fully anchored stable-version pattern, and use a quoted variable:

```yaml
env:
  VERSION: ${{ inputs.version }}

run: npx tsx scripts/publish.ts --version "$VERSION"
```

Use the same validated variable in the commit message. This keeps the version as data instead of inserting it into the shell script.

mdroidian (changes requested):

.github/workflows/obsidian-release.yaml
trangdoan982: the Github action will do this. Because `publish.ts` only changes manifest.json and package.json when the version is set to stable (see `isExternalRelease`)

mdroidian: So should we change the README to reflect this?

.github/workflows/stable_obsidian_release.yaml
trangdoan982: Fixed — renamed to `OBSIDIAN_LINEAR_RELEASE_KEY` in both the Sync and Complete steps.

.github/workflows/obsidian-release.yaml
trangdoan982: Renamed to `obsidian-release.yaml` to match the kebab-case convention used by the other workflow files.

.github/workflows/obsidian-release.yaml
trangdoan982: Yes — updated the README to remove the manual version-bump step. The release process now just says: ensure the Linear release exists, then run the workflow. The workflow handles bumping `package.json` and `manifest.json` and committing them.

.github/workflows/stable_obsidian_release.yaml
trangdoan982: Fixed — `inputs.version` is now only referenced once at the job level (`VERSION: ${{ inputs.version }}` in `env:`) and all shell steps use `"$VERSION"`. Also added an anchored semver validation step (`^[0-9]+\.[0-9]+\.[0-9]+$`) that runs before install.

apps/obsidian/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "@discourse-graph/obsidian",
   "name": "Discourse Graph",
-  "version": "0.1.0",
+  "version": "1.5.2",
```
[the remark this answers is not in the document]
trangdoan982: 1.5.2 is the latest version

.github/workflows/obsidian-release.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Sync Linear release
+        uses: linear/linear-release-action@v0
+        with:
+          access_key: ${{ secrets.LINEAR_RELEASES_ACCESS_KEY }}
+          command: sync
+          version: ${{ inputs.version }}
+          include_paths: "apps/obsidian/**,packages/database/**,packages/utils/**"
+
+      - name: Complete Linear release
+        uses: linear/linear-release-action@v0
```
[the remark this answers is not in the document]
trangdoan982: Good call — fixed. `createGithubRelease` now checks for an existing release by tag before attempting to create one; if it already exists it logs and returns early, so re-running the workflow after a transient Linear failure won't abort at a duplicate-release 422.

The Linear steps themselves (`sync` and `complete`) are third-party actions, so if one of those fails in isolation the worst case is moving the release stage manually in Linear.

.github/workflows/obsidian-release.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Publish stable release
+        run: cd apps/obsidian && npx tsx scripts/publish.ts --version "$VERSION"
+
+      - name: Commit version bump
+        run: |
+          if ! git diff --quiet -- apps/obsidian/package.json apps/obsidian/manifest.json; then
+            git config user.name "github-actions[bot]"
+            git config user.email "github-actions[bot]@users.noreply.github.com"
+            git add apps/obsidian/package.json apps/obsidian/manifest.json
+            git commit -m "chore: release obsidian ${VERSION} [skip ci]"
```
trangdoan982, line 59: @mdroidian I'm personally not sure if main can take a direct commit from yaml workflow. It seems like with L10, adding `permission: content: write`, would allow for this action. But want to confirm with you here

mdroidian: Personally, I would be very hesitant of automating a commit like this.  I'd say let's do this manually for awhile first before jumping to this.

trangdoan982: from meeting 1-1: dev will need to manually create a commit to update the versioning in monorepo. Once we iron out the edge cases we can add back

mdroidian (approved):

PR #1255, opened 2026-07-27
ENG-2038 Add admin-flagged node card context menu to Obsidian tldraw

sid597:
https://github.com/user-attachments/assets/2beba1cb-1c55-43d0-a71b-b4021b43f893

## Summary

- add a default-off admin flag for the Obsidian node-card context menu
- replace tldraw's style panel only when a single `discourse-node` is selected
- show Context by default and reuse tldraw's native styling controls in the Styling tab
- embed the existing relation-panel actions in Context
- show incoming and outgoing relation groups in the new menu, including configured empty groups and deduplicated linked files
- preserve the existing relation overlay and regular tldraw shape behavior
- add one CSS rule to widen this menu because tldraw hard-caps its style panel at 148 px

## Behavior boundaries

- Flag off, non-discourse selections, and multi-select continue to render tldraw's stock `DefaultStylePanel`.
- `includeAllDirections` is explicit at the new menu call site; `embedded` controls layout only.
- The legacy relations panel keeps its prior behavior: `matchingRelations.slice(0, 1)` uses the same predicate and first match as the previous `.find(...)`, and its counterpart lookup is unchanged.
- The new menu opts into all configured directions without changing the legacy overlay.
- The localized `ComponentType` casts are required because tldraw is typed against React 19 while the Obsidian app runs React 18; plain JSX adds `TS2786` diagnostics. The repository-wide React type mismatch is intentionally out of scope.

## Scope

This PR changes seven Obsidian production files only. It adds one load-bearing CSS rule and no test command, test framework, dependency, lockfile change, or standalone model layer.

Linear: https://linear.app/discourse-graphs/issue/ENG-2038/add-admin-flagged-node-card-context-menu-to-obsidian-tldraw

## Validation

- `pnpm --filter @discourse-graphs/obsidian build` — built with 0 errors
- package-local ESLint over all changed TypeScript files — passed with 0 warnings/errors
- Prettier and `git diff --check` — passed
- built CSS contains the width override and the defined tldraw `--color-text-3` token

## Known validation gaps

- the interaction has not yet been driven in a live Obsidian vault; no running Obsidian process, configured plugin mirror, or project-local vault was available in this environment
- the vault drive must include discourse node → plain rectangle → discourse node in both themes to inspect the selection-conditional 148 px ↔ 320 px width change for jitter or poor content reflow; add a transition only if that runtime behavior reads badly
- ENG-2038's automated-test Done-When is not met: `apps/obsidian` has no test harness, and a deferral ticket plus the ticket owner's acceptance are still required
- package-wide `check-types` remains red with the same 20 pre-existing tldraw/React JSX type-identity diagnostics; no diagnostic is introduced in `NodeCardContextMenu` or `RelationPanel`
- local validation used Node 20.20.2 while the repository declares Node >=22; pnpm emitted an engine warning


sid597:
@codex review

apps/obsidian/src/components/canvas/NodeCardContextMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+                : "border-transparent text-[var(--color-text-3)]",
+            ].join(" ")}
+            onClick={() => setActiveTab(id)}
+          >
+            {label}
+          </button>
+        ))}
+      </div>
+
+      {activeTab === "context" ? (
```
trangdoan982, line 84: Image

I'd add styling to show the active state of each tab as well.

apps/obsidian/src/components/canvas/NodeCardContextMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  useEffect(() => {
+    setActiveTab("context");
+  }, [selectedNode?.id]);
+
+  if (!selectedNode) {
+    return createElement(DefaultStylePanelComponent, { isMobile });
+  }
+
+  return createElement(
```
trangdoan982, line 64: Image

Another thing is that we might need to hide this panel when user select a different tool (hmm I'm not too sure about this solution either). But when user clicks on DG tool to create a new node, if there's a previously selected DG node, both panel will show up -> confusing and cramped

apps/obsidian/src/components/canvas/overlays/RelationPanel.tsx
```diff
@@ -57,7 +57,9 @@ export type RelationsPanelProps = {
   plugin: DiscourseGraphPlugin;
   canvasFile: TFile;
   nodeShape: DiscourseNodeShape;
-  onClose: () => void;
+  onClose?: () => void;
+  embedded?: boolean;
+  includeAllDirections?: boolean;
```
trangdoan982: idk if this flag/refactor with this `includeAllDirections` is necessary. RelationPanel already computed all relations? it adjusts the label based on whether the direction is incoming (complementary lable) or outcoming (just label)

apps/obsidian/src/components/canvas/overlays/RelationPanel.tsx
```diff
@@ -460,21 +467,33 @@ export const RelationsPanel = ({
...
-        <div className="text-sm font-medium text-gray-700">{headerTitle}</div>
-      </div>
+    <div
+      className={
+        embedded
+          ? "p-3"
+          : "min-w-80 max-w-md rounded-lg border bg-white p-4 shadow-lg"
+      }
+    >
+      {!embedded && (
```
trangdoan982: i also don't think embedded need to be a state. We can extract it as a var since it's used only once

trangdoan982 (changes requested):

apps/obsidian/src/components/canvas/overlays/RelationPanel.tsx
```diff
@@ -459,6 +456,52 @@ export const RelationsPanel = ({
     }
   };
 
+  return loading ? (
+    <div className="text-center text-gray-500">Loading relations...</div>
+  ) : error ? (
+    <div className="text-center text-red-600">{error}</div>
+  ) : groups.length === 0 ? (
+    <div className="text-center text-gray-500">No relations found.</div>
```
mdroidian, line 464: This should show when there are no relation instances, not no relation types.

apps/obsidian/src/components/canvas/NodeCardContextMenu.tsx
mdroidian: image

apps/obsidian/src/components/canvas/NodeCardContextMenu.tsx
mdroidian: I still see this issue

apps/obsidian/src/components/canvas/NodeCardContextMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <RelationsPanel
+          plugin={plugin}
+          canvasFile={canvasFile}
+          nodeShape={selectedNode}
+          embedded
+          includeAllDirections
+        />
+      ) : (
+        createElement(DefaultStylePanelContentComponent, { styles })
+      )}
```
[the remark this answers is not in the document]
mdroidian: @sid597

apps/obsidian/src/components/AdminPanelSettings.tsx
```diff
@@ -72,6 +83,30 @@ export const AdminPanelSettings = () => {
 
   return (
     <div className="general-settings">
+      <div className="setting-item">
+        <div className="setting-item-info">
+          <div className="setting-item-name">(BETA) Node card context menu</div>
```
mdroidian, line 88: let's change this to discourse context

mdroidian (changes requested):
Some things I noticed

Image

1) the styling menu is now stretched.  Let's try to keep it the same size/proportion it was before.  Likely minimizing the space between the headers then centering the remaining menu will work

Original

Image

Wide

Image

2) We're missing the purple color/highlight of the header showing which tab is selected

Image

3) Hide relation types that have no relation instances

4) Immediately after creating a node I see "Linked file not found" until I click away from the node and click back

Image

apps/obsidian/src/components/AdminPanelSettings.tsx
trangdoan982: i agree, node card context is a bit confusing

sid597:
https://www.loom.com/share/ad68de3446ba4492ad809c7cb059590d

trangdoan982 (approved):
tested again, all is good. I'm still a bit bothered a bit styling-wise about the Context and Styling tab having different sizing -> when you switch between them, there's an UI jump. But I also understand the native Styling panel width might be too small. maybe an in-btw size fixed for both of them?

sid597:
Yeah I am not happy about the jump as well but currently we have 2 options and neither are good for the other's case

mdroidian:
cc @jsmorabito

mdroidian (approved):

jsmorabito:
sorry im late to this, i agree with you all its a little awkward, but id prefer to move forward with the implementation as-is then iterate since moving this forward will unblock some other tldraw obsidian work like [Canvas drawer (Obsidian)](https://linear.app/discourse-graphs/project/canvas-drawer-obsidian-26ff23cf201c/overview)

PR #1161, opened 2026-06-26
[ENG-1856] Store source identity metadata for Roam imported nodes

sid597:
https://www.loom.com/share/cc332f43f68342efb10b35764bf866b3

## What

Adds the Roam-side storage layer for imported cross-app node identity (ENG-1856, M2 "Bidirectional node import"). When Roam imports an Obsidian-origin shared node, it must remember the node's stable source identity so it can (a) prevent duplicate imports and (b) refresh it later.

`apps/roam/src/utils/importedSourceIdentity.ts`:
- **`ImportedSourceIdentity`** — 6 app-neutral fields mirroring the `CrossAppNode` contract: `sourceApp`, `sourceSpaceId`, `sourceNodeId`, `sourceNodeRid`, `sourceTitle`, `sourceModifiedAt`.
- **`toImportedSourceIdentity(node)`** — derive the stored identity from a `CrossAppNode` (title from the `direct` content variant).
- **`writeImportedSourceIdentity(uid, identity)` / `readImportedSourceIdentity(uid)`** — persist/read under the page's `:block/props` → `discourse-graph` → `importedFrom`, merging alongside any sibling discourse-graph props (mirrors `migrateRelations` / `createReifiedBlock`).
- **`findImportedNodeUidByRid(rid)`** — datalog lookup returning an already-imported page's uid (duplicate prevention), keyed on the canonical `sourceNodeRid`.

Storage is block-props (not page content), so identity survives content overwrite on import/refresh and is datalog-queryable.

## Scope / boundaries (pre-flight approved)

This is a foundational helper PR — the exported helpers **are** the deliverable; their consumers are downstream M2 issues, so there is no in-PR caller by design:
- **ENG-1855** marks already-imported nodes via `findImportedNodeUidByRid`
- **ENG-1858** materializer writes identity via `writeImportedSourceIdentity(uid, toImportedSourceIdentity(node))`
- **ENG-1859** import action dedups; **ENG-1860 / ENG-1861** refresh read identity + `sourceModifiedAt`

Deliberately out of scope (named owners): node-type mapping → ENG-1858; author/provenance → ENG-1875; discovery UI → ENG-1855.

## Notes for review

- `IMPORTED_FROM_PROP_KEY` is exported following the existing `DISCOURSE_GRAPH_PROP_NAME` precedent (wire-key constants exported from their owning module); ENG-1861 (refresh-all) / ENG-1875 (provenance) will enumerate imported pages by it.
[...]

PR #1266, opened 2026-07-29
ENG-1857 Import Roam-origin shared nodes through the existing Obsidian importer

sid597:
https://github.com/user-attachments/assets/f3ad7bb3-86ad-4eac-9056-6e22b7187748

## Problem

The producer side of the M1 contract is merged ([FEE-892](https://linear.app/discourse-graphs/issue/FEE-892/add-roam-full-markdown-content-variant-for-shared-nodes) `full` variant, [ENG-1849](https://linear.app/discourse-graphs/issue/ENG-1849/sync-roam-node-type-schema-metadata-for-shared-nodes) schema metadata), and Roam-origin shared nodes already show up in Obsidian's import modal — but importing one was guaranteed to fail. Roam's `full` variant is body-only markdown (`# Title` + body, no YAML — see `buildFullMarkdown` in `apps/roam/src/utils/roamToCrossAppConverters.ts`), while `processFileContent` hard-required `nodeTypeId` + `nodeInstanceId` frontmatter *inside the content*: every Roam node hit "importedNode missing sourceNodeTypeId" and the created file was deleted. Roam-imported files also never got `nodeInstanceId` written to frontmatter, so duplicate prevention (`findExistingImportedFile`, `getLocalNodeInstanceIds`) could never match them and they would re-list as importable forever.

## What changed

* `fetchNodeTypeSchemasForInstances` (importNodes.ts): batched per-space lookup of each instance's node-type schema (`source_local_id` + name) from the source space's Concept rows. This consolidates the two inline queries `computeImportPreview` already ran — the importer now uses the same resolution instead of duplicating it. Both queries carry `arity = 0` per the data-layer invariant (node instances and node-type schemas; relation-type schemas have `roles` → arity 2 and must be excluded).
* `processFileContent`: now takes `nodeInstanceId` (the caller always has it) and `nodeTypeIdFromConcept`. Content frontmatter still wins when present, so Obsidian-origin imports behave exactly as before; the Concept-derived id is only reached by frontmatter-less (Roam-origin) content. For Obsidian-origin the two values are identical by construction (publish writes `schema_represented_by_local_id` from the same frontmatter). It also writes `nodeInstanceId` into the imported file's frontmatter — idempotent for Obsidian-origin, and what makes stable-identity dedup work for Roam-origin.
* `computeImportPreview`: inline instance→schema queries replaced with the shared helper (also drops two `as Array<...>` casts).
* **Test**: Roam-origin fixture for `buildSharedNodes` — URL-based rid (`https://roamresearch.com/#/app/<graph>/<uid>`), platform `Roam`, production-shape (Z-less) timestamps in, normalized ISO out.

## Removals (dead code inside the touched signatures)

* `originalFilePath` param on `processFileContent` — declared and passed, never read (was an eslint warning on main).
* `if (mappedNodeTypeId !== undefined)` — `mapNodeTypeIdToLocal` always returns `string`.
* The "importedNode missing nodeInstanceId" error path — the caller supplies `nodeInstanceId` directly now; the frontmatter read it guarded is gone.
* `ParsedFrontmatter.nodeInstanceId/publishedToGroups/authorId` — never read from this parse (the reads in `refreshImportedFile` are metadata-cache frontmatter, a different object).

## Extraction parity (preview queries → shared helper)

| Old (inline in `computeImportPreview`) | New (`fetchNodeTypeSchemasForInstances`) | Delta |
| -- | -- | -- |
| instance query: `space_id, is_schema=false, in(source_local_id)` | same + `arity=0` | invariant conformance |
| schema query: `select source_local_id, name` | `select id, source_local_id, name` + `arity=0` | `id` keys the instance→schema map |
| `if (!conceptRows) continue` (silent) | `console.error` + empty map | failures logged, same control flow |
| iterate unique schema rows | iterate per-instance map values | identical outcomes — dedup preserved by existing `seenNodeTypeIds` / `nodeTypeIdToName.has` |
| `as Array<{...}>` casts ×2 | none (null-guarded narrowing) | honest types against nullable view columns |

## Identity notes

* `importedFromRid` semantics are unchanged for Roam spaces: their space URL is `https://roamresearch.com/#/app/<graph>`, and the https branch of `spaceUriAndLocalIdToRid` ignores the `"note"` subtype — matching the rid `buildSharedNodes` produces for Roam platforms exactly.
* Duplicate prevention now works end-to-end for Roam-origin: `nodeInstanceId` + `importedFromRid` frontmatter → `findExistingImportedFile` (re-import updates in place) and `getLocalNodeInstanceIds` (already-imported nodes drop out of the modal).

## Markdown fidelity gaps (documented per ticket; follow-up ticket candidates)

1. **Node-type format not persisted** — Roam schema `literal_content` is `{label, template?}` only; a Roam type with no name match in Obsidian gets a synthesized format (`CLA - {content}`), not Roam's real one.
2. **Template semantics mismatch** — Roam persists template *text* under `template`; Obsidian treats `nodeType.template` as a template *filename*, so an unmatched Roam type with a template gets an unusable value.
3. `# Title` **H1 duplication** — Roam's `full` embeds the title as H1 and the Obsidian filename carries it too (producer shape from [FEE-892](https://linear.app/discourse-graphs/issue/FEE-892/add-roam-full-markdown-content-variant-for-shared-nodes), not changed here).
4. **Roam syntax passthrough** — `{{[[query]]}}`, `{{[[table]]}}`, `{{[[kanban]]}}`, attributes (`foo::`), `^^highlights^^` (≠ Obsidian `==`), `#[[multi-word tags]]` arrive verbatim; unresolved `[[links]]` become alias-form dead links.
5. **Nodes without** `full` (optional per [ENG-2016](https://linear.app/discourse-graphs/issue/ENG-2016/adjustment-to-cross-app-node-contract-make-full-content-optional)) fail import with a counted error — pre-existing importer behavior, unchanged.

## Out of scope (per ticket)

New import UI, materializer rewrite, relations/assets/refresh behavior. `refreshAllImportedFiles` still refreshes file-by-file (pre-existing loop); the new per-space lookup rides that pattern.

## Testing

* `packages/database` vitest 8/8 (new Roam fixture included); eslint delta 0 vs main (the 17 remaining warnings pre-date this PR, minus one this PR removes); tsc clean on touched files; plugin builds with 0 errors.
* Runtime demo of the real flow (Roam-origin rows → import → dedup on re-open → re-import updates in place) to follow.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

---

Open in Devin Review

### DIFF
--- a/apps/obsidian/src/utils/importNodes.ts
+++ b/apps/obsidian/src/utils/importNodes.ts
@@ -322,6 +322,76 @@ const fetchNodeContentForImport = async ({
   };
 };
 
+type NodeTypeSchemaForInstance = {
+  nodeTypeId: string;
+  name: string;
+};
+
+export const fetchNodeTypeSchemasForInstances = async ({
+  client,
+  spaceId,
+  nodeInstanceIds,
+}: {
+  client: DGSupabaseClient;
+  spaceId: number;
+  nodeInstanceIds: string[];
+}): Promise<Map<string, NodeTypeSchemaForInstance>> => {
+  const result = new Map<string, NodeTypeSchemaForInstance>();
+
+  const { data: instanceRows, error: instanceError } = await client
+    .from("my_concepts")
+    .select("source_local_id, schema_id")
+    .eq("space_id", spaceId)
+    .eq("is_schema", false)
+    .eq("arity", 0)
+    .in("source_local_id", nodeInstanceIds);
+
+  if (instanceError || !instanceRows) {
+    console.error("Error fetching node instance concepts:", instanceError);
+    return result;
+  }
+
+  const schemaIds = [
+    ...new Set(
+      instanceRows
+        .map((row) => row.schema_id)
+        .filter((id): id is number => id !== null),
+    ),
+  ];
+  if (schemaIds.length === 0) return result;
+
+  const { data: schemaRows, error: schemaError } = await client
+    .from("my_concepts")
+    .select("id, source_local_id, name")
+    .eq("space_id", spaceId)
+    .eq("is_schema", true)
+    .eq("arity", 0)
+    .in("id", schemaIds);
+
+  if (schemaError || !schemaRows) {
+    console.error("Error fetching node type schemas:", schemaError);
+    return result;
+  }
+
+  const schemasById = new Map<number, NodeTypeSchemaForInstance>();
+  for (const row of schemaRows) {
+    if (row.id !== null && row.source_local_id !== null && row.name !== null) {
+      schemasById.set(row.id, {
+        nodeTypeId: row.source_local_id,
+        name: row.name,
+      });
+    }
+  }
+
+  for (const row of instanceRows) {
+    if (row.source_local_id === null || row.schema_id === null) continue;
+    const schema = schemasById.get(row.schema_id);
+    if (schema) result.set(row.source_local_id, schema);
+  }
+
+  return result;
+};
+
 /**
  * Fetches created/last_modified from the source space Content (my_contents) for an imported node.
  * Used by the discourse context view to show "last modified in original vault".
@@ -939,9 +1009,6 @@ const sanitizePathForImport = (path: string): string => {
 
 type ParsedFrontmatter = {
   nodeTypeId?: string;
-  nodeInstanceId?: string;
-  publishedToGroups?: string[];
-  authorId?: number;
   [key: string]: unknown;
 };
 
@@ -1096,22 +1163,24 @@ const processFileContent = async ({
   sourceSpaceId,
   sourceSpaceUri,
   rawContent,
-  originalFilePath,
   filePath,
   importedCreatedAt,
   importedModifiedAt,
   authorId,
+  nodeInstanceId,
+  nodeTypeIdFromConcept,
 }: {
   plugin: DiscourseGraphPlugin;
   client: DGSupabaseClient;
   sourceSpaceId: number;
   sourceSpaceUri: string;
   rawContent: string;
-  originalFilePath?: string;
   filePath: string;
   importedCreatedAt?: number;
   importedModifiedAt?: number;
   authorId?: number;
+  nodeInstanceId: string;
+  nodeTypeIdFromConcept?: string;
 }): Promise<
   { file: TFile; error?: never } | { file?: never; error: string }
 > => {
@@ -1134,18 +1203,14 @@ const processFileContent = async ({
   // 2. Parse frontmatter from rawContent (metadataCache is updated async and is
   //    often empty immediately after create/modify), then map nodeTypeId and update frontmatter.
   const { frontmatter } = parseFrontmatter(rawContent);
-  const sourceNodeTypeId = frontmatter.nodeTypeId;
-  if (typeof sourceNodeTypeId !== "string") {
+  const sourceNodeTypeId =
+    typeof frontmatter.nodeTypeId === "string"
+      ? frontmatter.nodeTypeId
+      : nodeTypeIdFromConcept;
+  if (sourceNodeTypeId === undefined) {
     await plugin.app.vault.delete(file);
     return {
       error: "importedNode missing sourceNodeTypeId",
-    };
-  }
-  const sourceNodeId = frontmatter.nodeInstanceId;
-  if (typeof sourceNodeId !== "string") {
-    await plugin.app.vault.delete(file);
-    return {
-      error: "importedNode missing nodeInstanceId",
     };
   }
 
@@ -1161,12 +1226,11 @@ const processFileContent = async ({
     file,
     (fm) => {
       const record = fm as Record<string, unknown>;
-      if (mappedNodeTypeId !== undefined) {
-        record.nodeTypeId = mappedNodeTypeId;
-      }
+      record.nodeTypeId = mappedNodeTypeId;
+      record.nodeInstanceId = nodeInstanceId;
       record.importedFromRid = spaceUriAndLocalIdToRid(
         sourceSpaceUri,
-        sourceNodeId,
+        nodeInstanceId,
         "note",
       );
       record.lastModified = importedModifiedAt;
@@ -1244,6 +1308,12 @@ export const importSelectedNodes = async ({
       spaceName,
     });
 
+    const nodeTypeSchemasByInstance = await fetchNodeTypeSchemasForInstances({
+      client,
+      spaceId,
+      nodeInstanceIds: nodes.map((n) => n.nodeInstanceId),
+    });
+
     // Process each node in this space
     for (const node of nodes) {
       try {
@@ -1318,11 +1388,14 @@ export const importSelectedNodes = async ({
           sourceSpaceId: spaceId,
           sourceSpaceUri: spaceUri,
           rawContent: content,
-          originalFilePath: contentFilePath,
           filePath: finalFilePath,
           importedCreatedAt: createdAt,
           importedModifiedAt: modifiedAt,
           authorId,
+          nodeInstanceId: node.nodeInstanceId,
+          nodeTypeIdFromConcept: nodeTypeSchemasByInstance.get(
+            node.nodeInstanceId,
+          )?.nodeTypeId,
         });
 
         if (result.error) {
--- a/apps/obsidian/src/utils/importNodes.ts
+++ b/apps/obsidian/src/utils/importNodes.ts
@@ -1184,7 +1184,21 @@ const processFileContent = async ({
 }): Promise<
   { file: TFile; error?: never } | { file?: never; error: string }
 > => {
-  // 1. Create or update the file with the fetched content first.
+  // 1. Parse frontmatter from rawContent (metadataCache is updated async and is
+  //    often empty immediately after create/modify) and resolve the node type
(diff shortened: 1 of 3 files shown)


apps/obsidian/src/utils/importNodes.ts
```diff
@@ -322,6 +322,76 @@ const fetchNodeContentForImport = async ({
   };
 };
 
+type NodeTypeSchemaForInstance = {
+  nodeTypeId: string;
+  name: string;
+};
+
+export const fetchNodeTypeSchemasForInstances = async ({
```
sid597, line 330: Roam's `full` variant is body-only markdown (`buildFullMarkdown` in `apps/roam/src/utils/roamToCrossAppConverters.ts` emits `# Title` + body, no YAML), so there is no embedded frontmatter to read identity from — this resolves each instance's node type from the source space's Concept rows instead. It consolidates the two queries `computeImportPreview` was already running inline (parity table in the description). `name` is consumed by the preview caller only; the import path uses `nodeTypeId`. Both queries carry `arity = 0` so relation instances / relation-type schemas (arity 2) can't slip in regardless of what ids a caller passes.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -939,9 +1009,6 @@ const sanitizePathForImport = (path: string): string => {
 
 type ParsedFrontmatter = {
```
sid597, line 1010: Trimmed to what this parse actually feeds: only `nodeTypeId` is read from content frontmatter now. The `nodeInstanceId`/`publishedToGroups`/`authorId` reads further down this file (`refreshImportedFile`) come from the metadata cache — a different object — not this parse.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1134,20 +1203,16 @@ const processFileContent = async ({
   // 2. Parse frontmatter from rawContent (metadataCache is updated async and is
   //    often empty immediately after create/modify), then map nodeTypeId and update frontmatter.
   const { frontmatter } = parseFrontmatter(rawContent);
-  const sourceNodeTypeId = frontmatter.nodeTypeId;
-  if (typeof sourceNodeTypeId !== "string") {
+  const sourceNodeTypeId =
+    typeof frontmatter.nodeTypeId === "string"
+      ? frontmatter.nodeTypeId
+      : nodeTypeIdFromConcept;
```
sid597: Ordering: content frontmatter wins when present, so Obsidian-origin imports behave exactly as before; the Concept-derived id is only reached by frontmatter-less (Roam-origin) content. For Obsidian-origin the two values are identical by construction — publish writes `schema_represented_by_local_id` from the same frontmatter it embeds in the content.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1161,12 +1226,11 @@ const processFileContent = async ({
     file,
     (fm) => {
       const record = fm as Record<string, unknown>;
-      if (mappedNodeTypeId !== undefined) {
-        record.nodeTypeId = mappedNodeTypeId;
-      }
+      record.nodeTypeId = mappedNodeTypeId;
+      record.nodeInstanceId = nodeInstanceId;
```
sid597, line 1230: This line is what makes duplicate prevention work for Roam-origin nodes: `findExistingImportedFile` (re-import updates in place) and `getLocalNodeInstanceIds` (already-imported nodes drop out of the modal) both match on `nodeInstanceId` + `importedFromRid` frontmatter. Obsidian-origin content already carries this value, so there it's an idempotent rewrite.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1161,12 +1226,11 @@ const processFileContent = async ({
...
       record.importedFromRid = spaceUriAndLocalIdToRid(
```
sid597, line 1231: Now built from the caller-supplied `nodeInstanceId` — same value the removed frontmatter read produced. Rid shape is unchanged for Roam spaces: their space URL is https-based and the https branch of `spaceUriAndLocalIdToRid` ignores the `"note"` subtype, matching what `buildSharedNodes` produces for Roam platforms. Identity stays carried once as the RID (per the #1161/#1214 direction).

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1244,6 +1308,12 @@ export const importSelectedNodes = async ({
       spaceName,
     });
 
+    const nodeTypeSchemasByInstance = await fetchNodeTypeSchemasForInstances({
```
sid597, line 1311: Batched once per space, outside the per-node loop — mirrors where the preview step does the same resolution.

sid597 (commented):
Inline notes to orient review — decisions and parity accounting are in the PR description.

apps/obsidian/src/utils/importPreview.ts
```diff
@@ -82,69 +82,34 @@ export const computeImportPreview = async ({
...
-      .in("id", schemaIds);
-
-    if (!schemaRows) continue;
-
-    for (const schema of schemaRows as Array<{
-      source_local_id: string;
-      name: string;
-    }>) {
-      const sourceNodeTypeId = schema.source_local_id;
+    const nodeTypeSchemasByInstance = await fetchNodeTypeSchemasForInstances({
```
sid597, line 85: Drop-in replacement for the two inline queries that lived here — same filters plus `arity = 0`, `select` adds `id` (keys the instance→schema map), and null-data now logs instead of silently continuing. Dedup behavior is preserved by the existing `seenNodeTypeIds` / `nodeTypeIdToName.has` checks: the map yields one entry per instance rather than unique schemas, but outcomes are identical.

packages/database/src/lib/__tests__/sharedNodes.test.ts
```diff
@@ -78,6 +78,58 @@ describe("buildSharedNodes", () => {
     ]);
   });
 
+  it("builds a Roam-origin shared node with a URL-based rid", () => {
```
sid597, line 81: Roam-origin fixture: for an https space URL the rid is `<spaceUrl>/<uid>` (no `orn:` subtype), platform `Roam`; timestamps go in as production-shape Z-less naive UTC and come out normalized ISO.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1134,20 +1203,16 @@ const processFileContent = async ({
...
   //    often empty immediately after create/modify), then map nodeTypeId and update frontmatter.
   const { frontmatter } = parseFrontmatter(rawContent);
-  const sourceNodeTypeId = frontmatter.nodeTypeId;
-  if (typeof sourceNodeTypeId !== "string") {
+  const sourceNodeTypeId =
+    typeof frontmatter.nodeTypeId === "string"
+      ? frontmatter.nodeTypeId
+      : nodeTypeIdFromConcept;
+  if (sourceNodeTypeId === undefined) {
     await plugin.app.vault.delete(file);
```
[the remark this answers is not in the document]
sid597: Good catch — fixed in f44ec994. The node-type resolution (frontmatter parse + Concept-derived fallback) now runs before any vault write, and the `vault.delete` is gone entirely: a failed lookup returns the error without creating, overwriting, or deleting anything, for both new imports and refreshes of existing files. The node just counts as failed for that attempt.

apps/obsidian/src/utils/importNodes.ts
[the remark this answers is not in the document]
sid597: Declining for this PR: `apps/obsidian` has no unit-test infra today (no runner config, zero test files), and this path is Obsidian-API-coupled end to end (vault, metadataCache, fileManager) — meaningful coverage means standing up a harness and mocking the plugin surface, which is its own piece of work rather than part of this validation ticket. The unit-testable seam (the `buildSharedNodes` contract) is covered by the added Roam-origin fixture; the importer path itself gets a runtime demo (import → dedup on re-open → re-import updates in place).

sid597:
@codex review

apps/obsidian/src/utils/importNodes.ts
maparent, line 1311: Database access inside a loop: to be avoided. But the whole function needs to go.

apps/obsidian/src/utils/importNodes.ts
maparent, line 330: I think the first part of this function is re-fetching data that we already fetched, and could be removed altogether. It's used in `importSelectedNodes` which receives `ImportableNode` from `ImportNodeModal`; those are composed from `PublishedNodes` obtained from `getPublishedNodesForGroups`; and those are made from `SharedNode`s. `SharedNode` has the `schema_id`, it should be possible to add that field back to `PublishedNodes` and `ImportableNode`, and you would have a list of schema ids to fetch. Fetching the schemas themselves (2nd part) could be done outside of the loop at 1293.
Now... I realize that this is code you inherited and refactored. That refactoring may be out of scope. (I cannot believe I missed that database loop the first time around.) We can discuss whether this should be done before or after this PR.

maparent (approved):
Ok. I would do a major refactor (see line notes), but I realize it's an optimization of previous code, and solving this is not the scope of this PR. This PR does what it sets out to do, and otherwise lgtm. But we may want to talk about that optimization at some point.

PR #1267, opened 2026-07-29
[ENG-1858] Materialize Obsidian-origin markdown into Roam

sid597:
https://github.com/user-attachments/assets/4bd7744e-37d6-445d-8b04-258c3f6dfbb6

## What this adds

- **`materializeSharedNode({ client, sharedNode })`** — consumes one Obsidian-origin `SharedNode` payload:
  - validates the boundary (platform, RID shape via `isRid`, modified-time validity, non-empty title),
  - fetches the `full` content variant at import-action time and strips frontmatter,
  - materializes via **Roam's native markdown import** — `data.page.fromMarkdown` on create, `data.block.fromMarkdown` on update — so Roam owns markdown→blocks conversion,
  - when a page with the same `sourceNodeRid` already exists, updates it in place (replacement blocks land before old ones are deleted; rename included) — re-importing never duplicates,
  - writes `importedFrom` identity via the ENG-1856 helpers only after content writes succeed, and **awaits** it,
  - returns the source identity plus `{ success: true, action }` or `{ success: false, error: { message, stage } }` so ENG-1859 can report per-node outcomes; a created page whose identity cannot be stored is removed again (no untracked partial import).
- **`stripFrontmatter`** in `@repo/content-model` next to `normalizeLineEndings` — pure text logic both apps care about.
- **`setBlockProps` now returns its `block.update` promise** so the identity write is observable; the 9 existing call sites are prefixed with `void`, preserving their fire-and-forget behavior exactly. Whether any of them should await is pre-existing and unchanged here.
- **25 unit tests** (16 materializer, 6 identity, 3 text) covering: create, update-not-duplicate (with content-before-delete ordering), rename, both collision refusals, non-Obsidian/RID/timestamp/content-type rejections, fetch errors, identity-write failure cleanup + orphan reporting, title-only page when `full` is absent.

## Decisions to check in review

- **Native `fromMarkdown` API.** `data.page.fromMarkdown` / `data.block.fromMarkdown` are not in roamjs-components' types, so their signatures are declared locally and tests mock that declaration — green tests do not prove the real API behaves as declared. Runtime demo owed before merge (see below).
- **No production call site — deliberate.** ENG-1859 (import action) is blocked by this ticket and is the consumer. Failure results are returned, not toasted/PostHog'd here — reporting belongs to the action surface, matching `DiscoverSharedNodesDialog`'s `internalError` pattern.
- **Missing `full` variant → title-only page** rather than failure (`full` is optional per ENG-2016), via typed `data.page.create`.
- **Only `text/obsidian+markdown` is accepted for `full`** — Roam-produced `text/markdown` embeds the title as an H1 (`buildFullMarkdown`), which would duplicate the title into the body.
- **Title collision = refuse with an actionable reason, write nothing** — on create, and pre-checked on rename-during-update before any mutation. Final collision UX is out of scope per the ticket.
- **Node-type mapping deferred — open question.** MVP0 keeps the source title verbatim; a title matching a local node format is recognized by Roam's existing format matching. Verbatim vs reformat-to-local-format vs auto-create (like Obsidian's `mapNodeTypeIdToLocal`) needs a call; per ENG-1856's boundary note this decision belongs to ENG-1858.
- **Timestamps:** `sourceModifiedAt` is validated and canonicalized with `toISOString()` before storage — no naive parsing of DB values (the only PostgREST reads here are `text`/`content_type`).
- Pre-existing eslint warnings in `DiscourseNodeConfigPanel.tsx` and `supabaseContext.ts` sit on lines this PR does not touch.

## Out of scope (per ticket)

Import discovery UI / multi-select (ENG-1859), relations & assets (ENG-1872), refresh-all (ENG-1860), markdown fidelity documentation (ENG-1882).

## Verification

- `eslint --max-warnings 0` on changed files, `prettier`, `tsc --noEmit --skipLibCheck`: clean (roam + content-model).
- vitest: roam 67/67 (12 files), content-model 9/9.
- **Runtime demo against real Roam + Supabase owed before merge** — the native `fromMarkdown` behavior (markdown fidelity, caller-supplied uid handling) can only be proven on a live graph.


apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,133 @@
...
+}): Promise<{ markdown: string } | { error: string }> => {
+  const { data, error } = await client
+    .from("my_contents")
+    .select("text")
+    .eq("space_id", sharedNode.spaceId)
+    .eq("source_local_id", sharedNode.sourceLocalId)
+    .eq("variant", "full")
+    .maybeSingle();
+  if (error) return { error: error.message };
+  return { markdown: data?.text ?? "" };
```
sid597: Fetch happens at import-action time, deliberately not during discovery (per the listing-vs-import rework on #1214). Same lookup shape as Obsidian's `fetchNodeContent` in `importNodes.ts`, including the one-row-per-variant assumption its `maybeSingle` makes. A missing row materializes a title-only page rather than failing — `full` is optional per ENG-2016. `text` itself is NOT NULL in `content.sql`; the nullable type (and the `??`) comes from `my_contents` view typegen, so the live path through `??` is row absence.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,133 @@
...
+
+const createImportedPage = async ({
+  sharedNode,
+  tree,
+}: {
+  sharedNode: SharedNode;
+  tree: InputTextNode[];
+}): Promise<MaterializeSharedNodeResult> => {
+  const collidingPageUid = getPageUidByPageTitle(sharedNode.title);
+  if (collidingPageUid) {
```
sid597: MVP0 title-collision behavior: fail with an actionable reason and write nothing (same on the rename path in `updateImportedPage`). Final collision UX is out of scope per the ticket. Related: per ENG-1856's boundary note, node-type mapping is ENG-1858's decision — MVP0 keeps the source title verbatim, so a title matching a local node format is picked up by Roam's existing format recognition. Verbatim vs reformat vs auto-create is an open question in the PR description.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,133 @@
...
+      page: { uid: pageUid, title: sharedNode.title },
+    });
+  }
+  for (const { uid } of getShallowTreeByParentUid(pageUid)) {
+    await deleteBlock(uid);
+  }
+  for (const [order, node] of tree.entries()) {
+    await createBlock({ parentUid: pageUid, order, node });
+  }
+  writeImportedSourceIdentity({
```
sid597: Identity is written only after content writes succeed — on this update path nothing touches the existing `importedFrom` props until then, so a mid-write failure can't drop source identity (ticket Done-When #3). Same ordering on the create path above.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,133 @@
...
+    pageUid,
+    sourceModifiedAt: sharedNode.lastModified,
+    sourceNodeRid: sharedNode.rid,
+  });
+  return { status: "updated", pageUid };
+};
+
+export const materializeSharedNode = async ({
```
sid597, line 273: No production caller yet by design: ENG-1859 (import action) is blocked by this ticket and consumes this. That's also why failures are returned as `reason` strings instead of toast/PostHog here — reporting belongs to the action surface, matching how `DiscoverSharedNodesDialog` handles it with `internalError`.

apps/roam/src/utils/markdownToRoamBlocks.ts
```diff
@@ -0,0 +1,111 @@
...
+  const closingIndex = lines.findIndex(
+    (line, index) => index > 0 && FRONTMATTER_DELIMITER.test(line),
+  );
+  return closingIndex === -1 ? lines : lines.slice(closingIndex + 1);
+};
+
+const indentWidth = (whitespace: string): number =>
+  [...whitespace].reduce((width, char) => width + (char === "\t" ? 4 : 1), 0);
+
+export const markdownToRoamBlocks = (markdown: string): InputTextNode[] => {
```
sid597: Searched for an existing markdown→blocks path before writing this: `roamjs-components/marked` is the opposite direction (Roam text → HTML), `importDiscourseGraph` consumes already-structured JSON, and the export utils all go Roam→markdown. `marked`'s lexer would be a new direct dependency and would still need this tree-building on top of its flat tokens (nesting content under headings is a Roam-import convention, not a token property). Known fidelity limits (tables, setext headings, indented fences, ordered-list numbering, continuation lines) are ENG-1882's documentation scope.

sid597 (commented):
Inline notes to orient the review — one per non-obvious decision.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,326 @@
...
+  };
+  page: {
+    fromMarkdown: (args: {
+      page: { title: string; uid: string };
+      "markdown-string": string;
+    }) => Promise<void>;
+  };
+};
+
+const getRoamMarkdownApi = (): RoamMarkdownApi =>
```
sid597, line 67: `data.page.fromMarkdown` / `data.block.fromMarkdown` exist on the runtime `roamAlphaAPI` but not in roamjs-components' types, hence this local declaration. Preference order applied here: native Roam API > roamjs-components > hand-rolling — Roam owns markdown→blocks conversion, so fidelity follows Roam's importer rather than code we maintain (remaining fidelity questions are ENG-1882's scope).

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,326 @@
...
+  ...identity,
+  success: false,
+  ...(pageUid ? { pageUid } : {}),
+  error: {
+    message: error ? `${message}: ${getErrorMessage(error)}` : message,
+    stage,
+  },
+});
+
+const validateSharedNode = (
```
sid597, line 95: Boundary validation before any I/O: platform from the `Platform` enum (not a RID prefix test, which fails for https-shaped space URIs), RID shape via `isRid`, and the modified time canonicalized with `toISOString()` so what lands in block props is always UTC-canonical.

sid597 (commented):
Refreshed inline notes for the native-API design (the earlier review's anchors are outdated).

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,326 @@
...
+      error: `Source modified time "${sharedNode.lastModified}" is invalid`,
+    };
+
+  const title = sharedNode.title.trim();
+  if (!title) return { error: "Source node title is required" };
+
+  return { sourceModifiedAt: modifiedAt.toISOString(), title };
+};
+
+const fetchFullMarkdown = async ({
```
sid597, line 118: Content fetched at import-action time, deliberately not during discovery (per the listing-vs-import rework on #1214). A missing `full` row materializes a title-only page — `full` is optional per ENG-2016. Only `text/obsidian+markdown` is accepted: Roam-produced `text/markdown` embeds the title as an H1 (`buildFullMarkdown`), which would duplicate the title into the body. `text` is NOT NULL in `content.sql`; the nullable type comes from the `my_contents` view typegen.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,326 @@
...
+const createImportedPage = async ({
+  identity,
+  markdown,
+  title,
+}: {
+  identity: SourceIdentity;
+  markdown: string;
+  title: string;
+}): Promise<MaterializeSharedNodeResult> => {
+  if (getPageUidByPageTitle(title))
```
sid597, line 152: MVP0 collision behavior: a page we did not import holds this title, so creating over it is the only destructive outcome available — refuse with an actionable reason and write nothing. Final collision UX is out of scope per the ticket. Node-type mapping is deliberately deferred (title kept verbatim; open question in the PR description).

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,326 @@
...
+      stage: "create-page",
+    });
+  }
+
+  try {
+    await writeImportedSourceIdentity({ pageUid, ...identity });
+  } catch (error) {
+    let cleanupError: unknown;
+    try {
+      await window.roamAlphaAPI.data.page.delete({ page: { uid: pageUid } });
```
sid597, line 185: If the identity write fails after page creation, the page is removed again so no untracked partial import is left behind — re-importing stays clean. If the cleanup itself fails, the result carries the orphaned `pageUid` so the caller can surface it.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,326 @@
...
+      identity,
+      message: `Cannot rename the imported page "${localTitle}" to "${title}": another page already has that title. Rename or remove that page, then import again`,
+      pageUid,
+      stage: "title-collision",
+    });
+
+  try {
+    const previousChildren = getShallowTreeByParentUid(pageUid);
+    if (markdown) {
+      await getRoamMarkdownApi().block.fromMarkdown({
```
sid597, line 228: Replacement blocks are appended before the previous children are deleted, so the page is never empty mid-flight; the rename was already collision-checked before any mutation. Identity is refreshed last — a failure at any stage returns that stage plus the identity, and never erases the stored `importedFrom` props.

apps/roam/src/utils/setBlockProps.ts
```diff
@@ -33,10 +33,11 @@ const setBlockProps = (
         ? (deNormalizeProps(newProps) as Record<string, json>)
         : newProps),
     } as Record<string, json>;
-    window.roamAlphaAPI.data.block.update({ block: { uid, props } });
-    return props;
+    return window.roamAlphaAPI.data.block
```
sid597, line 36: This was a floating promise — callers could not observe whether the props write landed, which the materializer needs before reporting success. Returning the update promise makes it awaitable; the 9 existing call sites are prefixed with `void` to preserve their fire-and-forget behavior exactly. Whether any of them should await is pre-existing and unchanged by this PR.

packages/content-model/src/text/index.ts
```diff
@@ -1,2 +1,20 @@
 export const normalizeLineEndings = (text: string): string =>
   text.replace(/\r\n?/g, "\n");
+
+const FRONTMATTER_DELIMITER = "---";
+
+export const stripFrontmatter = (markdown: string): string => {
```
sid597, line 9: Obsidian's `full` variant is the raw note file including YAML frontmatter (`upsertNodesAsContentWithEmbeddings.ts` stores `vault.read(file)`); Roam has no frontmatter concept, so it must be stripped before `fromMarkdown`. Placed here next to `normalizeLineEndings` as pure text logic both apps care about. Open question: should the Obsidian producer strip at publish time instead? This is needed either way for rows already published.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,133 @@
...
+}): Promise<MaterializeSharedNodeResult> => {
+  const collidingPageUid = getPageUidByPageTitle(sharedNode.title);
+  if (collidingPageUid) {
+    return {
+      status: "failed",
+      reason: `A page titled "${sharedNode.title}" already exists in this graph but was not imported from "${sharedNode.spaceName}". Rename or remove that page, then import again.`,
+    };
+  }
+  const pageUid = await createPage({ title: sharedNode.title, tree });
+  writeImportedSourceIdentity({
```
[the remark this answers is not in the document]
sid597: Fixed as of 17cc8f97 (this thread anchors to the first commit): `setBlockProps` now returns the `block.update` promise, `writeImportedSourceIdentity` is async and awaited, and an identity-write failure on create deletes the new page again (staged `write-source-identity` failure, orphan uid reported if cleanup also fails) — so the stuck-on-title-collision scenario can't occur.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,133 @@
...
+  client: DGSupabaseClient;
+  sharedNode: SharedNode;
+}): Promise<{ markdown: string } | { error: string }> => {
+  const { data, error } = await client
+    .from("my_contents")
+    .select("text")
+    .eq("space_id", sharedNode.spaceId)
+    .eq("source_local_id", sharedNode.sourceLocalId)
+    .eq("variant", "full")
+    .maybeSingle();
```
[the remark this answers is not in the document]
sid597: Good catch — confirmed against `content.sql`: non-original rows carry `original = NULL`, so only `(space_id, source_local_id, variant, original)` is unique and derived `full` rows would break `maybeSingle()`. Fixed in 9e553499 with `.eq("original", true)`; the post-fetch content-type validation stays so a non-Obsidian original still fails with an actionable message. Note: Obsidian's `fetchNodeContent`/`fetchNodeContentForImport` carry the same latent issue — follow-up material, out of scope here.

sid597:
@codex review

packages/content-model/src/text/index.ts
sid597: Resolving the open question above with maparent's ruling from #1224 ([comment](https://github.com/DiscourseGraphs/discourse-graph/pull/1224#issuecomment-5104725814)): the dual-home design stays — frontmatter lives in `literal_content.source_data` (jsonb, indexed, key order lost) **and** in the `full` body (lossless), so the producer will not strip it and consumer-side stripping is the accepted MVP0 shape (what this helper does). He flags Roam-side stripping itself as acknowledged debt: "Roam should not have to care about Obsidian specifics… it should not have to strip the frontmatter either" — to be revisited with the atJSON transfer, where strippable regions get marked platform-agnostically (ENG-1882's v1 handoff).

Related future scope from the same comment: translating frontmatter into the Roam properties block is a separate piece of work and should read from `source_data`, not from re-parsing the full text. Deliberately **not ticketed yet** — under discussion first.

apps/roam/src/utils/materializeSharedNode.ts
sid597: Context from maparent on #1224 ([comment](https://github.com/DiscourseGraphs/discourse-graph/pull/1224#issuecomment-5104243454)): he proposes augmenting `ridToSpaceUriAndLocalId` to also return the `Platform` (the info is present in the RID forms we design). No change here — for discovery-driven imports the platform arrives server-sourced on the `SharedNode` (`my_spaces.platform`), which stays the source of truth. Where the augmented helper pays off is ENG-1860's refresh flow, which starts from a stored `importedFrom.sourceNodeRid` with no Space row in hand (today that costs a Space-table lookup via `getSpaceNameIdFromRid`). Under discussion — adopt there if/when he lands it.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,327 @@
...
+    .eq("variant", "full")
+    .eq("original", true)
+    .maybeSingle();
+  if (error) return { error: error.message };
+  if (!data?.text) return { markdown: "" };
+  if (data.content_type !== contentTypes.obsidianMarkdown)
+    return {
+      error: `Unsupported full content type "${data.content_type}" — expected "${contentTypes.obsidianMarkdown}"`,
+    };
+  return { markdown: stripFrontmatter(data.text).trim() };
```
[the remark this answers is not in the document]
sid597: Fixed in f6dc8f8f: `trim()` replaced with `trimBlankLines` (in `@repo/content-model` next to the other text helpers), which strips only surrounding blank lines so a leading indented code block keeps its indentation; a whitespace-only body still collapses to the title-only path. Both behaviors are pinned by tests ("preserves the indentation of a leading indented code block", "creates a title-only page when the body is only whitespace").

sid597:
@codex review

packages/content-model/src/text/index.ts
```diff
@@ -1,2 +1,23 @@
 export const normalizeLineEndings = (text: string): string =>
```
maparent, line 1: Flagging: I would spontaneously have put this in packages/utils. Not contesting, but I realize I do not have a strong heuristic as to why we should have things in content-model vs utils (or even database.) May be worth discussing.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,332 @@
...
+}: {
+  identity: SourceIdentity;
+  markdown: string;
+  pageUid: string;
+  title: string;
+}): Promise<MaterializeSharedNodeResult> => {
+  const localTitle = getPageTitleByPageUid(pageUid);
+  const needsRename = localTitle !== title;
+  if (needsRename && getPageUidByPageTitle(title))
+    return failure({
```
maparent, line 218: Should we fail here? I think it would make sense to both update the content _and_ give an error message to the user. Note that either partial success (title or body) should be counted as a failure in terms of updating lastModified.
Negotiable: that's my instinct but I think we should discuss this.
(Even thinking we could insert a block with the error message, and linking to the conflicting page to help the user act on the issue. But that would certainly be another task.)

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,332 @@
...
+import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageUid";
+import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
+import getShallowTreeByParentUid from "roamjs-components/queries/getShallowTreeByParentUid";
+import deleteBlock from "roamjs-components/writes/deleteBlock";
+import {
+  findImportedNodeUidBySourceRid,
+  writeImportedSourceIdentity,
+} from "./importedSourceIdentity";
+
+type MaterializationStage =
```
maparent, line 18: Nit: Those are error types, but a few of them can happen at different moments (eg create vs update) so I'm not 100% on the word "Stage". How do you feel about MaterializationErrorType?

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,332 @@
+import {
+  contentTypes,
+  stripFrontmatter,
+  trimBlankLines,
+} from "@repo/content-model";
+import type { DGSupabaseClient } from "@repo/database/lib/client";
+import { isRid } from "@repo/database/lib/rid";
+import type { SharedNode } from "@repo/database/lib/sharedNodes";
```
maparent, line 8: Made a comment in discord PM: I think we should use CrossAppNode. But that will almost certainly come later.

maparent (approved):
Ok, many comments, and nothing I want to block on as such, so approving now; but there's a few things I'd like us to discuss tomorrow before it's merged, mainly whether we should fail if we cannot rename.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -0,0 +1,332 @@
...
+    .from("my_contents")
+    .select("text, content_type")
+    .eq("space_id", sharedNode.spaceId)
+    .eq("source_local_id", sharedNode.sourceLocalId)
+    .eq("variant", "full")
+    .eq("original", true)
+    .maybeSingle();
+  if (error) return { error: error.message };
+  if (!data?.text) return { markdown: "" };
+  if (data.content_type !== contentTypes.obsidianMarkdown)
```
maparent, line 135: I let that pass and approved originally because of the PR title and linear ticket, which was scoped to Obsidian. But upon looking at ENG-1859, I see that there is not a separate path for doing the same materialization for imported Roam nodes, and I don't see another task for importing roam nodes. So either this task's scope should expand, or another prerequisite should be added to ENG-1959 (whose scope encompasses node import in general, as I read it.)

maparent (changes requested):
This only allows import of Obsidian nodes, and is deemed sufficient for 1859, i.e. general node import.
Now proposing either this or eng-1859 should be blocked until we can import roam nodes.

maparent (approved):

PR #1269, opened 2026-07-30
[ENG-1859] Add Roam import action for selected shared nodes

sid597:
https://github.com/user-attachments/assets/76d5f826-7e9c-45bd-a570-cf8f409aef13

Implements [ENG-1859](https://linear.app/discourse-graphs/issue/ENG-1859/add-roam-import-action-for-selected-shared-nodes): wires the shared-node discovery dialog (ENG-1855) to the materializer (ENG-1858) so a Roam user can select visible Obsidian-origin nodes and import them, with imported / skipped / failed counts and per-node failure messages.

**Stacked on** `eng-1858-materialize-obsidian-origin-markdown-into-roam-v3` — only the `[ENG-1859]` commits are new here.

## What's in the diff

- **`DiscoverSharedNodesDialog.tsx`** — checkbox column (enabled only for Obsidian-origin rows, since the materializer only supports those), select-all header checkbox over visible importable rows, primary `Import selected (N)` button that reads `Importing X of N…` while running, and a results Callout with the three counts plus per-node failure messages. Failures are also reported to PostHog via `internalError`. Inherits the existing `isSyncEnabled()` gate on the dialog command.
- **`importSharedNodes.ts`** (new) — sequential loop over the selected `SharedNode`s; each item is try/caught so partial failures never abort the batch (done-when #3); returns per-node `imported | skipped | failed(+message)` items. Mirrors Obsidian's `importSelectedNodes` shape (`onProgress(current, total)`, sequential, counts).
- **`materializeSharedNode.ts`** — gains the `"skipped"` action this ticket requires: when the existing imported page's stored `sourceModifiedAt` is at least as recent as the source's, it returns without fetching content or touching the page.
- **`discoverSharedNodes.ts`** — `DiscoveredSharedNode` gains one flat key, `sourceSpaceDbId` (the numeric Space id), plus a `toSharedNode` mapper that rebuilds the `SharedNode` the materializer takes; the listing shape from ENG-1855/ENG-2019 is otherwise unchanged.

## Decisions a reviewer would ask about

- **Why `sourceSpaceDbId` and `toSharedNode`?** The materializer's content query filters by the numeric `space_id` + `source_local_id` — and the RID can't replace them without an extra Space lookup per node (the URI→id resolution that `getSpaceNameIdFromRid` does for RID-only flows like refresh). Discovery already holds the numeric id, so it's passed through as one flat key, consistent with the existing flattened view model. `toSharedNode` rebuilds the `SharedNode` input at import time; `created`/`directMetadata` are null-filled there because the materializer never reads them (verified against every use) — if a future flow starts reading them, they need to be carried through instead.
- **Why does skip live in the materializer, not the dialog/util?** The materializer already owns the find-imported-node lookup; deciding skip there avoids a second datalog lookup per node and keeps all source-identity logic in one place. The refresh flow (ENG-1860) gets the same semantics for free. `"skipped"` slots into the established staged-result shape (`{ success: true, action }`).
- **Skip predicate**: stored `sourceModifiedAt >=` source `lastModified`, compared via `Date.parse` — both values are canonical UTC written/normalized by this same module. Missing or unparseable stored identity falls through to update (never skips on bad data).
- **Stage reorder** (find-imported-node now before fetch-content): skipped nodes shouldn't fetch the heavy `full` payload. No caller observes the old order — ENG-1858 is unreleased on this same stack, and all its tests still pass unchanged.
- **After an import, failed nodes stay selected** (successful ones are cleared) so retrying after fixing e.g. a title collision is one click.
- **The dialog can't be closed mid-import** (escape / outside click / X / Close all disabled) because the underlying Roam writes can't be cancelled; the lock lifts when the run finishes.
- **`getErrorMessage` is now exported** from `materializeSharedNode.ts` and reused by `importSharedNodes.ts` instead of adding a third private copy of the same helper.

## Known deferral

- The loop fetches each node's `full` content in a per-node query (inside the materializer). Batching the content fetch for a multi-select import would require the materializer to accept prefetched content; payloads are heavy and the skip path avoids the fetch entirely for up-to-date nodes, so this is deferred rather than reshaped here.

## Out of scope (per ticket)

Refresh after import (ENG-1860), relations and assets, advanced title-collision UX (collisions fail with an actionable message naming the page to rename/remove).

## Testing

- `eslint --max-warnings 0`, `prettier --check`, `tsc --noEmit --skipLibCheck` all pass.
- 74 unit tests pass in `apps/roam/src/utils`, including new coverage: skip-when-unchanged (no content fetch, no writes), update-when-changed, update-when-stored-identity-invalid, per-node outcomes + progress reporting, and continue-after-throw.
- Runtime demo of the real flow (select → import → counts/failures) to follow before review.


sid597 (commented):
Guidepost comments on the non-obvious decisions (details in the PR description).

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -317,6 +326,31 @@ export const materializeSharedNode = async ({
     });
   }
 
+  if (
+    importedPageUid &&
+    storedIdentity &&
+    isImportUpToDate({
```
sid597, line 332: The skip decision lives here (not in the import loop) so the find-imported-node lookup happens once and all source-identity logic stays in this module — the ENG-1860 refresh flow inherits the same semantics. Predicate: stored `sourceModifiedAt >=` source `lastModified`, both canonical UTC written by this module; missing/unparseable stored identity falls through to update, never to skip.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -295,19 +310,13 @@ export const materializeSharedNode = async ({
...
-      identity,
-      message: `Could not fetch the content of "${sharedNode.title}" from "${sharedNode.spaceName}": ${content.error}`,
-      stage: "fetch-content",
-    });
-
   let importedPageUid: string | null;
+  let storedIdentity: ImportedSourceIdentity | undefined;
   try {
     importedPageUid = await findImportedNodeUidBySourceRid(sharedNode.rid);
+    storedIdentity = importedPageUid
```
sid597, line 317: find-imported-node moved ahead of fetch-content so an up-to-date node skips without fetching the heavy `full` payload. No caller observes the old order (ENG-1858 is unreleased on this stack) and all existing stage tests pass unchanged.

apps/roam/src/utils/importSharedNodes.ts
```diff
@@ -0,0 +1,53 @@
...
+      const result = await materializeSharedNode({ client, sharedNode });
+      items.push(
+        result.success
+          ? {
+              sharedNode,
+              status: result.action === "skipped" ? "skipped" : "imported",
+            }
+          : { sharedNode, status: "failed", message: result.error.message },
+      );
+    } catch (error) {
```
sid597, line 43: materializeSharedNode is designed to return failures rather than throw, but a few sync roamjs queries sit outside its try blocks — this catch is what guarantees the ticket's "partial failures do not abort the whole selected import" even if one materialization throws.

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -123,21 +204,112 @@ const DiscoverSharedNodesDialog = ({ onClose }: { onClose: () => void }) => {
...
+      );
+      setNodes((previous) =>
+        previous.map((node) =>
+          importedRids.has(node.sharedNode.rid)
+            ? { ...node, alreadyImported: true }
+            : node,
+        ),
+      );
+      const failedImports = results.filter(isFailedSharedNodeImport);
+      setSelectedRids(
```
sid597, line 264: Selection is not fully cleared after a run: failed nodes stay selected so a retry after fixing the reported problem (e.g. renaming a colliding page) is one click; successful ones are deselected.

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -123,21 +204,112 @@ const DiscoverSharedNodesDialog = ({ onClose }: { onClose: () => void }) => {
...
+      setImportProgress(null);
+    }
+  };
+
   return (
     <Dialog
       autoFocus={false}
-      canEscapeKeyClose
-      canOutsideClickClose
+      canEscapeKeyClose={!importing}
```
sid597, line 299: All close paths (escape, outside click, X, Close button) are disabled while an import runs because the underlying Roam writes can't be cancelled — closing would just hide the progress and the failure report.

sid597:
@codex review

apps/roam/src/utils/discoverSharedNodes.ts
```diff
@@ -3,48 +3,21 @@ import {
...
-  }));
-
 export const discoverSharedNodes = async ({
   client,
   currentSpaceId,
 }: {
   client: DGSupabaseClient;
   currentSpaceId: number;
-}): Promise<DiscoveredSharedNode[]> => {
+}): Promise<{
```
sid597, line 14: Discovery now returns its two sources separately instead of pre-joining them into the `DiscoveredSharedNode` shape this file had since #1214. That shape was a field-renamed copy of `SharedNode` plus a baked-in `alreadyImported` flag — and wiring up import meant mapping it back into a `SharedNode` while fabricating the fields the renamed copy didn't carry (`created`, `directMetadata`). The dialog is the shape's only consumer, so the parallel type is gone: rows stay `SharedNode` end to end (list → selection → `importSharedNodes`), and "already imported" is derived from `importedSourceRids` at render.

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -77,26 +108,72 @@ const SharedNodeRow = ({ node }: { node: DiscoveredSharedNode }) => (
...
+        </ul>
+      )}
+    </Callout>
+  );
+};
+
 const DiscoverSharedNodesDialog = ({ onClose }: { onClose: () => void }) => {
-  const [nodes, setNodes] = useState<DiscoveredSharedNode[]>([]);
+  const [nodes, setNodes] = useState<SharedNode[]>([]);
+  const [importedRids, setImportedRids] = useState<Set<string>>(new Set());
```
sid597, line 147: Single source of truth for imported-ness. Rows don't carry an `alreadyImported` flag anymore — the badge derives from `importedRids.has(node.rid)` at render, and a successful import just adds rids to this set (previously this meant walking the nodes array and patching a stored flag on each matching row).

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -123,21 +200,106 @@ const DiscoverSharedNodesDialog = ({ onClose }: { onClose: () => void }) => {
...
-        node.sourceSpaceName,
-        node.sourceSpaceId,
+        node.platform,
+        node.spaceName,
+        node.spaceUri,
         node.title,
-        node.sourceNodeId,
-      ].some((value) => value?.toLocaleLowerCase().includes(normalizedSearch)),
+        node.sourceLocalId,
+      ].some((value) => value.toLocaleLowerCase().includes(normalizedSearch)),
```
sid597, line 208: No `?.` here on purpose: all five searched fields are non-optional `string`s on `SharedNode`.

apps/roam/src/utils/__tests__/discoverSharedNodes.test.ts
```diff
@@ -1,48 +0,0 @@
-import { describe, expect, it } from "vitest";
-import { toDiscoveredSharedNodes } from "~/utils/discoverSharedNodes";
```
sid597, line 2: Deleted with its subject: this file only tested the `SharedNode` → `DiscoveredSharedNode` field renaming, which no longer exists. The replacement join is `importedRids.has(node.rid)` at render — nothing left to unit-test.

sid597 (commented):
Inline notes on the shape change in the latest commit: the `DiscoveredSharedNode` type from #1214 is removed in favor of using `SharedNode` directly. Each note explains what replaced the removed piece.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -41,7 +43,7 @@ type MaterializationFailure = SourceIdentity & {
 
 type MaterializationSuccess = SourceIdentity & {
   success: true;
-  action: "created" | "updated";
+  action: "created" | "updated" | "skipped";
```
sid597, line 46: `skipped` extends the staged-result shape ENG-1858 established (identity echo + success/action or failure/stage). Consumers report per-item outcomes from `action`/`stage`; `importSharedNodes` collapses `created`/`updated` to a single "imported" status because the dialog doesn't distinguish them.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -67,9 +69,22 @@ type RoamMarkdownApi = {
 const getRoamMarkdownApi = (): RoamMarkdownApi =>
   window.roamAlphaAPI.data as unknown as RoamMarkdownApi;
 
-const getErrorMessage = (error: unknown): string =>
+export const getErrorMessage = (error: unknown): string =>
   error instanceof Error ? error.message : String(error);
 
+const isImportUpToDate = ({
```
sid597, line 75: Timestamp provenance for the `Date.parse` here: both sides are canonical ISO-with-Z strings — `sourceModifiedAt` comes from `SharedNode.lastModified`, which `packages/database` normalizes (`normalizeUtcTimestamp` → `toISOString()`), and `storedModifiedAt` is that same value echoed back from the block props we wrote at import time. So this is not naive DB-timestamp parsing; the Z-less-PostgREST hazard is handled a layer down. The `NaN` guard covers hand-edited or legacy props (block props are user-editable), in which case we fall through to a normal re-import.

apps/roam/src/utils/importSharedNodes.ts
```diff
@@ -0,0 +1,53 @@
...
+  client,
+  sharedNodes,
+  onProgress,
+}: {
+  client: DGSupabaseClient;
+  sharedNodes: SharedNode[];
+  onProgress: (current: number, total: number) => void;
+}): Promise<SharedNodeImportItem[]> => {
+  const items: SharedNodeImportItem[] = [];
+  for (const sharedNode of sharedNodes) {
```
sid597, line 32: Sequential on purpose, not an overlooked batch: each iteration is a full materialization (skip check → content fetch → page write) with per-item staged failure reporting and the `onProgress` callback driving the dialog's progress label. Batching the content fetches would front-load every node's full markdown (the payload discovery deliberately doesn't fetch) and lose per-item progress. Selections are user-sized (a handful), not data-scale.

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -13,34 +14,64 @@ import {
...
 import { getLoggedInClient, getSupabaseContext } from "~/utils/supabaseContext";
 
+const IMPORT_ERROR_TYPE = "Shared node import failed";
+const IMPORT_ERROR_OPERATION = "import-shared-nodes";
+
 const formatModifiedAt = (modifiedAt: string): string =>
   new Date(modifiedAt).toLocaleString();
 
-const SharedNodeRow = ({ node }: { node: DiscoveredSharedNode }) => (
+const isImportableSharedNode = (node: SharedNode): boolean =>
```
sid597, line 33: Obsidian-only by MVP scope: Obsidian is currently the only producer of importable `full` content, so its rows are the only ones with something to materialize. Roam-origin rows still show in the list (visibility of what's shared) but aren't selectable. When another producer lands, this predicate is the one place to widen.

sid597 (commented):
Second set of inline notes from the pre-PR pass: decisions a reviewer would otherwise have to ask about (staged-result shape, timestamp provenance, sequential loop, Obsidian-only gate).

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -67,9 +69,22 @@ type RoamMarkdownApi = {
 const getRoamMarkdownApi = (): RoamMarkdownApi =>
   window.roamAlphaAPI.data as unknown as RoamMarkdownApi;
 
-const getErrorMessage = (error: unknown): string =>
+export const getErrorMessage = (error: unknown): string =>
```
maparent, line 72: nit: utils?

apps/roam/src/utils/materializeSharedNode.ts
maparent: Hmm... I'd personally want this to be filtered upstream in the dialog, so the user does not even see the up-to-date nodes. It's what we're doing on Obsidian.

apps/roam/src/utils/importSharedNodes.ts
```diff
@@ -0,0 +1,53 @@
...
+    try {
+      const result = await materializeSharedNode({ client, sharedNode });
```
maparent, line 34: I think we need to either extend 1858 to handle roam, or make another prerequisite task to this one, so it handles both roam and obsidian. Of course we could do Roam later, but that's not how I understand this task.

maparent (changes requested):
I'd like for up-to-date nodes not to show up, following Obsidian pattern.
I think this PR should handle both roam and obsidian import.
We agreed that title rewriting according to target node type format is a different scope.
All to be discussed.

maparent (approved):

PR #1276, opened 2026-08-06
[ENG-2102] Implement status bar help menu for Obsidian

jsmorabito:
https://www.loom.com/share/372a9118730046c78827d36d077d5106

## Summary
- Adds a Discourse Graph icon to the Obsidian status bar (off by default) that opens a help menu with **Send feedback**, **Docs**, **Community**, and **Settings** items.
- Toggle to show/hide the icon lives in General settings.
- Extracted a shared `ToggleSetting` component in `GeneralSettings.tsx`, used by this toggle and the pre-existing "Show IDs in frontmatter" toggle.

## Scope check
- [x] Reasoned through scope against ENG-2102 and the final diff (no `$scope-check` tool available in this session, so this was done manually).
- Scope beyond `Done When`: None — the ticket has no description/Done When criteria beyond its title, and the diff is confined to `apps/obsidian/src` implementing/refining the status bar help menu.

## Test plan
- [x] Status bar icon hidden by default; toggling "Show help menu icon in status bar" in General settings shows/hides it live.
- [x] Clicking the icon opens a menu with Send feedback, Docs, Community, Settings.
- [x] Send feedback opens the existing feedback modal.
- [x] Docs/Community open the correct external URLs.
- [x] Settings opens the plugin's settings tab.
- [x] Icon sizing/color match other status-bar icons; hover state and tooltip placement match core Obsidian icons (e.g. Live Preview).
- [x] `check-types`, `eslint`, and `build` all pass.

Before/after screenshots aren't included here — I don't have file access to the ones shared in chat during development. @johnny1093, feel free to add them.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1276" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added a Help menu with links to feedback, documentation, community resources, and plugin settings.
  * Added an optional status-bar Help icon that opens the Help menu.
  * Added settings to show or hide the Help icon and frontmatter ID controls.
  * Help icon visibility now updates immediately when the setting changes.


jsmorabito:
@coderabbitai full review

jsmorabito:
Addressed Devin's 3 findings:
- innerHTML for the status bar logo → now registered via `addIcon()` + rendered with `setIcon()` (matches AGENTS.md's no-innerHTML rule for store review; kept the same viewBox but dropped the fixed width/height, which is what actually caused the earlier tiny-icon bug — not the nesting itself).
- Click listener not registered for auto-cleanup → switched to `this.registerDomEvent()`.

Confirmed working in app

apps/obsidian/src/components/GeneralSettings.tsx
```diff
@@ -4,10 +4,39 @@ import { setIcon } from "obsidian";
...
+      <div
+        className={`checkbox-container ${checked ? "is-enabled" : ""}`}
+        onClick={() => onChange(!checked)}
+      >
+        <input
+          type="checkbox"
+          checked={checked}
+          aria-label={name}
+          onChange={(e) => onChange(e.target.checked)}
+        />
```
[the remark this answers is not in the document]
jsmorabito: > Below is from Claude

> Good catch on the double-invocation — that's a real, if low-cost, side effect. CodeRabbit made this same suggestion originally, and I applied it in 53e6b85c: https://github.com/DiscourseGraphs/discourse-graph/commit/53e6b85c04be79f0a45d52efaf3c073a9736b334

> That broke both toggles in testing (worked once, then got stuck). Once the switch animates, the knob likely ends up covering the native input's hit area, so clicks landing anywhere else in the switch stop reaching the input and its onChange never fires — the div's onClick is what catches those (via bubbling, regardless of what's visually on top). Restored it in f80ab68d: https://github.com/DiscourseGraphs/discourse-graph/commit/f80ab68da7598618571efa3733a16660bc3709ae

> So keeping both is intentional: the double-fire when a click happens to land squarely on the input is harmless (same resulting boolean, one redundant no-op save), whereas relying on onChange alone breaks the toggle after the first click. Leaving this as-is.

jsmorabito:
@trangdoan982 I think everything is addressed now and this is ready for your review

apps/obsidian/src/icons.ts
```diff
@@ -1,5 +1,16 @@
...
+
+export const DISCOURSE_GRAPH_LOGO_ICON_ID = "discourse-graph-logo";
+
+// Same artwork as WHITE_LOGO_SVG, without fixed width/height: addIcon()'s
```
trangdoan982: if this is the case can you just remove the fixed heights from WHITE_LOGO_SVG and reuse it instead? we try not to have repeated code

apps/obsidian/src/types.ts
```diff
@@ -67,6 +67,7 @@ export type Settings = {
   canvasFolderPath: string;
   canvasAttachmentsFolderPath: string;
   nodeTagHotkey: string;
+  showHelpMenuStatusBarIcon: boolean;
```
trangdoan982, line 70: i'd also test to see if anything might break if i upgrade from a plugin version without this field -> having this required field. You can try by removing this field from `data.json` -> reload the plugin with this dev build -> see what happens

apps/obsidian/src/components/GeneralSettings.tsx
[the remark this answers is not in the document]
trangdoan982: can you clarify "that broke toggles" what "that" means exactly? you mean removing the `onClick`?

trangdoan982 (changes requested):
good work JM. just a few small comments. also for PR reviews that have been addressed, please resolve them @jsmorabito

apps/obsidian/src/icons.ts
jsmorabito: Yes, ill take a note about trying not to have repeated code, thanks for the tip

> From claude:

> done in 36f4e5dc: https://github.com/DiscourseGraphs/discourse-graph/commit/36f4e5dc22c26a4ce18b6e3adf24ee2b6cb1c4de

> Stripped the fixed width/height from WHITE_LOGO_SVG itself and reuse it for addIcon(). Checked its other consumer (the tldraw toolbar icon data URI in TldrawViewComponent.tsx) — it renders as a CSS background-image inside a fixed-size button, so it doesn't depend on those attributes either. Deleted the duplicate constant.

apps/obsidian/src/components/GeneralSettings.tsx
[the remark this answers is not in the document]
jsmorabito: I was able to toggle on or off once then the toggle appeared to be stuck (would not show as enabled or disabled after subsequent clicks)

> From Claude:
Yes, — "that" = removing the div's onClick (CodeRabbit's original suggestion, applied in 53e6b85c). Once I removed it and relied solely on the input's onChange, both toggles worked once then got stuck. Restored the div's onClick in f80ab68d and it's worked reliably since.

apps/obsidian/src/types.ts
jsmorabito: Followed the instructions and tested this out. All the functions worked as before. Claude said it tested it below, but it was actually me 😎 
> From Claude: Tested this: removed showHelpMenuStatusBarIcon from the dev vault's data.json to simulate a pre-upgrade user, then reloaded the plugin. No crash, no issue. loadSettings()'s Object.assign({}, DEFAULT_SETTINGS, loadedData) backfilled it to false (the safe default — icon hidden), and hasNewFields() correctly detected the missing key and persisted it back to data.json on the next save. Same migration path every other settings field added to this plugin already relies on.

jsmorabito:
> good work JM. just a few small comments. also for PR reviews that have been addressed, please resolve them @jsmorabito

Will do, noted for future work as well thank you for this advice. Is the broken toggles question [here](https://github.com/DiscourseGraphs/discourse-graph/pull/1276#discussion_r3730464914) the same as the [one above](https://github.com/DiscourseGraphs/discourse-graph/pull/1276#discussion_r3730464914) that I responded to?

trangdoan982 (approved):

PR #1278, opened 2026-08-06
ENG-2105 Add shared engineering writing style skill to the repo

mdroidian:
## Summary

- Add the shared `$discourse-engineering-writing-style` skill to the repository.
- Include interface metadata so the skill is discoverable by supported agents.

## Why

The skill was only available locally. Adding it under `skills/` makes the same concise engineering writing guidance available to the team and keeps it version-controlled.

## Validation

- `quick_validate.py skills/discourse-engineering-writing-style`
- `pnpm exec prettier --check skills/discourse-engineering-writing-style/SKILL.md skills/discourse-engineering-writing-style/agents/openai.yaml`
- Confirmed `policy.products` uses the schema-supported `CHAT` and `CODEX` values.

## Scope check

- [x] Ran `$scope-check` against ENG-2105 and the final diff.

### DIFF
--- a/skills/discourse-engineering-writing-style/SKILL.md
+++ b/skills/discourse-engineering-writing-style/SKILL.md
@@ -1,0 +1,32 @@
+---
+name: discourse-engineering-writing-style
+description: Draft and revise engineering emails, chat messages, requests, follow-ups, status updates, PR comments, issue reports, release notes, and other short-form team writing in the shared Discourse Graphs style. Use whenever a Discourse Graphs engineer asks to write, rewrite, phrase, shorten, clarify, or polish a message, unless they request a different style.
+---
+
+# Discourse Engineering Writing Style
+
+Write in a direct, concise, plainspoken style that preserves important context and qualification.
+
+## Core style
+
+- Lead with the relevant point and move quickly to the concrete question, request, or action.
+- Prefer the shortest version that remains clear and preserves necessary nuance.
+- Use natural contractions and everyday wording. Avoid corporate jargon, inflated phrasing, and generic professional filler.
+- Avoid em dashes as sentence-level punctuation. Use a period, comma, colon, semicolon, or parentheses to separate thoughts instead. Preserve em dashes only when they appear in exact quotations, titles, source text, or technical content supplied by the requester.
+- Keep warmth light: use a brief greeting, thanks, or friendly closing when appropriate, without extended pleasantries.
+- When identifying a problem or disagreement, state it plainly and pivot toward clarification, resolution, criteria, or next steps.
+- Use short paragraphs. Use bullets only when several distinct items genuinely scan better as a list.
+- Ask one clear question when one will do. Combine related questions only when separating them would add needless length.
+
+## Adapt to the channel
+
+- For a quick reply or chat message, omit the greeting and sign-off when they add no value.
+- For technical issues, describe the observed behavior, the relevant contrast or expectation, and the requested clarification or next step.
+
+## Drafting rules
+
+- Preserve facts, commitments, dates, names, links, and technical meaning supplied by the requester.
+- Do not invent availability, decisions, commitments, or emotional language.
+- Treat explicit directions such as "shorter," "warmer," "more forceful," "broader," or "one sentence" as overrides for the current draft.
+- When the requester asks for alternatives, vary the framing or emphasis rather than making superficial synonym swaps.
+- Return a ready-to-use draft first. Add explanation only when requested or when an ambiguity materially affects the wording.
--- a/skills/discourse-engineering-writing-style/SKILL.md
+++ b/skills/discourse-engineering-writing-style/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: discourse-engineering-writing-style
-description: Draft and revise engineering emails, chat messages, requests, follow-ups, status updates, PR comments, issue reports, release notes, and other short-form team writing in the shared Discourse Graphs style. Use whenever a Discourse Graphs engineer asks to write, rewrite, phrase, shorten, clarify, or polish a message, unless they request a different style.
+description: Draft and revise engineering emails, chat messages, requests, follow-ups, status updates, engineering tickets, PR comments, issue reports, release notes, technical documentation, and other team writing in the shared Discourse Graphs style. Use whenever a Discourse Graphs engineer asks to write, rewrite, phrase, shorten, clarify, or polish team communication or documentation, unless they request a different style.
 ---
 
 # Discourse Engineering Writing Style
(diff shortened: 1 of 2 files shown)


skills/discourse-engineering-writing-style/agents/openai.yaml
```diff
@@ -0,0 +1,11 @@
+interface:
+  display_name: "Discourse Engineering Writing Style"
+  short_description: "Draft concise, direct engineering team messages"
+  default_prompt: "Use $discourse-engineering-writing-style to draft a concise engineering team message."
+policy:
+  products:
```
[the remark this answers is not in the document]
mdroidian: Thanks, confirmed that a change is needed. The current OpenAI schema does support policy.products, but it only accepts CHAT, CODEX, or both. Our chatgpt, api, and atlas values are invalid, so we will replace them with CHAT and CODEX. Schema: https://developers.openai.com/plugins/deploy/submission-errors.md#skill-agent-metadata-errors

skills/discourse-engineering-writing-style/SKILL.md
```diff
@@ -0,0 +1,32 @@
...
+
+Write in a direct, concise, plainspoken style that preserves important context and qualification.
+
+## Core style
+
+- Lead with the relevant point and move quickly to the concrete question, request, or action.
+- Prefer the shortest version that remains clear and preserves necessary nuance.
+- Use natural contractions and everyday wording. Avoid corporate jargon, inflated phrasing, and generic professional filler.
+- Avoid em dashes as sentence-level punctuation. Use a period, comma, colon, semicolon, or parentheses to separate thoughts instead. Preserve em dashes only when they appear in exact quotations, titles, source text, or technical content supplied by the requester.
+- Keep warmth light: use a brief greeting, thanks, or friendly closing when appropriate, without extended pleasantries.
```
sid597: I think Adapt to the Channel section covers it

skills/discourse-engineering-writing-style/SKILL.md
```diff
@@ -0,0 +1,32 @@
...
+- When identifying a problem or disagreement, state it plainly and pivot toward clarification, resolution, criteria, or next steps.
+- Use short paragraphs. Use bullets only when several distinct items genuinely scan better as a list.
+- Ask one clear question when one will do. Combine related questions only when separating them would add needless length.
+
+## Adapt to the channel
+
+- For a quick reply or chat message, omit the greeting and sign-off when they add no value.
+- For technical issues, describe the observed behavior, the relevant contrast or expectation, and the requested clarification or next step.
+
+## Drafting rules
```
sid597, line 31: Who is the requester in this section? is it the dev working with agent or a reviewer? 

I think the skill should be on a team level and individuals can have their own drafting rules and different  weightage, otherwise this will silently be changing the workflow on an individual level.

sid597 (approved):

skills/discourse-engineering-writing-style/SKILL.md
mdroidian: Could you clarify what "it" is referring to here?  Are you suggesting we remove this line?

skills/discourse-engineering-writing-style/SKILL.md
mdroidian: The 'requester' means the developer asking the agent to draft or revise something.

The goal of this skill is to give our team-facing writing a consistent voice and style. It shouldn't define or change anyone's workflow.  While I'd be open for separate individual drafting layers, I'd first be curious to see what situation would use it.

Could you point to where you see it directing workflow, or give an example of an individual rule you'd want to layer on top?

skills/discourse-engineering-writing-style/SKILL.md
```diff
@@ -0,0 +1,32 @@
...
+# Discourse Engineering Writing Style
+
+Write in a direct, concise, plainspoken style that preserves important context and qualification.
+
+## Core style
+
+- Lead with the relevant point and move quickly to the concrete question, request, or action.
+- Prefer the shortest version that remains clear and preserves necessary nuance.
+- Use natural contractions and everyday wording. Avoid corporate jargon, inflated phrasing, and generic professional filler.
+- Avoid em dashes as sentence-level punctuation. Use a period, comma, colon, semicolon, or parentheses to separate thoughts instead. Preserve em dashes only when they appear in exact quotations, titles, source text, or technical content supplied by the requester.
```
trangdoan982, line 16: haha yes!

trangdoan982 (approved):
lgtm. I wholeheartedly agree with the Core styles but not sure how the Drafting rules will help with my use cases. but open to try it out first then twitch the prompt if needed

skills/discourse-engineering-writing-style/SKILL.md
maparent, line 16: I'm a bit surprised to see punctuation-related directives.
Real question: what do you two have against em-dashes (other than thinking it's a marker of AI...)?
But I think the issue is parsing long run-in thought, which I would handle with a higher-level directive:  What about "Separate thoughts clearly, preferably in hierarchical outline form. A sentence should only explain a new idea, and possibly connect it to an earlier idea."
(I see this conflicts with line 18, which also surprises me.)

skills/discourse-engineering-writing-style/SKILL.md
maparent: I agree that "for a quick reply or chat message" below is a narrower context than stated in this line. Not sure I'd even allow greetings or thanks.

skills/discourse-engineering-writing-style/SKILL.md
```diff
@@ -0,0 +1,32 @@
...
+- Keep warmth light: use a brief greeting, thanks, or friendly closing when appropriate, without extended pleasantries.
+- When identifying a problem or disagreement, state it plainly and pivot toward clarification, resolution, criteria, or next steps.
+- Use short paragraphs. Use bullets only when several distinct items genuinely scan better as a list.
```
maparent: I think as roam users, we should favour hierarchical bullets, no?

maparent (commented):
I would add a high-level directive about helping the author identify anything that is ambiguous, or can be misconstrued.

skills/discourse-engineering-writing-style/SKILL.md
mdroidian: This was co-opted from a general writing style that applied to many situations.  It could still apply, but likely to be rare.  I'll remove it.

skills/discourse-engineering-writing-style/SKILL.md
mdroidian: Most AI's overuse em-dashes

PR #1279, opened 2026-08-07
[ENG-1860] Add Roam manual refresh for imported nodes

sid597:
https://github.com/user-attachments/assets/0511dd9b-513a-4d77-b2e3-03885330a275

Adds a manual refresh path for imported cross-app nodes in Roam. Pages carrying stored `importedFrom` source identity get a `Refresh` button in the page-title additions (gated on the same node-sharing flag as the import entry point). Clicking it re-fetches the latest shared content by the stored `sourceNodeRid` — never by the local title — and overwrites the local page through the existing `materializeSharedNode` update path, which preserves identity props, renames on title drift, and never duplicates. This mirrors the Obsidian per-note refresh (`refreshImportedFile`).

Two pieces of infra: `getSharedNodeByRid` in `packages/database` fetches one shared node by rid with targeted queries (space by url, then the concept/direct/full rows), reusing `buildSharedNodes`; and `materializeSharedNode` gains an opt-in `force` that bypasses the up-to-date skip, since a manual refresh must restore source content even when stored timestamps say current (MVP0: imported local edits are overwritten by refresh). The outcome surfaces as one toast; materialization failures and unexpected errors also go to PostHog. If the stored rid resolves to a different page holding the same identity, refresh reports that instead of claiming success on the clicked page.

The button check runs outside the discourse-node gate in the H1 observer so imported pages whose titles don't match local node formats still get the button. Refreshing a source with no full content yields a title-only page, matching import's existing (tested) behavior. Refresh-all is ENG-1861, provenance indicators ENG-1875.

## Scope check

- [x] Ran `$scope-check` against ENG-1860 and the final diff.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1279" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/RefreshImportedNodeTitleButton.tsx
+++ b/apps/roam/src/components/RefreshImportedNodeTitleButton.tsx
@@ -1,0 +1,62 @@
+import { Button } from "@blueprintjs/core";
+import posthog from "posthog-js";
+import React, { useState } from "react";
+import renderToast from "roamjs-components/components/Toast";
+import { handleTitleAdditions } from "~/utils/handleTitleAdditions";
+import { refreshImportedNode } from "~/utils/refreshImportedNode";
+
+const REFRESH_TITLE_BUTTON_ATTRIBUTE =
+  "data-roamjs-refresh-imported-node-title-button";
+
+const RefreshImportedNodeTitleButton = ({
+  uid,
+}: {
+  uid: string;
+}): JSX.Element => {
+  const [refreshing, setRefreshing] = useState(false);
+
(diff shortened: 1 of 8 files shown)


maparent (approved):
LGTM.
We're accruing technical debt by not using CrossApp here, but I agree that refactoring will wait.
Flagging:
We probably need a new task so that refreshing the node will also refresh related relations at the same time.
I suspect refreshing schemas will be a distinct task.

maparent:
s/distinct task/distinct process/

PR #1280, opened 2026-08-07
[ENG-2022] Make Roam publishing sync-independent with complete node upserts

sid597:
https://github.com/user-attachments/assets/e3ff85e3-0a01-45b6-a95e-79339229b2ec

Makes Roam publishing self-sufficient instead of piggybacking on background sync: publish now upserts the complete node — concept, direct title, and full markdown — through `upsert_concepts`, rather than only granting access to nodes sync had already pushed. The synced-nodes pre-filter is gone; an RPC failure grants nothing, a per-row failure withholds that node's grant (and any relation touching it) and surfaces as `failedSyncedUids`, replacing `skippedUnsyncedUids`.

Node sharing gets its own feature flag (`Enable node sharing`), decoupled from the suggestive-mode/sync flag: the Publish tab, the page-title Publish button, and `DG: Share current node` move to it, while discovery/import (`DG: Discover shared nodes`) stays on the sync flag.

This is the local [ENG-2022](https://linear.app/discourse-graphs/issue/ENG-2022/make-roam-publishing-sync-independent-with-complete-node-upserts) branch rebased onto current main (clean, no conflicts). DiscourseGraphs/discourse-graph#1279 ([ENG-1860](https://linear.app/discourse-graphs/issue/ENG-1860/add-roam-manual-refresh-for-imported-nodes)) is stacked on top.

## [ENG-2121](https://linear.app/discourse-graphs/issue/ENG-2121) check

- [X] Ran `$scope-check` against [ENG-2022](https://linear.app/discourse-graphs/issue/ENG-2022/make-roam-publishing-sync-independent-with-complete-node-upserts) and the final diff.

---

Open in Devin Review

### DIFF
--- a/apps/roam/src/utils/publishNodesToGroups.ts
+++ b/apps/roam/src/utils/publishNodesToGroups.ts
@@ -356,6 +356,15 @@ export const publishNodesToGroups = async ({
       }
     }
   });
+  for (const node of nodesByUid.values()) {
+    if (
+      !nodeSchemaUids.has(node.nodeType) &&
+      upsertedNodeUids.has(node.localId)
+    ) {
+      upsertedNodeUids.delete(node.localId);
+      result.failedUpsertUids.push(node.localId);
+    }
+  }
   result.syncedNodeSchemaUids = [...syncedNodeSchemaUids];
   result.syncedRelationTripleSchemaUids = [...syncedRelationTripleSchemaUids];
   result.syncedRelationUids = [...syncedRelationUids];
--- a/apps/roam/src/utils/publishNodesToGroups.ts
+++ b/apps/roam/src/utils/publishNodesToGroups.ts
@@ -207,7 +207,7 @@ type PublishNodesResult = {
   syncedNodeSchemaUids: string[];
   syncedRelationTripleSchemaUids: string[];
   syncedRelationUids: string[];
-  failedSyncedUids: string[];
+  failedUpsertUids: string[];
   okGroupIds: string[];
   failedGroupIds: string[];
 };
@@ -240,7 +240,7 @@ export const publishNodesToGroups = async ({
     syncedNodeSchemaUids: [],
     syncedRelationUids: [],
     syncedRelationTripleSchemaUids: [],
-    failedSyncedUids: [],
+    failedUpsertUids: [],
     okGroupIds: [],
     failedGroupIds: [],
   };
@@ -339,7 +339,7 @@ export const publishNodesToGroups = async ({
   }
 
   response.data.forEach((v, i) => {
-    if (v === -1) {
+    if (v < 0) {
       const localId = upsertConcepts[i].source_local_id;
       if (localId) {
         if (syncedNodeSchemaUids.has(localId)) {
@@ -352,7 +352,7 @@ export const publishNodesToGroups = async ({
         } else if (syncedRelationUids.has(localId)) {
           syncedRelationUids.delete(localId);
         }
-        result.failedSyncedUids.push(localId);
+        result.failedUpsertUids.push(localId);
       }
     }
   });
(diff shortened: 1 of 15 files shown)


apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -286,7 +287,6 @@ export const publishNodesToGroups = async ({
 
   const neededUids = [
     ...nodeSchemaUids,
-    ...nodeUids,
```
maparent, line 289: What is the point of removing this? If I understand well, the consequence is always upsetting all concerned nodes. What if they had been upserted previously?
In my understanding, someone may re-request publishing a node, esp. to a new group.

apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -337,50 +329,64 @@ export const publishNodesToGroups = async ({
     missingNodeSchemas.map((s) => s.localId),
   );
 
-  if (upsertConcepts.length > 0) {
```
maparent, line 340: If we do not always upsert all nodes, the test stands.

apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -337,50 +329,64 @@ export const publishNodesToGroups = async ({
...
+          syncedNodeSchemaUids.delete(localId);
+          nodeSchemaUids.delete(localId);
+        } else if (upsertedNodeUids.has(localId)) {
+          upsertedNodeUids.delete(localId);
+        } else if (syncedRelationTripleSchemaUids.has(localId)) {
+          syncedRelationTripleSchemaUids.delete(localId);
+        } else if (syncedRelationUids.has(localId)) {
+          syncedRelationUids.delete(localId);
         }
+        result.failedUpsertUids.push(localId);
```
maparent, line 355: nice addition!

maparent (changes requested):
Looks good overall, but for one issue.

apps/roam/src/utils/publishNodesToGroups.ts
sid597: The query "which of these UIDs already exist in my_concepts?" did two different jobs for two different ID sets:

- For schemas / relation-schemas / relations: a dedup filter — figure out which are missing so publish can upload just those. Publish always had an upload path for these.
- For nodes: a gatekeeper — decide which selected nodes are allowed to be published at all. Present → grant access. Absent → skip, report "not synced yet." Publish had no upload path for nodes; it fully trusted sync to have put them there.


 The fix made publish upload every selected node itself, which means publish no longer needs sync's permission — the gatekeeper job for nodes ceased to exist. nodeUids came out of the query because its answer would no longer drive anything. The removal is fallout of the fix, not part of it.

I think the case that you put up is 1/3 cases 


| Case | State in DB | Skip upload safe? |
| --- | --- | --- |
| 1. Previously published | Complete row, kept<br>fresh by background<br>loop | Yes — upload is<br>redundant (map's case) |
| 2. Only touched by sync<br>(sync on, first publish) | Row exists but it's the<br>title-only stub | No — group gets granted<br>a stub with no content |
| 3. Never uploaded (sync<br>off, first publish) | No row | Moot — uploaded either<br>way (the ENG-2022 bug) |

maparent: Good point that I missed case 2, but are you saying that you are choosing to not handle case 1?

sid597: case 1 is handled by rewriting the same values with the node's own timestamps and never touches embeddings, so re-publishing is idempotent.

To skip safely, "row exists" isn't enough — publish would have to answer two more questions:

\- Is it complete? Does a variant='full' Content row exist, or is it just the direct-variant stub regular sync uploads?

\- Is it fresh? The loop is eventually consistent on a \~5-min cadence, so a just-edited node can be stale at the moment of publish.<br><br> Answering those ties publish's correctness back to sync's state — effectively rebuilding sync's diff logic inside publish — to save one idempotent write on a hand-picked selection.

maparent: I accept this argument. Approving.

maparent (approved):

PR #1284, opened 2026-08-07
ENG-2108 Add ranked search to QueryEngine with a vault-iteration fallback

trangdoan982:
## Scope check

- [x] Ran `$scope-check` against ENG-2108 and the final diff.

<details>
<summary>Mapping</summary>

One file, one concern. Three deviations from the ticket's literal wording, all narrowing rather than expanding: `createScorer` was dropped as a meaningless one-line wrapper over `prepareFuzzySearch`; the `Scorer` type and the `scorer?` injection parameter were dropped as unused indirection, so `prepareFuzzySearch` is called directly; and `filterCandidatesByNodeTypeIds` is exported instead, because the ticket requires callers to memoize the filtered array — which is only possible if the filter is callable on its own. No new dependencies, no migrations, no unrelated refactors.

</details>

## What this does

Adds the search seam for the Obsidian advanced node search panel (F1 of [Advanced Node Search (Obsidian)](https://linear.app/discourse-graphs/project/advanced-node-search-obsidian-db616dd9c62e)). No UI — the panel is #1285, stacked on this.

`QueryEngine.ts`:

| Export | Runs | Purpose |
| --- | --- | --- |
| `DiscourseNodeCandidate`, `RankedDiscourseNode` | — | types |
| `getDiscourseNodeCandidates()` | once per open | every DG node as `{ file, title, nodeTypeId }` |
| `filterCandidatesByNodeTypeIds(...)` | on filter change | node-type narrowing, OR across types |
| `rankDiscourseNodesByTitle({...})` | per keystroke | filter, score, sort |

### Three decisions worth reviewing

**Datacore first, vault iteration as fallback — via delegation, so it is not visible in this diff.** The call chain:

```
getDiscourseNodeCandidates()          ← new here
  └─ getFilesWithNodeTypeId()         ← existing, unchanged, so not in the diff
       ├─ Datacore installed  → `@page and exists(nodeTypeId)`
       └─ absent or throws    → fallbackGetFilesWithNodeTypeId()  (vault iteration)
```

Reusing that method rather than writing a second fallback here is why NF3 holds for free. 

## Verification

`check-types` clean, `eslint` 0 problems on the changed file.

**The reviewable demo is #1285**, stacked on this branch: https://www.loom.com/share/877db7d2de834548a81bca1750bbbecb

### DIFF
--- a/apps/obsidian/src/services/QueryEngine.ts
+++ b/apps/obsidian/src/services/QueryEngine.ts
@@ -1,4 +1,4 @@
-import { TFile, App } from "obsidian";
+import { TFile, App, prepareFuzzySearch, type SearchResult } from "obsidian";
 import type DiscourseGraphPlugin from "~/index";
 import { BulkImportPattern, BulkImportCandidate, DiscourseNode } from "~/types";
 import { getDiscourseNodeFormatExpression } from "~/utils/getDiscourseNodeFormatExpression";
@@ -21,6 +21,17 @@ type DatacorePage = {
   $path?: string;
 };
 
+export type DiscourseNodeCandidate = {
+  file: TFile;
+  /** Scored and rendered as-is: `renderResults` re-slices whatever was scored. */
+  title: string;
+  nodeTypeId: string;
+};
+
+export type RankedDiscourseNode = DiscourseNodeCandidate & {
+  match: SearchResult;
+};
+
 export class QueryEngine {
   private app: App;
   private dc:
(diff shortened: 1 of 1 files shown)


apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -21,6 +21,17 @@ type DatacorePage = {
   $path?: string;
 };
 
+export type DiscourseNodeCandidate = {
+  file: TFile;
+  /** Scored and rendered as-is: `renderResults` re-slices whatever was scored. */
```
mdroidian: I'm sorry, what is this comment trying to convey?  What does `renderResults` have to do with either `file` or `title`?

apps/obsidian/src/services/QueryEngine.ts
```diff
@@ -602,6 +633,54 @@ export class QueryEngine {
   }
 }
 
+/** Exported so callers can memoise the filtered array against their selected ids. */
+export const filterCandidatesByNodeTypeIds = (
```
mdroidian: `export` only when it needs to be exported (eg when it is used by callers.

Or make an inline comment saying that it will be used in xyz PR.

mdroidian (approved):

PR #1286, opened 2026-08-10
[ENG-1861] Add Roam refresh-all for imported nodes

sid597:
https://github.com/user-attachments/assets/74a1ec0a-bf44-40cb-92ec-3173a0dbc06c

Original setup is in #1279 

Adds a `DG: Refresh all imported nodes` command palette command (gated on node sharing, like the other sharing entry points) that refreshes every imported node with stored source identity and reports the outcome in a toast: `N refreshed, N skipped, N failed`.

Each node goes through the ENG-1860 single-node refresh path, so one failure doesn't abort the batch. To let the batch tally the three outcomes, `refreshImportedNode` now returns `status: "refreshed" | "skipped" | "failed"` instead of a boolean `success`, and takes a `force` option. The batch passes `force: false` so up-to-date imports are skipped instead of rewritten — that's where the skipped count comes from. The title button keeps `force: true` (default): an explicit click on one node still overwrites local edits even when the source is unchanged, per ENG-1860's behavior.

Nodes are fetched one at a time via `getSharedNodeByRid` rather than one bulk list because the ticket asks to reuse the single-node refresh path, and imported-node counts are small. Failure reporting is intentionally counts-only (each failure is still logged via `internalError`); actionable per-node reporting is ENG-1877.

The second commit fixes an adjacent violation of "imported nodes must not be synced": full sync selects nodes by title format + edit time only, so an imported page whose title matches a local node format was uploaded to Supabase as this space's own content, embeddings, and concepts — and import/refresh touching those pages re-triggered it. The fix filters pages with stored `importedFrom` identity out of the sync's node selection before the mode split, covering initial and incremental sync in both modes (shared-nodes-only mode was already safe via the shared source-local-id filter). Not covered here: the full-content path keys off explicit grants, which an imported page only enters via the publish title button that today also renders on imported pages — that publish-side exclusion is a separate follow-up.

Stacked on #1279 (ENG-1860).

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1286" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


maparent:
Ok watched the video. Two things before I review the code itself.

1. We should indeed not send imported nodes to the database; but I think we need to send imported schemas, as we want to allow limited edits of imported schemas. Either way, maybe that should be a separate PR?
2. The conservative refresh surprised me. The equivalent function on Obsidian is more aggressive. Now, I'm not necessarily against it; the current behaviour aggressively destroys user data, and yours may be an improvement. But it also introduces an inconsistency between the two "refresh import" mechanisms, that may also be confusing to the user. Was that a decision that was discussed? If not I think this is a great example of something where we want to review pros and cons together.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -1277,16 +1278,26 @@ export const createOrUpdateDiscourseEmbedding = async (
...
           sharedSourceLocalIds.has(node.source_local_id),
         )
-      : changedNodeInstances;
+      : nonImportedNodeInstances;
     const nodesToBackfillCoreTitle = (
       coreTitleBackfill?.nodesToBackfill ?? []
     ).filter(
       (node) =>
-        !sharedNodesOnlySync || sharedSourceLocalIds.has(node.source_local_id),
+        !importedNodeUids.has(node.source_local_id) &&
```
sid597, line 1298: Added while resolving the merge with main, not a mechanical conflict fix. `nodesToBackfill` is built from every discourse node in the graph, not from `changedNodeInstances`, so the `nonImportedNodeInstances` filter above does not cover it. Without this line an imported page already present in Supabase would be re-written by this space's sync through the core_title backfill, which is the invariant this PR exists to enforce. Side effect: imported nodes now count under `deferred` in the backfill PostHog event, since that is `nodesToBackfill.length - nodesToBackfillCoreTitle.length`. Telemetry only.

sid597:
1. Schemas are not affected. The exclusion filters node instances only (`syncDgNodesToSupabase.ts:1286-1293` and the backfill list at `:1298`). Schema upload goes through `convertDgToSupabaseConcepts` and `upsertNodeSchemaToContent`, which this PR doesn't touch, so schemas keep syncing as before. Roam also has no imported schemas yet: only imported instance pages get the `importedFrom` identity the filter keys on, and `materializeSharedNode` never creates node types. Limited edits of imported schemas would be new work on both sides. On splitting: I bundled the exclusion because refresh itself re-triggers the leak (refreshing an imported page bumps its edit time, and incremental sync then uploads it as this space's content via the title-format match). I can split it out if you prefer, but #1286 would then depend on that PR.

2. Partly specified, partly my call, so yes, let's review it. ENG-1861's Done When asks for refreshed, skipped, and failed counts, which needs a skip criterion. I reused the one ENG-1860 put in `materializeSharedNode`: skip when the stored source timestamp is at or after the remote `last_modified`. That's in the PR body, but we never weighed pros and cons. How I see it:
    - The two behaviors only differ for nodes whose source is unchanged. If the source changed, both Roam and Obsidian overwrite local edits (conflict resolution is out of scope in the ticket).
    - Conservative: safe to run any time, local edits survive unless the source actually changed, and the skipped count means something. Cost: inconsistent with Obsidian's refresh-all, and with the Roam title button, which still forces (a click on one node reads as "restore this one").
    - Aggressive: one mental model (refresh makes it match the source), but every run silently rewrites every imported node, with no undo.

    My preference: keep the conservative batch and align Obsidian's refresh-all to it in a follow-up ticket (skip when the source is unchanged), rather than make Roam aggressive. Happy to go the other way if you'd rather keep one behavior now and revisit data loss later.

maparent:
1. Good, thank you.
2. Thank you for your rationale for the conservative strategy. For the record, I said it was worth discussing, and was not in opposition. I agree with your preference. If @mdroidian also agrees, we may want to propose a ticket to align Obsidian to Roam here. (Or should we surface to the team?)

I will start the review asap.

mdroidian (approved):
Conservative refresh is the right approach. A user-facing “Refresh all” should only update imported nodes whose source has changed, rather than rewriting every imported node.

We should change Obsidian to match this Roam behavior. We can discuss the broader refresh behavior tomorrow using [these notes](https://roamresearch.com/#/app/discourse-graphs/page/cYTcRU9qb).

PR #1272, opened 2026-07-31
ENG-2037 Add tabbed node card context menu to Roam tldraw

sid597:
Overrides tldraw's StylePanel for Roam: selecting a single node card shows a Context tab (default) listing all discourse relations grouped by relation type, with +/− buttons that add or remove the related node on the canvas, and a Styling tab that reuses the stock styling controls. Regular tldraw shapes and multi-selection keep the default style panel. Adding a node also draws its existing relation arrows via createExistingRelations; removing deletes the node shape and its arrows — relation data in Roam is never modified.

---

Open in Devin Review

PR #1288, opened 2026-08-10
[ENG-2096] Remove the discourse context overlay from Roam canvas

sid597:
Removes the canvas discourse context overlay and the "(BETA) Overlay in canvas" personal setting. Stacked on ENG-2037 (#1272) — the node card context menu is now the single way to view relations on canvas. Replacement decision per ENG-2037 and the Roam decision page linked in ENG-2096; FEE-881 (which proposed keeping the overlay) was canceled.

Also drops the setting from the zod schema, PERSONAL_KEYS, and the legacy key migration map. Old stored values are ignored harmlessly since the schema is non-strict, so no data migration is needed. Outlines/linked-references overlays are untouched (out of scope per the ticket).

No tests covered the canvas overlay path, so none were removed.


apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -220,10 +220,6 @@ const PERSONAL_SCHEMA_PATH_TO_LEGACY_KEY = new Map<string, string>([
   [pathKey([PERSONAL_KEYS.disableSidebarOpen]), "disable-sidebar-open"],
   [pathKey([PERSONAL_KEYS.hideFeedbackButton]), "hide-feedback-button"],
   [pathKey([PERSONAL_KEYS.autoCanvasRelations]), "auto-canvas-relations"],
-  [
```
sid597, line 223: Legacy migration iterates `Object.values(PERSONAL_KEYS)`, so with the key gone this map entry is never looked up — old graphs simply skip migrating the dead setting.

apps/roam/src/components/settings/utils/zodSchema.ts
```diff
@@ -248,7 +248,6 @@ export const PersonalSettingsSchema = z.object({
     .default({ modifiers: 0, key: "" }),
   "Discourse context overlay": z.boolean().default(false),
   "Reified relation triples": z.boolean().default(true),
-  "Overlay in canvas": z.boolean().default(false),
```
sid597, line 251: `PersonalSettingsSchema` is a non-strict `z.object`, so a persisted `"Overlay in canvas"` prop is stripped on parse. Writes path-patch rather than rewrite the object, so the stale boolean stays in old users' block props — harmless residue, not worth a cleanup migration.

apps/roam/src/components/canvas/DiscourseNodeUtil.tsx
```diff
@@ -449,16 +448,9 @@ export class DiscourseNodeUtil extends BaseBoxShapeUtil<DiscourseNodeShape> {
...
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    const isOverlayEnabled = useMemo(
-      () => getPersonalSetting<boolean>([PERSONAL_KEYS.overlayInCanvas]),
-      [],
-    );
 
     const isEditing = this.editor.getEditingShapeId() === shape.id;
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    const [overlayMounted, setOverlayMounted] = useState(false);
-    // eslint-disable-next-line react-hooks/rules-of-hooks
```
sid597, line 461: `overlayMounted` + the `onPointerEnter` handler below existed solely to defer overlay mounting until hover; they go with the overlay.

mdroidian:
@codex review

mdroidian (approved):
For future: a UI video on something like this is still a good idea. Just double-checking that everything still works, doesn't break, and we get the expected result.

PR #1292, opened 2026-08-12
ENG-2113 Add footer action bar with open in new tab and split

trangdoan982:
https://www.loom.com/share/85492db69f9f4c0d92f09d3234d64eb7

## Scope check

- [x] Ran `$scope-check` against ENG-2113 and the final diff.
- Scope beyond `Done When`: two deliberate deviations, both cosmetic-to-small.
  1. **Enter opens a new tab in the main panel, not the active pane.** `Done When` only requires that "both actions open the correct note and close the modal", so this stays inside the acceptance boundary — but it contradicts the ticket's `Solution`, which specified `app.workspace.getLeaf(false)`, and the ticket title. Replacing the note the user was already reading loses their place, which is the opposite of what a lookup surface should do. The footer label reads "open in new tab" to match.
  2. **An `esc close` hint.** A third, non-clickable footer item for Escape, which Obsidian's modal scope already handles. Included so the footer matches Roam's `AdvancedSearchFooter` and the native quick switcher, both of which show one.
- Required now: (1) yes — it is the behaviour actually wanted from the surface, and shipping `getLeaf(false)` first would mean changing it immediately. (2) No, easily dropped.
- Anyone affected or consulted: requested directly by @trang-doan during review of the working build.
- Decision: not recorded in Linear. **The ticket's `Solution` and title should be updated to say "new tab" rather than "active pane"** — flagging rather than editing the ticket myself.

Stacked on #1285 (ENG-2109) — review that first. Base is its branch, not `main`.

## What this does

F8. `Enter` opens the active result in a new tab in the main panel, `Shift+Enter` opens it in a split, both close the modal, and both are clickable in a new footer.

`Mod+Enter` and `Alt+Enter` deliberately fall through untouched so ENG-2114's insert-at-cursor can claim `Mod+Enter`, the same combo Roam uses. `Shift+Enter` has no native conflict: it means "create new file" only inside Obsidian's own quick switcher, which this modal does not offer.

The Enter branch lives in the existing wrapper `onKeyDown` — ENG-2109 moved that handler off the input specifically so result actions would have one place to live. It also guards `isComposing`, so committing an IME candidate never opens a file.

Both open helpers (`openFileInNewTab`, `openFileInNewLeaf`) already existed and are reused unchanged, so no file outside the search feature is modified.

## Avoiding ENG-2000

Roam's footer hardcodes Blueprint's macOS glyph icons per action (`keyIcons={["key-option", "key-enter"]}`), so Windows users see ⌘/⌥ while the handler actually accepts Ctrl. Rather than port that shape, every hint here goes through one `keyboardHints.ts` map keyed on `Platform.isMacOS`. No call site can name a platform-specific symbol.

`formatHintKeys` takes `isMacOS` as a parameter so the Windows/Linux branch is verifiable without that platform:

```
macOS       ["↵","⇧ ↵","⌘ ↵","⌥ ↵","esc"]
win/linux   ["Enter","Shift Enter","Ctrl Enter","Alt Enter","Esc"]
```

## Native styling

- The footer reuses Obsidian's own `prompt-instructions` / `prompt-instruction` / `prompt-instruction-command` classes — what `SuggestModal.setInstructions()` emits.
- Matching closely with Roam design

Beyond that it only needed `flex-shrink: 0` (the `.modal-content` column is fixed-height with `overflow: hidden`) and a reset for Obsidian's global `button` chrome.

### DIFF
--- a/apps/obsidian/src/components/NodeSearchFooter.tsx
+++ b/apps/obsidian/src/components/NodeSearchFooter.tsx
@@ -14,12 +14,15 @@ type FooterActionProps = {
   onClick: () => void;
 };
 
+// Roam's footer renders each key as a bordered cap rather than bold text, which
+// keeps `esc` reading as one of the set instead of standing out. Deliberately
+// not `prompt-instruction-command`, whose weight is what makes it stand out.
 const KeyHints = ({ keys }: { keys: HintKey[] }): ReactElement => (
   <>
     {getHintKeys(keys).map((symbol) => (
-      <span className="prompt-instruction-command" key={symbol}>
+      <kbd className="dg-search-footer-key" key={symbol}>
         {symbol}
-      </span>
+      </kbd>
     ))}
   </>
 );
@@ -40,13 +43,12 @@ const FooterAction = ({
     onMouseDown={(event) => event.preventDefault()}
   >
     <KeyHints keys={keys} />
-    <span>{label}</span>
+    <span className="dg-search-footer-label">{label}</span>
   </button>
 );
 
-// Reuses Obsidian's own `prompt-instruction` markup, the same classes
-// `SuggestModal.setInstructions()` emits, so the footer matches the native
-// quick switcher. This modal extends plain `Modal`, so that API is unavailable.
+// Sits in Obsidian's `prompt-instructions` container for its type and spacing,
+// but styles the keys as Roam's search footer does.
 export const NodeSearchFooter = ({
   canAct,
   onOpenInNewTab,
@@ -66,9 +68,9 @@ export const NodeSearchFooter = ({
       onClick={onOpenInSplit}
     />
     {/* Escape is handled by Obsidian's modal scope, so this is a hint only. */}
-    <span className="prompt-instruction">
+    <span className="prompt-instruction dg-search-footer-hint">
       <KeyHints keys={["Escape"]} />
-      <span>close</span>
+      <span className="dg-search-footer-label">close</span>
     </span>
   </div>
 );
--- a/apps/obsidian/src/components/NodeSearchFooter.tsx
+++ b/apps/obsidian/src/components/NodeSearchFooter.tsx
@@ -3,7 +3,7 @@ import { getHintKeys, type HintKey } from "~/utils/keyboardHints";
 
 type NodeSearchFooterProps = {
   canAct: boolean;
-  onOpenInActivePane: () => void;
+  onOpenInNewTab: () => void;
   onOpenInSplit: () => void;
 };
 
@@ -49,15 +49,15 @@ const FooterAction = ({
 // quick switcher. This modal extends plain `Modal`, so that API is unavailable.
 export const NodeSearchFooter = ({
   canAct,
-  onOpenInActivePane,
+  onOpenInNewTab,
   onOpenInSplit,
 }: NodeSearchFooterProps): ReactElement => (
   <div className="prompt-instructions dg-search-footer">
     <FooterAction
       disabled={!canAct}
       keys={["Enter"]}
-      label="open"
-      onClick={onOpenInActivePane}
+      label="open in new tab"
+      onClick={onOpenInNewTab}
     />
     <FooterAction
       disabled={!canAct}
--- a/apps/obsidian/src/components/NodeSearchFooter.tsx
+++ b/apps/obsidian/src/components/NodeSearchFooter.tsx
@@ -1,0 +1,74 @@
+import { type ReactElement } from "react";
+import { getHintKeys, type HintKey } from "~/utils/keyboardHints";
+
+type NodeSearchFooterProps = {
+  canAct: boolean;
+  onOpenInActivePane: () => void;
+  onOpenInSplit: () => void;
+};
+
+type FooterActionProps = {
+  disabled: boolean;
+  keys: HintKey[];
+  label: string;
+  onClick: () => void;
+};
+
+const KeyHints = ({ keys }: { keys: HintKey[] }): ReactElement => (
+  <>
+    {getHintKeys(keys).map((symbol) => (
+      <span className="prompt-instruction-command" key={symbol}>
+        {symbol}
+      </span>
+    ))}
+  </>
+);
+
+const FooterAction = ({
+  disabled,
+  keys,
+  label,
+  onClick,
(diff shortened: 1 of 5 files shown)


mdroidian:
@codex review

apps/obsidian/src/components/NodeSearchFooter.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+type FooterActionProps = {
+  disabled?: boolean;
+  keys: HintKey[];
+  label: string;
+  onClick: () => void;
+};
+
+// Roam's footer renders each key as a bordered cap rather than bold text, which
```
mdroidian, line 18: I'm not sure what this comment is supposed to be informing me.  Was there a different choice made in Obsidian as opposed to Roam that this comment is trying to infer about?

apps/obsidian/src/components/NodeSearchFooter.tsx
```diff
@@ -0,0 +1,76 @@
...
+  keys,
+  label,
+  onClick,
+}: FooterActionProps): ReactElement => (
+  <button
+    type="button"
+    className="prompt-instruction dg-search-footer-action"
+    disabled={disabled}
+    onClick={onClick}
+    // Same reason as the result rows: keep focus in the query input so arrow-key
```
mdroidian, line 42: What is the same reason?

apps/obsidian/src/components/NodeSearchModal.tsx
```diff
@@ -395,11 +402,36 @@ const NodeSearch = ({
...
+
+    if (event.key !== "Enter") return;
+    // Enter also commits an IME candidate, which must not open a file.
+    if (event.nativeEvent.isComposing) return;
+    // Mod+Enter and Alt+Enter are left alone for the insert and dock actions.
+    if (event.metaKey || event.ctrlKey || event.altKey) return;
+
     event.preventDefault();
-    moveActiveIndex(event.key === "ArrowDown" ? 1 : -1);
+    openActiveResult(event.shiftKey ? openFileInNewLeaf : openFileInNewTab);
```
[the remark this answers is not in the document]
mdroidian: @trangdoan982

apps/obsidian/src/styles/style.css
```diff
@@ -3921,3 +3921,64 @@ kbd.tlui-kbd {
   border-radius: var(--radius-s);
   padding: 0 1px;
 }
+
+/* Obsidian's own `.prompt-instructions` rules supply the type and colour. They
+   centre the row for the quick switcher, which is too narrow to need alignment;
+   this footer sits under a full-width result list, so the actions line up with
+   its left edge instead. `flex-shrink` keeps the footer from collapsing inside
+   the fixed-height flex column above it. */
+.dg-node-search-modal .dg-search-footer {
```
mdroidian, line 3930: prefer tailwind where we can

apps/obsidian/src/utils/keyboardHints.ts
```diff
@@ -0,0 +1,36 @@
+import { Platform } from "obsidian";
+
+export type HintKey = "Mod" | "Alt" | "Shift" | "Enter" | "Escape";
+
+// Obsidian shows glyphs on macOS and spelled-out words everywhere else, so the
+// same shortcut has to render two ways. Roam's search footer hardcoded the macOS
```
mdroidian, line 6: Some unnecessary comments here

mdroidian (approved):

apps/obsidian/src/components/NodeSearchModal.tsx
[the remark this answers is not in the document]
trangdoan982: Valid, and worse than described now that close is also a button — fixed in 48c234e3.

Enter on a focused footer button bubbled to the modal handler, whose \`preventDefault()\` suppressed the button's native click, so Tab→close and Tab→split both opened a new tab instead. Rather than scope the shortcut to the input (ENG-2109 deliberately moved it to the wrapper so arrow keys work anywhere in the modal), the Enter branch now returns early when the event originates inside a button:

```ts
if (event.target instanceof HTMLElement && event.target.closest("button") !== null) return;
```

Arrow keys are unaffected — they return before this guard — and the buttons stay Tab-reachable.

apps/obsidian/src/components/NodeSearchFooter.tsx
trangdoan982: Fair — it was explaining Roam history rather than this code, so I removed it. No Obsidian-vs-Roam decision was hiding in there; the keys are just rendered as bordered caps instead of bold text so `esc` reads as one of the set. That is now evident from the `dg-search-footer-key` class and its CSS rule.

apps/obsidian/src/components/NodeSearchFooter.tsx
trangdoan982: Unclear reference on my part — it pointed at the result rows in NodeSearchModal.tsx, which do the same thing. Reworded to say the actual reason: clicking must not move focus out of the query input, or the arrow keys stop reaching the result list.

apps/obsidian/src/styles/style.css
trangdoan982: Done in 48c234e3 — the footer layout is now utilities (`inline-flex h-auto cursor-pointer items-center gap-1 rounded-none border-0 p-0 disabled:*`, and `shrink-0 justify-start px-0 pb-0 text-left` on the row). Net 16 lines of CSS removed.

Four properties have to stay in CSS, and it is a specificity limit rather than preference: Obsidian sets `color`, `background-color`, and `box-shadow` in `button:not(.clickable-icon)` and `button:hover`, which are (0,1,1) and outrank a single utility class (0,1,0). I tried `text-inherit` first and it loses — the label renders in `--text-normal` on an interactive-grey pill. `font-size: inherit` stays because Tailwind has no inherit-size utility, and without it the button takes `--font-ui-small` instead of the row's smaller type. The remaining `dg-search-footer-key` rule is kept for the inherited font and em-based sizing.

apps/obsidian/src/utils/keyboardHints.ts
trangdoan982: Trimmed to one line. The ENG-2000 background lives in the PR description instead, which is the right place for it.

PR #1293, opened 2026-08-13
ENG-2147 Upgrade eslint and typescript-eslint

trangdoan982:
## Scope check

- Scope beyond `Done When`: **None.** The diff is the upgrade and the minimum config changes needed to make it lint.

---

Carries out the upgrade recommended by the [ENG-2118](https://linear.app/discourse-graphs/issue/ENG-2118/investigate-whether-to-upgrade-repo-eslint-from-version-8-to) investigation. That ticket scoped the upgrade itself out; this is the follow-up that lands it.

> This PR previously carried the ENG-2118 investigation writeup. It has been rebased on latest `main`, re-verified, and rewritten as the actual upgrade under ENG-2147.

## Why

[ENG-1832](https://linear.app/discourse-graphs/issue/ENG-1832/add-obsidian-eslint-to-the-project) needs `eslint-plugin-obsidianmd@0.4.1`, whose peer ranges `main` could not satisfy:

| Peer | Required | Before | After |
|---|---|---|---|
| `eslint` | `>=9.19.0` | 8.57.1 ❌ | 9.39.5 ✅ |
| `typescript-eslint` | `^8.35.1` | 7.18.0 ❌ | 8.67.0 ✅ |
| `@eslint/js` | `^9.30.1` | 9.34.0 ✅ | 9.34.0 ✅ |

## Testing

```bash
git fetch origin && git checkout eng-2118-investigate-whether-to-upgrade-repo-eslint-from-version-8-to
pnpm install
```

Confirm the upgrade actually resolved:

```bash
pnpm --dir packages/eslint-config exec eslint --version
```
image

Lint everything. Use `--continue` so one workspace failing does not mask the rest:

```bash
npx turbo lint --continue
```

image

These are same errors that are appearing on main branch. Lint errors not caused by this upgrade

```bash
npx turbo check-types
```
image

CI is unaffected either way: the `lint-changed-files` job runs `eslint` directly against changed files and never invokes `turbo lint`.

## What changed

**Dependencies** (`pnpm-workspace.yaml` catalog + `packages/eslint-config/package.json`)

- `eslint` 8.57.1 → `^9.39.5`
- `typescript-eslint` `^7.18.0` → `^8.67.0`
- `@types/eslint` 8.56.12 → `^9.6.1`
- Dropped `@vercel/style-guide` and `eslint-config-turbo`. Both were declared but never imported by any config, and `@vercel/style-guide` was the last hard pin holding eslint below 9.
- Dropped the now-unneeded `eslint: ^8.57` entry from `allowedDeprecatedVersions`.

**Config** — three v8/v9 breaking changes had to be handled for the repo to lint at all:

1. **`tsconfigRootDir` must be absolute.** typescript-eslint v8 rejects `"."`, and every file fails to parse. Now uses `import.meta.dirname`, which is safe because the repo already requires Node >=22.
2. **Type-aware linting must be scoped to TS files.** With `project: true` applied to every file, v8 *errors* on `.mjs` files that belong to no tsconfig, where v7 silently skipped them. The `files` glob must include `.mts`/`.cts` or ESLint crashes outright on `packages/database/scripts`.
3. **Plain JS needs `disableTypeChecked`** for the same reason — added as a dedicated block in `packages/eslint-config/base.js`.

No source files are touched. The diff is ESLint config, the dependency catalog, and `pnpm-lock.yaml` (which shrinks by ~1,100 lines as the two dropped packages' trees go away).

## Appendix: Results

Measured on this branch rebased onto `main` @ `da2861f3`.

| Check | Result |
|---|---|
| `turbo lint` (7 workspaces) | **0 errors**, 754 warnings |
| `turbo check-types` | **8 / 8 pass** |

Per workspace:

| Workspace | Errors | Warnings |
|---|---|---|
| `apps/roam` | 0 | 538 |
| `apps/obsidian` | 0 | 111 |
| `apps/website` | 0 | 86 |
| `packages/database` | 0 | 15 |
| `packages/ui` | 0 | 4 |
| `packages/content-model` | 0 | 0 |
| `packages/utils` | 0 | 0 |

### Warnings: 573 → 754

The repo runs `eslint-plugin-only-warn`, so nothing here can fail a build — but the count moves and reviewers should know why. **62 of the +181 are ESLint 9 reporting unused suppression directives**, which ESLint 8 ignored by default:

| New warning | Count |
|---|---|
| Unused `eslint-disable` directive | 43 |
| Unused `eslint-enable` directive | 19 |

The remainder comes from typescript-eslint v8 broadening its type-aware rules. Top contributors in the current run:

| Rule | Count |
|---|---|
| `@typescript-eslint/no-floating-promises` | 94 |
| `@typescript-eslint/no-unnecessary-type-assertion` | 79 |
| `@typescript-eslint/no-unused-vars` | 61 |
| `@typescript-eslint/no-unsafe-assignment` | 58 |
| `@typescript-eslint/no-unsafe-member-access` | 55 |
| `react/prop-types` | 48 |
| `react-hooks/exhaustive-deps` | 31 |
| `@typescript-eslint/no-misused-promises` | 30 |
| `@typescript-eslint/naming-convention` | 30 |
| `@typescript-eslint/no-explicit-any` | 30 |

⚠️ **Do not run `eslint --fix` over this branch.** 103 of the warnings in `apps/roam` alone are reported as auto-fixable, but the fixers here are not safe to apply blind: `no-unnecessary-type-assertion` and the unused-directive fixers delete code that exists for a reason in places, and a sweeping autofix would bury this dependency upgrade under hundreds of unrelated source changes. Cleaning up warnings belongs in its own ticket.

### DIFF
--- a/apps/obsidian/eslint.config.mjs
+++ b/apps/obsidian/eslint.config.mjs
@@ -3,9 +3,10 @@ import { config } from "@repo/eslint-config/react-internal";
 export default [
   ...config,
   {
+    files: ["**/*.{ts,tsx,mts,cts}"],
     languageOptions: {
       parserOptions: {
-        tsconfigRootDir: ".",
+        tsconfigRootDir: import.meta.dirname,
         project: true,
         ecmaFeatures: {
           jsx: true,
--- a/apps/roam/eslint.config.mjs
+++ b/apps/roam/eslint.config.mjs
@@ -5,9 +5,10 @@ import { config } from "@repo/eslint-config/react-internal";
 export default [
   ...config,
   {
+    files: ["**/*.{ts,tsx,mts,cts}"],
     languageOptions: {
       parserOptions: {
-        tsconfigRootDir: ".",
+        tsconfigRootDir: import.meta.dirname,
         project: true,
         ecmaFeatures: {
           jsx: true,
--- a/apps/website/eslint.config.mjs
+++ b/apps/website/eslint.config.mjs
@@ -4,9 +4,10 @@ export default [
   ...nextJsConfig,
   // additions
   {
+    files: ["**/*.{ts,tsx,mts,cts}"],
     languageOptions: {
       parserOptions: {
-        tsconfigRootDir: ".",
+        tsconfigRootDir: import.meta.dirname,
         project: true,
         ecmaFeatures: {
           jsx: true,
--- a/packages/content-model/eslint.config.mjs
+++ b/packages/content-model/eslint.config.mjs
@@ -3,10 +3,11 @@ import { config as base } from "@repo/eslint-config/base";
 export default [
   ...base,
   {
+    files: ["**/*.{ts,tsx,mts,cts}"],
     languageOptions: {
       parserOptions: {
         project: true,
-        tsconfigRootDir: ".",
+        tsconfigRootDir: import.meta.dirname,
       },
     },
   },
--- a/packages/database/eslint.config.mjs
+++ b/packages/database/eslint.config.mjs
@@ -3,10 +3,11 @@ import { config as base } from "@repo/eslint-config/base";
 export default [
   ...base,
   {
+    files: ["**/*.{ts,tsx,mts,cts}"],
     languageOptions: {
       parserOptions: {
         project: true,
-        tsconfigRootDir: ".",
+        tsconfigRootDir: import.meta.dirname,
       },
     },
   },
--- a/packages/eslint-config/base.js
+++ b/packages/eslint-config/base.js
@@ -130,6 +130,12 @@ export const config = [
     },
   },
   {
+    // Plain JS files are not part of the TypeScript projects, and typescript-eslint v8
+    // errors instead of silently skipping them.
+    files: ["**/*.js", "**/*.mjs", "**/*.cjs"],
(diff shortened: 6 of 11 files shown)


maparent:
Really great work!
Two questions: Is it easy to fix the `typescript-eslint@7.18.0` constraint? Latest version is 8.67.0, maybe it forces us to upgrade typescript?

Re `lint --fix`. 
My reflex would be to run it with the old ESLint (no negative impact) before migrating.

trangdoan982:
thanks for the review MAP. no need to upgrade typescript. upgrading typescript-eslint already widens support for typescript.<br><br>re `lint --fix`. I think the reason we don't run it already is that there are simply too many warnings already. So whether to run before or after the upgrade we'll face the same issue: too many changes with no way to track if it breaks certain functionalities or not. I'd say we we upgrade and fix per file as the PR CI touches changed files

maparent:
fair enough.

trangdoan982:
[ENG-2147: Upgrade eslint and typescript-eslint](https://linear.app/discourse-graphs/issue/ENG-2147/upgrade-eslint-and-typescript-eslint)

maparent (approved):

mdroidian (approved):

PR #1298, opened 2026-08-17
ENG-2097 Exclude Page, Block, and Any from relation endpoint selectors

sid597:
https://github.com/user-attachments/assets/37da761e-788b-4a83-97e5-889fad335595

Builds the relation editor's Source and Destination options from configured discourse node types only (via the existing `excludeDefaultNodes`), so `Page`, `Block`, and `Any` can no longer be saved as new relation endpoints.

The full node map still includes Page/Block/Any because the relations table, sorting, and the selector's current-value button use it to display existing relations with generic endpoints — migrating those is out of scope per the ticket. `MenuItemSelect` renders `activeItem` independently of `items`, so editing such a relation still shows its current endpoint; it just can't be selected anew.

Closes ENG-2097

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1298" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/settings/DiscourseRelationConfigPanel.tsx
+++ b/apps/roam/src/components/settings/DiscourseRelationConfigPanel.tsx
@@ -44,7 +44,9 @@ import { render as renderToast } from "roamjs-components/components/Toast";
 import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageUid";
 import updateBlock from "roamjs-components/writes/updateBlock";
 import getTextByBlockUid from "roamjs-components/queries/getTextByBlockUid";
-import getDiscourseNodes from "~/utils/getDiscourseNodes";
+import getDiscourseNodes, {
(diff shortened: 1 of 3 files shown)


mdroidian (approved):
Some video suggestions
1) It'd be neat if the videos had some transition (very quick, very light) and/or a toast to notify when something changes (eg, in the top right or left, overtop of everything).

Right now that is all done via voice and can be hard to follow with the non-standard language.

So in this case it was "This is the build before the fix" and "Same graph, same relations, now on the branch"

2) Let's remove redundant, unnecessary language.  It significantly reduces the effectiveness of the video.

EG: 

"The graph has several discourse node types and a set of relations between them" (I would never have guessed ... 🙄)
"Every relation has a source and destination" (this is the definition of a relation, yes, thank you 😑)

PR #1301, opened 2026-08-17
ENG-2145 Utility function to discover importable relations and schemas before node import

maparent:
https://linear.app/discourse-graphs/issue/ENG-2145/utility-function-to-discover-importable-relations-and-schemas-before

https://www.loom.com/share/df8155a2b15446e7b650574591b6a89c


mdroidian (approved):
See graphite's comments

PR #1315, opened 2026-08-19
ENG-2110 Add node type filter dropdown menu

trangdoan982:
https://www.loom.com/share/f8349cfbe61f447c80986c968f7b9547

## What this does

Adds a node type filter beside the search input in the node search modal. Selecting types narrows results immediately, the active filter shows as a count badge on the trigger, clearing restores the full set, and multiple types return the union.

## Changes

**`apps/obsidian/src/utils/discourseNodeTypeFilter.ts`** (new) — Roam's semantics ported, narrowed to `string[]` ids and Obsidian's `DiscourseNode` (`.id`/`.name` rather than `.type`/`.text`). The piece worth attention is `toPanelSelectedIds`/`fromPanelSelectedIds`: the stored empty set means "no filter", which the panel must render as *every* row checked, and selecting all normalises back to empty. That keeps the stored state, the count badge, and `filterCandidatesByNodeTypeIds` agreeing on what counts as a filter.

**`apps/obsidian/src/components/NodeTypeFilterMenu.tsx`** (new) — Obsidian-native trigger (`clickable-icon` + `setIcon`) with a count badge, over a panel with checkbox rows, colour dots, a per-row "Only" button, a `Clear filter (n)` button while a filter is active, and a type-search input past 7 types. Reuses `getAllDiscourseNodeColors` for the dots so palette fallbacks stay index-stable. Outside-click uses `activeDocument`, following `InlineNodeTypePicker`, so it works in a popout window.

**`apps/obsidian/src/components/NodeSearchModal.tsx`** — `selectedNodeTypeIds` state as the single source of truth so ENG-2111's chips can read and write the same state; trigger inline-right of the input, leaving the row beneath free for those chips; focus returns to the input when the panel closes.

## Two behaviours worth understanding before reviewing

**`Clear filter` rather than `Select all`.** Roam has a tri-state "Select all" checkbox here, and it is the one place this PR knowingly diverges. Because an empty selection and a full one are the same state by design, that checkbox sits checked and inert whenever no filter is active, so clicking it appears to do nothing. Clearing is the control's only real function, so it now says so and appears only when there is something to clear.

### DIFF
--- a/apps/obsidian/src/components/NodeSearchModal.tsx
+++ b/apps/obsidian/src/components/NodeSearchModal.tsx
@@ -434,18 +434,6 @@ const NodeSearch = ({
     });
   };
 
-  // Focus stays in the search input while the panel is open, so Escape arrives
-  // here rather than at the panel. Obsidian's Modal closes on Escape from its own
-  // keymap scope, so the native event has to stop or the modal goes too.
-  const closeTypeFilterOnEscape = (
-    event: KeyboardEvent<HTMLDivElement>,
-  ): void => {
-    event.preventDefault();
-    event.nativeEvent.stopImmediatePropagation();
-    setIsTypeFilterOpen(false);
-    inputRef.current?.focus();
-  };
-
   const handleTypeFilterOpenChange = (nextOpen: boolean): void => {
     setIsTypeFilterOpen(nextOpen);
     // Returns the keyboard path to the results the moment the panel closes.
@@ -453,11 +441,6 @@ const NodeSearch = ({
   };
 
   const handleKeyDown = (event: KeyboardEvent<HTMLDivElement>) => {
-    if (event.key === "Escape" && isTypeFilterOpen) {
-      closeTypeFilterOnEscape(event);
-      return;
-    }
-
     if (event.key === "ArrowDown" || event.key === "ArrowUp") {
       // Otherwise the caret jumps to the start or end of the query.
       event.preventDefault();
@@ -499,6 +482,7 @@ const NodeSearch = ({
           className="min-w-0 flex-1"
         />
         <NodeTypeFilterMenu
+          app={app}
           isOpen={isTypeFilterOpen}
           nodeTypes={plugin.settings.nodeTypes}
           onOpenChange={handleTypeFilterOpenChange}
--- a/apps/obsidian/src/components/NodeSearchModal.tsx
+++ b/apps/obsidian/src/components/NodeSearchModal.tsx
@@ -329,8 +329,7 @@ const NodeSearch = ({
   const [query, setQuery] = useState("");
   const [debouncedQuery, setDebouncedQuery] = useState("");
   const [activeIndex, setActiveIndex] = useState(0);
-  // The single source of truth for active type filters: F6's chips will read and
-  // write this same state, so either surface can manage them.
+  // Single source of truth: ENG-2111's tag chips will read and write this too.
   const [selectedNodeTypeIds, setSelectedNodeTypeIds] = useState<string[]>([]);
   const [isTypeFilterOpen, setIsTypeFilterOpen] = useState(false);
   const inputRef = useRef<HTMLInputElement | null>(null);
--- a/apps/obsidian/src/components/NodeSearchModal.tsx
+++ b/apps/obsidian/src/components/NodeSearchModal.tsx
@@ -21,6 +21,7 @@ import {
 import { createRoot, Root } from "react-dom/client";
 import type DiscourseGraphPlugin from "~/index";
 import { NodeSearchFooter } from "~/components/NodeSearchFooter";
+import { NodeTypeFilterMenu } from "~/components/NodeTypeFilterMenu";
 import {
   openFileInNewLeaf,
   openFileInNewTab,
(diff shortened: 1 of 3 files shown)


apps/obsidian/src/components/NodeTypeFilterMenu.tsx
```diff
@@ -0,0 +1,292 @@
...
+        if (event.key !== "Escape" || !isOpen) return;
+        // Obsidian's Modal closes on Escape from its own keymap scope, so the
+        // native event has to stop here or the whole modal goes with the panel.
+        event.preventDefault();
+        event.nativeEvent.stopImmediatePropagation();
+        // The modal's own Escape handler is an ancestor of this one; stopping the
+        // synthetic event too keeps it from closing the panel a second time.
+        event.stopPropagation();
+        onOpenChange(false);
+      }}
```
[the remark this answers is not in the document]
trangdoan982: Valid and fixed in 34dec053. The container's `onKeyDown` only stopped propagation for Escape, so every other key reached the modal's handler — Enter ran the `openActiveResult` branch and the arrows moved the result selection. It now calls `stopPropagation()` for all keys while the panel is open, before the Escape-specific handling. Thanks — this was the failure mode I was most worried about and you found a path I had missed.

apps/obsidian/src/components/NodeTypeFilterMenu.tsx
```diff
@@ -0,0 +1,292 @@
...
+    if (showTypeSearch) searchRef.current?.focus();
+  }, [showTypeSearch]);
+
+  const handleSelectAll = (): void => {
+    if (selectAllState === "off") {
+      onSelectedIdsChange(nodeTypes.map((nodeType) => nodeType.id));
+      return;
+    }
+    onSelectedIdsChange([]);
+  };
```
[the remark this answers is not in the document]
trangdoan982: Correct diagnosis, and fixed in 34dec053 — though by changing the control rather than the semantics.

The round-trip you traced is deliberate: an empty selection and a full one genuinely are the same state, because "zero types selected" would mean zero results. That canonicalisation is what keeps the stored state, the count badge, and `filterCandidatesByNodeTypeIds` agreeing, so it stays. What was wrong was pairing it with a checkbox, which then had no meaningful off position and sat checked-and-inert whenever no filter was active.

Since clearing is the only thing that row ever did, it is now a "Clear filter (n)" button rendered only while a filter is active. No dead state left to click.

apps/obsidian/src/components/NodeSearchModal.tsx
```diff
@@ -329,6 +330,10 @@ const NodeSearch = ({
   const [query, setQuery] = useState("");
   const [debouncedQuery, setDebouncedQuery] = useState("");
   const [activeIndex, setActiveIndex] = useState(0);
+  // The single source of truth for active type filters: F6's chips will read and
```
mdroidian: F6?

apps/obsidian/src/utils/discourseNodeTypeFilter.ts
```diff
@@ -0,0 +1,64 @@
+import { DiscourseNode } from "~/types";
+
+/**
+ * Node type filtering for the search modal. An empty `selectedTypeIds` means no
+ * filter, matching `filterCandidatesByNodeTypeIds` in QueryEngine — so "nothing
+ * selected" and "every type selected" are the same state and both show all nodes.
+ * Ported from Roam's advanced search so both apps filter alike.
+ */
+
+/** Below this many types the list is short enough to scan without a search box. */
```
mdroidian: I'm guessing this doesn't take mobile into account?

apps/obsidian/src/utils/discourseNodeTypeFilter.ts
```diff
@@ -0,0 +1,64 @@
...
+export const NODE_TYPE_FILTER_SEARCH_THRESHOLD = 7;
+
+export const hasActiveTypeFilter = ({
```
mdroidian: This probably doesn't need to be an exported function.  Could it just be an inline variable?

apps/obsidian/src/utils/discourseNodeTypeFilter.ts
```diff
@@ -0,0 +1,64 @@
...
+}: {
+  selectedTypeIds: string[];
+  allTypeIds: string[];
+}): boolean =>
+  selectedTypeIds.length > 0 && selectedTypeIds.length < allTypeIds.length;
+
+/**
+ * The stored empty set means "no filter", which the panel shows as every row
+ * checked — otherwise an unfiltered search would render as an empty checklist.
+ */
```
mdroidian: Same with this

apps/obsidian/src/utils/discourseNodeTypeFilter.ts
```diff
@@ -0,0 +1,64 @@
...
+  selectedTypeIds,
+  allTypeIds,
+}: {
+  selectedTypeIds: string[];
+  allTypeIds: string[];
+}): boolean =>
+  selectedTypeIds.length > 0 && selectedTypeIds.length < allTypeIds.length;
+
+/**
+ * The stored empty set means "no filter", which the panel shows as every row
```
mdroidian: What is "the stored empty set" referring to?

mdroidian (approved):

PR #1316, opened 2026-08-19
ENG-2158 Publish node type format with schemas

sid597:
image
This PR is the format edge of the diagram: publishing the node type format that both directions depend on.

Roam now publishes a node type's format with its schema row, so an importer can seed a new local type from it (ENG-2156, ENG-2157).

- `CrossAppNodeSchema` gains `format?: string`, mapped to `literal_content.format` in `crossAppNodeSchemaToDbConcept`.
- Both Roam producers write it: `nodeSchemaToCrossApp` (publish) and `discourseNodeSchemaToLocalConcept` (periodic sync). `upsert_concepts` assigns `literal_content` wholesale, so a producer that omits the key erases it.
- Obsidian is unchanged. Its schema parser already reads flat `format` as a fallback.
- No backfill. Old rows get the key on their next re-upsert.

Tests cover both producers, key preservation with a template, and the publish path end to end.

Closes ENG-2158


packages/database/src/crossAppContracts.ts
```diff
@@ -24,6 +24,7 @@ export type CrossAppNodeSchema = CrossAppSchemaBase & {
   label: string;
   template?: string;
   templateTitle?: string;
+  format?: string;
```
sid597, line 28: Flat `literal_content.format`, per the decision on the ticket (MG, team chat 2026-08-19). Obsidian's schema parser already reads flat `format` as a fallback, so the Obsidian producer stays unchanged. A separate key also avoids `template`, which the sync producer already uses for the block template.

packages/database/src/lib/crossAppConverters.ts
```diff
@@ -90,6 +90,7 @@ export const crossAppNodeSchemaToDbConcept = (
   const literalInfo = filterUndefined({
     template: node.templateTitle,
     template_content: node.template,
+    format: node.format,
```
sid597, line 100: The publish path writes only `format` here, no `label` or `template`. That asymmetry with the sync producer is pre-existing: `publishNodesToGroups` only upserts schemas that are not yet in `my_concepts`, and the next periodic sync rewrites the row with the full key set. Unifying the two shapes is out of scope for this ticket.

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -79,22 +79,21 @@ export const discourseNodeSchemaToLocalConcept = (
 ): LocalConceptDataInput => {
   const titleParts = node.text.split("/");
   const label = titleParts[titleParts.length - 1] ?? node.text;
+  const literalContent: { [key: string]: Json } = {
```
sid597: `literal_content` is now built once, then the template branch adds its key. The old shape replaced the whole object in that branch, so every new key had to be written in two places. No behavior change for `label` and `template`.

apps/roam/src/utils/roamToCrossAppConverters.ts
```diff
@@ -200,5 +200,6 @@ export const nodeSchemaToCrossApp = (
     label: s.text,
     authorId: userUid,
     createdAt: new Date(relData[":create/time"] || Date.now()),
+    format: s.format,
```
sid597, line 223: `DiscourseNode.format` is marked deprecated in favor of `specification`, but that deprecation is about node identification. `format` is still the title pattern, and the title pattern is what ENG-2156 and ENG-2157 read from the schema row.

sid597 (commented):
Notes on the non-obvious decisions, one per thread.

sid597:
@codex review

apps/roam/src/utils/roamToCrossAppConverters.ts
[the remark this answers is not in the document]
sid597: Valid finding, deliberately deferred: publishing upserts only schemas missing from my_concepts, and the periodic sync only re-selects types edited since the last watermark, so rows synced before this PR keep a formatless literal_content. Backfilling them is now tracked in ENG-2175 (https://linear.app/discourse-graphs/issue/ENG-2175/backfill-format-on-existing-node-type-schema-rows), following the same client-side pattern as the core_title migration (ENG-2155). Doing it in this PR would also force the literal_content shape question — upsert_concepts replaces the jsonb wholesale and crossAppNodeSchemaToDbConcept omits label — which is scoped there.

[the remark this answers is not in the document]
sid597: Backfill PR is up: #1346, stacked on this one.

maparent (changes requested):
What is there looks good, but now that [ENG-2146](https://linear.app/discourse-graphs/issue/ENG-2146/bulk-convert-some-db-structures-to-crossapp-structures-as-needed-for) has landed, would you mind also handling format in `dbNodeSchemaToCrossApp` ?

mdroidian (approved):
Feel free to merge as is and add a subsequent ticket to handle `dbNodeSchemaToCrossApp`, or handle `dbNodeSchemaToCrossApp` here and re-request review.

If you merge as is, make sure to link to the ticket here.

maparent:
Agreed, can be separate or together, as long as it's noted.

sid597:
Going with merge-as-is: `dbNodeSchemaToCrossApp` format handling is deferred to [ENG-2209](https://linear.app/discourse-graphs/issue/ENG-2209/read-node-type-format-in-dbnodeschematocrossapp). A draft of it briefly landed on this branch and was reverted in ca4b3306.

maparent (approved):

PR #1317, opened 2026-08-19
ENG-2153 Undecorate Roam node titles into core_title on publish

sid597:
https://github.com/user-attachments/assets/599351d6-e35c-466b-9e8b-f2be90871399

image
This PR is the top path of the diagram: undecorate on publish, Roam.

Roam publishes the decorated page title (`[[CLM]] - my claim`) as the node's content value, so an importing app receives Roam's title grammar instead of the content. This PR extracts `{content}` from the title with the node type's format and writes it as `literal_content.core_title` in all three concept producers (publish, periodic sync, full-content). The decorated title stays as the direct content value.

If the type has no format, or the title does not match it, `core_title` equals the title. `coreTitle` is required on `CrossAppNode`: `upsert_concepts` replaces `literal_content` wholesale, so a producer that omits the key would erase `core_title` on the next sync. Required turns that mistake into a compile error.

No DB migration: `literal_content` jsonb already exists (MAP confirmed on the ticket). Tests cover formatted, unformatted, and non-matching titles, a format with `{Source}`, content containing `" - "` (non-greedy shortest match, accepted for v0), and the round trip.

Backfill of existing rows is ENG-2155. Decorating on import is ENG-2156 (Roam) and ENG-2157 (Obsidian).

From review of #1318: `getDiscourseNodeFormatExpression` escapes only `[ ] ? . +`, so a format containing other regex metacharacters (e.g. parens) breaks the placeholder-to-capture-group mapping this extraction relies on; those titles fall back to the decorated title in `core_title`. Cross-app fix (both builders plus placeholder-grammar unification) is tracked in ENG-2176.

The db-to-CrossApp direction for `core_title` is not in this PR. No converter-file consumer exists today: Obsidian reads `core_title` in its import select and decorates with `decorateTitle` (#1330), and Roam's import (ENG-2156) reads it through `sharedNodes.ts`. A `dbToCrossAppConverters.ts` counterpart lands in its own ticket, not here and not in ENG-2156.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1317" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/database/src/crossAppContracts.ts
```diff
@@ -74,6 +74,10 @@ type InlineCrossAppTypedContent = InlineCrossAppContent & {
 // A node instance
 export type CrossAppNode = CrossAppBase & {
   nodeType: LocalId;
+  // The title stripped of the node type's title format ("[[CLM]] - {content}"
+  // -> the {content} part). Equals the title when the type has no format or
+  // the title does not match it.
+  coreTitle: string;
```
sid597, line 82: Required on purpose. `upsert_concepts` replaces `literal_content` wholesale and defaults it to `{}`, so a producer that omits the key erases `core_title` on the next re-upsert. Required makes "every producer writes it" a compile error.

packages/database/src/lib/crossAppConverters.ts
```diff
@@ -75,6 +75,9 @@ export const crossAppNodeToDbConcept = (
     name: node.content.direct.value,
     author_local_id: node.authorId,
     schema_represented_by_local_id: node.nodeType,
+    literal_content: {
+      core_title: node.coreTitle,
```
sid597, line 84: No migration needed: `literal_content` jsonb already exists (MAP confirmed on the ticket). `name` and the direct content keep the decorated title; `core_title` is added next to it.

apps/roam/src/utils/roamToCrossAppConverters.ts
```diff
@@ -15,6 +15,14 @@ import { toMarkdown } from "./pageToMarkdown";
 import getFullTreeByParentUid from "roamjs-components/queries/getFullTreeByParentUid";
 import getPageViewType from "roamjs-components/queries/getPageViewType";
 import { contentTypes } from "@repo/content-model";
+import getDiscourseNodes from "./getDiscourseNodes";
+import extractContentFromTitle from "./extractContentFromTitle";
+
+const getCoreTitle = (title: string, nodeTypeUid: string): string => {
```
sid597: Callers only carry `{uid, type}`, so the format lookup lives in the converter. Calling `getDiscourseNodes()` per node is cheap: the accessor is version-cached.

apps/roam/src/utils/roamToCrossAppConverters.ts
```diff
@@ -73,6 +81,7 @@ export const fullContentNodeToCrossApp = (
     createdAt: new Date(node.created || Date.now()),
     modifiedAt: new Date(node.last_modified || Date.now()),
     nodeType: node.node_type_id,
+    coreTitle: getCoreTitle(title, node.node_type_id),
```
sid597: Nothing reads this `coreTitle` yet: this producer feeds `upsert_content`, not `upsert_concepts`. It is set so any `CrossAppNode` that reaches concept conversion carries it (the ticket lists all three producers).

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -667,11 +668,17 @@ export const convertDgToSupabaseConcepts = async ({
...
 
+  const formatByNodeTypeUid = new Map(
+    allNodeTypes.map((nodeType) => [nodeType.type, nodeType.format]),
+  );
   const nodeBlockToLocalConcepts = nodesSince.map((node) => {
     const localConcept = discourseNodeBlockToLocalConcept(context, {
       nodeUid: node.source_local_id,
       schemaUid: node.type,
       text: node.node_title ? `${node.node_title} ${node.text}` : node.text,
+      coreTitle: extractContentFromTitle(node.node_title ?? node.text, {
```
sid597: Inlined instead of reusing the publish-side helper because `allNodeTypes` is already in scope; the helper would refetch via `getDiscourseNodes()`. `node_title ?? node.text` matters for block-backed types: `node_title` is the format-matched page title, `text` is the block string.

apps/roam/src/utils/roamToCrossAppConverters.ts
```diff
@@ -73,6 +88,7 @@ export const fullContentNodeToCrossApp = (
     createdAt: new Date(node.created || Date.now()),
     modifiedAt: new Date(node.last_modified || Date.now()),
     nodeType: node.node_type_id,
+    coreTitle: getCoreTitle(title, node.node_type_id, formatByNodeTypeUid),
```
sid597: Nothing reads this `coreTitle` yet: this producer feeds `upsert_content`, not `upsert_concepts`. It is set so any `CrossAppNode` that reaches concept conversion carries it (the ticket lists all three producers).

sid597:
Both bot findings addressed in 7d91389: a matched-but-empty `{content}` capture now yields an empty `core_title` instead of falling back to the decorated title, and the node-format lookup is built once per batch (`getFormatByNodeTypeUid`) instead of calling `getDiscourseNodes()` per node.

maparent (changes requested):
LGTM but same idea: Please also handle `dbToCrossAppConverters.ts`

mdroidian:
> LGTM but same idea: Please also handle `dbToCrossAppConverters.ts`

@maparent Could you clarify what you mean here?

apps/roam/src/utils/convertRoamNodeToFullContent.ts
```diff
@@ -16,10 +19,11 @@ export const convertRoamNodeToFullContent = ({
...
 }: {
   nodes: RoamFullContentNode[];
-}): LocalContentDataInput[] =>
-  nodes.flatMap((node) => {
+}): LocalContentDataInput[] => {
+  const formatByNodeTypeUid = getFormatByNodeTypeUid();
+  return nodes.flatMap((node) => {
     try {
-      const crossAppNode = fullContentNodeToCrossApp(node);
+      const crossAppNode = fullContentNodeToCrossApp(node, formatByNodeTypeUid);
```
mdroidian: `getFormatByNodeTypeUid()` is a second lookup. These nodes are built in `getSharedRoamNodesWithFullContentUpdatesSince`, and `matchingNodeType.format` is already there when you set `node_type_id`.

Could we put `format` on `RoamFullContentNode` at that construction site and skip the map here?

mdroidian (approved):

maparent:
This PR is adding a field; it should handle that field in both translation direction (dbToCrossApp as well as crossAppToDb)

mdroidian:
Is that in this tasks scope or in the scope of [ENG-2156](https://linear.app/discourse-graphs/issue/ENG-2156/decorate-imported-node-titles-in-roam-from-core-title)?

apps/roam/src/utils/convertRoamNodeToFullContent.ts
sid597: Done in cc5370ea: `format` now rides `RoamFullContentNode` from the construction site in `getSharedRoamNodesWithFullContentUpdatesSince`, and the map lookup is gone.

sid597:
Added the reverse direction in 8cedea1f: `dbNodeToCrossApp` / `dbNodesToCrossApp` convert a Concept row back into a `CrossAppNode`, reading `literal_content.core_title` into `coreTitle` (falling back to `name` for rows written before this field existed). Also merged main to resolve the conflicts with the slots work and the `nodeSchemaToCrossApp` `modifiedAt` fix.

sid597:
Correction to my earlier comment: the reverse-direction converters are reverted (00aced39, 09a277ff) to keep this PR at the reviewed shape. `dbToCrossAppConverters.ts` has no node-instance consumer today: Obsidian reads `core_title` in its import select plus `decorateTitle` (#1330), and Roam's import (ENG-2156) reads it through `sharedNodes.ts`. So the db-to-CrossApp handling of `core_title` moves to its own ticket rather than this PR or ENG-2156. Remaining changes since the last review round: `format` carried on `RoamFullContentNode` from its construction site (mdroidian's comment) and the merge of main.

maparent (approved):

PR #1318, opened 2026-08-19
ENG-2154 Undecorate Obsidian node titles into core_title on publish

sid597:
https://github.com/user-attachments/assets/76bbd00a-f12b-410a-acb5-f70314737743

image
This PR is the top path of the diagram: undecorate on publish, Obsidian.

Obsidian publishes the decorated file name, so importers receive Obsidian's title grammar. This writes the undecorated content as `literal_content.core_title` on publish (contract from ENG-2153).

- `discourseNodeInstanceToLocalConcept` extracts `{content}` from the file basename with the node type's format and writes it as `core_title`. The existing keys stay.
- A type without a format, or a non-matching basename, gets `core_title` equal to the basename. Importers fall back to the direct content text when the key is absent, so old rows stay valid. Backfill is ENG-2155.
- The key is written on every upsert because `upsert_concepts` replaces `literal_content` wholesale. Publish and periodic sync both flow through this one converter.

No tests: apps/obsidian has no test infrastructure and we decided not to add it in this PR; ENG-2176 moves the extractor where a test runner exists.

From review:

- `extractContentFromTitle` now returns the `{content}` capture by placeholder position and preserves a matched-but-empty capture as `""`. The format's placeholders map 1:1 to the expression's capture groups, so the capture index is the placeholder index. This replaces the ticket's accepted first-capture limitation: ENG-2155's backfill treats any non-null `core_title` as already set, so a wrong value written before a later fix would never be repaired — cheaper to fix before rows are written. Same extraction rule as the Roam side (#1317), with one asymmetry tracked in ENG-2176: Roam's placeholder scan (`{[\w\d-]+}`) is wider than its builder's grammar, Obsidian's matches its builder exactly.
- The change is publish-path-only: the two pre-existing callers (QueryEngine) store the result in `BulkImportCandidate.extractedContent`, which nothing reads.
- `discourseNodeInstanceToLocalConcept` takes named parameters, matching the sibling converters in the file and the AGENTS.md rule.
- The unused `accountLocalId` parameter of `convertDgToSupabaseConcepts` is removed — pre-existing on main, removed here because this PR touches the signature and `eslint --max-warnings 0` flags the file.
- Formats containing regex metacharacters the expression builder does not escape still mis-extract (fallback to the decorated basename); deferred to ENG-2176 since the same gap exists in Roam's builder, so it is one cross-app fix. ENG-2176 also covers unifying the three copies of this transform (Roam, here, and `DiscourseContextView`'s inline copy — which can render a different Content string than what gets published) into a shared, testable package.

Notes for review:

- A format without `{content}` publishes the first placeholder's capture (format validation only requires some `{var}`) — same rule as Roam.
- `core_title` recomputes when a file's title changes and on full sync. A format edit alone does not rewrite existing rows: it also does not rename existing files, so stored values stay consistent with the basenames they came from.
- `label` stays alongside `core_title`: `label` is the decorated basename, `core_title` the undecorated content — an importer cannot undecorate a foreign app's title grammar, so both cross the boundary.


apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -160,12 +161,17 @@ export const discourseRelationTripleSchemaToLocalConcept = ({
 export const discourseNodeInstanceToLocalConcept = (
   context: SupabaseContext,
   nodeData: ObsidianDiscourseNodeData,
+  nodeTypesById: Record<string, DiscourseNode>,
```
sid597: The converter only had the type id, and the format lives on the type. The single caller already builds this map for `discourseRelationTripleSchemaToLocalConcept`, so it comes in the same way.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -160,12 +161,17 @@ export const discourseRelationTripleSchemaToLocalConcept = ({
...
 ): LocalConceptDataInput => {
   const extraData = getNodeExtraData(nodeData.file, context.userId);
   const { nodeInstanceId, nodeTypeId, importedFromRid, ...otherData } =
     nodeData.frontmatter;
   const literal_content: Record<string, Json> = {
     label: nodeData.file.basename,
+    core_title: extractContentFromTitle(
+      nodeTypesById[nodeData.nodeTypeId]?.format ?? "",
+      nodeData.file.basename,
+    ),
```
sid597, line 178: `extractContentFromTitle` already returns the title when the format is empty or does not match, which is the fallback ENG-2153 specifies. The `?? ""` covers a file whose `nodeTypeId` no longer matches a configured type.

The write is unconditional on purpose: `upsert_concepts` replaces `literal_content` wholesale, so a conditional write would let a later sync erase the key.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -160,12 +161,17 @@ export const discourseRelationTripleSchemaToLocalConcept = ({
 export const discourseNodeInstanceToLocalConcept = (
   context: SupabaseContext,
   nodeData: ObsidianDiscourseNodeData,
+  nodeTypesById: Record<string, DiscourseNode>,
 ): LocalConceptDataInput => {
```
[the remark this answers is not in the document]
sid597: Done in 7fe37bb1 — named parameters, matching the sibling converters.

apps/obsidian/src/utils/conceptConversion.ts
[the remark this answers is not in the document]
sid597: Deliberate, per the PR description: apps/obsidian has no test runner and we decided not to add the infrastructure in this PR, so the ticket's test bullet is deferred. The extraction semantics were verified against six format/title cases during review.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -160,12 +161,17 @@ export const discourseRelationTripleSchemaToLocalConcept = ({
...
   const extraData = getNodeExtraData(nodeData.file, context.userId);
   const { nodeInstanceId, nodeTypeId, importedFromRid, ...otherData } =
     nodeData.frontmatter;
   const literal_content: Record<string, Json> = {
     label: nodeData.file.basename,
+    core_title: extractContentFromTitle(
+      nodeTypesById[nodeData.nodeTypeId]?.format ?? "",
+      nodeData.file.basename,
```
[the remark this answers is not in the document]
sid597: Valid — deferred to ENG-2176 (https://linear.app/discourse-graphs/issue/ENG-2176/escape-regex-metacharacters-in-discourse-node-format-expressions). The same unescaped-metacharacter gap exists in Roam's getDiscourseNodeFormatExpression, and the builder has five call sites here plus Roam's matchers, so it's one cross-app fix in its own PR. Until then an affected format publishes core_title equal to the decorated basename.

apps/obsidian/src/utils/extractContentFromTitle.ts
```diff
@@ -1,10 +1,25 @@
...
   const regex = getDiscourseNodeFormatExpression(format);
-  const match = title.match(regex);
+  const match = regex.exec(title);
+  if (!match) return title;
 
-  return match?.[1]?.trim() || title;
+  const contentIndex = placeholders.findIndex(
+    (name) => name.toLowerCase() === "content",
+  );
+  const capture = contentIndex >= 0 ? match[contentIndex + 1] : match[1];
```
sid597, line 22: getDiscourseNodeFormatExpression turns every `{[a-zA-Z]+}` placeholder into one `(.*?)` group, in order, and the scan above uses the same pattern, so the capture index is the placeholder index + 1. A matched-but-empty capture stays `""` on purpose: the old `|| title` fallback wrote the decorated basename as core_title, a non-null value ENG-2155's backfill probe would treat as already migrated. Formats with metacharacters the expression builder does not escape still fall back to the basename — ENG-2176.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -157,15 +158,24 @@ export const discourseRelationTripleSchemaToLocalConcept = ({
...
+  nodeData: ObsidianDiscourseNodeData;
+  nodeTypesById: Record<string, DiscourseNode>;
+}): LocalConceptDataInput => {
   const extraData = getNodeExtraData(nodeData.file, context.userId);
   const { nodeInstanceId, nodeTypeId, importedFromRid, ...otherData } =
     nodeData.frontmatter;
   const literal_content: Record<string, Json> = {
     label: nodeData.file.basename,
+    core_title: extractContentFromTitle(
+      nodeTypesById[nodeData.nodeTypeId]?.format ?? "",
```
sid597, line 176: `nodeData.nodeTypeId` is the typed copy of `frontmatter.nodeTypeId` (same value; `getDiscourseNodes` copies it out). The destructured `nodeTypeId` on line 171 exists to strip the key from `otherData` before it becomes `source_data`, and line 187 keeps using it with the pre-existing `as string` — switching that to `nodeData.nodeTypeId` would drop an assertion, left as is to avoid churn on a PR under review.

maparent (approved):
LGTM

mdroidian (approved):

PR #1329, opened 2026-08-22
ENG-2128 Express source as a `sourceDocument` slot/reference of Evidence

maparent:
https://linear.app/discourse-graphs/issue/ENG-2128/express-source-as-a-sourcedocument-slotreference-of-evidence

code walkthrough: https://www.loom.com/share/aacdf3b6a2cb4abe9e7063026f67d521

demo: https://www.loom.com/share/4e6249bca9754bfea7159c6a078e9c02

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1329" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/utils/conceptConversion.ts
+++ b/apps/roam/src/utils/conceptConversion.ts
@@ -109,15 +109,11 @@ export const discourseNodeBlockToLocalConcept = (
   {
     nodeUid,
     schemaUid,
-    text,
     title,
     schema,
   }: {
     nodeUid: string;
     schemaUid: string;
-    text: string;
-    // The node's title, which is where a {source} placeholder is filled in. It differs
-    // from text for a node whose text lives in a block below its page.
     title?: string;
     schema?: DiscourseNode;
   },
@@ -125,7 +121,7 @@ export const discourseNodeBlockToLocalConcept = (
   const sourceUid = title ? sourceUidOfNode(title, schema) : undefined;
   return {
     space_id: context.spaceId,
-    name: text,
+    name: title,
     source_local_id: nodeUid,
     schema_represented_by_local_id: schemaUid,
     is_schema: false,
--- a/apps/roam/src/utils/conceptConversion.ts
+++ b/apps/roam/src/utils/conceptConversion.ts
@@ -1,6 +1,12 @@
 import { InputTextNode } from "roamjs-components/types";
 import getBlockProps from "./getBlockProps";
 import { DiscourseNode } from "./getDiscourseNodes";
+import {
+  SOURCE_SLOT,
+  schemaHasSourceSlot,
+  sourceSlotSchemaId,
+  sourceUidOfNode,
+} from "./sourceSlot";
 import getDiscourseRelations from "./getDiscourseRelations";
 import type { DiscourseRelation } from "./getDiscourseRelations";
 import type { SupabaseContext } from "~/utils/supabaseContext";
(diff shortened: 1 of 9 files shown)


mdroidian:
@codex review

apps/roam/src/utils/sourceSlot.ts
```diff
@@ -0,0 +1,67 @@
...
+// reference. Better to leave the slot out than to fill it with that.
+//
+// Which node type it is, we do not check, and that leniency is part of the hack: the
+// Source type is recognised here by being named "source", and a graph coming from
+// another app may well name it otherwise. Rejecting anything but a local "Source" node
+// would silently drop those. This goes away with slots as a real node type setting.
+const isDiscourseNodeTitle = (
+  title: string,
+  allNodes: DiscourseNode[],
+): boolean => allNodes.some((node) => matcherFor(node.format).test(title));
```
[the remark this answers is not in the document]
mdroidian: @maparent

apps/roam/src/utils/sourceSlot.ts
```diff
@@ -0,0 +1,67 @@
...
+// The page a node's {source} placeholder resolves to, when there is one. The
+// placeholder is usually filled with a page reference, and a title holding a slash is
+// a namespaced page rather than a source, so it is left alone.
+export const sourceUidOfNode = (
+  title: string,
+  schema: NodeFormat | undefined,
+  allNodes?: DiscourseNode[],
+): string | undefined => {
+  if (!schemaHasSourceSlot(schema)) return undefined;
+  const sourceTitle = extractFieldFromTitle(title, schema!, "source")
```
mdroidian: Avoid `!` whenever possible.  If not possible, please add a comment explaining why it is not possible.

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -104,18 +110,28 @@ export const discourseNodeBlockToLocalConcept = (
     nodeUid,
     schemaUid,
     text,
+    title,
+    schema,
   }: {
     nodeUid: string;
     schemaUid: string;
     text: string;
+    // The node's title, which is where a {source} placeholder is filled in. It differs
```
mdroidian: Can you give me an example of the problem this is solving?

mdroidian (changes requested):
See the codex comment, let's make sure we default Page and Block matchers.
And there's a couple other comments we should address before merging.

apps/roam/src/utils/conceptConversion.ts
maparent: Discussed in meeting; removed text.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -667,11 +667,16 @@ export const convertDgToSupabaseConcepts = async ({
...
 
+  const schemasByUid = new Map(
+    allNodeTypes.map((nodeType) => [nodeType.type, nodeType]),
+  );
+
   const nodeBlockToLocalConcepts = nodesSince.map((node) => {
     const localConcept = discourseNodeBlockToLocalConcept(context, {
       nodeUid: node.source_local_id,
       schemaUid: node.type,
-      text: node.node_title ? `${node.node_title} ${node.text}` : node.text,
```
mdroidian, line 674: It looks like this breaks concept syncing for "page-backed nodes" (read regular nodes). The page query stores the page title in `node.text` and does not set `node.node_title`, so this passes `undefined`. That becomes an undefined concept `name`, but `Concept.name` is required, causing the upsert to fail with `-2`.

Please restore the fallback:

```ts
title: node.node_title ?? node.text,
```

For page-backed nodes, `node.text` is the page title. For "block-backed nodes" (read nodes that include an embedding block ref), `node.node_title` is the containing page’s title. 

We should probably add a test that covers both cases.

mdroidian (approved):
Approving this but the requested change is blocking and requires fixing.  If you choose to go with code other than `title: node.node_title ?? node.text,`, please re-request a review.

apps/roam/src/utils/syncDgNodesToSupabase.ts
maparent: Ah, good catch. I looked at getAllDiscourseNodesSince but not getDiscourseNodeTypeWithSettingsBlockNodes.
I think this is worth a refactor, where node-title is always the title, and optional embed-text is the block if different from the title.
But that's a different scope.

maparent: (And I prefer that to a test that would tend to enshrine the current inconsistency)

PR #1330, opened 2026-08-23
ENG-2157 Decorate imported node titles in Obsidian from core_title

sid597:
https://github.com/user-attachments/assets/71c3b7a5-54f2-4312-8c68-ad2889afa6e3

image
This PR is the bottom path of the diagram: decorate on import, Obsidian.

Obsidian's import used the incoming title verbatim, so a node published from Roam arrived carrying `[[CLM]] - ` and ignored what the local vault calls that node type. This rebuilds the file name from `core_title` with the local node type's format, so the same node lands as `CLM - sleep improves memory` in a vault that formats claims that way. Fallback is the incoming title, used when `core_title` is absent (rows published before ENG-2153/2154), the local type has no format, or the format cannot be rebuilt from the core title alone (no `{content}` placeholder, or a second placeholder such as `{Source}` whose value the database does not hold). That last case is why Roam's shipped `[[EVD]] - {content} - {Source}` keeps the incoming title: substituting the empty string dropped the source from a name that used to carry it and left a dangling separator.

Three pieces:

- The instance query in `importNodes.ts` reads the single JSON key by path (`core_title:literal_content->>core_title`), so the import-preview path that shares it stays free of the payload column. postgrest-js types that projection as `string`, but it is null for rows published before ENG-2153, so the `?? undefined` is load-bearing.
- Local node type resolution moves out of `processFileContent` into the per-node loop, because the local format has to be known before the file name is built. `processFileContent` now only writes the file and frontmatter. Its only error came from that resolution, so it returns a `TFile` instead of a result union.
- The name build applies the format with `decorateTitle`, a pure helper in `packages/database/src/lib` because ENG-2156 needs the same transform in Roam. It returns `null` when the format cannot be rebuilt, and the caller falls back. I didn't reuse `formatNodeName`: it splits the compiled regex source on the first `(.*?)`, so a multi-placeholder format produces a name ending in a literal `(.*?)`. The placeholder pattern is exported as `FORMAT_PLACEHOLDER` and `getDiscourseNodeFormatExpression` now uses it, so decorate and match cannot drift apart.

`decorateTitle` is deterministic and one `sanitizedFileName` feeds both the create path and the rename guard, so once a file carries the decorated name, re-import and refresh leave it alone. Files imported before this change carry the incoming title, so their next refresh renames them once (and rewrites inbound wikilinks); that is the intended migration. A related fix falls out: Obsidian-origin nodes that lived in a subfolder used to be created at their full source path, skipping the decorated name, and a later refresh would then rename them. The create path now keeps the folders and replaces the last segment.

Merge order: nothing on `main` writes `core_title` yet (#1317 / #1318 are open). Merged alone, every `coreTitle` is undefined and the import keeps today's names; the decoration in the demo needs those producers. The subfolder create-path fix is live regardless.

Refactor parity (what moved and what changed on the way):

| Old | New | Parity |
|---|---|---|
| `fetchNodeTypeSchemasForInstances` → `Map<id, Schema>` | `fetchNodeImportInfoForInstances` → `Map<id, {schema?, coreTitle?}>` | changed: an entry per visible instance, not only schema-resolved ones |
| early return on empty `schemaIds` / schema-fetch error | fall through with `schema: undefined` | changed: core titles survive a schema-fetch failure |
| node-type derivation inside `processFileContent` | same code in the `importSelectedNodes` loop | identical logic; errors go to `console.error` + `continue` instead of a result union |
| `mapNodeTypeIdToLocal` after the vault write | before the vault write | changed: a failed import can leave a created node type behind |
| `processFileContent` → `{file} \| {error}` | → `TFile` | changed: the one error source moved out, the caller's `result.file!` is gone |
| create path `sanitizePathForImport(contentFilePath)` | source folder + derived basename | changed: create and refresh agree on the name |

Tests for the helper live in `packages/database` (apps/obsidian has no test runner), so the ticket's test bullet is covered for the pure part and the `importNodes` wiring is not. That part is on the demo.

Known and deferred, each pre-existing and left as it was:

- `core_title` is a bare `literal_content` key on both sides: `CrossAppNode.coreTitle` is the contract field (ENG-2153), the serialized key has no shared constant. The producers live on #1317/#1318 and this PR is based on `main`, so the constant is a follow-up once those land.
- `formatNodeName` (local node creation) and `decorateTitle` (import) now disagree on multi-placeholder formats; converging them needs a ticket because `formatNodeName`'s `null` doubles as form validation.
- `mapNodeTypeIdToLocal` runs one `my_concepts` query per imported node, as it did inside `processFileContent` before; batching it is not this ticket.
- A type created from a Roam schema keeps Roam's `[[CLM]] - {content}` format without going through `checkInvalidChars`, so it decorates as `[[CLM]] - x.md`; the incoming title carried the same brackets before this change.
- The create path has never allocated a unique name when the target file exists; decorated names are unique whenever the incoming ones are (one placeholder, distinct type prefixes), so this change does not widen that gap.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1330" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/obsidian/src/utils/importNodes.ts
+++ b/apps/obsidian/src/utils/importNodes.ts
@@ -21,6 +21,8 @@ import {
 } from "./importRelations";
 import { createTemplateFile } from "./templates";
 import { resolveFolderForSpaceUri } from "./importFolderMetadata";
+import { getNodeTypeById } from "./typeUtils";
+import { decorateTitle } from "@repo/database/lib/decorateTitle";
 
 type PublishedNode = {
   source_local_id: string;
@@ -327,7 +329,12 @@ type NodeTypeSchemaForInstance = {
   name: string;
 };
 
-export const fetchNodeTypeSchemasForInstances = async ({
+type NodeInstanceImportInfo = {
+  schema?: NodeTypeSchemaForInstance;
+  coreTitle?: string;
+};
+
+export const fetchNodeImportInfoForInstances = async ({
   client,
   spaceId,
   nodeInstanceIds,
@@ -335,12 +342,14 @@ export const fetchNodeTypeSchemasForInstances = async ({
   client: DGSupabaseClient;
   spaceId: number;
   nodeInstanceIds: string[];
-}): Promise<Map<string, NodeTypeSchemaForInstance>> => {
-  const result = new Map<string, NodeTypeSchemaForInstance>();
+}): Promise<Map<string, NodeInstanceImportInfo>> => {
+  const result = new Map<string, NodeInstanceImportInfo>();
 
   const { data: instanceRows, error: instanceError } = await client
     .from("my_concepts")
-    .select("source_local_id, schema_id")
+    .select(
+      "source_local_id, schema_id, core_title:literal_content->>core_title",
+    )
     .eq("space_id", spaceId)
     .eq("is_schema", false)
     .eq("is_relation", false)
@@ -358,35 +367,42 @@ export const fetchNodeTypeSchemasForInstances = async ({
         .filter((id): id is number => id !== null),
     ),
   ];
-  if (schemaIds.length === 0) return result;
-
-  const { data: schemaRows, error: schemaError } = await client
-    .from("my_concepts")
-    .select("id, source_local_id, name")
-    .eq("space_id", spaceId)
-    .eq("is_schema", true)
-    .eq("is_relation", false)
-    .in("id", schemaIds);
-
-  if (schemaError || !schemaRows) {
-    console.error("Error fetching node type schemas:", schemaError);
-    return result;
-  }
 
   const schemasById = new Map<number, NodeTypeSchemaForInstance>();
-  for (const row of schemaRows) {
-    if (row.id !== null && row.source_local_id !== null && row.name !== null) {
-      schemasById.set(row.id, {
-        nodeTypeId: row.source_local_id,
-        name: row.name,
-      });
+  if (schemaIds.length > 0) {
+    const { data: schemaRows, error: schemaError } = await client
+      .from("my_concepts")
+      .select("id, source_local_id, name")
+      .eq("space_id", spaceId)
+      .eq("is_schema", true)
+      .eq("is_relation", false)
+      .in("id", schemaIds);
+
+    if (schemaError || !schemaRows) {
+      console.error("Error fetching node type schemas:", schemaError);
+    } else {
+      for (const row of schemaRows) {
+        if (
+          row.id !== null &&
+          row.source_local_id !== null &&
+          row.name !== null
+        ) {
+          schemasById.set(row.id, {
+            nodeTypeId: row.source_local_id,
+            name: row.name,
+          });
+        }
+      }
     }
   }
 
   for (const row of instanceRows) {
-    if (row.source_local_id === null || row.schema_id === null) continue;
-    const schema = schemasById.get(row.schema_id);
-    if (schema) result.set(row.source_local_id, schema);
+    if (row.source_local_id === null) continue;
+    result.set(row.source_local_id, {
+      schema:
+        row.schema_id === null ? undefined : schemasById.get(row.schema_id),
+      coreTitle: row.core_title ?? undefined,
+    });
   }
 
   return result;
@@ -1159,8 +1175,6 @@ export const mapNodeTypeIdToLocal = async ({
 
 const processFileContent = async ({
   plugin,
-  client,
-  sourceSpaceId,
   sourceSpaceUri,
   rawContent,
   filePath,
@@ -1168,11 +1182,9 @@ const processFileContent = async ({
   importedModifiedAt,
   authorId,
   nodeInstanceId,
-  nodeTypeIdFromConcept,
+  nodeTypeId,
 }: {
   plugin: DiscourseGraphPlugin;
-  client: DGSupabaseClient;
-  sourceSpaceId: number;
   sourceSpaceUri: string;
   rawContent: string;
   filePath: string;
@@ -1180,25 +1192,9 @@ const processFileContent = async ({
   importedModifiedAt?: number;
   authorId?: number;
   nodeInstanceId: string;
-  nodeTypeIdFromConcept?: string;
-}): Promise<
-  { file: TFile; error?: never } | { file?: never; error: string }
-> => {
-  // 1. Parse frontmatter from rawContent (metadataCache is updated async and is
-  //    often empty immediately after create/modify) and resolve the node type
-  //    before any vault write, so a failed lookup leaves existing files untouched.
-  const { frontmatter } = parseFrontmatter(rawContent);
-  const sourceNodeTypeId =
-    typeof frontmatter.nodeTypeId === "string"
-      ? frontmatter.nodeTypeId
-      : nodeTypeIdFromConcept;
-  if (sourceNodeTypeId === undefined) {
-    return {
-      error: "importedNode missing sourceNodeTypeId",
-    };
-  }
-
-  // 2. Create or update the file with the fetched content.
+  nodeTypeId: string;
+}): Promise<TFile> => {
+  // Create or update the file with the fetched content.
   // On create, set file metadata (ctime/mtime) to original vault dates via vault adapter.
   let file: TFile | null = plugin.app.vault.getFileByPath(filePath);
   const stat =
@@ -1214,19 +1210,11 @@ const processFileContent = async ({
     await plugin.app.vault.process(file, () => rawContent, stat);
   }
 
-  const mappedNodeTypeId = await mapNodeTypeIdToLocal({
-    plugin,
-    client,
-    sourceSpaceId,
-    sourceSpaceUri,
-    sourceNodeTypeId,
-  });
-
   await plugin.app.fileManager.processFrontMatter(
     file,
     (fm) => {
       const record = fm as Record<string, unknown>;
-      record.nodeTypeId = mappedNodeTypeId;
+      record.nodeTypeId = nodeTypeId;
       record.nodeInstanceId = nodeInstanceId;
       record.importedFromRid = spaceUriAndLocalIdToRid(
         sourceSpaceUri,
@@ -1239,7 +1227,7 @@ const processFileContent = async ({
     stat,
   );
 
-  return { file };
+  return file;
 };
 
 export const importSelectedNodes = async ({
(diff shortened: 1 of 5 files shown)


apps/obsidian/src/utils/importNodes.ts
```diff
@@ -327,20 +329,27 @@ type NodeTypeSchemaForInstance = {
   name: string;
 };
 
-export const fetchNodeTypeSchemasForInstances = async ({
+type NodeInstanceImportInfo = {
```
sid597, line 332: `NodeInstanceImportInfo` nests the existing `NodeTypeSchemaForInstance` instead of flattening its fields. `nodeTypeId` and `name` only ever get set together from one schema row, so two independent optional fields would describe a state this code can't produce. The map now holds an entry for every visible instance rather than only the schema-resolved ones. That's what lets a core title reach the caller when the schema lookup comes back empty.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -327,20 +329,27 @@ type NodeTypeSchemaForInstance = {
...
-}): Promise<Map<string, NodeTypeSchemaForInstance>> => {
-  const result = new Map<string, NodeTypeSchemaForInstance>();
+}): Promise<Map<string, NodeInstanceImportInfo>> => {
+  const result = new Map<string, NodeInstanceImportInfo>();
 
   const { data: instanceRows, error: instanceError } = await client
     .from("my_concepts")
-    .select("source_local_id, schema_id")
+    .select(
+      "source_local_id, schema_id, core_title:literal_content->>core_title",
```
sid597, line 351: This reads the single JSON key by path instead of selecting the whole `literal_content` column. `computeImportPreview` shares this query and only wants node type names, so the preview path stays free of the jsonb payload. The key is the bare `core_title` that ENG-2153 writes.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -358,35 +367,42 @@ export const fetchNodeTypeSchemasForInstances = async ({
...
-  }
 
   const schemasById = new Map<number, NodeTypeSchemaForInstance>();
-  for (const row of schemaRows) {
-    if (row.id !== null && row.source_local_id !== null && row.name !== null) {
-      schemasById.set(row.id, {
-        nodeTypeId: row.source_local_id,
-        name: row.name,
-      });
+  if (schemaIds.length > 0) {
```
sid597, line 372: The schema fetch moved inside an `if` instead of guarding two early returns. Both former early exits, no schema ids and a failed schema fetch, now fall through to the merge loop so core titles survive them. On a schema error the caller gets titles with no `nodeTypeId`, which it already guards for.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -358,35 +367,42 @@ export const fetchNodeTypeSchemasForInstances = async ({
...
+      }
     }
   }
 
   for (const row of instanceRows) {
-    if (row.source_local_id === null || row.schema_id === null) continue;
-    const schema = schemasById.get(row.schema_id);
-    if (schema) result.set(row.source_local_id, schema);
+    if (row.source_local_id === null) continue;
+    result.set(row.source_local_id, {
```
sid597, line 401: One write per instance instead of two passes over `instanceRows`. `schema` is genuinely undefined here when `schema_id` is null or the schema row isn't visible under RLS.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -358,35 +367,42 @@ export const fetchNodeTypeSchemasForInstances = async ({
...
+      schema:
+        row.schema_id === null ? undefined : schemasById.get(row.schema_id),
+      coreTitle: row.core_title ?? undefined,
```
sid597, line 404: The `?? undefined` looks redundant because postgrest-js types a `->>` projection as plain `string`. It doesn't model a JSON path extraction as nullable, and this one is. The key is absent on every row published before ENG-2153, so Postgres returns NULL and PostgREST sends `null`. I pinned the declared type down with a temporary `Exact<A, B>` assertion before relying on it, so the guard stays.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1159,46 +1175,26 @@ export const mapNodeTypeIdToLocal = async ({
...
-      : nodeTypeIdFromConcept;
-  if (sourceNodeTypeId === undefined) {
-    return {
-      error: "importedNode missing sourceNodeTypeId",
-    };
-  }
-
-  // 2. Create or update the file with the fetched content.
+  nodeTypeId: string;
+}): Promise<TFile> => {
```
sid597, line 1196: `processFileContent` returns a `TFile` instead of a result union. Its only error came from the node type derivation, which moved up to the caller. That left the error branch and the caller's `result.file!` unreachable, so both are gone.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1355,19 +1343,58 @@ export const importSelectedNodes = async ({
...
-        // Sanitize file name
-        const sanitizedFileName = sanitizeFileName(fileName);
+        const nodeImportInfo = nodeImportInfoByInstance.get(
+          node.nodeInstanceId,
+        );
+
+        // Parse frontmatter from content (metadataCache is updated async and is
+        // often empty immediately after create/modify) and resolve the node type
+        // before any vault write, so a failed lookup leaves existing files untouched.
+        const { frontmatter } = parseFrontmatter(content);
```
sid597, line 1353: This derivation moved up from `processFileContent`, and its comment came with it. We parse the raw content rather than `metadataCache` because the cache is often empty right after a write. Resolving before any vault write still means a failed lookup leaves existing files untouched.

sid597 (commented):
Annotations on the non-obvious decisions in this diff.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1355,19 +1343,58 @@ export const importSelectedNodes = async ({
...
+            `Error processing file content for node ${node.nodeInstanceId}:`,
+            "importedNode missing sourceNodeTypeId",
+          );
+          failedCount++;
+          processedCount++;
+          onProgress?.(processedCount, totalNodes);
+          continue;
+        }
+
+        const mappedNodeTypeId = await mapNodeTypeIdToLocal({
```
sid597, line 1369: `mapNodeTypeIdToLocal` now runs before the file is written rather than after. That's the point of the move, since the local node type has to exist before we can read its format. The function itself is unchanged. One consequence worth knowing: it can create a node type as a side effect, so a failure later in the loop can leave a type behind for a node that didn't import.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1355,19 +1343,58 @@ export const importSelectedNodes = async ({
...
+          plugin,
+          client,
+          sourceSpaceId: spaceId,
+          sourceSpaceUri: spaceUri,
+          sourceNodeTypeId,
+        });
+
+        const localNodeType = getNodeTypeById(plugin, mappedNodeTypeId);
+        const coreTitle = nodeImportInfo?.coreTitle;
+        const titleForFileName =
```
sid597: The decoration. We rebuild the title from `core_title` using the local type's format, so a Roam-origin `[[CLM]] - x` lands as whatever this vault calls a claim. `decorateTitle` returns `null` when the format cannot be rebuilt from the core title alone: no `{content}` placeholder, or a second placeholder such as `{Source}` that the database has no value for. Roam's shipped Evidence format is that case, so an Evidence note keeps its incoming title rather than landing as `[[EVD]] - x -` with the source dropped. `decorateTitle` is pure, so a re-import or a refresh computes the same name and the rename guard below leaves the file alone.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1355,19 +1343,58 @@ export const importSelectedNodes = async ({
...
           // Update existing file - use its current path
           finalFilePath = existingFile.path;
         } else {
           // Preserve source vault folder structure under import/{vaultName} when we have filePath from Content
-          const pathUnderImport =
+          const sourceFolder =
             contentFilePath && contentFilePath.includes("/")
-              ? sanitizePathForImport(contentFilePath)
-              : `${sanitizedFileName}.md`;
+              ? sanitizePathForImport(contentFilePath.replace(/\/[^/]*$/, ""))
```
sid597, line 1393: Obsidian-origin nodes that lived in a subfolder used to be created at their full source path, which skipped the decorated name entirely. A later refresh would then compare against the decorated basename and rename the file. Keeping the folders but replacing the last segment makes create and refresh agree.

apps/obsidian/src/utils/importPreview.ts
```diff
@@ -82,13 +82,16 @@ export const computeImportPreview = async ({
...
-    const nodeTypeSchemasByInstance = await fetchNodeTypeSchemasForInstances({
+    const nodeImportInfoByInstance = await fetchNodeImportInfoForInstances({
       client,
       spaceId,
       nodeInstanceIds: nodes.map((n) => n.nodeInstanceId),
     });
 
-    for (const { nodeTypeId, name } of nodeTypeSchemasByInstance.values()) {
+    for (const { schema } of nodeImportInfoByInstance.values()) {
+      if (!schema) continue;
```
sid597, line 92: Preview only wants node type names, so it skips entries whose schema didn't resolve. One check instead of two, since nesting makes the co-presence of `nodeTypeId` and `name` structural.

apps/obsidian/src/utils/importNodes.ts
```diff
@@ -1355,19 +1343,58 @@ export const importSelectedNodes = async ({
...
+              : "";
+          const pathUnderImport = sourceFolder
+            ? `${sourceFolder}/${sanitizedFileName}.md`
+            : `${sanitizedFileName}.md`;
```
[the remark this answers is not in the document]
sid597: Confirmed, and the case you describe is closed by the placeholder gate in 94ea2ded: a format with a placeholder other than `{content}` no longer decorates, so two nodes can only share a derived name when their incoming titles would have collided too (one placeholder, distinct type prefixes, unique source titles). The create path has never allocated a unique name when the target exists; that gap predates this PR and is listed in the description as deferred rather than fixed here.

maparent (approved):

maparent:
I agree that not changing formats of incoming schemas is the right call in this scope, but maybe this is worth calling out in another task?

mdroidian:
> I agree that not changing formats of incoming schemas is the right call in this scope, but maybe this is worth calling out in another task?

[ENG-2195: Normalize imported node type formats for Obsidian filenames](https://linear.app/discourse-graphs/issue/ENG-2195/normalize-imported-node-type-formats-for-obsidian-filenames)

mdroidian (approved):

PR #1331, opened 2026-08-23
ENG-2156 Decorate imported node titles in Roam from core_title

sid597:
https://github.com/user-attachments/assets/4de08041-0a38-49a8-9174-9bef1ad60c7c

image
This PR is the bottom path of the diagram: decorate on import, Roam.

Imported nodes used to land in Roam with the source's title verbatim, so an Obsidian `CLM - This is a test Claim` became a page Roam's format regex didn't recognize as a `[[CLM]]`. The importer now rebuilds the title from the publisher's `core_title` with the local node type's format, and falls back to the incoming title whenever either piece is missing or the format cannot be rebuilt from the core title alone (`decorateTitle` returns `null` for a format without `{content}` or with a second placeholder such as `{Source}`). That is why this graph's Evidence type keeps Obsidian's title in the demo: substituting an empty source produced `[[EVD]] - x -`, which Roam's own format regex does not match once the trailing space is trimmed. Filling the `{Source}` slot is [ENG-2142](https://linear.app/discourse-graphs/issue/ENG-2142/map-source-to-the-roam-referenced-node-on-pull) (map `source` to the Roam referenced node on pull); once that lands, Evidence titles decorate like the rest.

The pull query already fetched `schema_id` and threw it away. It now carries `schemaId` and `coreTitle` onto `SharedNode`, both read as JSON scalars by path (`literal_content->>core_title`) rather than by pulling `literal_content`, so discovery still selects no payload columns. postgrest-js types those projections as `string`; they are null at runtime for rows without the key, which is what the `?? undefined` handles. `schemaId` is the row id, used only to correlate schema rows within one request; it is never persisted.

`resolveSharedNodeTypes` maps each schema id to a local node type in one batched query: match a local type whose uid equals the remote schema's `source_local_id`, else match on name, else create a local type from the schema. `Page`, `Block`, and `Any` are reserved local vocabulary: built-ins are excluded from the name match, and a remote schema carrying one of those names is never matched to a built-in by name and never created as a configured type, so its nodes keep their incoming titles. What such a schema should resolve to instead is [ENG-2210](https://linear.app/discourse-graphs/issue/ENG-2210/investigate-conflict-resolution-for-remote-schemas-named-page-block-or). A created type reuses the remote id as its page uid, so the next import from that space matches on id. After a create the resolver calls `refreshConfigTree()`, as the settings panel does: `createDiscourseNodeType` only invalidates the new-store cache, and with `Use new settings store` off (the default) `getDiscourseNodes` reads `discourseConfigRef.nodes`, which only that call repopulates. Without it the created type stayed invisible for the session and every later import or refresh re-entered the create branch. This is the same algorithm Obsidian runs in `mapNodeTypeIdToLocal`. The format is read contract-first through the shared `resolveSchemaFormat` helper in `packages/utils`: top-level `literal_content.format` is authoritative, and `source_data.format` is a temporary fallback until Obsidian writes the top-level key and existing rows are backfilled ([ENG-2208](https://linear.app/discourse-graphs/issue/ENG-2208/publish-obsidian-node-type-format-at-literal-contentformat)). One deliberate difference: when a schema carries no format, Obsidian invents `ABC - {content}` and we store an empty format instead, which leaves those titles verbatim until ENG-2158's format reaches the schema row. Types created this way get an empty shortcut; the user can set one in settings. Resolution is best-effort: a schema row RLS hides, a failed query, or a failed type creation reports to PostHog and leaves those nodes with their incoming title rather than failing the import.

I did not reuse `CrossAppNodeSchema` for the schema rows: its converter (`dbNodeSchemasToCrossApp`) selects the whole row including `literal_content` and needs space and account maps, and the contract has no `format` field. The local `SharedNodeSchema` is a five-column projection of `my_concepts`, not a cross-app payload. Adding `format` to the contract is ENG-2158's call.

Refresh is unchanged in shape: it resolves the type for its one node and passes it through, and `updateImportedPage` still renames only when the title differs, so refreshing an already-decorated page rewrites nothing. Pages imported before this change get decorated on their next Refresh; selecting them again in the discovery dialog while they are up to date stays a no-op, as it was.

Both producers have merged (#1318 for Obsidian, #1317 for Roam), so rows from either app carry `core_title` and decorate on import. The demo ran against a local database with both producers applied.

Previously stacked on `eng-2157-decorate-imported-node-titles-in-obsidian-from-core_title` (#1330) for the `decorateTitle` helper; #1330 has merged and this PR's diff is against `main`.

Deferred: resolving `{Source}` into a real source (ENG-2142), prop-based node identification (ENG-2133), a review/accept flow for types created during import (v0 creates them directly), and an `importedFrom` breadcrumb on created types (Roam's `DiscourseNode` has no such field yet). Until ENG-1861's imported-uid filter (#1286) lands, periodic sync treats decorated imported pages as local nodes. Created types from Obsidian schemas get 26-char page uids (`node_<nanoid>`); nothing in our code reads uid shape, but it shows in URLs. `createDiscourseNodeType` has no cleanup if the props write fails after the page is created (the settings panel has the same exposure); the import path reports the failure and leaves the node undecorated.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1331" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/database/src/lib/sharedNodes.ts
```diff
@@ -55,7 +57,7 @@ export type SharedNodeRows = {
 };
 
 const CONCEPT_COLUMNS =
-  "is_schema, last_modified, schema_id, source_local_id, space_id";
+  "core_title:literal_content->>core_title, is_schema, last_modified, schema_id, source_local_id, space_id";
```
sid597: `core_title` lives inside `literal_content`, so this selects the one JSON key by path instead of the whole column. Listing queries must not pull payload columns, and `literal_content` is named in db-domain-invariants as one. postgrest-js types the alias as `string`; it is null at runtime when the key is absent, which is why `SharedConcept` widens it to `string | null` and `buildSharedNodes` maps null to undefined.

packages/database/src/lib/sharedNodes.ts
```diff
@@ -188,11 +190,13 @@ export const buildSharedNodes = ({
         {
           rid,
           sourceLocalId: node.source_local_id,
+          schemaId: node.schema_id,
```
sid597, line 229: The query already fetched `schema_id` and dropped it. Roam needs it to find the node type schema the title format comes from. The null guard above this block is what lets the field be required rather than optional.

packages/database/src/lib/sharedNodes.ts
```diff
@@ -188,11 +190,13 @@ export const buildSharedNodes = ({
...
           spaceId: node.space_id,
           spaceName: space.name,
           spaceUri: space.url,
           platform: space.platform,
           title: direct.text,
+          coreTitle: node.core_title ?? undefined,
```
sid597, line 235: Optional on purpose. Nodes published before ENG-2153 carry no `core_title`, and those keep their incoming title verbatim.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+import type { Tables } from "@repo/database/dbTypes";
+import type { SharedNode } from "@repo/database/lib/sharedNodes";
+import { createDiscourseNodeType } from "~/components/settings/utils/accessors";
+import getDiscourseNodes, {
+  excludeDefaultNodes,
+  type DiscourseNode,
+} from "./getDiscourseNodes";
+import internalError from "./internalError";
+
+const SCHEMA_COLUMNS =
```
sid597, line 14: Roam publishes the node type format flat in `literal_content` (ENG-2158). Obsidian nests it under `literal_content.source_data`, because its frontmatter is dual-homed there. Both are read by JSON path, so this select carries no payload column.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+
+type SharedNodeSchema = Pick<
+  Tables<"my_concepts">,
+  "id" | "name" | "source_local_id"
+> & {
+  format: string | null;
+  source_data_format: string | null;
+};
+
+const findOrCreateNodeType = async (
```
sid597, line 30: Match order is id, then name, then create (MG, team chat 2026-08-19). Obsidian already runs the same algorithm in `mapNodeTypeIdToLocal` (apps/obsidian/src/utils/importNodes.ts:1059). Built-in types are in the list on purpose: a remote schema named `Page` resolves to Roam's Page rather than creating a user type that `getDiscourseNodes` would then prefer over the built-in. A created type reuses the remote schema's `source_local_id` as its page uid, so the next import from that space matches on id instead of matching on name again.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (matchedByName) return matchedByName;
+  if (!schema.name || !schema.source_local_id) return undefined;
+
+  posthog.capture("Discourse Node: Type Created From Import", {
+    label: schema.name,
+  });
+  return createDiscourseNodeType({
+    text: schema.name,
+    shortcut: "",
+    format: schema.format ?? schema.source_data_format ?? "",
```
sid597: A schema published before ENG-2158 has no format, and an empty format leaves the imported title verbatim. Obsidian's reader invents `ABC - {content}` in that case. We deliberately do not, so nobody gets a format they never configured.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export const resolveSharedNodeTypes = async ({
+  client,
+  sharedNodes,
+}: {
+  client: DGSupabaseClient;
+  sharedNodes: SharedNode[];
+}): Promise<Map<string, DiscourseNode>> => {
+  const schemaIds = [...new Set(sharedNodes.map(({ schemaId }) => schemaId))];
+  const { data, error } = await client
```
sid597, line 69: One query for the whole batch instead of one per node. `is_relation` is computed from the row's own roles (concept.sql:35-43), so `is_relation = false` selects node type schemas. Schema ids are global, so no space filter is needed. A row RLS hides yields no map entry, and we never create a local type from a row we could not read.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+    .from("my_concepts")
+    .select(SCHEMA_COLUMNS)
+    .eq("is_schema", true)
+    .eq("is_relation", false)
+    .in("id", schemaIds);
+  if (error) {
+    internalError({
```
sid597, line 76: Decoration is cosmetic, so neither a failed schema query nor a failed type creation aborts the import. The affected nodes land with their incoming title and we hear about it in PostHog.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+      error,
+      type: RESOLVE_ERROR_TYPE,
+      context: { operation: RESOLVE_ERROR_OPERATION, schemaIds },
+      sendEmail: false,
+    });
+    return new Map<string, DiscourseNode>();
+  }
+
+  const nodeTypeBySchemaId = new Map<number, DiscourseNode>();
+  for (const schema of data) {
```
sid597, line 86: Sequential on purpose. Two remote schemas can share a name, so the second iteration has to see what the first created, and that only works once the legacy cache is refreshed: `createDiscourseNodeType` invalidates the new-store cache, `refreshConfigTree()` in `findOrCreateNodeType` repopulates `discourseConfigRef.nodes` for the default (store flag off) path.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -292,10 +294,12 @@ const updateImportedPage = async ({
 export const materializeSharedNode = async ({
   client,
   sharedNode,
+  nodeType,
   force = false,
 }: {
   client: DGSupabaseClient;
   sharedNode: SharedNode;
+  nodeType?: Pick<DiscourseNode, "format">;
```
sid597, line 302: Only the format is used, so the param asks for only that. Resolution happens in the callers, which keeps this function free of a new I/O stage. Passing nothing reproduces today's behavior exactly.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -315,6 +319,10 @@ export const materializeSharedNode = async ({
     sourceModifiedAt: validated.sourceModifiedAt,
     sourceNodeRid: sharedNode.rid,
   };
+  const pageTitle =
```
sid597, line 322: Computed once, before the create/update branch, so both paths and the title collision checks agree. `validateSharedNode` stays pure and still validates the incoming title. `decorateTitle` returns `null` for a format it cannot rebuild from the core title (no `{content}`, or a second placeholder such as `{Source}`), so those nodes keep the incoming title rather than landing as `[[EVD]] - x -`. Until #1317/#1318 land nothing on `main` writes `core_title`, and every import takes the fallback arm.

apps/roam/src/components/settings/utils/accessors.ts
```diff
@@ -1090,13 +1090,16 @@ export const createDiscourseNodeType = async ({
   text,
   shortcut,
   format,
+  uid,
 }: {
   text: string;
   shortcut: string;
   format: string;
+  uid?: string;
```
sid597, line 1115: The settings panel must not pass a uid, since Roam generates one. The importer must, to reuse the remote type id so later imports match on id. `createPage` already accepts an optional `uid` and falls back to `generateUID()`.

sid597 (commented):
Guideposts for the non-obvious decisions in this diff.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  posthog.capture("Discourse Node: Type Created From Import", {
+    label: schema.name,
+  });
+  return createDiscourseNodeType({
+    text: schema.name,
+    shortcut: "",
+    format: schema.format ?? schema.source_data_format ?? "",
+    uid: schema.source_local_id,
+  });
```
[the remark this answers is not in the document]
sid597: The symptom is real but the cause is the in-session cache, not persistence. `createDiscourseNodeType` writes the legacy page tree (`discourse-graph/nodes/<name>` with Shortcut/Tag/Format children) and `refreshConfigTree` at `index.ts:58` rescans it on load, so the type survives a reload under either store. What was missing was the refresh within the session: with the store flag off, `getDiscourseNodes` reads `discourseConfigRef.nodes`, which only `refreshConfigTree()` repopulates. Fixed in 4463f2a9 by calling it after a successful create, mirroring `DiscourseNodeConfigPanel.createNodeType`.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -315,6 +319,10 @@ export const materializeSharedNode = async ({
...
+    sharedNode.coreTitle && nodeType?.format
+      ? decorateTitle(nodeType.format, sharedNode.coreTitle)
+      : validated.title;
```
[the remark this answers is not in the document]
sid597: Intentional, and now stated in the description: the up-to-date skip predates this PR and stays timestamp-only. A page imported before this change gets its decorated title on its next Refresh (force path, which goes through `updateImportedPage` and renames when the title differs); re-selecting an up-to-date page in the discovery dialog remains a no-op, as before. Folding a title comparison into the skip would make the dialog rename pages the user did not ask to change, which is a separate decision from decoration.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,96 @@
...
+  "id" | "name" | "source_local_id"
+> & {
+  format: string | null;
+  source_data_format: string | null;
+};
+
+const findOrCreateNodeType = async (
+  schema: SharedNodeSchema,
+): Promise<DiscourseNode | undefined> => {
+  const localNodeTypes = getDiscourseNodes();
```
sid597, line 33: No `excludeDefaultNodes` filter here on purpose. The built-in Page and Block types are synthesized by `getDiscourseNodes` only while no configured type shares their name, so creating a user type named `Page` from a remote schema would shadow the built-in (and its `has title` specification) for every consumer. Letting the built-ins take part in the name match resolves such a schema to the built-in instead.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,96 @@
...
+  const matchedByName = localNodeTypes.find(
+    (nodeType) => nodeType.text === schema.name,
+  );
+  if (matchedByName) return matchedByName;
+  if (!schema.name || !schema.source_local_id) return undefined;
+
+  const nodeType = await createDiscourseNodeType({
+    text: schema.name,
+    shortcut: "",
+    format: schema.source_data_format || schema.format || "",
```
sid597: Same precedence and operator as Obsidian's `parseSchemaLiteralContent` (importNodes.ts:1044): `source_data.format` first, then the flat key, with `||` so an empty string falls through. No producer writes both keys today, so the order only matters for the parity claim.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,96 @@
...
+    uid: schema.source_local_id,
+  });
+  refreshConfigTree();
```
sid597, line 54: Mirrors `DiscourseNodeConfigPanel.createNodeType`, which pairs `createDiscourseNodeType` with `refreshConfigTree()`. The accessor only invalidates the new-store cache; with `Use new settings store` off (the default) `getDiscourseNodes` reads `discourseConfigRef.nodes`, which this call repopulates. Without it the created type was invisible for the rest of the session and every later import or refresh re-entered the create branch, where `createPage` then failed on the existing title. The PostHog capture sits after the create so a failed creation is not counted as a created type.

apps/roam/src/utils/resolveSharedNodeTypes.ts
```diff
@@ -0,0 +1,96 @@
...
+          operation: RESOLVE_ERROR_OPERATION,
+          schemaId: schema.id,
+          schemaName: schema.name,
+        },
+        sendEmail: false,
+      });
+    }
+  }
+
+  return nodeTypeBySchemaId;
```
sid597, line 105: Keyed by schema id rather than re-keyed by rid: both callers already hold `sharedNode.schemaId`, so the rid map was a second map that only renamed the first.

apps/roam/src/utils/getDiscourseNodeFormatExpression.ts
```diff
@@ -1,7 +1,9 @@
+import { FORMAT_PLACEHOLDER } from "@repo/database/lib/decorateTitle";
+
 export const getDiscourseNodeFormatInnerExpression = (format: string): string =>
   `${format
     .replace(/(\[|\]|\?|\.|\+)/g, "\\$1")
-    .replace(/{[a-zA-Z]+}/g, "(.*?)")}`;
+    .replace(FORMAT_PLACEHOLDER, "(.*?)")}`;
```
sid597, line 6: The placeholder pattern now comes from `decorateTitle`, so the regex that decides what `extractContentFromTitle` captures and the one `decorateTitle` rebuilds from are the same object. #1330 does the same for the Obsidian helper.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -315,6 +319,10 @@ export const materializeSharedNode = async ({
     sourceModifiedAt: validated.sourceModifiedAt,
     sourceNodeRid: sharedNode.rid,
   };
+  const pageTitle =
+    (sharedNode.coreTitle && nodeType
+      ? decorateTitle(nodeType.format, sharedNode.coreTitle)
+      : null) ?? validated.title;
```
sid597, line 325: `decorateTitle` returns `null` when the format carries a placeholder other than `{content}` (this graph's Evidence format has `{Source}`, and the shared row holds nothing to put there), so the incoming title is kept rather than writing a half-filled name the format regex would not match. Filling that slot is [ENG-2142](https://linear.app/discourse-graphs/issue/ENG-2142/map-source-to-the-roam-referenced-node-on-pull); once it lands, these titles decorate too.

maparent (commented):
Should this not come after @eng-2128? (And sorry I'm late there, the migration was trickier than I hoped.) You will have to rewrite this to include the source. So I vote to postpone.

maparent:
Ah, no I see that it's a separate task. Ok will look.

apps/roam/src/utils/resolveSharedNodeTypes.ts
maparent: That makes sense, but thinking forward I think that the format is the one that we extract from the contract, and as such should have precedence. If we decide to modify the format (which you pointed out elsewhere) the modified one would be this one, and the source_data_format (if retained) would be unprocessed. So we'd probably want format there.
Does that mean we should also make the change in `parseSchemaLiteralContent`? I would tend to.

maparent (approved):
One nit that requires judgement. You may or may not agree with me, and not blocking imho.

mdroidian:
We'll need to discuss and have a pointed to how we're solving the fact that we are storing `format` in two different places: https://roamresearch.com/#/app/discourse-graphs/page/eFt8yo8Zq

apps/roam/src/utils/resolveSharedNodeTypes.ts
mdroidian: We need to treat `Page`, `Block`, and `Any` as reserved local vocabulary, eg
- We should never resolve a remote schema to a built-in solely by name
- We should never create a configured type using a reserved name. 

For example, a remote schema named `Block` may have its own identity, format, and semantics. Mapping it to the local built-in would silently discard that information, while creating it locally would shadow behavior that downstream consumers rely on.  There are probably at least a 6-10 other downstream problems that would show up.

For v0, let's block these schema names and create a low-priority ticket to investigate proper conflict resolution.

mdroidian (changes requested):
Blocked on handling Roam's reserved node type vocabulary

apps/roam/src/utils/resolveSharedNodeTypes.ts
sid597: Done in 43b9d4ae: precedence is now contract-first through a shared helper, `packages/utils/src/resolveSchemaFormat.ts`. Its header comment states the top-level key is authoritative and the nested read is a temporary fallback. On `parseSchemaLiteralContent`: agreed. That change rides the follow-up PR that moves Obsidian to top-level `literal_content.format` (Post-V0 Roam–Obsidian sync architecture), which also switches its reader to this helper.

apps/roam/src/utils/resolveSharedNodeTypes.ts
sid597: Done in fad4f742: built-ins are excluded from the name match, and a schema named `Page`, `Block`, or `Any` is never matched to a built-in by name and never created as a configured type (`RESERVED_NODE_TYPE_NAMES`). A user-configured type carrying such a name still matches by name. Conflict resolution ticket: [ENG-2210](https://linear.app/discourse-graphs/issue/ENG-2210/investigate-conflict-resolution-for-remote-schemas-named-page-block-or).

apps/roam/src/utils/resolveSharedNodeTypes.ts
sid597: Partly superseded by fad4f742: built-ins no longer take part in the name match. `Page`, `Block`, and `Any` are reserved: a remote schema carrying one of those names never resolves to a built-in by name and is never created as a configured type, so its nodes keep their incoming titles. The rest of this note (match order id then name then create, id reuse for later imports) still holds. Conflict resolution for reserved-name schemas is [ENG-2210](https://linear.app/discourse-graphs/issue/ENG-2210/investigate-conflict-resolution-for-remote-schemas-named-page-block-or).

sid597:
@mdroidian @maparent both items are addressed on the current head.

**Reserved node type names.** `fad4f742` does what you asked. Built-ins are excluded from the name match (`resolveSharedNodeTypes.ts:39`), and a remote schema named `Page`, `Block`, or `Any` is never matched to a built-in by name and never created as a configured type (`RESERVED_NODE_TYPE_NAMES`, `resolveSharedNodeTypes.ts:17,43`). Those nodes keep their incoming title. `Any` is in the set because it is the wildcard `*` label (`DiscourseRelationConfigPanel.tsx:1007`, `getDiscourseContextResults.ts:210`), not because a built-in carries that name. A user-configured type that happens to be named `Page` still matches by name, which is unchanged behavior. Conflict resolution for reserved-name schemas is [ENG-2210](https://linear.app/discourse-graphs/issue/ENG-2210/investigate-conflict-resolution-for-remote-schemas-named-page-block-or).

**`format` in two places.** The decision is the [August 28 sync](https://roamresearch.com/#/app/discourse-graphs/page/Toww-3UYl), and it matches the recommendation on your block: standardize on `literal_content.format` and keep the dual read only during the migration. It is split across five tickets:

- [ENG-2158](https://linear.app/discourse-graphs/issue/ENG-2158/publish-node-type-format-with-schemas) (merged, #1316): Roam writes `literal_content.format`.
- [ENG-2207](https://linear.app/discourse-graphs/issue/ENG-2207/add-shared-node-type-format-resolver-to-repoutils): shared resolver in `@repo/utils`. This PR adds the resolver and moves the Roam reader onto it. The Obsidian reader is still on its own precedence, so ENG-2207 should not close with this PR.
- [ENG-2208](https://linear.app/discourse-graphs/issue/ENG-2208/publish-obsidian-node-type-format-at-literal-contentformat): Obsidian writes the top-level key and stops writing `source_data.format`.
- [ENG-2209](https://linear.app/discourse-graphs/issue/ENG-2209/read-node-type-format-in-dbnodeschematocrossapp): `dbNodeSchemaToCrossApp` exposes `format` instead of leaving it in `metadata`.
- [ENG-2175](https://linear.app/discourse-graphs/issue/ENG-2175/backfill-format-on-existing-node-type-schema-rows): backfills rows published before ENG-2158. One thing to settle before it lands: its "Done When" still says Obsidian-origin rows stay on `source_data.format`, which contradicts ENG-2208. Those two need reconciling.

@maparent on `parseSchemaLiteralContent`: agreed, and that is the remaining half of ENG-2207. Doing it here puts an Obsidian reader change in a Roam PR that is already over the size guideline, and the flip is a no-op until ENG-2208 writes the key, so I would rather it ride with 2208. The one hard constraint is ordering: the reader has to prefer the top-level key before or at the same time as the writer starts emitting it, never after.

**One correction on this head.** The `resolveSchemaFormat` header comment said nothing writes the top-level key yet. That stopped being true when #1316 merged on Sep 1, so the comment now says Roam writes it and names ENG-2208 as the ticket that ends the dual read. Same reason, the description's line about #1317 being open is stale: Roam-origin rows carry `core_title` on `main` now, so those imports decorate rather than take the fallback arm.

mdroidian:
@sid597 does a remote type named `Page`, `Block`, or `Any` match on name or not?  It seems like there are conflicting answers:

> a remote schema named `Page`, `Block`, or `Any` is never matched to a built-in by name and never created as a configured type 

vs

> A user-configured type that happens to be named `Page` still matches by name

mdroidian:
@sid597 In the future, could you split responses into separate threads.  One thread per response item.  That way we don't have multiple conversation merging into one large block of text.

mdroidian:
> `format` **in two places.** The decision is the [August 28 sync](<https://roamresearch.com/#/app/discourse-graphs/page/Toww-3UYl>)

I feel like there is some text/context missing here.  It doesn't relate to the text block right above it, I don't think.  What does "`format` in two places" …. OH … that is referencing a previously stated problem.  "he fact that we are storing `format` in two different places".  from here: https://github.com/DiscourseGraphs/discourse-graph/pull/1331#issuecomment-5434300930<br><br>Yeah, we definitely need multiple threads.

apps/roam/src/utils/resolveSharedNodeTypes.ts
sid597: > @sid597 does a remote type named Page, Block, or Any match on name or not? It seems like there are conflicting answers:

> a remote schema named Page, Block, or Any is never matched to a built-in by name and never created as a configured type

> vs

> A user-configured type that happens to be named Page still matches by name



@mdroidian Both, and they're about different things.

- The built-in `Page` and `Block` types (`backedBy: "default"`, `getDiscourseNodes.ts:44-79`) are filtered out of the name match (`resolveSharedNodeTypes.ts:38-41`). A remote schema named `Page` never resolves to the built-in.
- If the graph has a user-configured type named `Page`, that type has already replaced the built-in inside `getDiscourseNodes` (`getDiscourseNodes.ts:164-167`) and is `backedBy: "user"`, so it matches by name like any other user type. The schema resolves to the user's `Page`, not the built-in.
- If neither matches, the schema is not created (`resolveSharedNodeTypes.ts:43`) and its nodes keep their incoming title.

So the rule is: never the built-in, never a new type. A user type that shadows a built-in counts as the user's own type. If you'd rather reserved names skip the name match entirely, that's a one-line move of the reserved check above the name match. I'd keep it as is, since the user chose to define that type.

mdroidian (approved):

PR #1332, opened 2026-08-23
ENG-2155 Migrate existing database titles to undecorated core_title

sid597:
https://github.com/user-attachments/assets/afc4a904-1c7c-4f0a-b2c0-20a46df6ade7

image
This PR backfills the diagram's core_title for rows published before the top path existed.

Rows published before #1317 and #1318 have no `core_title` in `literal_content`, so importers fall back to the foreign decorated title. This backfills them client-side, per app, over each app's own space. It has to be client-side: `core_title` is `{content}` extracted with the node type's format, and the database never stores the format (`crossAppNodeSchemaToDbConcept` persists `template`/`template_content` only), so the undecoration can only be computed where the graph or vault is.

Each app already pages every own-space instance concept during a full sync to work out which concepts went missing. This extends that same query to also project `literal_content->>core_title` and partitions the rows in JS: rows with a null value get forced into the batch that the #1317/#1318 producers recompute from the local graph or vault. No extra requests and no `literal_content` payload, just the one key. The partition is client-side because the same paged result computes `missingConcepts`; filtering it server-side would break that consumer.

- Roam probes only on the first sync after load, so a re-run finds nothing. Backfilled nodes join the concept batch only, never the content/embedding batch, so a backfill costs no re-embedding. When sync is off, only shared nodes are backfilled and the toast says how many rows are waiting for sync to be enabled. Reported with a toast plus a `Sync core_title backfill` PostHog event.
- Obsidian probes on every full sync, which is a no-op after the first pass. Reported with a Notice. The asset pass after the sync now runs only for nodes with a detected file change; backfill-only nodes (and nodes re-pushed for a missing concept row) have unchanged files, so their file references are left alone instead of being re-read and re-written per node.
- Both stay silent when there was nothing to do. Counts are backfilled, already set, and rows that no longer match a discourse node locally (a deleted page, or one renamed so its title no longer matches its type's format). Those are reported, never guessed at. `cleanupOrphanedNodes` removes only rows whose block is gone, so a renamed page's row stays and keeps being reported.

One producer fix rides along because the backfill made it urgent: `getNodeExtraData` returned the Roam creator uid as `author_uid`, a key `concept_local_input` does not have, so every instance the Roam sync pushed landed with `author_id = NULL`, and a later re-push of a row that publish had given an author blanked it. Re-pushing every pre-`core_title` row through that producer would have done that to the whole published corpus at once. The key is now `author_local_id`, which `_local_concept_to_db_concept` resolves against `PlatformAccount`; rows whose author has no account row stay NULL as before. This also means ordinary Roam syncs now set `author_id` where they did not.

Not rewritten: stored decorated titles, relations, assets. Pushing a Roam row back through the sync producer also recomputes the concept `name`, which for block-backed node types is `node_title + " " + text`. That's inherent to recomputing locally rather than patching one key. `upsertConceptBatches` keeps its existing behaviour of throwing on any negative row id, so one row failure aborts the sync with the earlier rows committed; that predates this PR (a first-ever sync has the same shape) and is not changed here. `orderConceptsByDependency`'s "dependencies not in batch" warning fires whenever a node's type was not itself edited since the last sync, so it also fires on this pass; pre-existing noise, not a correctness issue (`upsert_concepts` resolves `schema_id` from the database).

Three known gaps, all from open bot findings on #1317/#1318 and all producing a non-null but wrong `core_title`: an empty `{content}` capture, unescaped regex metacharacters in a node format, and Obsidian's `extractContentFromTitle` returning capture 1 rather than the `{content}` capture (so `{Source} - {content}` stores the source). This migration treats any non-null value as already set, so those rows will need a separate value-aware pass once those bugs are fixed. The serialized key `core_title` is a bare literal on the producer and probe sides; a shared constant is a follow-up once #1317/#1318 and the readers on #1330/#1331 land, since they sit on different bases.

`partitionByCoreTitle` and the probe select live in `packages/database/src/lib` because both apps need the identical definition of "missing a core title", and because Obsidian has no test runner, so the shared test is the only coverage over logic Obsidian runs. It's a PostgREST projection shape, not cross-app data, so it doesn't extend `crossAppContracts.ts`. supabase-js types the projection as `string`; it's null at runtime for rows without the key, so the row type declares `string | null`.

Stacked on #1317; the diff range also includes #1318's 7-line commit. This branch predates main's switch from `arity` to `is_relation` on `my_concepts` node queries, so the two extended selects keep `arity` to match this tree and need the switch when #1317 rebases onto main.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1332" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

Demo note: the Roam half of the video shows the two rows before and after the first sync after extension load (the rows are read back over REST); the 5 s toast fires during the recorder's extension-load phase, so it is not on camera, and the Obsidian half shows the equivalent notice. While recording, the dev graph's first sync on this branch ended with a "Discourse node embeddings sync failed" toast after the concept rows were written. What I could establish afterwards: every targeted row had `core_title` written, the local Postgres log (`log_min_messages = warning`) shows no `upsert_concepts` row-failure warning and no statement timeout in that window, and everything after the concept upsert in `createOrUpdateDiscourseEmbedding` is wrapped (`cleanupOrphanedNodes`, `endSyncTask`) or non-throwing (the report). So the throw is client-side and I could not reproduce it offline; the browser console line `createOrUpdateDiscourseEmbedding: Process failed:` from a live run is what pins it. Worth a look before merging.


packages/database/src/lib/coreTitleBackfill.ts
```diff
@@ -0,0 +1,23 @@
+export const CORE_TITLE_PROBE_SELECT =
+  "source_local_id, core_title:literal_content->>core_title";
```
sid597, line 2: `literal_content` is `jsonb NOT NULL DEFAULT '{}'`, so the column is never null but the `core_title` key is absent on every row published before #1317/#1318. PostgREST can't tell an absent key from a JSON null. Both come back as `null`, and both mean the same thing here, so we partition client-side instead of filtering with `.is(...)`.

packages/database/src/lib/coreTitleBackfill.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export type CoreTitleProbeRow = {
+  source_local_id: string | null;
+  core_title: string | null;
```
sid597, line 6: supabase-js types `core_title:literal_content->>core_title` as `string`, not `string | null`. That's wrong at runtime. Declaring the row type with `string | null` widens the response safely, so there's no cast and no guard, and the null branch stays honest.

packages/database/src/lib/coreTitleBackfill.ts
```diff
@@ -0,0 +1,23 @@
...
+};
+
+export const partitionByCoreTitle = (
```
sid597, line 9: Shared because both apps need the identical definition of "this row is missing a core title". Obsidian has no test runner, so this is the only place that behavior is covered by a test. It describes a PostgREST projection, not cross-app data, which is why it doesn't extend `crossAppContracts.ts`.

apps/roam/src/utils/coreTitleBackfill.ts
```diff
@@ -0,0 +1,44 @@
...
+}): CoreTitleBackfill => {
+  const { missingCoreTitleIds, withCoreTitleCount } =
+    partitionByCoreTitle(conceptRows);
+  const localIds = new Set(localNodes.map((node) => node.source_local_id));
+  return {
+    nodesToBackfill: localNodes.filter((node) =>
+      missingCoreTitleIds.has(node.source_local_id),
+    ),
+    withCoreTitleCount,
+    orphanedIds: [...difference(missingCoreTitleIds, localIds)],
```
sid597: `orphanedCount` is missing minus allNodes: rows whose `source_local_id` is not a discourse node in this graph any more, because the page was deleted or renamed so its title no longer matches its type's format. We report them, we never guess a title for them. `cleanupOrphanedNodes` later in the sync deletes only rows whose block is gone, so a renamed page's row stays and keeps being counted here. Only the count is kept: nothing reads the ids, and Obsidian stores a number too.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -870,6 +876,35 @@ export const setSyncActivity = (active: boolean) => {
   }
 };
 
+const reportCoreTitleBackfill = ({
```
sid597, line 877: The probe only runs on `isInitialSync`, so the second sync finds nothing and this goes silent. The toast is deliberately independent of `showToast`. That flag gates sync failure toasts, and a one-time migration result should still show if the initial sync got retried by the timer.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -896,7 +934,7 @@ const getAllMissingOrNewDiscourseNodes = async ({
   const existingConceptIdsReq = await getAllPages(
     supabaseClient
       .from("my_concepts")
-      .select("source_local_id")
+      .select(CORE_TITLE_PROBE_SELECT)
```
sid597, line 925: Extends the query that already pages every own-space instance concept rather than adding a second probe. One request does both jobs now: the `missingConcepts` diff and the core_title partition. It projects the key, not `literal_content`, so no payload column is pulled. Note this branch predates main's switch from `arity` to `is_relation`, so this chain needs that swap when #1317 rebases onto main.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -1238,6 +1289,16 @@ export const createOrUpdateDiscourseEmbedding = async (
           sharedSourceLocalIds.has(node.source_local_id),
         )
       : changedNodeInstances;
+    const nodesToBackfillCoreTitle = (
```
sid597, line 1285: Backfill nodes are merged into the concept batch only. They never reach `upsertNodesToSupabaseAsContentWithEmbeddings`, so a backfill costs no re-embedding. When sync is off we only backfill nodes that are actually shared, matching what `nodeInstancesToSync` does one line up.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -1316,14 +1377,21 @@ export const createOrUpdateDiscourseEmbedding = async (
...
-          nodesSince: nodeInstancesToSync,
+          nodesSince: conceptNodesToSync,
           since: sinceTime,
           allNodeTypes: allDgNodeTypes,
           sharedNodeTypeIds,
           supabaseClient: activeSupabaseClient,
           context: activeContext,
         }),
     });
+    if (coreTitleBackfill !== null) {
```
sid597, line 1381: Reported after the concept upsert resolves. `upsertConceptBatches` throws on any batch error or negative row id, so reaching this line means the rows really were written (a throw leaves the earlier rows committed and skips this report; that is the pre-existing per-row semantics of `upsert_concepts`, not changed here). `backfilled` is counted after the `sharedNodesOnlySync` filter and `skipped` before it, so with sync off the non-shared rows missing a core title are in neither number; `deferred` carries them so the toast does not read as a complete pass. One thing to flag: pushing a row back through this producer also recomputes the concept `name`, which for block-backed types is `node_title + " " + text`. That is inherent to recomputing locally instead of patching one key.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -344,11 +375,14 @@ const buildChangedNodesFromNodes = async ({
...
   let missingConcepts: Set<string> | undefined;
+  let coreTitleProbe:
+    | { missingCoreTitleIds: Set<string>; skipped: number; orphaned: number }
+    | undefined;
   if (fullSync) {
     const existingConceptIds = await getAllPages(
       supabaseClient
         .from("my_concepts")
-        .select("source_local_id")
+        .select(CORE_TITLE_PROBE_SELECT)
```
sid597, line 360: Same trick as the Roam side. This query already pages every own-space instance row for `missingConcepts`, so adding the projected key costs zero extra requests. It projects the key rather than `literal_content`, unlike the older `template_content` probe below, which pulls the whole column and re-filters in JS. That one is worth a follow-up ticket, but it's out of scope here.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -389,7 +430,8 @@ const buildChangedNodesFromNodes = async ({
 
     if (
       finalChangeTypes.length === 0 &&
-      !missingConcepts?.has(node.nodeInstanceId)
+      !missingConcepts?.has(node.nodeInstanceId) &&
+      !coreTitleProbe?.missingCoreTitleIds.has(node.nodeInstanceId)
```
sid597: Same widening the `missingConcepts` mechanism already uses. Forced-in nodes carry `changeTypes: []`, and `getVariantsToCreate` turns that into no content and no embedding work, so only the concept gets re-upserted. Imported nodes still `continue` above this, so their rows stay for the source space to backfill.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -458,6 +512,10 @@ export const syncAllNodesAndRelations = async (
       fullSync: true,
     });
 
+    if (coreTitleBackfill !== null) {
```
sid597: The Notice fires after `convertDgToSupabaseConcepts`, which throws on upsert error, so the counts describe rows that were actually written. It stays silent when nothing was backfilled and nothing was orphaned, which is every run after the first.

sid597 (commented):
Annotations on the non-obvious decisions in this backfill.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -160,12 +161,17 @@ export const discourseRelationTripleSchemaToLocalConcept = ({
...
 ): LocalConceptDataInput => {
   const extraData = getNodeExtraData(nodeData.file, context.userId);
   const { nodeInstanceId, nodeTypeId, importedFromRid, ...otherData } =
     nodeData.frontmatter;
   const literal_content: Record<string, Json> = {
     label: nodeData.file.basename,
+    core_title: extractContentFromTitle(
+      nodeTypesById[nodeData.nodeTypeId]?.format ?? "",
+      nodeData.file.basename,
+    ),
```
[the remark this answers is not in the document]
sid597: Agreed on the defect; it is in #1318's producer (`apps/obsidian/src/utils/extractContentFromTitle.ts`), which this PR's diff range includes but does not change. Added to the description's known-gaps list alongside the two other #1317/#1318 findings: all three store a non-null but wrong `core_title`, and this migration treats any non-null value as already set, so those rows need a value-aware pass once the extractor is fixed.

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -50,7 +50,7 @@ const getNodeExtraData = (
   const created = new Date(created_t).toISOString();
   const last_modified = new Date(last_modified_t).toISOString();
   return {
-    author_uid,
+    author_local_id: author_uid,
```
sid597, line 60: The key this object is spread into is `concept_local_input.author_local_id`; `author_uid` was silently dropped by `jsonb_populate_record`, so every instance the Roam sync pushed had `author_id = NULL` and a re-push of a published row blanked the author publish had set. With the backfill re-pushing every pre-`core_title` row in one pass this stopped being a slow leak, hence the fix here. `_local_concept_to_db_concept` only looks the account up (no creation), so rows whose author has no `PlatformAccount` stay NULL as before; the datalog variable keeps its name.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -458,8 +514,15 @@ export const syncAllNodesAndRelations = async (
...
+      noticeCoreTitleBackfill(coreTitleBackfill);
+    }
+
     // When synced nodes are already published, ensure non-text assets are in storage.
-    await syncPublishedNodesAssets(plugin, changedNodeInstances);
+    await syncPublishedNodesAssets(
+      plugin,
+      changedNodeInstances.filter((node) => node.changeTypes.length > 0),
```
sid597, line 475: `syncPublishedNodeAssets` reads and rewrites `FileReference` rows per published node unconditionally, so without this filter every published backfill-only node cost two round-trips on the first full sync after this ships. Nodes forced in with `changeTypes: []` (core-title backfill, and the pre-existing missing-concept case) have unchanged files, so their asset references are already right.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -870,6 +876,44 @@ export const setSyncActivity = (active: boolean) => {
...
+    deferred,
+    skipped,
+    orphaned,
+  });
+  if (backfilled === 0 && deferred === 0 && orphaned === 0) return;
+  const messages = [
+    `Backfilled core title for ${backfilled} node${backfilled === 1 ? "" : "s"}.`,
+    `${skipped} already had one.`,
+  ];
+  if (deferred > 0) {
```
sid597: With sync off only shared nodes are pushed, so `backfilled` is the post-filter count while `skipped` comes from the whole probe. `deferred` is the rows that were neither, so the toast does not claim a completed pass; they are picked up by the next initial sync once sync is on.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -320,15 +335,33 @@ const detectNodeChanges = (
...
 };
 
+const noticeCoreTitleBackfill = ({
+  backfilled,
+  skipped,
+  orphaned,
+}: CoreTitleBackfillCounts): void => {
+  if (backfilled === 0 && orphaned === 0) return;
+  const messages = [
+    `Backfilled core title for ${backfilled} node${backfilled === 1 ? "" : "s"}.`,
```
mdroidian: This is not something we need (or should) pass on to the user.  A normal user will have no idea what "backfilled" means or why it matters to them.

We can remove this toast altogether and only report it to postHog in Roam.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -458,8 +514,15 @@ export const syncAllNodesAndRelations = async (
       fullSync: true,
     });
 
+    if (coreTitleBackfill !== null) {
+      noticeCoreTitleBackfill(coreTitleBackfill);
```
mdroidian: As mentioned above, drop this notice to the user.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -870,6 +876,44 @@ export const setSyncActivity = (active: boolean) => {
...
+  orphaned: number;
+}): void => {
+  posthog.capture("Sync core_title backfill", {
+    backfilled,
+    deferred,
+    skipped,
+    orphaned,
+  });
+  if (backfilled === 0 && deferred === 0 && orphaned === 0) return;
+  const messages = [
```
mdroidian: As noted above, drop the message to the user

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -870,6 +876,44 @@ export const setSyncActivity = (active: boolean) => {
...
+  ];
+  if (deferred > 0) {
+    messages.push(`${deferred} waiting for sync to be enabled.`);
+  }
+  if (orphaned > 0) {
+    messages.push(
+      `${orphaned} no longer match a discourse node in this graph.`,
+    );
+  }
+  renderToast({
```
mdroidian: As noted above, drop the message to the user

mdroidian (approved):
Generally speaking, if a flag like "sync" is off, the user probably expects no communication with the db.  We should respect that (again, generally speaking).  In this instance with alpha only users, we can push this change through.

Let's make sure we remove the notification to the user before merging.

PR #1346, opened 2026-08-24
ENG-2175 Backfill format on existing node type schema rows

sid597:
https://github.com/user-attachments/assets/551ab26c-73fe-4994-b4f5-5f9b538376bf

Schema rows synced before #1316 never get `literal_content.format`: publish skips schemas that already exist, and the periodic sync only re-selects types edited since the last watermark. This backfills them through the sync producer. #1316 has merged; this PR's diff is against `main`.

How it works:

- On the first sync cycle per session, probe own-space schema rows (`is_schema = true`, `is_relation = false`) for a missing `format`.
- Force the matching node types into the concept batch, so `discourseNodeSchemaToLocalConcept` recomputes the complete row.
- Report backfilled / already set / orphaned counts. Silent when there is nothing to do. Re-runs are no-ops.

Decisions, annotated inline on the diff:

- Sync producer, not publish path: `upsert_concepts` replaces `literal_content` wholesale, and the publish converter writes a different shape.
- No "waiting for sync" bucket, unlike the core_title backfill: the schema batch is not shared-gated today, and the probe result is intersected with rows that already exist, so nothing new can be published.
- The `getNodeExtraData` → `author_local_id` rider this backfill depends on landed in `main` via #1332; only the test asserting it (`conceptConversion.test.ts`) remains in this diff.

Known gaps, shared with the core_title backfill on #1332: the orphan toast repeats on each load, and the posthog event has no space attribution. Worth one joint fix later rather than forking the two reports now.

### DIFF
--- a/apps/roam/src/utils/conceptConversion.ts
+++ b/apps/roam/src/utils/conceptConversion.ts
@@ -20,7 +20,7 @@ import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageU
 const getNodeExtraData = (
   node_uid: string,
 ): {
-  author_uid: string;
+  author_local_id: string;
   created: string;
   last_modified: string;
   page_uid: string;
@@ -57,7 +57,7 @@ const getNodeExtraData = (
   const created = new Date(created_t).toISOString();
   const last_modified = new Date(last_modified_t).toISOString();
   return {
-    author_uid,
+    author_local_id: author_uid,
     created,
     last_modified,
     page_uid,
@@ -218,7 +218,7 @@ export const discourseRelationDataToLocalConcept = (
   const created = new Date(
     Math.max(...nodeData.map((nd) => new Date(nd.created).getTime())),
   ).toISOString();
-  const author_local_id: string = nodeData[0].author_uid; // take any one; again until I get the relation object
+  const author_local_id: string = nodeData[0].author_local_id; // take any one; again until I get the relation object
   return {
     space_id: context.spaceId,
     source_local_id: relationUid,
--- a/apps/roam/src/utils/schemaFormatBackfill.ts
+++ b/apps/roam/src/utils/schemaFormatBackfill.ts
@@ -1,0 +1,41 @@
+import { difference, intersection } from "@repo/utils/setOperations";
+import { type DiscourseNode } from "./getDiscourseNodes";
+
+export const SCHEMA_FORMAT_PROBE_SELECT =
+  "source_local_id, format:literal_content->>format";
+
+type SchemaFormatProbeRow = {
+  source_local_id: string | null;
+  format: string | null;
+};
+
+export type SchemaFormatBackfill = {
+  nodeTypeIdsToBackfill: Set<string>;
+  withFormatCount: number;
+  orphanedCount: number;
+};
+
+export const buildSchemaFormatBackfill = ({
+  conceptRows,
+  nodeTypes,
+}: {
+  conceptRows: SchemaFormatProbeRow[];
+  nodeTypes: DiscourseNode[];
+}): SchemaFormatBackfill => {
+  const missingFormatIds = new Set<string>();
+  let withFormatCount = 0;
+  for (const row of conceptRows) {
+    if (row.source_local_id === null) continue;
+    if (row.format === null) {
+      missingFormatIds.add(row.source_local_id);
+    } else {
+      withFormatCount += 1;
+    }
+  }
+  const localTypeIds = new Set(nodeTypes.map((nodeType) => nodeType.type));
+  return {
+    nodeTypeIdsToBackfill: intersection(missingFormatIds, localTypeIds),
+    withFormatCount,
+    orphanedCount: difference(missingFormatIds, localTypeIds).size,
+  };
+};
--- a/apps/roam/src/utils/syncDgNodesToSupabase.ts
+++ b/apps/roam/src/utils/syncDgNodesToSupabase.ts
@@ -1129,19 +1129,19 @@ const probeSchemaFormatBackfill = async ({
   spaceId: number;
   nodeTypes: DiscourseNode[];
 }): Promise<SchemaFormatBackfill> => {
-  const probeRes = await supabaseClient
-    .from("my_concepts")
-    .select(SCHEMA_FORMAT_PROBE_SELECT)
-    .eq("space_id", spaceId)
-    .eq("is_schema", true)
-    .eq("is_relation", false);
-  if (probeRes.error) {
-    throw new Error(
-      `Schema format probe failed: ${JSON.stringify(probeRes.error, null, 2)}`,
-    );
-  }
+  const probeRows = await getAllPages(
+    supabaseClient
+      .from("my_concepts")
+      .select(SCHEMA_FORMAT_PROBE_SELECT)
+      .eq("space_id", spaceId)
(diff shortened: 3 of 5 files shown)


apps/roam/src/utils/schemaFormatBackfill.ts
```diff
@@ -0,0 +1,41 @@
+import { difference, intersection } from "@repo/utils/setOperations";
+import { type DiscourseNode } from "./getDiscourseNodes";
+
+export const SCHEMA_FORMAT_PROBE_SELECT =
+  "source_local_id, format:literal_content->>format";
```
sid597, line 5: PostgREST reads an absent jsonb key and a JSON null the same way: both come back as null through `->>`. The Roam producers never write a null format, so both mean "backfill", and we partition client-side instead of filtering with `.is(...)`. supabase-js types the projection `string`; the declared row type widens it to `string | null`, which is what actually arrives.

This mirrors eng-2155's core_title helper but stays in apps/roam: Obsidian schema rows carry `source_data.format` from the start, so only Roam needs it. Obsidian's `template_content` probe pushes the null test into the query instead; here we need the "already set" count, so we fetch all own-space schema rows (tens) and split.

apps/roam/src/utils/schemaFormatBackfill.ts
```diff
@@ -0,0 +1,41 @@
...
+      missingFormatIds.add(row.source_local_id);
+    } else {
+      withFormatCount += 1;
+    }
+  }
+  const localTypeIds = new Set(nodeTypes.map((nodeType) => nodeType.type));
+  return {
+    nodeTypeIdsToBackfill: intersection(missingFormatIds, localTypeIds),
+    withFormatCount,
+    orphanedCount: difference(missingFormatIds, localTypeIds).size,
```
sid597, line 39: Orphans are probed rows that no longer match a configured node type. We report the count and never guess a format. They can persist: `cleanupOrphanedNodes` deletes rows whose Roam block is gone, but a type removed from the DG config keeps its page and its row. The core_title backfill accepted the same permanence for renamed pages.

An empty format counts as already set. The producer wrote the key (`format` defaults to ""), so there is nothing to recompute and the backfill converges.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -650,7 +657,10 @@ export const convertDgToSupabaseConcepts = async ({
   );
 
   allNodeTypes.forEach((nodeType) => {
-    if (sharedNodeTypeIds.has(nodeType.type)) {
+    if (
+      sharedNodeTypeIds.has(nodeType.type) ||
+      backfillNodeTypeIds.has(nodeType.type)
```
sid597, line 669: A second force-include beside `sharedNodeTypeIds` instead of smuggling the backfill through it; they mean different things. The set is an intersection with ids already in `my_concepts`, so the backfill can only rewrite existing rows. It cannot publish a schema the user never synced, including in shared-content-only mode.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -1069,6 +1079,62 @@ const getSharedRoamNodesWithFullContentUpdatesSince = async ({
   });
 };
 
+const probeSchemaFormatBackfill = async ({
```
sid597, line 1123: Runs only on `isInitialSync`, so the second cycle probes nothing and the report goes silent. Paged through `getAllPages` like the other `my_concepts` reads in this file; the expected scale is node types per space, so one page in practice. A probe failure fails the cycle like any other phase; `initialSync` stays true, so the next attempt probes again.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -1069,6 +1079,62 @@ const getSharedRoamNodesWithFullContentUpdatesSince = async ({
...
+      `Schema format probe failed: ${JSON.stringify(probeRes.error, null, 2)}`,
+    );
+  }
+  return buildSchemaFormatBackfill({
+    conceptRows: probeRes.data,
+    nodeTypes,
+  });
+};
+
+const reportSchemaFormatBackfill = ({
```
sid597, line 1149: Deliberately independent of `showToast`, same as the core_title report on #1332: it announces a one-time data migration, not cycle status. `skipped` keeps the property name from the core_title posthog event; the toast calls the same number "already had one".

apps/roam/src/utils/conceptConversion.ts
```diff
@@ -13,7 +13,7 @@ import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageU
 const getNodeExtraData = (
   node_uid: string,
 ): {
-  author_uid: string;
+  author_local_id: string;
```
sid597, line 23: Required by the backfill, not opportunistic. `concept_local_input` has no `author_uid` field, so `jsonb_populate_record` dropped it and every sync-produced concept carried a NULL `author_id`. `upsert_concepts` sets `author_id` unconditionally on conflict, so re-upserting publish-origin schema rows (which do have an author) would blank them, and `dbNodeSchemaToCrossApp` then throws "Missing author". Identical to a4936129 on eng-2155; whichever merges second sees a no-op conflict.

`page_uid` below is the same dead-key class but inert: nothing maps to it and nothing breaks because of it. Removing it belongs with the literal_content shape unification.

sid597 (commented):
Notes on the non-obvious decisions, one per thread.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -1069,6 +1079,62 @@ const getSharedRoamNodesWithFullContentUpdatesSince = async ({
...
+  spaceId,
+  nodeTypes,
+}: {
+  supabaseClient: DGSupabaseClient;
+  spaceId: number;
+  nodeTypes: DiscourseNode[];
+}): Promise<SchemaFormatBackfill> => {
+  const probeRes = await supabaseClient
+    .from("my_concepts")
+    .select(SCHEMA_FORMAT_PROBE_SELECT)
```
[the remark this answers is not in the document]
sid597: Switched to `getAllPages` with `.order("id")` in 29d605af, matching the other `my_concepts` reads in this file. Expected scale is node types per space, so one page in practice, but you're right that a truncation at `max_rows = 1000` would have been silent and would never converge.

apps/roam/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -1069,6 +1079,62 @@ const getSharedRoamNodesWithFullContentUpdatesSince = async ({
...
+    orphaned,
+  });
+  if (backfilled === 0 && orphaned === 0) return;
+  const messages = [
+    `Backfilled format for ${backfilled} node type${backfilled === 1 ? "" : "s"}.`,
+    `${skipped} already had one.`,
+  ];
+  if (orphaned > 0) {
+    messages.push(`${orphaned} no longer match a node type in this graph.`);
+  }
```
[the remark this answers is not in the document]
sid597: Known shape, kept deliberately: it mirrors `reportCoreTitleBackfill` on #1332, which reads the same way in the orphaned-only case. The PR body tracks a joint fix for the two reports (recurring orphan toast, message composition) so they don't fork.

mdroidian (approved):

PR #1350, opened 2026-08-24
[ENG-2055] Handle duplicate node shortcuts

sid597:
## Reviewer brief

- Result: Creating a node type walks the label for the first character not already used as a shortcut, instead of leaving the shortcut unset on collision. If E is taken, EXPERIMENT gets X. If every character of the label is taken, the shortcut stays unset. This supersedes the blank-on-collision behavior from ENG-1741 (#1020).
- Review focus: The taken set spans all node types including the built-in Page/Block pseudo-nodes, whose p/b are live canvas tool shortcuts, so P and B are not auto-assigned. That is deliberate.
- Risk or follow-up: Nothing here surfaces or fixes shortcuts that are already duplicated in an existing graph; auto-assign only prevents new collisions. Pre-existing: on the legacy settings store, `importSharedRelations.ts` creates node types in a loop without refreshing the config tree, so auto-assign there reads stale shortcuts and can assign the same letter twice in one import.

## Verification

`pnpm ci:validate` green. 12 unit tests on `getFirstAvailableShortcut` cover taken-letter skipping, case-insensitive matching, accented Latin and Cyrillic labels, ASCII and non-ASCII digits, leading and interior punctuation, and the all-taken and empty-label fallbacks.

## Loom video

https://github.com/user-attachments/assets/d16451ea-620d-4831-ba34-6348ac7dad99

Recorded before the Shortcut column was removed, so the node list in the video shows a column that is no longer there. The auto-assign behavior it demonstrates is current.

## Scope check

- [x] Ran `$scope-check` against ENG-2055 and the final diff.

ENG-2055's `Done When` is the unfilled template placeholder, so the boundary is the Problem section plus Matt's comment on the ticket.

- [x] Ran a comprehensive review of the entire final diff in a subagent with a fresh context (2026-09-16). No blockers; its one should-fix, the util trusting its caller to uppercase the taken set, is fixed in 8c2a19e4.


mdroidian (approved):
The additional column was quite out of scope, will be replaced with the incoming settings redesign, and was misaligned.

Image

Please remove it before merging.

mdroidian:
@sid597 also make sure to run `$scope-check` for each PR

sid597:
Removed the Shortcut column in c57de52a. `DiscourseNodeConfigPanel.tsx` is byte-identical to the merge base again, so the PR is down to three files: the auto-assign walk in `accessors.ts`, the new `getFirstAvailableShortcut` util, and its tests. `pnpm ci:validate` green.

One gap named in the body's Risk line: with the column gone, nothing surfaces shortcuts that are already duplicated in an existing graph. Auto-assign only prevents new collisions.

sid597:
Ran `$scope-check`; the result is in the PR body under Scope check. ENG-2055's `Done When` is still the unfilled template placeholder, so I scoped against the Problem section plus Matt's comment on the ticket (if E is taken, EXPERIMENT gets X). The Shortcut column was the only thing past that line.

PR #1351, opened 2026-08-24
[ENG-2088] Add tag to block button on canvas block cards

sid597:
https://github.com/user-attachments/assets/b0b89b1f-ff35-4563-b549-21278e4ce614

Adds a `# Add tag` button to the header of block cards on the Roam canvas, per the DES-359 mock. It reuses `NodeMenu` in tag mode, so selecting a tag appends it to the end of the block text; a new optional `onTagAdded` callback on `NodeMenu` lets the shape update its own props (title + recalculated size) after the write, since nothing on the canvas watches block strings. The header buttons (sidebar, add tag, convert) now sit in one flex container instead of fixed absolute offsets — the large Convert-button hunk is indentation-only (`git diff -w`). The button only shows on live block shapes when at least one node type has a tag configured, and the key-remount on close mirrors `renderImageToolsMenu`.

Known limitation, left out of scope: each block card mounts a trigger-mode `NodeMenu`, which attaches document-level key listeners that no-op unless that menu is open; avoiding that would change shared `NodeMenu` behavior that the textarea flows depend on.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1351" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


apps/roam/src/components/canvas/DiscourseNodeUtil.tsx
```diff
@@ -619,114 +648,145 @@ export class DiscourseNodeUtil extends BaseBoxShapeUtil<DiscourseNodeShape> {
...
+            />
+
+            {/* Add Tag to Block Button */}
+            {extensionAPI && showAddTagButton && (
+              <NodeMenu
+                key={addTagMenuKey}
+                blockUid={shape.props.uid}
+                extensionAPI={extensionAPI}
+                onClose={() => setAddTagMenuKey((k) => k + 1)}
+                onTagAdded={handleTagAdded}
```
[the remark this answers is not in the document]
sid597: Addressed the feasible portion in 515d01b5: the pure tag-insertion logic is extracted to `insertTagIntoText` in `src/utils/formatUtils.ts` with unit tests in `src/utils/__tests__/formatUtils.test.ts`. The async block update and the shape title/size refresh are not unit-tested because `apps/roam`'s vitest harness is node-env, utils-only (no jsdom/component harness) — standing that up is out of scope here, so those paths are covered by the PR demo instead.

apps/roam/src/components/canvas/DiscourseNodeUtil.tsx
```diff
@@ -619,114 +662,127 @@ export class DiscourseNodeUtil extends BaseBoxShapeUtil<DiscourseNodeShape> {
...
-              </span>
-            </Button>
-          )}
+              title="Open in sidebar (Shift+Click)"
+            />
+
+            {/* Add Tag to Block Button */}
+            {extensionAPI &&
+              showAddTagButton &&
+              (isAddTagMenuOpen ? (
```
sid597, line 691: Lazy mount, per the NodeMenu-per-card comment: a closed card renders only the trigger button, so it registers no `document` key listeners and never runs NodeMenu's `getDiscourseNodes` memo. Unmount on close replaces the earlier `key`-bump remount; both reset `isOpen`, `showNodeTypes`, and `activeIndex` on the next open. The trigger is a single hoisted element (`addTagTrigger`, L632) used in both branches so the two states can't drift.

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -79,7 +86,7 @@ const NodeMenu = ({
   );
   const menuRef = useRef<HTMLUListElement>(null);
   const [activeIndex, setActiveIndex] = useState(0);
-  const [isOpen, setIsOpen] = useState(!trigger);
+  const [isOpen, setIsOpen] = useState(defaultIsOpen ?? !trigger);
```
sid597, line 90: `defaultIsOpen` exists so the canvas can mount NodeMenu already-open after its own trigger button is clicked. It is read only for the initial state; the three existing callers omit it and keep the previous `!trigger` behavior.

apps/roam/src/components/DiscourseNodeMenu.tsx
```diff
@@ -161,20 +168,18 @@ const NodeMenu = ({
...
 
-          void updateBlock({ text: newText, uid: targetBlockUid });
+          await updateBlock({ text: newText, uid: targetBlockUid });
           posthog.capture("Discourse Tag: Created via Node Menu", {
             tag,
           });
+          onTagAdded?.(newText);
         };
         // timeout required to ensure the block is updated
         setTimeout(() => void addTagToBlock(), 100);
```
sid597, line 189: This 100ms timer outlives NodeMenu's unmount by design (`onClose()` below fires immediately). The closure captures values only and calls back into the shape component, which stays mounted, so there is no setState-after-unmount.

apps/roam/src/utils/formatUtils.ts
```diff
@@ -180,3 +180,18 @@ export const findReferencedNodeInText = ({
 export const escapeCljString = (str: string) => {
   return str.replace(/\\/g, "\\\\").replace(/"/g, '\\"').toLowerCase();
 };
+
+export const insertTagIntoText = ({
```
sid597, line 184: Extracted from NodeMenu's `onSelect` so the insertion rules are unit-testable. One deliberate difference from the inline original: the `tag.replace(/^#/, "")` strip is dropped — the only caller reads `data-tag`, which is already stripped at the source (`DiscourseNodeMenu.tsx` MenuItem), so the strip was dead code.

sid597 (commented):
Annotations for the two follow-up commits (515d01b5, c8d92185) addressing the review round.

apps/roam/src/utils/formatUtils.ts
```diff
@@ -180,3 +180,18 @@ export const findReferencedNodeInText = ({
...
+  text,
+  tag,
+  selectionStart,
+}: {
+  text: string;
+  tag: string;
+  selectionStart: number;
+}) => {
+  const textToInsert = `${selectionStart === 0 ? "" : " "}#${tag}`;
```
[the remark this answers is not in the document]
sid597: Deliberate removal, not a miss — the strip was dead through the only call path. The sole caller reads `menuItem.getAttribute("data-tag")` (`DiscourseNodeMenu.tsx`), and that attribute is written already stripped: `data-tag={item.tag?.replace(/^#/, "")}`. So `insertTagIntoText` never receives a leading `#`, and the old inline strip could never fire. Removing it also makes the pathological `##Foo` tag config consistent: the menu label renders `##Foo` and insertion now produces `##Foo` too, where the double-strip previously inserted `#Foo` — a silent mismatch with what the user selected. See the annotation on this line in review 5075210227.

apps/roam/src/components/canvas/DiscourseNodeUtil.tsx
```diff
@@ -487,6 +490,33 @@ export class DiscourseNodeUtil extends BaseBoxShapeUtil<DiscourseNodeShape> {
       return null;
     }, [shape]);
 
+    const showAddTagButton =
+      getDiscourseNodeTypeId({ shape }) === "blck-node" &&
+      isLiveBlock(shape.props.uid) &&
+      !editor.isShapeOrAncestorLocked(shape) &&
```
sid597: Lock guard: tldraw's `updateShapes` silently skips locked shapes (inline `continue` in Editor.js 2.4.6, ~L5302–5309 — it checks `shape.isLocked` and `isShapeOrAncestorLocked` unless `_shouldIgnoreShapeLock`), and this toolbar is a DOM-level control that bypasses tldraw's interaction gating — so without this check the Roam write would succeed while the locked card stayed visibly stale. Hiding the affordance keeps lock semantics: a locked card offers no mutation. The guard passes `shape.id` (not the record) so the check does a captured store read in both lock states and the button reappears reactively on unlock. (The pre-existing Convert button has a sibling exposure — `deleteShapes` on a locked card is silently dropped, so converting duplicates the card; left out of scope as it predates this PR and deserves its own ticket.)

sid597:
@codex review

apps/roam/src/utils/formatUtils.ts
```diff
@@ -180,3 +180,21 @@ export const findReferencedNodeInText = ({
...
+  selectionStart,
+}: {
+  text: string;
+  tag: string;
+  selectionStart: number;
+}) => {
+  const textToInsert = `${selectionStart === 0 ? "" : " "}#${tag.replace(
+    /^#+/,
+    "",
+  )}`;
```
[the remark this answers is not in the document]
sid597: Fixed in 372f58c1. The separator is skipped when the character before the cursor is whitespace, with a unit test for the trailing-space case.

apps/roam/src/utils/formatUtils.ts
```diff
@@ -180,3 +180,21 @@ export const findReferencedNodeInText = ({
...
+
+export const insertTagIntoText = ({
+  text,
+  tag,
+  selectionStart,
+}: {
+  text: string;
+  tag: string;
+  selectionStart: number;
+}) => {
```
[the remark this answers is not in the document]
sid597: Fixed in 372f58c1, the helper now declares `: string`.

mdroidian (changes requested):
https://www.loom.com/share/3a76704bc5cf4e34b3a787eddc794c16

sid597:
Addressed the Loom in fdb011a7 (three commits):

- Restored the original plus icon and `Convert to [node type]` text. The Add tag trigger keeps the `# Add tag` label from DES-359.
- Add tag is hidden when the block already has a recognized discourse tag, so a card shows either Add tag or Convert to. After a tag is saved, the card title updates before resizing, so a sizing failure cannot leave Add tag visible.
- Mouse selection works. The menu's pointer-down no longer reaches the canvas through the React portal; tldraw was capturing the pointer before the item got its click. A DOM test fails without the fix and passes with it, and keyboard selection is covered too.

`pnpm ci:validate` green. A live Roam check of the mouse path is still to record.

sid597:
https://www.loom.com/share/5ba79384f7dd487990e267b62f2803d7

mdroidian (approved):

PR #1352, opened 2026-08-25
ENG-2000 footer defaults to mac icon for dock

trangdoan982:
`Object.defineProperty(navigator, "platform", { get: () => "Win32" })` to mock platform into Window results in
image

Refresh and my Mac correctly shows mac-appropriate icons
image

The advanced search footer showed the mac-only glyphs for option and command on every platform, so Windows and Linux users were told to press a key their keyboard does not have. 

| Action | Before (all platforms) | Now on mac | Now on Windows/Linux |
|---|---|---|---|
| dock results | ⌥⏎ | ⌥⏎ | `Alt`⏎ |
| insert | ⌘⏎ | ⌘⏎ | `Ctrl`⏎ |
| open / sidebar / close | ⏎ / ⇧⏎ / ⎋ | unchanged | unchanged |

New `keyboardHints` util mirrors the Obsidian plugin's [`keyboardHints.ts`](https://github.com/DiscourseGraphs/discourse-graph/blob/main/apps/obsidian/src/utils/keyboardHints.ts), including its testability trick — the pure function takes `isMac` so the non-mac branch is verifiable without that platform.

## Verification

- `pnpm test:unit` — 110 passed / 20 files, including 5 new cases covering both branches
- `pnpm check-types`, `pnpm lint` (0 errors), prettier — clean
- Rendered the real component through `react-dom/server` with `navigator.platform` stubbed, since the non-mac branch is unreachable from a Mac:

```
MacIntel   ⌥⏎ dock results | ⌘⏎ insert | ⏎ open | ⇧⏎ sidebar | ⎋ close
Win32      "Alt"⏎ dock results | "Ctrl"⏎ insert | ⏎ open | ⇧⏎ sidebar | ⎋ close
```

Mac output is byte-identical to before the change, which was the main risk.

**Not verified:** loading the built extension in a running Roam graph. Chrome's directory picker for the developer extension needs a real user gesture, so I could not complete that step. The rendered markup above is strong evidence but is not the real stylesheet — the pill widths and the `text-transform` are the parts that only a live load will fully confirm.

## Scope check

- [ ] Ran `$scope-check` against the ENG ticket and final diff.
- Scope beyond `Done When`: the ticket has an empty `Done When`. Beyond the footer fix, the diff touches `DiscourseNodeMenu.tsx` to deduplicate the platform check into a shared util rather than add a second copy, and adds a CSS rule for styling Tailwind cannot express here. Both are in service of the fix; say the word and I'll split either out.

Closes ENG-2000


apps/roam/src/styles/styles.css
```diff
@@ -221,3 +221,11 @@ body.roamjs-query-column-resizing * {
   .bp3-button.bp3-minimal[data-roamjs-canvas-page-ref-drag-handle="true"] {
   background-color: #394b59;
 }
+
+/* Not Tailwind classes: the extension borrows Roam's utilities and has no build
+   of its own, so arbitrary values never resolve. Keeps icon and text key hints
+   the same width, and opts out of the footer label's lowercasing. */
+.roamjs-footer-kbd {
+  min-width: 1.25rem;
```
mdroidian: `min-w-5`

apps/roam/src/styles/styles.css
```diff
@@ -221,3 +221,11 @@ body.roamjs-query-column-resizing * {
   .bp3-button.bp3-minimal[data-roamjs-canvas-page-ref-drag-handle="true"] {
   background-color: #394b59;
 }
+
+/* Not Tailwind classes: the extension borrows Roam's utilities and has no build
```
mdroidian: What?

apps/roam/src/styles/styles.css
```diff
@@ -221,3 +221,11 @@ body.roamjs-query-column-resizing * {
...
+   of its own, so arbitrary values never resolve. Keeps icon and text key hints
+   the same width, and opts out of the footer label's lowercasing. */
+.roamjs-footer-kbd {
+  min-width: 1.25rem;
+  text-transform: none;
```
mdroidian: `normal-case`

apps/roam/src/utils/platform.ts
```diff
@@ -0,0 +1,8 @@
+// Mirrors Blueprint's hotkeyParser so our hints match its hotkey labels.
+// roamAlphaAPI.platform can't substitute: it has no macOS flag. navigator.platform
+// is deprecated, but userAgentData is missing in Safari and Firefox.
+export const isMacOS = (): boolean => {
```
mdroidian, line 4: reuse `isMac`, etc, from `DiscourseNodeMenu`

mdroidian (approved):
Before merging
1) Fix the tailwind 
2) reuse isMac (move the isMac and it's required vars/functions into a util)

And maybe ask the agent if the cause of this was the new tailwind skill.
If so, exactly why?
And if not, send it a disappointed face and tell it you are thinking of switching to Codex.

apps/roam/src/styles/styles.css
trangdoan982: 👍 Switched.

apps/roam/src/styles/styles.css
trangdoan982: 👍 Switched.

apps/roam/src/styles/styles.css
trangdoan982: Gone — the rule it described is now `min-w-5` + `normal-case`.

apps/roam/src/utils/platform.ts
trangdoan982: That's what this is — `isMac` moved out of `DiscourseNodeMenu` into here, and it now imports `isMacOS` from this file (its line 27).

PR #1354, opened 2026-08-26
[ENG-2181] Advanced node search filters menu UX update

jsmorabito:
## Summary

- Node type filter checkboxes in the Obsidian node search now start **unchecked** instead of all-checked-by-default; checking one filters results to it directly.
- Removed the per-row **Only** button (and its hover-reveal styling) — checking a single box now does the same thing.
- Simplified `discourseNodeTypeFilter.ts`: `hasActiveTypeFilter` is now a plain `selectedTypeIds.length > 0`, and `toPanelSelectedIds` / `fromPanelSelectedIds` are deleted — they existed only to translate an empty selection into an "all checked" display, which no longer applies.

Linear: https://linear.app/discourse-graphs/issue/ENG-2181/advanced-node-search-filters-menu-ux-update

## Video

https://www.loom.com/share/8d0c7af989684c49b5c44d7003b66efe

## Test plan

- [x] `pnpm run check-types` — no new errors (pre-existing unrelated tldraw/canvas type errors only)
- [x] `pnpm exec eslint` on both changed files — clean
- [x] Manually tested in dev vault via Open node search → filter by type menu
- [ ] Before/after screenshots attached
- [x] Loom walkthrough linked

## Self-review

Checked against `STYLE_GUIDE.md`:
- Explicit return types, named-object params where >2 args, no `any` — all hold.
- No stray comments left from the removed "Only" button; top-of-file doc comment updated to describe the new checkbox semantics.
- One gap: `discourseNodeTypeFilter.ts` has no unit tests (none existed before this change either). Flagging rather than adding tests unasked, since that'd be scope beyond this ticket — happy to add a follow-up if wanted.

## Scope check

- [ ] Ran `$scope-check` against ENG-2181 and the final diff.
- Scope beyond `Done When`: Not documented — ENG-2181 has no written `Done When` section, only a linked Loom video (from DES-286) I don't have the ability to watch. The diff matches the plain-language request this PR was built from: default-unchecked filter checkboxes, direct per-type filtering on check, and removal of the "Only" button. No other behavior was changed.

## Summary by CodeRabbit

- **Improvements**
  - Simplified node-type filtering: selecting any node types now activates the filter, while clearing all selections removes it.
  - Filter counts and activity indicators now directly reflect the selected node types.
  - Removed the “Only” action from individual node-type filter options.

### DIFF
--- a/apps/obsidian/src/components/NodeTypeFilterMenu.tsx
+++ b/apps/obsidian/src/components/NodeTypeFilterMenu.tsx
@@ -6,25 +6,21 @@ import { getAllDiscourseNodeColors } from "~/utils/colorUtils";
 import {
   NODE_TYPE_FILTER_SEARCH_THRESHOLD,
   filterNodeTypesByQuery,
-  fromPanelSelectedIds,
   hasActiveTypeFilter,
-  toPanelSelectedIds,
 } from "~/utils/discourseNodeTypeFilter";
 
 const NodeTypeFilterRow = ({
   color,
   isChecked,
   nodeType,
-  onSelectOnly,
   onToggle,
 }: {
   color: string | undefined;
   isChecked: boolean;
   nodeType: DiscourseNode;
-  onSelectOnly: () => void;
   onToggle: () => void;
 }): ReactElement => (
-  <div className="hover:bg-modifier-hover group flex items-center gap-2 px-3 py-1.5">
+  <div className="hover:bg-modifier-hover flex items-center gap-2 px-3 py-1.5">
     <label className="flex min-w-0 flex-1 cursor-pointer items-center gap-2">
       <input
         type="checkbox"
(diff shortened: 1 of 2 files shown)


jsmorabito:
@coderabbitai full review

trangdoan982:
I'm not sure if this is the desired design. When we made this proposal for the Node Summoning Menu in Roam, I remember being asked specifically for the ability to make it easy to filter "all but one node", "all", or "only one." Not sure if there's any user context that's changed since then? But if it has, and we agreed on this, the code looks fine to me.

jsmorabito:
> I'm not sure if this is the desired design. When we made this proposal for the Node Summoning Menu in Roam, I remember being asked specifically for the ability to make it easy to filter "all but one node", "all", or "only one." Not sure if there's any user context that's changed since then? But if it has, and we agreed on this, the code looks fine to me.

I haven't done a design pass on the node summoning menu's filter menu so I wont comment on its implementation, but in the context of this particular filter menu for advanced node search I agree with Michaels sentiment here https://linear.app/discourse-graphs/issue/ENG-2181/advanced-node-search-filters-menu-ux-update that it's more straightforward to reverse the behavior of filters.

I don't think I misinterpreted @mdroidian from that video, but tagging so he can double check my loom 😁

trangdoan982:
i'll defer to MG to review then

mdroidian:
Let's continue this discussion in Linear.

mdroidian:
Changing to draft as the conversation continues in linear.

PR #1356, opened 2026-08-27
FEE-86 Using the canvas Roam guide

KarolaKirsanow:
This adds a user doc in roam/guides on "using the canvas".

This PR also deletes the empty "Experiment tracking" stub, now replaced by a use-case article


jsmorabito (approved):
Guide content looks great — clean structure, good use of Callouts/Image components, matches the sentence-case style convention.

Pushed a follow-up commit removing a few things that looked accidental and shouldn't ship to the public docs site:

* `tech-writing-folder/` (an internal Claude writing-style skill, not user docs)
* `Tutorial-Using the Canvas.md` (a raw legacy draft referencing third-party Firebase-hosted images from unrelated graphs)
* `canvas-tut04.png` / `canvas-tut06.png` (orphaned, unreferenced by the .mdx guide)
* the unrelated `next.config.ts` `experimental.cpus` tweak

Also fixed a capitalization inconsistency in `getting-started.md` ("Discourse Graph" → "discourse graph" to match the rest of the list).

Approving.

---

Claude wrote the above 😁 I looked it over and its great! Claude just had these little catches which i had it implement

apps/website/content/roam/guides/using-the-canvas.mdx
```diff
@@ -0,0 +1,313 @@
...
+import Image from "next/image";
+import { Callout } from "nextra/components";
+
+The Canvas lets you lay out your work visually. It's great for synthesizing your work into a story: you can arrange nodes you've already created into a canvas layout, or generate new nodes directly from the canvas itself.
+
+## Creating a canvas
+
+### From anywhere
+
+Type `jj` followed by `canvas` anywhere in your graph to create and title a canvas. The canvas is created as a separate page in your graph.
```
mdroidian, line 17: @KarolaKirsanow This requires both SmartBlocks to be installed and the specific "canvas" smartblock to also be installed in someones graph, correct?

apps/website/content/roam/guides/using-the-canvas.mdx
```diff
@@ -0,0 +1,313 @@
...
+
+<Callout type="info" emoji="💡">
+  You can also use this SmartBlock button for one-click canvas creation:
+  `{{Create canvas:SmartBlock:Create New Canvas:RemoveButton=false,Icon=presentation}}`.
```
mdroidian, line 21: @KarolaKirsanow This would require that a SmartBlock named "Create New Canvas" exists in their graph.  Maybe we should mentioned/link to that?

Otherwise it won't work.

apps/website/content/roam/guides/using-the-canvas.mdx
```diff
@@ -0,0 +1,313 @@
...
+    alt="The Key Image setting in Discourse Graph Settings"
+    width={660}
+    height={644}
+    className="docs-bordered-image"
+    sizes="300px"
+    style={{ width: 300 }}
+  />
+</Callout>
+
+<Callout type="info" emoji="💡">
```
mdroidian, line 168: @KarolaKirsanow FYI, this field expects a reference to a saved query, rather than the image block itself. The query runs for each canvas node, and its first result determines where to look for the image.

For example, to use an image nested under an H1 block block on each node’s page:

  1. Create a query with these conditions:

```
FIND node WHERE
 node is in page with title :in NODETEXT
 node has parent h1
 h1 has heading 1
```


  2. Save the query with an alias such as canvas-key-image.
  3. Enable Key image in the node type’s Canvas settings, select Query builder reference, and enter that alias or a block reference to the saved query.

`:in NODETEXT` receives the current canvas node’s title, so the query finds the appropriate block on each node’s page. The first image within that block or its children becomes the key image.

I don't think this was ever documented, sorry!

apps/website/content/roam/guides/using-the-canvas.mdx
KarolaKirsanow: Good catch! The first use-case guides I wrote recommend installing SmartBlocks but there's no reason to expect a reader to have encountered those first. 🤦‍♀️Once again made foolish by the curse of knowledge …
 #iss-candidate: most of our workflows have an implied reliance on SmartBlocks and we should clarify whether that plugin is Strongly Recommended

apps/website/content/roam/guides/using-the-canvas.mdx
KarolaKirsanow: Flat-out  omission on my part. Will add this link to the template.

apps/website/content/roam/guides/using-the-canvas.mdx
KarolaKirsanow: opened a [new PR](<https://github.com/DiscourseGraphs/discourse-graph/pull/1461>) to address these changes and add cloud canvas documentation

PR #1369, opened 2026-08-31
ENG-2211 Fix Roam direct canvas load race

mdroidian:
Direct cloud canvas loads can crash when another extension changes the shared React hook during startup. DG now captures Roam's React exports once, keeping the initial useSyncExternalStore implementation stable across renders. Roam provides React 18.2.0, so DG no longer injects a React 17 fallback.

- Use the published roamjs-components 0.90.0 release, which removes the loader's global hook assignment. Remove the temporary dependency patch.
- Preserve Roam's React functions, internal dispatcher, and renderer.
- Remove the deferred canvas renderer and retain normal production minification.
- Test that the initial hook is retained and later global replacements do not affect DG.

Limitation: if another extension replaces the hook before DG initializes, DG captures that implementation. This change prevents later replacements; it does not recover an overwritten native hook.

Validation: pnpm install --frozen-lockfile, pnpm ci:validate (including 112 Roam tests), and the production build passed with published roamjs-components 0.90.0. Live verification on daeb12c0 passed with Breadcrumbs and SmartBlocks enabled: two direct cloud-canvas reloads, a direct local-canvas reload, and daily-note navigation into the cloud canvas. No console warnings or errors were recorded, and DG settings confirmed the loaded commit. Extension settings were preserved. GitHub format, lint, validate, and deployment checks passed.

Upstream: https://github.com/RoamJS/roamjs-components/pull/67 (merged and published).


mdroidian:
@codex review

PR #1373, opened 2026-08-31
ENG-2189 Reorganize Roam settings: replace the Personal/Global split with four topic-based sections

trangdoan982:
https://www.loom.com/share/d7c223af47a34951a4bd8bfca32ece51

## Reviewer brief

**Result:** the settings dialog groups settings by topic (Preferences / Features / Grammar / Advanced) instead of by who they affect. `Home` is gone, `Left sidebar` and `Canvas` appear once each, and nine legacy tab ids alias to their new homes so saved deep links keep resolving.

**Follow-up**

- The scope indicator in the ticket's Solution is not here. Deferred to ENG-2183 and ENG-2187. 
- Attributes and Overlay stay on the node page. My taxonomy missed this.
<details>
<summary> ## Verification </summary>

- `turbo check-types` 8/8 workspaces, eslint 0 errors (516 warnings, unchanged from base), 142/142 existing unit tests pass.
- Driven in Roam on `discourse-dev-graph`: four sections, no duplicate tab titles, the enable toggle gates the left sidebar sections, the Favorites gear menu opens `features-left-sidebar`, `Ctrl+Shift+A` still opens the admin panel.
- Legacy dual-write, with `Use new settings store = false`: editing `trigger` in Preferences > General and `Canvas Page Format` in Features > Canvas each updated both the block prop and the legacy block, and neither overwrote the other. Both restored afterwards.
</details>

## Scope check

- [x] Ran `$scope-check` against ENG-2189 and the final diff.
- Scope beyond `Done When`:
  - **Row anchors** (`utils/settingAnchor.ts`, written by 8 rows, read by nothing). Serves [ENG-2184: Add search to the settings panel](https://linear.app/discourse-graphs/issue/ENG-2184/add-search-to-the-settings-panel). Included now because this PR relocates every row, and addressing a row by setting key rather than by tab is what survives relocation. Deferring means ENG-2184 edits all the same rows again. Agreed with @trang while planning, not recorded in Linear.
  - **Left sidebar placement.** The taxonomy files it under `Preferences > General`. It ships under `Features` with its enable toggle on the tab, because a tab that hides itself when disabled cannot host the toggle that re-enables it. Agreed with @trang, not recorded in Linear.
  - **Copy changes**, which the ticket excludes but the merge forces: `Personal node menu trigger` becomes `Personal override` under a `Node trigger` heading, `Left Sidebar` becomes `Enable left sidebar` (legacy block key unchanged), and the Favorites gear menu's `Global Section` and `Personal Section` become one `Left sidebar` entry now that both open the same tab.
  - **New behaviour:** the personal override's placeholder shows the effective graph-wide trigger. This is the Solution's "show a graph-wide default beside its personal override", but it is new code rather than a move.
  - **`Nodes` section header renamed to `Node types`**, so nothing appears twice in the nav. The Grammar tab is also titled `Nodes`. Pre-existing collision, and the section disappears in ENG-2186.

## Summary by CodeRabbit

- **New Features**
  - Redesigned Settings with grouped sections for Preferences, Features, Grammar, and Advanced options.
  - Added dedicated settings for canvas behavior, discourse context overlays, left-sidebar configuration, and styling preferences.
  - Added customizable placeholder text for node trigger controls.
  - Added direct navigation to specific settings and improved settings organization.
- **Improvements**
  - Consolidated left-sidebar settings into a single, easier-to-find entry.
  - Preserved compatibility with existing settings links and navigation.
  - Improved settings headings, grouping, and visual layout.


apps/roam/src/components/settings/utils/useLegacyConfigUids.ts
```diff
@@ -0,0 +1,46 @@
...
+import discourseConfigRef from "~/utils/discourseConfigRef";
+import refreshConfigTree from "~/utils/refreshConfigTree";
+import {
+  getUidAndBooleanSetting,
+  getUidAndStringSetting,
+} from "~/utils/getExportSettings";
+import { DISCOURSE_CONFIG_PAGE_TITLE } from "~/data/constants";
+
+/** Shared so the three tabs writing under this parentUid cannot collide on order. */
+export const LEGACY_CONFIG_ORDER = {
```
trangdoan982: These three settings dual-write: block props plus a mirror into the legacy Roam block tree. The mirror only happens when a panel is given `uid`, `parentUid` and `order`, and `Use new settings store = false` (the shipping default) reads from that tree.

Before this PR all three were resolved in one `useMemo` in `GeneralSettings.tsx` with `order={0,1,2}`. They now render from three different files (Preferences > General, Features > Canvas, Features > Left sidebar) while still writing under the same `parentUid`, so the orders live here rather than in each file. Two panels sharing an order would create blocks at the same position and overwrite each other.

Verified live with the flag off: editing `trigger` and `Canvas Page Format` each updated both stores, and neither clobbered the other block.

apps/roam/src/components/settings/utils/settingsTabs.ts
```diff
@@ -0,0 +1,36 @@
...
+  advancedQueries: "advanced-queries",
+  advancedExport: "advanced-export",
+  admin: "secret-admin-panel",
+} as const;
+
+export const DEFAULT_SETTINGS_TAB_ID: TabId =
+  SETTINGS_TAB_IDS.preferencesGeneral;
+
+/** Tab ids from before the taxonomy. Saved deep links still carry these. */
+export const SETTINGS_TAB_ALIASES: Record<string, TabId> = {
```
trangdoan982, line 20: Nine legacy tab ids, one per tab that existed before the taxonomy. Deep links that users saved, plus the three in-repo call sites, resolve through here.

Both "Home" tabs collapse onto Preferences > General because that is where most of their rows landed. Both left sidebar tabs collapse onto the single merged tab.

No unit test covers this map. A wrong entry fails silently by dropping the user on the default tab, so it is worth reading against the tab list in `Settings.tsx`.

apps/roam/src/components/settings/Settings.tsx
```diff
@@ -191,93 +179,95 @@ export const SettingsDialog = ({
               tabId: String(id),
             });
           }}
-          selectedTabId={visibleTabId}
+          selectedTabId={activeTabId}
```
trangdoan982, line 172: This was `selectedTabId={visibleTabId}`. The `visibleTabId` guard existed only to bounce the selection off the two `hidden={!leftSidebarEnabled}` left sidebar tabs when the feature was off.

Those two tabs are now one tab that is never hidden, because it hosts the toggle that enables the feature. Nothing can strand the selection, so the guard has no remaining case to protect.

apps/roam/src/components/settings/LeftSidebarSettings.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const legacyUids = useLegacyConfigUids();
+  const [isAlertOpen, setIsAlertOpen] = useState(false);
+  return (
+    <div className="flex flex-col gap-6 p-1">
+      {/* `title` is the legacy block key; the visible label is set separately. */}
+      <FeatureFlagPanel
+        title="Left Sidebar"
+        label="Enable left sidebar"
+        description="Whether or not to enable the left sidebar."
+        featureKey="Enable left sidebar"
```
trangdoan982, line 32: This toggle used to live on Global > Home, one tab away from the settings it controls, because this tab hid itself when the feature was off.

Moving it here means the tab can no longer be feature-gated: a tab that hides itself cannot host the control that un-hides it. The sections it gates are still gated, so no setting becomes reachable that was not before.

apps/roam/src/components/settings/utils/settingAnchor.ts
```diff
@@ -0,0 +1,11 @@
+const SETTING_ANCHOR_ATTRIBUTE = "data-setting-id";
+
+/**
+ * Marks a setting row so it can be scrolled to by setting key. Addressing rows
+ * by key rather than by tab keeps the address stable when a setting moves tabs.
+ */
+export const settingAnchor = (
```
trangdoan982, line 4: Out of scope for ENG-2189, in service of ENG-2184 (settings search). Called out here so it is not reviewed against this ticket.

The decision worth checking is the addressing scheme: a row is addressed by its setting key path, never as `<tab>/<field>`. A tab-derived address would be invalidated by this very PR, and by every future relocation. Setting key paths already exist in `settingKeys.ts` and do not change when a setting moves tabs.

Nothing reads the attribute yet.

apps/roam/src/components/settings/GeneralSettings.tsx
```diff
@@ -1,102 +0,0 @@
...
-import posthog from "posthog-js";
-import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
-import {
-  getUidAndBooleanSetting,
-  getUidAndStringSetting,
-} from "~/utils/getExportSettings";
-import { DISCOURSE_CONFIG_PAGE_TITLE } from "~/data/constants";
-import { ROAM_DOCS, withDocsLink } from "./utils/docs";
-
-const DiscourseGraphHome = ({
```
trangdoan982, line 20: Deleted. Its three rows moved to their taxonomy homes:

- `trigger` to `PreferencesGeneral.tsx`, as the graph-wide half of the Node trigger group
- `Canvas Page Format` to `CanvasSettings.tsx`
- `Enable left sidebar` to `LeftSidebarSettings.tsx`

The `useMemo` resolving the three legacy block uids became `utils/useLegacyConfigUids.ts`, because all three panels still need it and they no longer share a file.

mdroidian:
@coderabbitai full review

apps/roam/src/components/settings/components/BlockPropSettingPanels.tsx
```diff
@@ -34,6 +35,9 @@ type RoamBlockSyncProps = {
   parentUid?: string;
   uid?: string;
   order?: number;
+  /** Legacy block text when it must differ from `title`; readers such as
```
mdroidian, line 38: Can you give an example of this?

apps/roam/src/components/settings/utils/settingsTabs.ts
```diff
@@ -0,0 +1,36 @@
...
+  grammarRelations: "grammar-relations",
+  advancedQueries: "advanced-queries",
+  advancedExport: "advanced-export",
+  admin: "secret-admin-panel",
+} as const;
+
+export const DEFAULT_SETTINGS_TAB_ID: TabId =
+  SETTINGS_TAB_IDS.preferencesGeneral;
+
+/** Tab ids from before the taxonomy. Saved deep links still carry these. */
```
mdroidian, line 19: Can you give an example of a "saved deep link"?

apps/roam/src/components/settings/utils/settingsTabs.ts
```diff
@@ -0,0 +1,36 @@
...
+  "canvas-shortcuts-personal-settings": SETTINGS_TAB_IDS.featuresCanvas,
+  "left-sidebar-personal-settings": SETTINGS_TAB_IDS.featuresLeftSidebar,
+  "left-sidebar-global-settings": SETTINGS_TAB_IDS.featuresLeftSidebar,
+  "discourse-graph-export": SETTINGS_TAB_IDS.advancedExport,
+  "discourse-nodes": SETTINGS_TAB_IDS.grammarNodes,
+  "discourse-relations": SETTINGS_TAB_IDS.grammarRelations,
+};
+
+/** Unknown ids pass through: per-node tabs are keyed by node page uid. */
+export const resolveSettingsTabId = (id?: TabId): TabId => {
```
mdroidian, line 33: Why is `id` undefined?  Can you give an example when we won't know the id?  If we always know the id, or just want an easy default, then set the default in the prop instead of (id === undefined)

apps/roam/src/components/settings/utils/useLegacyConfigBlocks.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/** Block text is what discourseConfigRef.ts reads back by; kept with `order` so
+ *  panels writing under one parentUid can neither collide nor drift. */
+const LEGACY_CONFIG_BLOCKS = {
+  trigger: { blockKey: "trigger", order: 0 },
+  canvasPageFormat: { blockKey: "Canvas Page Format", order: 1 },
+  leftSidebarFlag: { blockKey: "(BETA) Left Sidebar", order: 2 },
+} as const;
+
+/** Spreads straight onto a panel's block-sync props. */
```
mdroidian, line 18: ?

apps/roam/src/components/settings/utils/useLegacyConfigBlocks.ts
```diff
@@ -0,0 +1,57 @@
...
+  uid: string | undefined;
+  parentUid: string;
+};
+
+type LegacyConfigBlocks = Record<
+  keyof typeof LEGACY_CONFIG_BLOCKS,
+  LegacyConfigBlock
+>;
+
+/** `Use new settings store = false` (the default) still reads this tree. No
```
mdroidian, line 31: The default changed in [ENG-2173: Use props-based settings by default](https://linear.app/discourse-graphs/issue/ENG-2173/use-props-based-settings-by-default)

mdroidian:
@codex review

apps/roam/src/components/settings/CanvasShortcutSettings.tsx
```diff
@@ -103,45 +103,33 @@ const CanvasShortcutSettings = ({
...
-                    updateShortcut(node.type, { value })
-                  }
-                />
-              );
-            })}
-          </div>
-        }
-      />
-    </Tabs>
+    <div className="inline-grid grid-cols-[auto_auto] items-center gap-x-4 gap-y-2 p-1">
```
mdroidian, line 106: Is `grid-cols-[auto_auto]` recognized in roam?

apps/roam/src/components/settings/DiscourseContextSettings.tsx
```diff
@@ -0,0 +1,57 @@
...
+        onChange={(checked) => {
+          void setSetting("discourse-context-overlay", checked);
+          onPageRefObserverChange(overlayHandler)(checked);
+          posthog.capture("Personal Settings: Overlay Toggled", {
+            enabled: checked,
+          });
+        }}
+      />
+      <PersonalFlagPanel
+        title="(BETA) Overlay in canvas"
```
mdroidian, line 42: Note: This feature is being removed [ENG-2096: Remove the discourse context overlay from Roam canvas](https://linear.app/discourse-graphs/issue/ENG-2096/remove-the-discourse-context-overlay-from-roam-canvas)

apps/roam/src/components/settings/LeftSidebarSettings.tsx
mdroidian: One of the benefits we are losing would be hiding the `Left Sidebar` tab when disabled.
Also, this is likely very poor UX for non admin users, who go to the tab and see a blank screen when disabled. (as the "Enable left sidebar" should/will be admin gated).

I believe we should adhere to the original scoping, but I'm not sure if we re-clarified where "feature flags" should live.

apps/roam/src/components/settings/Settings.tsx
mdroidian: This is a feature we should keep (hiding feature tabs that are disabled).

apps/roam/src/styles/settingsStyles.css
```diff
@@ -18,3 +18,21 @@
   padding: 0 10px;
   width: 100%;
 }
+
+/* Outranks the 14px row labels by size and by the only rule on the page.
+   Hand-rolled: Roam ships no Tailwind preflight, so `border-b-2` never paints. */
```
mdroidian, line 23: "Roam ships no Tailwind preflight"?

apps/roam/src/styles/settingsStyles.css
```diff
@@ -18,3 +18,21 @@
...
+.dg-settings-heading {
```
mdroidian, line 24: Why are we not using tailwind?

apps/roam/src/styles/settingsStyles.css
```diff
@@ -18,3 +18,21 @@
...
+  font-size: 16px;
+  font-weight: 600;
+  line-height: 1.3;
+  letter-spacing: -0.01em;
+  color: var(--dg-neutral-dark, #1f1f1f);
+  padding-bottom: 8px;
+  border-bottom: 2px solid var(--dg-border, rgb(229, 231, 235));
+}
+
+.dg-settings-group {
```
mdroidian, line 34: Why are we not using tailwind?

apps/roam/src/styles/settingsStyles.css
```diff
@@ -18,3 +18,21 @@
...
+/* Outranks the 14px row labels by size and by the only rule on the page.
+   Hand-rolled: Roam ships no Tailwind preflight, so `border-b-2` never paints. */
+.dg-settings-heading {
+  font-size: 16px;
+  font-weight: 600;
+  line-height: 1.3;
+  letter-spacing: -0.01em;
+  color: var(--dg-neutral-dark, #1f1f1f);
+  padding-bottom: 8px;
+  border-bottom: 2px solid var(--dg-border, rgb(229, 231, 235));
```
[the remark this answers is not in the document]
mdroidian: @trangdoan982

mdroidian (changes requested):

apps/roam/src/components/settings/CanvasShortcutSettings.tsx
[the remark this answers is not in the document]
trangdoan982: Added `settingAnchor([PERSONAL_KEYS.canvasNodeShortcuts])` on the grid wrapper.

apps/roam/src/styles/settingsStyles.css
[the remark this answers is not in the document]
trangdoan982: Fixed: `--dg-neutral-dark` / `--dg-border` were defined nowhere, so only the fallbacks ever painted. The heading now uses the Blueprint palette with a `.bp3-dark` override, the pattern already used in `styles.css`.

apps/roam/src/components/settings/PreferencesGeneral.tsx
```diff
@@ -283,6 +179,34 @@ const HomePersonalSettings = ({
...
+              "Override the global trigger for the discourse node menu.",
+              ROAM_DOCS.creatingNodes,
+            )}
+          />
+          <NodeMenuTriggerComponent
+            extensionAPI={extensionAPI}
+            initialValue={
+              personalSettings[PERSONAL_KEYS.personalNodeMenuTrigger]
+            }
+            placeholder=""
```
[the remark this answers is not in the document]
trangdoan982: Not taking the suggested fix: the personal override is an `IKeyCombo` (key + modifiers), while `GLOBAL_KEYS.trigger` is a trigger character, so showing it in this input would be misleading. The blankness is real though — the `placeholder=""` and the prop that carried it are reverted, so the field shows "Click to set trigger" again, as it did before this PR.

apps/roam/src/components/settings/utils/settingsTabs.ts
trangdoan982: Undefined when the dialog opens without a target tab — the plain "Open settings" command palette entry passes no `selectedTabId`. Moved the default into the parameter as suggested.

apps/roam/src/components/settings/utils/useLegacyConfigBlocks.ts
trangdoan982: Right — ENG-2173 is on `main` but not yet in this stack’s base, so the comment was stale either way. Reworded to "Graphs with `Use new settings store` off still read this tree" rather than naming a default.

apps/roam/src/components/settings/utils/useLegacyConfigBlocks.ts
trangdoan982: The comment was too terse. Reworded: the type matches the Roam block-sync props (`parentUid`, `uid`, `order`, `blockKey`) that the BlockPropSettingPanels panels take, so a panel can spread one of these objects directly.

apps/roam/src/components/settings/LeftSidebarSettings.tsx
trangdoan982: I see your point about hiding the tab when the setting is disabled, but I'm not sure if we necessarily need to stick to it. My design opinion is:
- it's sufficient to hide the rest of the actual settings if it's disabled. It's easier to find one setting in one tab than to spread it into different tabs. I imagine if user wants to enable this, they have to find the source feature flag first, then navigate to the right setting tab -> too many steps 
    + this is also what Linear does with its settings. [Example](https://linear.app/docs/use-cycles)
- re your point about seeing blank screen when disabled: would it be better if we move the entire section into a subsection under General settings?
- when it's admin-gated we can just disable if the user isn't admin. or have a separate tab for all admin settings that are disabled for non-admin users.

open to hear your thoughts on these suggestions

apps/roam/src/components/settings/components/BlockPropSettingPanels.tsx
trangdoan982: this change is for the legacy setting, because the block text is the title. But we're renaming some of the settings, such as:
- Node trigger: the panel title is now Graph-wide default, but the block text stays trigger. 
- Enable left sidebar: the panel title is Enable left sidebar, the block text is (BETA) Left Sidebar.

apps/roam/src/components/settings/LeftSidebarSettings.tsx
trangdoan982: #decided in 1-1 with MG on Sept 17: keep the "Left sidebar" toggle in General settings -> allow the Left sidebar tab to hide when disabled

trangdoan982:
@mdroidian will re-tag you on this PR once I push the changes for left sidebar up. No action required from you on this PR right now

trangdoan982:
@mdroidian Loom test for restoring the existing visibleTabId with left sidebar:

https://www.loom.com/share/e9261fdfdd784cb09b9100534f2cf74e

mdroidian (approved):

PR #1374, opened 2026-08-31
ENG-2185 Move export settings out of Settings into the Export panel

trangdoan982:
https://www.loom.com/share/0c24f4b276f049d29f5c21bbeb67083a

https://www.loom.com/share/d794a7965df54225bf5b5bb78ac1dadb

## Reviewer brief

- **Result:** the 7 export options are set from a collapsed `Export options` disclosure at the bottom of the Export dialog's `Export` tab. The Export tab is gone from Settings. Storage keys, the `getExportSettings` read path, and export behaviour are untouched, so saved values carry over.
- **Review focus:** `ExportOptions` was moved with `git mv`, so it reads as a rename. The only content changes are the import paths (it is no longer inside `settings/`) and the component name.

<details>
<summary> Verification </summary>

- `pnpm install --frozen-lockfile` then `pnpm ci:validate` from the repo root: exit 0. `check-types` across 12 packages, 142 tests in 23 files passing.
- ESLint on every changed `apps/roam` file: 0 errors. `Export.tsx` (40) and `BlockPropSettingPanels.tsx` (3) report the same warning counts before and after this change (stashed and re-ran to compare), so no new ones. One `exhaustive-deps` warning this change introduced in `Settings.tsx` is fixed rather than left.
- Dev bundle rebuilt clean; `dist/extension.js` contains the new disclosure and posthog event, and no longer contains the `discourse-graph-export` tab id.
</details>

## Scope check

- Scope beyond `Done When`:
     - **`apps/website/content/roam/guides/sharing-discourse-graph.md`** said the options live on the `Export` tab of the configuration, which this change makes wrong.

- [ ] Ran a comprehensive review of the entire final diff in a subagent with a fresh context. Use `$dg-delegated-full-review` when no other full-review workflow is available.

## Known gap: export can read a stale setting

The pending-write flush this PR used before (`await flushPendingSettingWrites()` in the export path) has been removed. It and its registry are tracked in ENG-2286, which **blocks this ticket**.

Until ENG-2286 lands, editing an export option and pressing Export within about 350 ms reads the option at its previous value. The export settings and the Export button share this dialog, so no unmount happens on that path and the commit-on-unmount landed in ENG-2186 (#1375) does not cover it.

The underlying debounce bug is ENG-2287.


apps/roam/src/components/settings/Settings.tsx
```diff
@@ -256,14 +274,6 @@ export const SettingsDialog = ({
...
           />
-          <Tab
-            id="discourse-graph-export"
-            title="Export"
-            className="overflow-y-auto"
-            panel={
-              <DiscourseGraphExport globalSettings={settings.globalSettings} />
-            }
-          />
           <Tab
```
trangdoan982, line 250: The `Export` tab was removed here. Its contents moved verbatim to `apps/roam/src/components/ExportOptions.tsx` (a `git mv` of `settings/ExportSettings.tsx`, so it shows as a rename in this diff) and now render inside the Export dialog's Export tab. No control, setting key, or default changed in the move.

apps/roam/src/components/Export.tsx
```diff
@@ -225,6 +231,22 @@ const ExportDialog: ExportDialogComponent = ({
...
+    [],
+  );
+
+  // The export option panels write legacy config blocks alongside block props, so
+  // the cached config tree has to be refreshed the way SettingsDialog does. Gated
+  // on the section having been opened because refreshConfigTree re-reads every
+  // node page and re-registers the datalog translators, which is too heavy to run
+  // on every close of a dialog that is opened for each export.
+  const closeDialog = (): void => {
+    if (exportOptionsOpened.current) refreshConfigTree();
```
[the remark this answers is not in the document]
trangdoan982: Valid, fixed in 28c63a179 — with two corrections to the scope.

Frontmatter is not affected: `BaseMultiTextPanel.handleAdd` calls `setter` synchronously before it touches Roam. `BaseTextPanel` (the 250ms one) is not used by any of the seven export controls. The two that were genuinely exposed are **max filename length** and **link type**, both `BaseNumberPanel`/`BaseSelectPanel`, which defer only the block-prop write by 100ms and cleared the timer on unmount. So the window was 100ms rather than 350ms, but the two failure modes you describe were both real: export within the window read the old value, and Cancel within it dropped the edit.

Fix is the flush you suggested rather than removing the debounce, since committing per keystroke would write to Roam on every event. New `apps/roam/src/utils/pendingSettingWrites.ts` holds a registry of pending commits; a `useDeferredWrite` hook keeps each panel timer and its registry entry in step so a commit runs exactly once whether it fires by timer, by flush, or by unmount; unmount now commits instead of cancelling. The Export dialog flushes before running the export. Five unit tests cover the registry, including self-deregistration during a flush and a commit that registers follow-up work.

This also fixed the same data loss in Settings, where closing the dialog right after editing a number or select could drop the write.

Separately, the `closeDialog` wrapper your prompt points at is gone. It only existed to call `refreshConfigTree` on close, and every panel already refreshes the tree after its own write, so it was redundant.

Still unverified in a live graph — noted in the PR body.

mdroidian:
@codex review

apps/roam/src/components/settings/utils/settingsTabs.ts
```diff
@@ -24,7 +23,8 @@ export const SETTINGS_TAB_ALIASES: Record<string, TabId> = {
   "canvas-shortcuts-personal-settings": SETTINGS_TAB_IDS.featuresCanvas,
   "left-sidebar-personal-settings": SETTINGS_TAB_IDS.featuresLeftSidebar,
   "left-sidebar-global-settings": SETTINGS_TAB_IDS.featuresLeftSidebar,
-  "discourse-graph-export": SETTINGS_TAB_IDS.advancedExport,
+  // Export options now live in the Export dialog.
```
mdroidian: This comment will not age well.  Let's just remove it.

apps/roam/src/components/Export.tsx
```diff
@@ -225,6 +231,21 @@ const ExportDialog: ExportDialogComponent = ({
     if (initialPanel) setSelectedTabId(INITIAL_PANEL_TO_TAB_ID[initialPanel]);
   }, [initialPanel, sharingEnabled]);
   const [includeDiscourseContext, setIncludeDiscourseContext] = useState(false);
+  const [exportOptionsOpen, setExportOptionsOpen] = useState(false);
+  const exportOptionsOpened = useRef(false);
+  // Collapse unmounts the panels while closed, so each open seeds them from current values.
+  const exportGlobalSettings = useMemo(
+    () => bulkReadSettings().globalSettings,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
mdroidian, line 239: Could we read the current settings explicitly when opening the options and store that snapshot in state? That would make the intent clearer and avoid the extra useMemo dependency and lint suppression.

apps/roam/src/components/Export.tsx
```diff
@@ -225,6 +231,21 @@ const ExportDialog: ExportDialogComponent = ({
...
+    [exportOptionsOpen],
+  );
+
+  // Option panels also write legacy config blocks, so refresh the tree as SettingsDialog
+  // does, but only if they were opened: refreshConfigTree re-reads every node page.
```
mdroidian: This comment should explain *WHY* we need to do this.  And that reason is because we need export function to re-read the new values, because we are not passing the changed values to the export function.

apps/roam/src/components/Export.tsx
```diff
@@ -951,41 +972,53 @@ const ExportDialog: ExportDialogComponent = ({
           />
         </Label>
 
-        <div className="flex items-end justify-between">
+        <div className="mt-2 flex items-center justify-between gap-4">
+          <Button
+            minimal={true}
+            small={true}
+            icon={exportOptionsOpen ? "chevron-down" : "chevron-right"}
```
mdroidian, line 979: We could use `rightIcon` to fix the alignment issue

Image

apps/website/content/roam/guides/sharing-discourse-graph.md
```diff
@@ -17,9 +17,7 @@ Demo:
 
 ## Export Options
 
-We have a range of options for customizing the markdown export. These can be found on the `Export` tab of the discourse graph configuration.
-
-![](/docs/roam/settings-export.png)
+We have a range of options for customizing the markdown export. These can be found under `Export options` on the `Export` tab of the export dialog, alongside the export you are about to run.
```
mdroidian, line 20: Maybe change one or two of these "Export"s, otherwise it reads like "export ... export ... export ... export".

"Change export dialog to "share dialog" and remove ", alongside the export you are about to run"

mdroidian (approved):

PR #1375, opened 2026-08-31
ENG-2186 Grammar › Nodes drill-down and settings navigation primitive

trangdoan982:
## Reviewer brief

- **Result:** the rail no longer grows one tab per node type. `Grammar › Nodes` opens a list of node types. Clicking one opens that node's page in place, with a back button and a `Grammar › Nodes › <node>` breadcrumb. The node page itself is unchanged here (still the tabbed `NodeConfig`); ENG-2188 (#1420) rebuilds it on top of this PR.
- **Review focus: the route.** Navigation state is a path array (`["grammar-nodes", nodeTypeUid]`) owned by a reducer with four actions: `select-tab`, `push`, `pop`, `truncate`. Blueprint `Tabs` only sees `path[0]`; the Nodes panel reads the rest through `SettingsNavContext`. Deep links still work: `resolveInitialSettingsPath` treats any id that is not a registered tab as a node type uid and opens it inside `Grammar › Nodes`. A uid that no longer matches a node falls back to the list.
- **No settings-write changes.** This PR is presentation only. The deferred-write rework, the template buffer flush and its delete ordering, and the node colour commit-on-unmount have all been reverted out. The only non-navigation change left is an optional `title` on the template panel, which the drill-down needs to render it without a label.
- **Known bug this PR makes easier to hit (ENG-2287):** settings panels debounce their write and cancel it on unmount. Back and the breadcrumbs sit next to the field and are keyboard-reachable, so typing and leaving within ~350ms drops the edit; between 250ms and 350ms it writes the legacy block but not the new store, leaving the two disagreeing. Previously the only exit was a deliberate tab click, reliably slower than that. ENG-2286 separately covers flush-before-read.

## Verification

- `tsc --noEmit` and eslint report no errors or warnings for `apps/roam`. Production build completes with 0 errors.
- Checked in Roam: the rail lists no node types; list → node page → Back and breadcrumb both return to the list; openers that pass a node type uid as `selectedTabId` land on that node's page; typing and then pressing Back immediately keeps the edit.

## Loom video

https://www.loom.com/share/ffd805d8bbfa47f5ae418d0a85c79906

## Scope check

- Scope beyond `Done When`: none. The write plumbing has been removed to ENG-2286 and ENG-2287.

### DIFF
--- a/apps/roam/src/components/settings/GrammarNodesRoute.tsx
+++ b/apps/roam/src/components/settings/GrammarNodesRoute.tsx
@@ -3,7 +3,6 @@ import { OnloadArgs } from "roamjs-components/types";
 import getDiscourseNodes, {
   excludeDefaultNodes,
 } from "~/utils/getDiscourseNodes";
-import { formatHexColor } from "./DiscourseNodeCanvasSettings";
 import { useSettingsNav } from "./navigation/SettingsNavContext";
 import SettingsPageHeader from "./navigation/SettingsPageHeader";
 import DiscourseNodeConfigPanel from "./DiscourseNodeConfigPanel";
@@ -39,9 +38,6 @@ const GrammarNodesRoute = ({
         ancestorLabels={NODES_ANCESTOR_LABELS}
         rootLabel="Nodes"
         resolveLabel={resolveLabel}
-        dotColor={
-          formatHexColor(node?.canvasSettings?.color ?? "") || undefined
-        }
       />
       <div className="dg-settings-route__body">
         {node ? (
--- a/apps/roam/src/components/settings/GrammarNodesRoute.tsx
+++ b/apps/roam/src/components/settings/GrammarNodesRoute.tsx
@@ -1,0 +1,59 @@
+import React, { useEffect } from "react";
+import { OnloadArgs } from "roamjs-components/types";
+import getDiscourseNodes, {
+  excludeDefaultNodes,
+} from "~/utils/getDiscourseNodes";
+import { formatHexColor } from "./DiscourseNodeCanvasSettings";
+import { useSettingsNav } from "./navigation/SettingsNavContext";
+import SettingsPageHeader from "./navigation/SettingsPageHeader";
+import DiscourseNodeConfigPanel from "./DiscourseNodeConfigPanel";
+import NodeConfig from "./NodeConfig";
+
+const NODES_ANCESTOR_LABELS = ["Grammar"] as const;
+
+const GrammarNodesRoute = ({
+  onloadArgs,
+}: {
+  onloadArgs: OnloadArgs;
+}): JSX.Element => {
+  const { segments, goToDepth } = useSettingsNav();
+  const nodes = getDiscourseNodes().filter(excludeDefaultNodes);
+
+  const [nodeTypeUid] = segments;
+  const node = nodeTypeUid
+    ? nodes.find((n) => n.type === nodeTypeUid)
+    : undefined;
+
+  // A deleted node type or stale deep link resolves to nothing; return to the list.
+  const isStalePath = Boolean(nodeTypeUid) && !node;
+  useEffect(() => {
+    if (isStalePath) goToDepth(0);
+  }, [isStalePath, goToDepth]);
+
+  const resolveLabel = (segment: string): string =>
+    nodes.find((n) => n.type === segment)?.text ?? segment;
+
+  return (
+    <div className="dg-settings-route">
+      <SettingsPageHeader
+        ancestorLabels={NODES_ANCESTOR_LABELS}
+        rootLabel="Nodes"
+        resolveLabel={resolveLabel}
+        dotColor={
+          formatHexColor(node?.canvasSettings?.color ?? "") || undefined
+        }
+      />
+      <div className="dg-settings-route__body">
+        {node ? (
+          <NodeConfig node={node} onloadArgs={onloadArgs} />
+        ) : (
+          <div className="p-1">
+            <DiscourseNodeConfigPanel />
+          </div>
+        )}
+      </div>
+    </div>
+  );
+};
+
+export default GrammarNodesRoute;
--- a/apps/roam/src/components/settings/components/EphemeralBlocksPanel.tsx
+++ b/apps/roam/src/components/settings/components/EphemeralBlocksPanel.tsx
@@ -16,7 +16,8 @@ import type { DiscourseNodeBaseProps } from "./BlockPropSettingPanels";
 const DEBOUNCE_MS = 250;
 const TEMPLATE_BUFFER_TEXT = "Template";
 
-type DualWriteBlocksPanelProps = DiscourseNodeBaseProps & {
+type DualWriteBlocksPanelProps = Omit<DiscourseNodeBaseProps, "title"> & {
+  title?: string;
   uid: string;
   defaultValue?: InputTextNode[];
 };
@@ -115,7 +116,7 @@ const DualWriteBlocksPanel = ({
     const newUid = window.roamAlphaAPI.util.generateUID();
     const dv = defaultValueRef.current;
     const seed: InputTextNode[] = dv && dv.length > 0 ? dv : [{ text: " " }];
-    void createBlock({
+    const created = createBlock({
       node: { text: TEMPLATE_BUFFER_TEXT, uid: newUid, children: seed },
       parentUid: nodeType,
       order: "last",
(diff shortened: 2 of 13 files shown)


apps/roam/src/components/settings/components/EphemeralBlocksPanel.tsx
```diff
@@ -181,13 +202,14 @@ const DualWriteBlocksPanel = ({
 
     return () => {
       cancelled = true;
-      window.clearTimeout(debounceRef.current);
+      // Navigating away lands right after a keystroke, and the buffer block is deleted next.
+      flushPendingChanges();
```
trangdoan982, line 207: Pre-existing bug this PR has to fix rather than inherit: the cleanup used to `clearTimeout` the pending 250ms write without running it, and the buffer effect then deleted the block carrying that edit. Only a deliberate tab click unmounted this before, so the window was rarely hit. A back button and a breadcrumb are fast, keyboard-adjacent unmount paths landing right after a keystroke, which turns it into a reproducible edit loss. Active only when `isNewSettingsStoreEnabled()`.

Known limit: this recovers an edit Roam has already committed and debounced. A keystroke Roam has not committed yet is still lost.

apps/roam/src/components/settings/utils/settingsNavigation.ts
```diff
@@ -0,0 +1,89 @@
...
+};
+
+const KNOWN_TAB_IDS: readonly string[] = Object.values(SETTINGS_TAB_IDS);
+
+/**
+ * Resolves the tab a caller asked for, including the aliases in `settingsTabs`. An id that
+ * is not a tab at all is a node type uid from when every node type had its own rail tab;
+ * those links now open the node's page inside Grammar > Nodes.
+ */
+export const resolveInitialSettingsPath = (
```
trangdoan982, line 50: This is the ENG-2186 "existing deep links still resolve" requirement. `resolveSettingsTabId` (ENG-2189) maps the renamed tabs; its passthrough comment assumes per-node tabs still exist, keyed by node page uid. This PR removes those tabs, so an id that is not a known tab is now read as a node type uid and opens that node inside Grammar › Nodes instead of selecting a tab that no longer exists.

Deliberately no separate deep-link API: `selectedTabId` plus this resolution covers it, and ENG-2189 already addresses individual rows via `settingAnchor`.

apps/roam/src/styles/settingsStyles.css
```diff
@@ -18,3 +18,70 @@
   padding: 0 10px;
   width: 100%;
 }
+
+:root {
+  /* Brand secondary, mirroring packages/tailwind-config. */
+  --dg-secondary: #5f57c0;
+}
+
+/* Hand-rolled, not Tailwind: apps/roam has no Tailwind build (tailwind.config.ts sets
```
trangdoan982: AGENTS.md says prefer Tailwind, so this needs justifying. `apps/roam/tailwind.config.ts` sets `content: []` and is commented "required for editor support" — there is no Tailwind build for this app, and the built `extension.css` contains no utilities. Classes only work when Roam's own stylesheet happens to ship them.

Found the hard way: a `<h3 className="m-0">` here fell back to host heading margins and opened a ~440px gap. `flex-shrink: 0` below is load-bearing for the same class of reason — without it the header is a shrinkable flex item and overlaps the scrolling body.

apps/roam/src/components/settings/GrammarNodesRoute.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const [nodeTypeUid, subPage] = segments;
+  const node = nodeTypeUid
+    ? nodes.find((n) => n.type === nodeTypeUid)
+    : undefined;
+
+  // A deleted node type or stale deep link resolves to nothing; return to the list.
+  const isStalePath = Boolean(nodeTypeUid) && !node;
+  useEffect(() => {
+    if (isStalePath) goToDepth(0);
```
trangdoan982, line 29: The path outlives this panel — `renderActiveTabPanelOnly` unmounts and remounts it on every tab switch — so it can point at a node type deleted in the meantime, or arrive stale from a saved link. `goToDepth(0)` rather than `pop()` so a depth-2 stale path converges in one dispatch instead of two.

apps/roam/src/components/settings/navigation/SettingsPageHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  return (
+    <div className="dg-settings-page-header">
+      <div className="dg-settings-page-header__back">
+        <Button
+          minimal
+          small
+          icon="chevron-left"
+          text={parent.label}
+          onClick={pop}
```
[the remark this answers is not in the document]
trangdoan982: Valid, fixed in b441825. `BaseTextPanel` and `DiscourseNodeColorSetting` now run the pending write during unmount cleanup instead of clearing the timer. Both keep their `errorRef` guard, so an invalid tag, format or shortcut is still not persisted.

The same defect in `EphemeralBlocksPanel` was already fixed earlier in this PR. You are right that Back generalises it: before this branch these editors only unmounted on a deliberate tab click, so the window was rarely hit.

apps/roam/src/utils/settingsNavigation.ts
```diff
@@ -0,0 +1,139 @@
...
+export const serializeSettingsPath = (path: SettingsPath): string =>
+  path.map(encodeURIComponent).join("/");
+
+export const parseSettingsPath = (raw: string): SettingsPath | null => {
+  const parts = raw
+    .split("/")
+    .filter(Boolean)
+    .map((part) => decodeURIComponent(part));
+  return parts.length === 0 ? null : parts;
+};
```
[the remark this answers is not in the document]
trangdoan982: No longer applies. This branch was rebased onto ENG-2189 and `apps/roam/src/utils/settingsNavigation.ts` no longer exists — the module moved to `apps/roam/src/components/settings/utils/settingsNavigation.ts` and the string-path API it flagged was removed entirely.

ENG-2189 supplies `resolveSettingsTabId` with an alias map, so there is no `parseSettingsPath` and no `decodeURIComponent` left to throw on a malformed escape.

apps/roam/src/components/settings/navigation/SettingsNavContext.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { createContext, useContext, useMemo } from "react";
+import {
+  depthOf,
+  segmentsOf,
+  type SettingsNavAction,
+  type SettingsPath,
+} from "../utils/settingsNavigation";
```
mdroidian, line 7: ~

apps/roam/src/components/settings/navigation/SettingsPageHeader.tsx
```diff
@@ -0,0 +1,69 @@
+import React from "react";
+import { Button } from "@blueprintjs/core";
+import { buildBreadcrumbTrail } from "../utils/settingsNavigation";
```
mdroidian, line 3: ~

apps/roam/src/styles/settingsStyles.css
```diff
@@ -36,3 +36,48 @@
   flex-direction: column;
   gap: 4px;
 }
+
+:root {
+  /* Brand secondary, mirroring packages/tailwind-config. */
+  --dg-secondary: #5f57c0;
+}
+
+/** Hand-rolled: apps/roam has no Tailwind build, so only utilities Roam ships resolve. */
```
mdroidian, line 45: "apps/roam has no Tailwind build"?

Please use tailwind.

apps/roam/src/utils/pendingSettingWrites.ts
```diff
@@ -0,0 +1,26 @@
+import { settleTrackedRoamWrites } from "./setBlockProps";
```
mdroidian: I was not expecting to see a writing utils file on a purely UI change that was explicitly stated as not a functional change.

As this is PR is over the limit and outside of scope, let's separate the functional change with the UI movement change and create a ticket specifically for the functional change explaining what specific problem it is solving.

mdroidian (changes requested):
I’m blocking this on scope and size. [ENG-2186](https://linear.app/discourse-graphs/issue/ENG-2186/grammar-nodes-drill-down-and-settings-navigation-primitive) covers the node list and drill-down navigation. The proposal explicitly limits M1 to presentation changes without changing how settings work.

Please remove the settings-write refactor and export flush infrastructure from this PR and track them in a separate ticket.

The persistence issue appears to be pre-existing. If the new Back button makes it easier to encounter, call that out in the separate ticket with the specific failure, reproduction steps, and acceptance criteria. That will make the problem and proposed fix much easier to review.

If any part is needed to prevent a regression introduced by this PR, please identify that dependency explicitly.

apps/roam/src/styles/settingsStyles.css
trangdoan982: You are right — the comment was wrong. Roam ships Tailwind, which is why the rest of `apps/roam` styles with utilities.

Dropped all the `dg-settings-*` rules and moved to Tailwind in `cade293c`:

- route: `flex h-full flex-col` / `min-h-0 flex-auto overflow-y-auto`
- header: `mb-5 block flex-shrink-0`, title `mb-1 flex items-center gap-2 text-2xl font-semibold leading-tight`, trail `text-sm leading-snug text-gray-500`, crumb `cursor-pointer`

`settingsStyles.css` is back to just the Blueprint tab override. The trail now uses `text-gray-500` (the repo idiom for muted text) instead of the old `opacity: 0.6`.

apps/roam/src/components/settings/navigation/SettingsNavContext.tsx
trangdoan982: Switched to the alias in `0ae32bc6`: `~/components/settings/utils/settingsNavigation`. These were the only two `../` imports this PR added — `apps/roam/src` is 593 `~/` to 7 `../`, so they were the outliers. (The `../utils/zodSchema` in `BlockPropSettingPanels.tsx` predates this PR, from #887, so I left it alone.)

apps/roam/src/components/settings/navigation/SettingsPageHeader.tsx
trangdoan982: Switched to the alias in `0ae32bc6`: `~/components/settings/utils/settingsNavigation`. These were the only two `../` imports this PR added — `apps/roam/src` is 593 `~/` to 7 `../`, so they were the outliers. (The `../utils/zodSchema` in `BlockPropSettingPanels.tsx` predates this PR, from #887, so I left it alone.)

apps/roam/src/components/settings/components/EphemeralBlocksPanel.tsx
trangdoan982: removed per review

trangdoan982:
@mdroidian i removed the deferred write settings. for this PR all changes are just cosmetic reorganization. I recorded the follow-up ticket: [ENG-2286: Settings writes: pending-write registry and flush-before-read](https://linear.app/discourse-graphs/issue/ENG-2286/settings-writes-pending-write-registry-and-flush-before-read)

PR #1377, opened 2026-08-31
ENG-2183 Build the new setting-item row component

trangdoan982:
## Reviewer brief

**Result:** one shared row renders every panel-based setting as scope badge · label · always-visible description · trailing control. Descriptions leave the hover popover, so their doc links are reachable (ENG-2080).

**Review focus.**

1. **The seam.** `SettingTitle` was the one helper all five `Base*Panel`s funnelled label and description through, so rewiring it converts ~70 rows with no call-site edits. The risk is that it also changes ~70 rows at once: flag rows now render `Switch` instead of `Checkbox`, and text/number/select controls move from below the label to trailing, losing `fill`.
2. **Why the row uses a raw `<label>` and flex.** Not cosmetic. Measured against live Roam CSS: `label.bp3-label .bp3-popover-wrapper` is `display:block; margin-top:5px` at specificity (0,2,1), and `.bp3-icon` is `vertical-align:text-bottom`. Inside a Blueprint `<Label>` the scope badge lands **19.25px** out of line; with the raw label plus flex it is **1px**. A comment in `SettingItemRow.tsx` records this — please keep it if you restructure the markup.
3. **Scope is derived, not passed.** It comes from the wrapper, since the wrapper already binds the setter that decides who a value is written for. Consequence worth checking: `PersonalTextPanel` and `PersonalNumberPanel` accept a `setter` override, so a caller overriding the setter must also pass `scope` or the badge will contradict where the value lands. Noted in code.
4. **Tailwind's reach in this app.** Utilities do resolve, because Roam itself bundles Tailwind — but on a 14px rem base, and **without preflight**. So `border-b` and `border` set a width that never paints, and this repo's `neutral-*` theme colours are absent from Roam's build entirely. Both are hand-rolled in `settingsStyles.css` rather than papered over. This corrects the comment added in #1375, which concluded utilities do not apply at all.

**Risk or follow-up**

- **No visual verification inside the extension.** The dev extension would not load in any tab automation could reach, because Roam needs its directory handle re-granted through a native picker. Alignment, badge fill and row dividers were instead verified by injecting the row's exact markup into live Roam and measuring; the rows as rendered *by the extension* are unverified. The `Checkbox`→`Switch` swap and the loss of `fill` on `NumericInput`/`HTMLSelect` are the parts most worth a human look.
- `withDocsLink`'s `stopPropagation` is deliberately kept. `CanvasShortcutSettings` still nests a description inside a `<label>`, so it remains load-bearing.
- Two pre-existing issues found, neither introduced nor fixed here: `shortcutValue` in `NodeConfig.tsx` is assigned and never used (arrived with #1375), and `EphemeralBlocksPanel`'s container uses `border border-gray-200`, which is inert for the preflight reason above — that box has no visible border today.

## Verification

- `turbo check-types` 8/8 workspaces; eslint 0 errors, 515 warnings (unchanged from base); prettier clean; 142/142 roam, 16/16 content-model, 39/39 database unit tests; dev build `0 errors` with the new code confirmed present in the bundle.
- Measured against live Roam CSS by injecting the row's rendered markup: badge-to-label-text centre offset 1px, versus 19.25px for the Blueprint `<Label>` structure; badge fill `rgb(241,241,241)` and `1px solid` border painting; row divider `0px` on the first row and `1px` on subsequent ones.
- Not covered: no unit tests. `apps/roam`'s vitest is `environment: "node"` and includes only `src/utils/__tests__`, so a render test would need jsdom and RTL added; deferred by agreement. The doc-link fix therefore has no automated guard.
- Not verified: the rows as the extension actually renders them (see above).

## Loom video

Not recorded.

## Scope check

- [x] Ran `$scope-check` against ENG-2183 and the final diff.
- Scope beyond `Done When`:
  - **Rewiring the five `Base*Panel`s.** ENG-2183 excludes migrating call sites, and no call site changes here — the edit is internal to `BlockPropSettingPanels.tsx`. Included because a row with no consumers cannot be verified at all: this repo has no Storybook. Agreed with @trang while planning.
  - **Migrating seven bespoke `<Label>` rows** (Personal override, Discourse tool shortcut, Default filters, Page groups, node Color, Specification, Template blocks). Requested by @trang so that every setting row carries a scope badge, which the ticket's Solution asks for but ENG-2187 owned. Not recorded in Linear.
  - **Deriving scope at the wrapper.** This delivers most of ENG-2187's "populate the scope indicator" for the ~70 panel rows, leaving that ticket the remaining bespoke rows. Cheaper here than a second pass over the same files.
  - **`Checkbox` → `Switch`** on flag rows, and controls moving to trailing. Both are what the mockups show; neither is stated in `Done When`.
  - **Superseding #1375's `SettingsDrillDownRow`.** Deleted, and its two callers in `NodeConfig.tsx` now compose the shared row. Requested by @trang to avoid two drill-down rows landing in the same stack. Its `buttonText` values were action labels rather than value summaries, so `SettingDrillDownSummary`'s contract documents both uses.
  - **`CanvasShortcutSettings` deliberately not migrated.** It is a two-column grid where each node contributes two grid cells; converting it would restructure a compact override table into stacked rows. Read as one personal setting with per-node overrides rather than N setting items. Flagged for @trang rather than changed.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1377" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


apps/roam/src/components/settings/components/SettingsDrillDownRow.tsx
```diff
@@ -1,31 +0,0 @@
-import React from "react";
```
trangdoan982: Deleted, not lost. This row is superseded by `SettingItemRow` + `SettingDrillDownSummary`, and its two callers in `NodeConfig.tsx` (Index, Template) now compose those instead.

Worth knowing while reviewing: its `buttonText` values were action labels (`See all Claim nodes`, `Edit template`) rather than value summaries. Rather than quietly stretch the new prop, `SettingDrillDownSummary`'s `summary` doc now covers both uses, since a list or a template has no single scalar value to show.

The visible change is the control going from `outlined` + `intent=PRIMARY` to `minimal` with a chevron, plus both rows gaining a `nodeType` scope badge.

apps/roam/src/components/settings/NodeConfig.tsx
```diff
@@ -116,38 +110,41 @@ const DiscourseNodeColorSetting = ({
...
+            <Tooltip content={color ? "Unset" : "Color not set"}>
+              <Icon
+                className={"ml-2 align-middle opacity-80"}
+                icon={color ? "delete" : "info-sign"}
+                onClick={() => {
+                  if (colorWriteTimeoutRef.current) {
+                    window.clearTimeout(colorWriteTimeoutRef.current);
+                    colorWriteTimeoutRef.current = null;
+                  }
+                  pendingColorRef.current = null;
```
trangdoan982, line 139: Conflict-resolution note. This line is #1375's (`Flush pending debounced writes when navigating away`); this PR restructured the surrounding `<Label>` into `SettingItemRow` at the same time, so the two collided here on rebase.

Both are kept deliberately: the row structure is mine, the `pendingColorRef` reset is #1375's. If you are diffing against that branch, this is the one hunk where the resolution was manual rather than automatic.

PR #1378, opened 2026-08-31
ENG-2184 Add search to the settings panel

trangdoan982:
https://www.loom.com/share/3a670cb80b624ac284e0d3e82e2826bc

## Reviewer brief

- **Result:** A search field at the top of the settings rail matches settings by name, section/group and description, and jumps to the matching row wherever it currently lives — including per-node-type settings behind the Grammar › Nodes drill-down. Tabs are offered as their own result kind, so "export" reaches the Export page even though no setting is named that.

- **Stacked on #1379 (ENG-2213)**, which records what every setting is called and where it lives. This PR is only the search layer that reads it: ranking, the field, and the jump.

- **Review focus — the jump retries by frame.** The target row is not in the DOM when the jump is dispatched: `renderActiveTabPanelOnly` means the destination panel mounts on a later commit, and a row inside a `Collapse` mounts later still. `useSettingAnchorScroll` therefore polls by animation frame until the row appears or a ~500ms budget runs out, then degrades to leaving the user on the right page. A single lookup would silently miss.

<details> <summary> **Risk or follow-up:** </summary>
  - **No tests.** Ranking tiers, node-type expansion and availability gating are unverified by CI.
  - **Not verified in Roam: the transparency fix itself.** Roam stopped re-registering the developer extension after the reload needed to pick up the new CSS. Worth one manual look — open settings, type anything, confirm the panel has a solid background.
  - Escape stops propagation only while the query is non-empty, so a first Escape clears the search and a second closes the dialog. Intentional, and easy to read as a bug.
</details>

<details> <summary> Verification </summary>

Manual and local only.

- `pnpm ci:validate` (check-types + 142 unit tests) and eslint on the changed files — 0 errors, no new warnings on added lines.
- Driven in a real Roam graph on this build (version stamp `…eng-2184-add-search-to-the-settings-panel`): the field renders in the rail and returns correct labels and breadcrumbs. That run is what caught the transparent popover.
- Position, contrast and keyboard scrolling were measured in a throwaway browser harness that mounted the real component inside a real vertical Blueprint `Tabs` carrying the same overflow rules as `Settings.tsx`. Measured: opaque 340px surface; opens rightward from the field's left edge; active-row breadcrumb at 4.53:1 on Blueprint's `#137CBD` (80% opacity measured 3.49:1, below AA — hence full white); ArrowDown scrolls the active row into view. The harness was not kept.
</details>

## Scope check

- Scope beyond `Done When`:
  - **Description matching.** The ticket scoped this out; requested during implementation. The text lives on the ENG-2213 entry, so search and the row read the same string.
  - **Dead `settingAnchorId` export removed** — added earlier in this branch's history, never used.


apps/roam/src/components/settings/utils/settingsCatalog.ts
```diff
@@ -0,0 +1,549 @@
...
+    path: rootPath(SETTINGS_TAB_IDS.featuresCanvas),
+  },
+  {
+    keys: [PERSONAL_KEYS.autoCanvasRelations],
+    label: "Auto canvas relations",
+    description:
+      "Automatically add discourse relations to canvas when a node is added",
+    docsLink: ROAM_DOCS.storedRelations,
+    keywords: ["personal", "tldraw"],
+    path: rootPath(SETTINGS_TAB_IDS.featuresCanvas),
```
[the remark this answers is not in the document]
trangdoan982: Canvas shortcut overrides are now authored in the catalog (#1379, `PERSONAL_KEYS.canvasNodeShortcuts`), with an anchor on the override grid so a result can land on it. Node attributes stay exempt deliberately: they are edited as a set by `DiscourseNodeAttributes` and there is no single row to jump to — recorded in `NonRowKeyId`.

apps/roam/src/components/settings/navigation/useSettingAnchorScroll.ts
```diff
@@ -0,0 +1,64 @@
...
+    const look = () => {
+      const target = document.querySelector(settingAnchorSelector(anchorId));
+      if (target) {
+        target.scrollIntoView({ block: "center", behavior: "smooth" });
+        target.classList.add(SETTING_ANCHOR_FLASH_CLASS);
+        flashed = target;
+        flashTimeout = setTimeout(() => {
+          target.classList.remove(SETTING_ANCHOR_FLASH_CLASS);
+        }, FLASH_DURATION_MS);
+        settle();
```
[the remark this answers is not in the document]
trangdoan982: Fixed: the flash now owns its own timeout instead of living in the effect cleanup, so settling the jump no longer strips it. Re-flashing a row that is still animating restarts the animation rather than riding out the first one. See `useSettingAnchorScroll.ts` on the current head.

apps/roam/src/components/settings/navigation/useSettingAnchorScroll.ts
```diff
@@ -0,0 +1,64 @@
...
+      rafId = requestAnimationFrame(look);
+    };
+
+    rafId = requestAnimationFrame(look);
+    return () => {
+      cancelAnimationFrame(rafId);
+      if (flashTimeout) clearTimeout(flashTimeout);
+      flashed?.classList.remove(SETTING_ANCHOR_FLASH_CLASS);
+    };
+  }, [anchorId, onSettled]);
```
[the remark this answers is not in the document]
trangdoan982: Confirmed and fixed. The flash class is now owned by `flashRow`, whose timeout is independent of the effect; the effect cleanup only cancels the frame-by-frame lookup. Re-flashing the same row restarts the animation via a forced reflow.

apps/roam/src/components/settings/utils/settingAnchor.ts
```diff
@@ -9,3 +9,9 @@ export const settingAnchor = (
 ): Record<string, string> => ({
   [SETTING_ANCHOR_ATTRIBUTE]: settingKeys.join("/"),
 });
+
+/** Setting keys are authored identifiers, so quoting is enough to build a selector. */
+export const settingAnchorSelector = (anchorId: string): string =>
+  `[${SETTING_ANCHOR_ATTRIBUTE}="${anchorId.replace(/"/g, '\\"')}"]`;
```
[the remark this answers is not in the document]
trangdoan982: Applied in 07f219a7: backslashes are escaped before quotes. Worth saying for the record that anchor ids are authored keys from `settingKeys.ts`, so nothing user-typed reaches this selector today — the escape is there so a future key cannot break it, not to fix a live case.

mdroidian:
@codex review

apps/roam/src/components/settings/navigation/SettingsSearchField.tsx
```diff
@@ -0,0 +1,175 @@
...
+  InputGroup,
+  Menu,
+  MenuItem,
+  Popover,
+  Position,
+} from "@blueprintjs/core";
+import {
+  buildSettingsCatalog,
+  type SearchableEntry,
+} from "../utils/settingsCatalog";
```
mdroidian, line 13: ~

apps/roam/src/components/settings/navigation/SettingsSearchField.tsx
```diff
@@ -0,0 +1,175 @@
...
+import { rankSettings } from "../utils/settingsSearch";
```
mdroidian, line 14: ~

apps/roam/src/components/settings/navigation/useSettingAnchorScroll.ts
```diff
@@ -0,0 +1,65 @@
+import { useEffect } from "react";
+import {
+  SETTING_ANCHOR_FLASH_CLASS,
+  settingAnchorSelector,
+} from "../utils/settingAnchor";
```
mdroidian, line 5: ~

apps/roam/src/styles/settingsStyles.css
```diff
@@ -109,3 +110,46 @@
...
+    0 0 0 1px rgba(16, 22, 26, 0.1),
+    0 2px 4px rgba(16, 22, 26, 0.2),
+    0 8px 24px rgba(16, 22, 26, 0.2);
+}
+
+.dg-settings-search__results .bp3-menu {
+  background-color: transparent;
+}
+
+.dg-settings-search__scroll {
```
mdroidian, line 135: please use tailwind

apps/roam/src/styles/settingsStyles.css
```diff
@@ -109,3 +110,46 @@
   text-align: center;
   font-family: monospace;
 }
+
+/** Inset to the tab titles' left edge. */
+.dg-settings-search {
```
mdroidian, line 115: please use tailwind

mdroidian (approved):
The codex comments are worth addressing.

Please ask your agent why it isn't adhering to the `~` requirement in AGENTS.md.

Here's some general feedback.
https://www.loom.com/share/7c3e6e7eab2944829ef5f5d11cce5d03

- [ ] Make repeated setting results distinguishable by including the node type in the main label, e.g. “Issue > Index.”
- [ ] Avoid matching shared breadcrumb terms on repeated node settings. Searching “Nodes” or “Grammar” should show the relevant page, not a list of Color and other repeated entries.
- [ ] Add a direct page result for each node type. Searching “Evidence” should show its settings page first; its individual settings can appear below.
- [ ] Adjust the highlight spacing so the background and colored edge have padding around the row content on both sides.  (or drop the color strip)
- [ ] Add ability to search for canvas shortcut overrides (a single entry is fine, not an entry for each shortcut)
- [ ] Index relation names so searching for a relation, such as “Informs,” returns a result that opens its settings.

Not all required for this PR, but I imagine we should create a place to capture this feedback from all team members.

PR #1380, opened 2026-09-02
ENG-2141 Map the Roam referenced node to `source` on push

sid597:
## Reviewer brief

- Result: publish drops the `sourceDocument` slot when the referenced Source is not a Concept in the space, warns with the page name, and orders the upsert so an in-batch Source precedes the node that references it. An imported Source publishes its origin RID instead of its local uid.
- Review focus: the existence check in `omitMissingSource` (`publishNodesToGroups.ts`). Decision points are marked as inline comments.
- Risk or follow-up:
  - An imported RID skips the existence check. The DB resolves it under the publisher's current read access, so a revoked grant or a deleted origin concept still fails the row. Open question: accept for v0, or add a client check via `rid_to_space_id_and_local_id`?
  - The warning is `console.warn` only. Open question: also count omitted sources on `PublishNodesResult` and show it in the Export toast?
  - Already-synced Evidence that points at an imported Source repoints from the Roam mirror concept to the origin concept on the next sync. Intended, but it changes existing rows.
  - Draft PR #1334 turns an unresolvable slot from a failed row into a silently empty one. This PR avoids sending unresolvable values, so it does not depend on either behavior.
  - Nothing reads `CrossAppNode.slots` on the Obsidian side yet. ENG-2140 owns that.

## Verification

- Roam: `tsc`, Prettier, and 173 unit tests pass. `turbo run test:unit` passes across workspaces.
- ESLint reports 7 pre-existing directive warnings on untouched lines, none new.
- New tests: source present, absent, in the same batch, imported RID, malformed RID, repeated publish, one lookup per source.
- Driven against a local database in the demo video: on the main build the Evidence row fails to publish, on this branch it publishes with an empty `reference_content` and a warning that names the Source, and after the Source is published the slot resolves to its concept row. The imported-RID path is covered by unit tests only.

##  video

https://github.com/user-attachments/assets/14538da1-db6f-4753-af71-ccbaf20052ad

## Scope check

- [x] Ran `$scope-check` against ENG-2141 and the final diff.

- A `pre-pr-review` pass ran before the last fixes (source uid dedupe, RID guard, RID parse check). The full pass on the pushed diff is still to run.


apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -311,14 +321,33 @@ export const publishNodesToGroups = async ({
...
 
-  const upsertConcepts = [
-    ...missingNodeSchemas.map((s) => crossAppNodeSchemaToDbConcept(s)),
-    ...[...nodesByUid.values()].map((node) => crossAppNodeToDbConcept(node)),
-    ...missingRelationTripleSchemas.map((rs3) =>
-      crossAppRelationTripleSchemaToDbConcept(rs3),
-    ),
-    ...missingRelations.map((r) => crossAppRelationToDbConcept(r)),
-  ].filter((r) => r !== undefined);
+  const omitMissingSource = (node: CrossAppNode): CrossAppNode => {
```
sid597, line 336: Decision point. A missing Source drops the slot instead of publishing the Source as a dependency (ticket default; scoping doc F1 and F5). If that decision changes, this closure is the one branch to replace.

apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -311,14 +321,33 @@ export const publishNodesToGroups = async ({
...
+    const sourceId = node.slots?.[SOURCE_SLOT];
+    if (
+      sourceId === undefined ||
+      isRid(sourceId) ||
```
sid597, line 340: Open question. A RID is trusted without a lookup. The `my_concepts` query above is filtered to this space, and checking a RID needs its origin space id (a `Space` lookup or the `rid_to_space_id_and_local_id` RPC). `rid_or_local_id_to_concept_db_id` runs under the publisher's read access, so a revoked grant or a deleted origin concept still fails the row with -2. Proposal: accept for v0 and file a follow-up.

apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -311,14 +321,33 @@ export const publishNodesToGroups = async ({
...
+      nodesByUid.has(sourceId) ||
+      syncedUids.has(sourceId)
+    )
+      return node;
+    console.warn(
```
sid597, line 345: Open question. This matches the sync path's `console.warn` for dependencies missing from a batch. `Export.tsx` already builds the publish toast from `PublishNodesResult`, so an omitted-source count there plus a PostHog property is about 10 lines. Do we want it in this PR? The title is re-read by uid because threading it needs a field on `CrossAppNode`, which the ticket puts out of scope.

apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -311,14 +321,33 @@ export const publishNodesToGroups = async ({
...
+      `Source "${getPageTitleByPageUid(sourceId)}" (${sourceId}) is not in this space yet; publishing "${node.content.direct.value}" without it.`,
+    );
+    return { ...node, slots: undefined };
```
sid597, line 348: `slots` carries only `sourceDocument` in v0; `roamToCrossAppConverters.ts:149` is the single producer. Dropping the object equals dropping the key until a second slot exists.

apps/roam/src/utils/publishNodesToGroups.ts
```diff
@@ -311,14 +321,33 @@ export const publishNodesToGroups = async ({
...
+  };
+
+  const { ordered: upsertConcepts } = orderConceptsByDependency(
```
sid597, line 351: Reuses the sync path's ordering so an in-batch Source is upserted before the Evidence that references it; `upsert_concepts` processes rows in order. The old order (schemas, nodes, relation schemas, relations) is preserved, so the index mapping to `response.data` below still holds. `missing` is ignored on purpose: already-synced schemas are not in the batch, so every node's schema would appear there.

apps/roam/src/utils/sourceSlot.ts
```diff
@@ -51,10 +53,11 @@ const isDiscourseNodeTitle = (
     .filter((n) => n.format !== "{content}") // exclude page and block
     .some((node) => matcherFor(node.format).test(title));
 
-// The page a node's {source} placeholder resolves to, when there is one. The
+// The page a node's {source} placeholder resolves to, when there is one: its uid, or
+// the RID it is known by elsewhere when it was imported from another app. The
 // placeholder is usually filled with a page reference, and a title holding a slash is
 // a namespaced page rather than a source, so it is left alone.
-export const sourceUidOfNode = (
+export const sourceIdOfNode = (
```
sid597, line 82: Renamed from `sourceUidOfNode` because it can now return a RID. Both callers renamed their local variable to match. No behavior change for the uid case.

sid597 (commented):
Inline notes on the decisions in this diff. Three are open questions for the reviewers.

apps/roam/src/utils/sourceSlot.ts
```diff
@@ -67,5 +70,8 @@ export const sourceUidOfNode = (
   if (!sourceTitle || sourceTitle.includes("/")) return undefined;
   if (!isDiscourseNodeTitle(sourceTitle, allNodes ?? getDiscourseNodes()))
     return undefined;
-  return getPageUidByPageTitle(sourceTitle) || undefined;
+  const sourceUid = getPageUidByPageTitle(sourceTitle);
+  if (!sourceUid) return undefined;
+  const sourceRid = readImportedSourceIdentity(sourceUid)?.sourceNodeRid;
+  return sourceRid !== undefined && isRid(sourceRid) ? sourceRid : sourceUid;
```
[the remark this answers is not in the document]
sid597: Fixed in df152b16. `sourceIdOfNode` now requires the value to parse to a space URI and a local id via `ridToSpaceUriAndLocalId`, else it keeps the page uid. Tests cover `not a rid`, `orn:bad`, and `orn:obsidian.note:vault-a/`.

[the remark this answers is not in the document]
sid597: Fixed in df152b16. `sourceIdOfNode` now requires the value to parse to a space URI and a local id via `ridToSpaceUriAndLocalId`, else it keeps the page uid. Tests cover `not a rid`, `orn:bad`, and `orn:obsidian.note:vault-a/`. Tightening the import path's own `isRid` check on incoming shared nodes is outside this ticket.

apps/roam/src/utils/sourceSlot.ts
```diff
@@ -67,5 +76,10 @@ export const sourceUidOfNode = (
   if (!sourceTitle || sourceTitle.includes("/")) return undefined;
   if (!isDiscourseNodeTitle(sourceTitle, allNodes ?? getDiscourseNodes()))
     return undefined;
-  return getPageUidByPageTitle(sourceTitle) || undefined;
+  const sourceUid = getPageUidByPageTitle(sourceTitle);
+  if (!sourceUid) return undefined;
+  const sourceRid = readImportedSourceIdentity(sourceUid)?.sourceNodeRid;
+  return sourceRid !== undefined && isWellFormedRid(sourceRid)
```
sid597, line 98: Per maparent on #1214, a `localId` may itself be a RID, so `CrossAppNode.slots` keeps its `LocalId` type; `rid_or_local_id_to_concept_db_id` accepts both. The value must parse to a space URI and a local id because block props are user-editable; anything else keeps the page uid. Existing rows: sync now writes the origin RID where it used to write the Roam mirror's uid, so already-synced Evidence repoints on the next sync.

sid597:
Decision: a missing referenced Source drops the `sourceDocument` slot with a warning and is not published or granted with the Evidence. Michael chose this (option 1) on [ENG-2141](https://linear.app/discourse-graphs/issue/ENG-2141) on 2026-09-02. The warn-and-publish-both option is [ENG-2232](https://linear.app/discourse-graphs/issue/ENG-2232), for after v0; `omitMissingSource` stays the single branch to replace.

apps/roam/src/utils/publishNodesToGroups.ts
sid597: Decided: option 1 stands (Michael on ENG-2141, 2026-09-02). Publishing the Source with the Evidence is ENG-2232, for after v0.

mdroidian (changes requested):
> Local delegated full review
 Ran a comprehensive review of the entire final diff in a subagent with a fresh context. Use $dg-delegated-full-review when no other full-review workflow is available.
A pre-pr-review pass ran before the last fixes (source uid dedupe, RID guard, RID parse check). The full pass on the pushed diff is still to run.

Please run the `$dg-delegated-full-review`

maparent:
My opinions on follow-ups:

1. An imported RID skips the existence check: not ideal but fine for v0. I don't expect revocation to be frequent. (Delete maybe more so.)
2. I think that, in most cases where a source is unpublished, it's a mistake that the user would want to be warned about, so they can correct it. I would push for a toast in that case. If the source is missing, I expect the user to know. (ETA: If it's hard to distinguish the cases, a toast in both cases is fine too.)
3. I agree it's intended. Not sure how it changes rows.
4. Good.
5. Fair.

maparent:
Otherwise LGTM.

maparent (changes requested):
Change the warn to a toast, otherwise LGTM.

maparent (approved):

mdroidian (approved):

PR #1258, opened 2026-07-28
Use vault fallback while Datacore initializes

NathanSkene:
## Summary

- check Datacore's explicit `api.core.initialized` state before using its query index
- use Obsidian's vault and metadata cache while Datacore is still initializing or when a query throws
- trust empty Datacore results after initialization, avoiding duplicate full-vault scans for legitimate no-match queries
- cover title search, compatible-node lookup, and node enumeration with focused regression tests

## Reproduction

Datacore exposes its API during plugin loading, but initializes its index asynchronously on `workspace.onLayoutReady`. During a first install, reindex, or cache rebuild, Discourse Graph can therefore see a loaded Datacore API whose `core.initialized` flag is still false. Queries in that window can return an empty or partial universe even though matching discourse nodes are already available through Obsidian's metadata cache.

The previous distribution-repository PR described the index as potentially “empty or stale.” The reproduced case is narrower: Datacore is present but has not finished initialization. This change does not treat an empty result from an initialized index as stale.

## Validation

- `pnpm --filter @discourse-graphs/obsidian test` — 5 tests pass
- `pnpm --filter @discourse-graphs/obsidian check-types`
- targeted ESLint on the changed source and tests
- Obsidian production build (`tsx scripts/build.ts`)
- `git diff --check`

Replaces DiscourseGraphs/discourse-graph-obsidian#9, which targeted the generated distribution repository.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1258" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #1381, opened 2026-09-02
ENG-2139 Map the Obsidian source relation to `source` on push

sid597:
## Reviewer brief

- Result: an Obsidian node pushed to Supabase now carries `local_reference_content.sourceDocument` when it has a relation to a node of the Source type. The value is the Source's `importedFromRid` when it was imported, else its `nodeInstanceId`. With several such relations, the earliest by `created` (then `id`) wins. With none, or when the database cannot resolve the Source to a concept, the key is omitted. Full sync re-upserts every node whose stored slot differs from the one its relations call for, so a relation added, accepted, or removed reaches the database at the next full sync or publish, and existing vaults backfill on their first full sync.
- Review focus: `apps/obsidian/src/utils/sourceSlot.ts`. `indexSourceSlotValues` does one pass over `relations.json` and returns `Record<nodeInstanceId, sourceDocumentId>`; `filterAvailableSourceSlotValues` drops, for the nodes being pushed, every value `rid_or_local_id_to_concept_db_id` cannot resolve; the node converter reads `nodeData.sourceDocument`. `findStaleSourceSlotNodeIds` compares that map with the stored slot, read through the full-sync probe that already backfills `core_title` (`buildChangedNodesFromNodes`), by local id. The Source type is matched by name, as Roam does in `apps/roam/src/utils/sourceSlot.ts` (#1329). Both group decisions are settled: `SOURCE_SLOT` lives in `apps/obsidian/src/constants.ts`, and an unresolvable value is omitted with a console warning (see the Decision comment).
- Risk or follow-up:
  - `publishNewRelation` syncs relations only, so a relation added between two published nodes reaches the slot at the next full sync or publish, not at creation.
  - A Source whose concept the database cannot resolve (created while sync was off, or imported and later unshared) is omitted from the slot and a console warning names it, so the node's row succeeds. The full-sync comparison still wants that value, so the node's concept row re-upserts on every full sync (no content or embeddings) until the Source resolves; then the slot lands. The check is one RPC per distinct Source among the pushed nodes, run in parallel. Relation instances still carry the `-2` failure for a missing endpoint; unchanged here.
  - No tests. `apps/obsidian` has no unit-test runner; #1258 and #1366 each add one. The Done When line "Tests cover zero, one, and multiple matching relations" is not met here and goes with the harness question to ENG-2143.
  - Read side is out of scope: `dbToCrossAppConverters.ts` emits no instance `slots`, and `sharedNodes.ts:208` builds the cross-space slot RID without the `note` subtype Obsidian stores in `importedFromRid`. ENG-2140 and ENG-2142.
  - Obsidian schemas write no `roles`, so Obsidian Evidence keeps `arity 0` while Roam Evidence declares `roles: ["sourceDocument"]`. Slot-schema work is out of scope per the ticket.

## Verification

- `eslint --max-warnings 0`, `prettier --check`, `tsc --noEmit --skipLibCheck` on the four files: clean for this change. The two eslint warnings in `apps/obsidian` are present on main at untouched lines; the 20 tsc errors seen on main come from that checkout's local type resolution and do not appear in a fresh install.
- `turbo run test:unit`: 32 files pass (roam 25, database 5, content-model 2).
- `pnpm ci:validate` is red locally on main as well (obsidian, ui, website type errors from local type resolution). CI on main is green at ef37b20d.
- Runtime: not yet driven. See the Loom section.

## Loom video

https://github.com/user-attachments/assets/17f9eb38-721e-45ca-bbcd-6466fe7d948f

## Scope check

- [x] Ran `$scope-check` against ENG-2139 and the final diff.
- Scope beyond `Done When`: a node whose Source is not a concept the database can resolve pushes without `sourceDocument` and logs a warning naming the Source. The full-sync comparison itself is what makes "a node with no matching source relation omits the source value" and "repeated pushes produce the same single source value" hold after a relation changes.
- Required now: this PR writes the slot for the first time, and its full-sync comparison re-queues an unresolvable value on every full sync, which failed that node's concept row and its title sync with `-2`. Same behavior as ENG-2141 (#1380).
- Anyone affected or consulted: Yes, maparent.
- Decision: https://github.com/DiscourseGraphs/discourse-graph/pull/1381#pullrequestreview-5182196750 and https://github.com/DiscourseGraphs/discourse-graph/pull/1381#issuecomment-5660701219

### DIFF
--- a/apps/obsidian/src/utils/conceptConversion.ts
+++ b/apps/obsidian/src/utils/conceptConversion.ts
@@ -11,6 +11,7 @@ import type { LocalConceptDataInput } from "@repo/database/inputTypes";
 import type { ObsidianDiscourseNodeData } from "./syncDgNodesToSupabase";
 import type { Json } from "@repo/database/dbTypes";
 import { extractContentFromTitle } from "./extractContentFromTitle";
+import { SOURCE_SLOT } from "./sourceSlot";
 
 /**
  * Get extra data (author, timestamps) from file metadata
@@ -162,10 +163,12 @@ export const discourseNodeInstanceToLocalConcept = ({
   context,
   nodeData,
   nodeTypesById,
+  sourceSlotByNodeId,
 }: {
   context: SupabaseContext;
   nodeData: ObsidianDiscourseNodeData;
   nodeTypesById: Record<string, DiscourseNode>;
+  sourceSlotByNodeId: Record<string, string>;
 }): LocalConceptDataInput => {
   const extraData = getNodeExtraData(nodeData.file, context.userId);
   const { nodeInstanceId, nodeTypeId, importedFromRid, ...otherData } =
@@ -180,6 +183,7 @@ export const discourseNodeInstanceToLocalConcept = ({
   };
   if (importedFromRid && typeof importedFromRid === "string")
     literal_content.importedFromRid = importedFromRid;
+  const sourceDocumentId = sourceSlotByNodeId[nodeData.nodeInstanceId];
   return {
     space_id: context.spaceId,
     name: nodeData.file.path,
@@ -187,6 +191,10 @@ export const discourseNodeInstanceToLocalConcept = ({
     schema_represented_by_local_id: nodeTypeId as string,
     is_schema: false,
     literal_content,
+    // A value the database cannot resolve to a concept fails this row's upsert (-2).
+    ...(sourceDocumentId
+      ? { local_reference_content: { [SOURCE_SLOT]: sourceDocumentId } }
+      : {}),
     ...extraData,
   };
 };
--- a/apps/obsidian/src/utils/sourceSlot.ts
+++ b/apps/obsidian/src/utils/sourceSlot.ts
@@ -1,4 +1,9 @@
-import { spaceUriAndLocalIdToRid } from "@repo/database/lib/rid";
+import {
+  isRid,
+  ridToSpaceUriAndLocalId,
+  spaceUriAndLocalIdToRid,
+} from "@repo/database/lib/rid";
+import type { Json } from "@repo/database/dbTypes";
 import type { DiscourseNode, RelationInstance } from "~/types";
 import type { DiscourseNodeInVault } from "./getDiscourseNodes";
 
@@ -78,4 +83,56 @@ export const indexSourceSlotValues = ({
       sourceDocumentIdOf(candidate.sourceDocumentNode),
     ]),
   );
+};
+
+export const SOURCE_SLOT_PROBE_SELECT =
+  "reference_content, concepts_of_relation(id, source_local_id)";
+
+export type SourceSlotProbeRow = {
+  source_local_id: string | null;
+  reference_content: Json | null;
+  concepts_of_relation: {
+    id: number | null;
+    source_local_id: string | null;
+  }[];
+};
+
+const storedSourceDocumentLocalId = (
+  row: SourceSlotProbeRow,
+): string | undefined => {
+  const content = row.reference_content;
+  if (typeof content !== "object" || content === null || Array.isArray(content))
+    return undefined;
+  const conceptId = content[SOURCE_SLOT];
+  if (typeof conceptId !== "number") return undefined;
+  return (
+    row.concepts_of_relation.find((concept) => concept.id === conceptId)
+      ?.source_local_id ?? undefined
+  );
+};
+
+const localIdOf = (sourceDocumentId: string): string =>
+  isRid(sourceDocumentId)
+    ? ridToSpaceUriAndLocalId(sourceDocumentId).sourceLocalId
+    : sourceDocumentId;
+
+// A relation added, accepted, or removed changes no note, so the stored slot is compared
+// with the wanted one on full sync. The comparison is by local id: the database stores
+// the Source's concept id, and an imported Source is wanted by its origin RID.
+export const findStaleSourceSlotNodeIds = ({
+  rows,
+  sourceSlotByNodeId,
+}: {
+  rows: SourceSlotProbeRow[];
+  sourceSlotByNodeId: Record<string, string>;
+}): Set<string> => {
+  const stale = new Set<string>();
+  for (const row of rows) {
+    if (row.source_local_id === null) continue;
+    const wanted = sourceSlotByNodeId[row.source_local_id];
+    const wantedLocalId = wanted === undefined ? undefined : localIdOf(wanted);
+    if (wantedLocalId !== storedSourceDocumentLocalId(row))
+      stale.add(row.source_local_id);
+  }
+  return stale;
 };
--- a/apps/obsidian/src/utils/sourceSlot.ts
+++ b/apps/obsidian/src/utils/sourceSlot.ts
@@ -1,0 +1,81 @@
+import { spaceUriAndLocalIdToRid } from "@repo/database/lib/rid";
+import type { DiscourseNode, RelationInstance } from "~/types";
+import type { DiscourseNodeInVault } from "./getDiscourseNodes";
+
+// Temporary hack, the Obsidian twin of apps/roam/src/utils/sourceSlot.ts: until slots
+// are a first-class node type setting, the Source type is the node type named "Source",
+// and a node's sourceDocument is the destination of its earliest relation to a node of
+// that type. Relation endpoints are stored as a nodeInstanceId, as this vault's RID for
+// it, or as an imported node's origin RID, so lookups go through an index of all three.
+
+export const SOURCE_SLOT = "sourceDocument";
+
+type NodesByEndpoint = Record<string, DiscourseNodeInVault>;
+
+const indexNodesByEndpoint = ({
+  nodes,
+  localSpaceUri,
+}: {
+  nodes: DiscourseNodeInVault[];
+  localSpaceUri: string;
+}): NodesByEndpoint => {
+  const index: NodesByEndpoint = {};
+  for (const node of nodes) {
+    index[node.nodeInstanceId] = node;
+    index[spaceUriAndLocalIdToRid(localSpaceUri, node.nodeInstanceId, "note")] =
+      node;
+    const importedFromRid = node.frontmatter.importedFromRid;
+    if (typeof importedFromRid === "string") index[importedFromRid] = node;
+  }
+  return index;
+};
+
+const isSourceNodeType = (nodeType: DiscourseNode | undefined): boolean =>
+  nodeType?.name.toLowerCase() === "source";
+
+const byCreatedThenId = (a: RelationInstance, b: RelationInstance): number =>
+  a.created - b.created || a.id.localeCompare(b.id);
+
+const sourceDocumentIdOf = (node: DiscourseNodeInVault): string => {
+  const importedFromRid = node.frontmatter.importedFromRid;
+  return typeof importedFromRid === "string"
+    ? importedFromRid
+    : node.nodeInstanceId;
+};
+
+type SourceCandidate = {
+  relation: RelationInstance;
+  sourceDocumentNode: DiscourseNodeInVault;
+};
+
+export const indexSourceSlotValues = ({
+  relations,
+  nodes,
+  localSpaceUri,
+  nodeTypesById,
+}: {
+  relations: RelationInstance[];
+  nodes: DiscourseNodeInVault[];
+  localSpaceUri: string;
+  nodeTypesById: Record<string, DiscourseNode>;
+}): Record<string, string> => {
+  const nodesByEndpoint = indexNodesByEndpoint({ nodes, localSpaceUri });
+  const earliestByNodeId: Record<string, SourceCandidate> = {};
+  for (const relation of relations) {
+    if (relation.tentative === false) continue;
+    const node = nodesByEndpoint[relation.source];
+    const sourceDocumentNode = nodesByEndpoint[relation.destination];
+    if (!node || !sourceDocumentNode) continue;
+    if (!isSourceNodeType(nodeTypesById[sourceDocumentNode.nodeTypeId]))
+      continue;
+    const current = earliestByNodeId[node.nodeInstanceId];
+    if (!current || byCreatedThenId(relation, current.relation) < 0)
+      earliestByNodeId[node.nodeInstanceId] = { relation, sourceDocumentNode };
+  }
+  return Object.fromEntries(
+    Object.entries(earliestByNodeId).map(([nodeInstanceId, candidate]) => [
+      nodeInstanceId,
+      sourceDocumentIdOf(candidate.sourceDocumentNode),
+    ]),
+  );
+};
--- a/apps/obsidian/src/utils/sourceSlot.ts
+++ b/apps/obsidian/src/utils/sourceSlot.ts
@@ -85,54 +85,80 @@ export const indexSourceSlotValues = ({
   );
 };
 
+// Appended to a my_concepts select that already carries source_local_id.
 export const SOURCE_SLOT_PROBE_SELECT =
-  "reference_content, concepts_of_relation(id, source_local_id)";
+  "reference_content, concepts_of_relation(id, space_id, source_local_id)";
 
 export type SourceSlotProbeRow = {
-  source_local_id: string | null;
   reference_content: Json | null;
   concepts_of_relation: {
     id: number | null;
+    space_id: number | null;
     source_local_id: string | null;
   }[];
 };
 
-const storedSourceDocumentLocalId = (
+type StoredSourceDocument = {
+  spaceId: number | null;
+  localId: string | null;
+};
+
+const storedSourceDocumentOf = (
   row: SourceSlotProbeRow,
-): string | undefined => {
+): StoredSourceDocument | undefined => {
   const content = row.reference_content;
   if (typeof content !== "object" || content === null || Array.isArray(content))
     return undefined;
   const conceptId = content[SOURCE_SLOT];
   if (typeof conceptId !== "number") return undefined;
-  return (
-    row.concepts_of_relation.find((concept) => concept.id === conceptId)
-      ?.source_local_id ?? undefined
+  const concept = row.concepts_of_relation.find(
+    (candidate) => candidate.id === conceptId,
   );
+  if (!concept) return undefined;
+  return { spaceId: concept.space_id, localId: concept.source_local_id };
 };
 
-const localIdOf = (sourceDocumentId: string): string =>
-  isRid(sourceDocumentId)
-    ? ridToSpaceUriAndLocalId(sourceDocumentId).sourceLocalId
-    : sourceDocumentId;
+const storedMatchesWanted = ({
+  stored,
+  wanted,
+  spaceId,
+}: {
+  stored: StoredSourceDocument | undefined;
+  wanted: string | undefined;
+  spaceId: number;
+}): boolean => {
+  if (wanted === undefined) return stored === undefined;
+  if (stored === undefined) return false;
+  if (isRid(wanted))
+    return (
+      stored.spaceId !== spaceId &&
+      stored.localId === ridToSpaceUriAndLocalId(wanted).sourceLocalId
+    );
+  return stored.spaceId === spaceId && stored.localId === wanted;
+};
 
-// A relation added, accepted, or removed changes no note, so the stored slot is compared
-// with the wanted one on full sync. The comparison is by local id: the database stores
-// the Source's concept id, and an imported Source is wanted by its origin RID.
+// A relation added, accepted, or removed changes no file, so the mtime check never
+// fires; full sync compares the stored slot with the wanted one instead. The database
+// stores the Source's concept id: a local Source must resolve to a concept in this
+// space, an imported Source (wanted by its origin RID) to one in another space.
 export const findStaleSourceSlotNodeIds = ({
   rows,
   sourceSlotByNodeId,
+  spaceId,
 }: {
-  rows: SourceSlotProbeRow[];
+  rows: (SourceSlotProbeRow & { source_local_id: string | null })[];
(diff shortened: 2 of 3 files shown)


apps/obsidian/src/utils/sourceSlot.ts
```diff
@@ -0,0 +1,81 @@
...
+import type { DiscourseNode, RelationInstance } from "~/types";
+import type { DiscourseNodeInVault } from "./getDiscourseNodes";
+
+// Temporary hack, the Obsidian twin of apps/roam/src/utils/sourceSlot.ts: until slots
+// are a first-class node type setting, the Source type is the node type named "Source",
+// and a node's sourceDocument is the destination of its earliest relation to a node of
+// that type. Relation endpoints are stored as a nodeInstanceId, as this vault's RID for
+// it, or as an imported node's origin RID, so lookups go through an index of all three.
+
+export const SOURCE_SLOT = "sourceDocument";
```
sid597, line 16: `SOURCE_SLOT` is also defined in `apps/roam/src/utils/sourceSlot.ts:12`, and `crossAppConverters.ts:92` and `sharedNodes.ts:212` depend on the same key. Its one home is `packages/database/src/crossAppContracts.ts`. Hoisting touches roam and database, so this PR stays inside `apps/obsidian`.

Decision for the group: hoist here, or ticket the hoist and merge with the duplicate?

apps/obsidian/src/utils/sourceSlot.ts
```diff
@@ -0,0 +1,81 @@
...
+    index[node.nodeInstanceId] = node;
+    index[spaceUriAndLocalIdToRid(localSpaceUri, node.nodeInstanceId, "note")] =
+      node;
+    const importedFromRid = node.frontmatter.importedFromRid;
+    if (typeof importedFromRid === "string") index[importedFromRid] = node;
+  }
+  return index;
+};
+
+const isSourceNodeType = (nodeType: DiscourseNode | undefined): boolean =>
```
sid597, line 38: Name match follows Roam's `sourceNodeType` (#1329). Obsidian node type ids are generated at load (`constants.ts:36`), so the name is the only stable handle. A vault that renames the Source type gets no slot until slots become a node type setting.

sid597 (commented):
Inline notes. Three are marked "Decision for the group".

apps/obsidian/src/utils/sourceSlot.ts
```diff
@@ -0,0 +1,81 @@
...
+  nodeType?.name.toLowerCase() === "source";
+
+const byCreatedThenId = (a: RelationInstance, b: RelationInstance): number =>
+  a.created - b.created || a.id.localeCompare(b.id);
+
+const sourceDocumentIdOf = (node: DiscourseNodeInVault): string => {
```
sid597, line 44: `importedFromRid` first, `nodeInstanceId` otherwise: the rule `relationInstanceToLocalConcept` already uses for relation endpoints. `rid_or_local_id_to_concept_db_id` resolves either form.

apps/obsidian/src/utils/sourceSlot.ts
```diff
@@ -0,0 +1,81 @@
...
+    ? importedFromRid
+    : node.nodeInstanceId;
+};
+
+type SourceCandidate = {
+  relation: RelationInstance;
+  sourceDocumentNode: DiscourseNodeInVault;
+};
+
+export const indexSourceSlotValues = ({
```
sid597, line 56: No schema gate. Roam writes the slot only for node types whose format has `{source}`. Obsidian formats have no placeholder, so any node with a relation to a Source gets the slot, and the earliest relation wins across relation types.

Nothing downstream reads the gate: `arity` and `is_relation` derive from the schema's `roles`, and `sharedNodes.ts` reads the keys. Obsidian schemas write no `roles`, which the ticket puts out of scope, so Obsidian Evidence stays `arity 0` while Roam Evidence is 1. Flagging the asymmetry in case it matters for the read side.

apps/obsidian/src/utils/sourceSlot.ts
```diff
@@ -0,0 +1,81 @@
...
+}: {
+  relations: RelationInstance[];
+  nodes: DiscourseNodeInVault[];
+  localSpaceUri: string;
+  nodeTypesById: Record<string, DiscourseNode>;
+}): Record<string, string> => {
+  const nodesByEndpoint = indexNodesByEndpoint({ nodes, localSpaceUri });
+  const earliestByNodeId: Record<string, SourceCandidate> = {};
+  for (const relation of relations) {
+    if (relation.tentative === false) continue;
```
sid597, line 70: Imported relations not yet accepted (`tentative === false`) do not define a node's source. Same filter the relation batch applies in `convertDgToSupabaseConcepts`.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -180,13 +183,18 @@ export const discourseNodeInstanceToLocalConcept = ({
...
     literal_content.importedFromRid = importedFromRid;
+  const sourceDocumentId = sourceSlotByNodeId[nodeData.nodeInstanceId];
   return {
     space_id: context.spaceId,
     name: nodeData.file.path,
     source_local_id: nodeInstanceId as string,
     schema_represented_by_local_id: nodeTypeId as string,
     is_schema: false,
     literal_content,
+    // A value the database cannot resolve to a concept fails this row's upsert (-2).
```
sid597, line 194: `upsert_concepts` resolves this value with `rid_or_local_id_to_concept_db_id`. When that returns NULL, the row's `reference_content` becomes NULL, the insert fails, and the function returns `-2` for the row. The client reads only `error`, so the node's title update is dropped with no signal.

Triggers: the Source was created while sync was off and the Evidence syncs alone; or an imported Source whose origin space later unshares it (the lookup runs under RLS). Relation instances already have this failure mode.

Decision for the group: keep as is and ticket reading the negative returns into PostHog, or check existence before writing the slot, as ENG-2141 plans for the Roam publish path?

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -605,18 +607,24 @@ const convertDgToSupabaseConcepts = async ({
     )
     .filter((n) => !!n);
 
+  const relationInstancesData = await loadRelations(plugin);
+  const relationInstances = Object.values(relationInstancesData.relations);
+  const sourceSlotByNodeId = indexSourceSlotValues({
```
sid597: The endpoint index inside `indexSourceSlotValues` covers `nodeInstanceId`, this vault's RID, and `importedFromRid`. `allNodesById` above covers bare ids only, and `relationInstanceToLocalConcept` and `ensurePublishedRelationsAccuracy` drop RID endpoints today. Widening them changes relation push, which this ticket does not own, so both indexes stay.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -605,18 +607,24 @@ const convertDgToSupabaseConcepts = async ({
...
+    relations: relationInstances,
+    nodes: allNodes,
+    localSpaceUri: getLocalSpaceUri(plugin.app),
+    nodeTypesById,
+  });
   const nodeInstanceToLocalConcepts = nodesSince.map((node) => {
```
sid597, line 667: When the slot reaches the database. A node concept is re-upserted only when its file changes: a title change on the incremental path; title, content, or missing concept on full sync. Linking an existing note to a Source edits `relations.json`, not the note, so the slot lands on the next sync after the note itself changes. Publish writes `publishedToGroups` into the frontmatter after its own sync, so the slot arrives one publish late. Plain sync waits for an edit.

Fix is about 15 lines here: add the source node of each relation changed since `lastRelationsSync` to `nodesSince`, with the predicate the relation batch uses and `changeTypes: []` so embeddings are not recomputed.

Decision for the group: in this PR, or a follow-up ticket?

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -605,18 +607,24 @@ const convertDgToSupabaseConcepts = async ({
...
     return discourseNodeInstanceToLocalConcept({
       context,
       nodeData: node,
       nodeTypesById,
+      sourceSlotByNodeId,
     });
```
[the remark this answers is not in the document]
sid597: Known. The sync trigger is the open decision in the thread on line 618: fix in this PR or a follow-up ticket. Adding your backfill point there: a fix keyed on `lastRelationsSync` does not touch relations older than the watermark, so existing vaults would need a one-time probe, as `missingCoreTitleIds` did for `core_title`.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
sid597: Addendum from the Devin thread: a fix keyed on `lastRelationsSync` covers relations added from now on. Relations already in existing vaults are older than the watermark, so a first-sync backfill needs a one-time probe, the way `missingCoreTitleIds` backfilled `core_title` in ENG-2155. Same decision: this PR, or the follow-up ticket.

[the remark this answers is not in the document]
sid597: Agreed on the gap. It is the open decision in the thread on line 618; your two extra cases (acceptance and removal) are added there. Removal deletes the record from `relations.json`, so a watermark on relation timestamps cannot see it; the design that covers all your cases compares the stored slot with the current relations on full sync, the way `missingCoreTitleIds` backfills `core_title`.

sid597: Codex adds two cases: a relation accepted (`tentative` flips to true, no timestamp changes) and a relation removed (`removeRelationById` deletes the record, nothing left to diff). Two designs:

1. Watermark. Add the source node of every relation newer than `lastRelationsSync` to the batch. About 15 lines. Covers additions only.
2. Compare on full sync. Full sync already probes every node concept in the space (`buildChangedNodesFromNodes`, `CORE_TITLE_PROBE_SELECT`). Add `reference_content` to that probe, compute the wanted slot for every local node, and re-upsert the nodes whose stored `sourceDocument` differs, including nodes whose slot must clear. About 40 to 60 lines. Covers add, accept, remove, retarget, and the backfill of existing vaults. Same pattern as the `core_title` backfill in ENG-2155. Still misses `publishNewRelation`, which syncs relations only; that slot lands on the next full sync or publish.

Design 2 is the one that keeps "a node with no matching source relation omits the source value" true after a removal.

sid597: Decided by sid: fixed in this PR, design 2. Commit 63a964ab adds `reference_content` and the `concepts_of_relation` embed to the full-sync probe, and `findStaleSourceSlotNodeIds` re-queues every node whose stored slot differs from the wanted one, including nodes whose slot must clear. Those nodes carry `changeTypes: []`, so the content upsert skips them and only the concept row is rewritten. `publishNewRelation` still syncs relations only; that slot lands on the next full sync or publish, stated in the PR body.

apps/obsidian/src/utils/sourceSlot.ts
```diff
@@ -0,0 +1,138 @@
...
+
+const localIdOf = (sourceDocumentId: string): string =>
+  isRid(sourceDocumentId)
+    ? ridToSpaceUriAndLocalId(sourceDocumentId).sourceLocalId
+    : sourceDocumentId;
+
+// A relation added, accepted, or removed changes no note, so the stored slot is compared
+// with the wanted one on full sync. The comparison is by local id: the database stores
+// the Source's concept id, and an imported Source is wanted by its origin RID.
+export const findStaleSourceSlotNodeIds = ({
```
sid597, line 144: Compared by local id. The database stores the Source's concept id; the `concepts_of_relation` embed gives that concept's `source_local_id`; an imported Source is wanted by its origin RID, reduced to its local part. A Source whose origin concept is hidden (unshared) yields no match, so that node re-queues on every full sync and its row fails with `-2` until the concept resolves. Same failure the converter comment names, now repeated per full sync rather than once.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -353,11 +361,12 @@ const buildChangedNodesFromNodes = async ({
   const changedNodes: ObsidianDiscourseNodeData[] = [];
   let missingConcepts: Set<string> | undefined;
   let missingCoreTitleIds: Set<string> | undefined;
+  let staleSourceSlotIds: Set<string> | undefined;
   if (fullSync) {
     const existingConceptIds = await getAllPages(
       supabaseClient
         .from("my_concepts")
-        .select(CORE_TITLE_PROBE_SELECT)
+        .select(`${CORE_TITLE_PROBE_SELECT}, ${SOURCE_SLOT_PROBE_SELECT}`)
```
sid597, line 370: Same probe ENG-2155 uses to backfill `core_title`, extended with the slot columns, so the comparison adds no query. The embed is the one `sharedNodes.ts` uses on the read side.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -380,6 +389,11 @@ const buildChangedNodesFromNodes = async ({
       missingConcepts = difference(nodeIds, dbConceptIds);
       missingCoreTitleIds =
         partitionByCoreTitle(existingConceptIds).missingCoreTitleIds;
+      if (sourceSlotByNodeId)
+        staleSourceSlotIds = findStaleSourceSlotNodeIds({
```
sid597, line 394: Stale nodes join `changedNodes` with `changeTypes: []`. `createNodeContentEntries` returns early for them, so only the concept row rewrites; no content or embeddings.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -420,6 +435,27 @@ const buildChangedNodesFromNodes = async ({
   return { changedNodes };
 };
 
+const indexSourceSlots = async ({
```
sid597: Built once here so full sync can compare, then passed on to `convertDgToSupabaseConcepts`. The incremental path has no `allNodes` at this point, so that function keeps an inline fallback.

sid597 (commented):
Notes on the full-sync comparison added in 63a964ab.

sid597 (commented):
Notes on a0160ac4.

apps/obsidian/src/utils/sourceSlot.ts
```diff
@@ -0,0 +1,164 @@
...
+  const conceptId = content[SOURCE_SLOT];
+  if (typeof conceptId !== "number") return undefined;
+  const concept = row.concepts_of_relation.find(
+    (candidate) => candidate.id === conceptId,
+  );
+  if (!concept) return undefined;
+  return { spaceId: concept.space_id, localId: concept.source_local_id };
+};
+
+const storedMatchesWanted = ({
```
sid597, line 121: The stored slot is compared by which side of the space boundary it resolves to, plus local id. A local Source must resolve to a concept in this space; an imported Source, wanted by its origin RID, to a concept in another space. `asSimpleLocalId` in `dbToCrossAppConverters.ts` makes the same distinction on the read side. Two foreign spaces sharing a local id would still compare equal; local ids are uuidv7 or Roam uids, so that case is not guarded.

apps/obsidian/src/utils/syncDgNodesToSupabase.ts
```diff
@@ -605,18 +658,22 @@ const convertDgToSupabaseConcepts = async ({
     )
     .filter((n) => !!n);
 
+  relationInstancesData =
+    relationInstancesData ?? (await loadRelations(plugin));
+  const relationInstances = Object.values(relationInstancesData.relations);
+  sourceSlotByNodeId =
+    sourceSlotByNodeId ??
```
sid597, line 665: The index built here keys nodes by all three endpoint forms because an imported Source is stored as its RID. `allNodesById` above stays keyed by bare `nodeInstanceId`: `relationInstanceToLocalConcept` and `ensurePublishedRelationsAccuracy` look up bare ids and drop RID endpoints today, and widening them changes relation push, which this ticket does not own.

sid597:
**Decision: when the slot reaches the database**

Question: a node concept is rewritten only when its own file changes, so a source relation added, accepted, or removed never reached `sourceDocument`. Raised by the pre-PR review, Devin, and Codex.

Options:
1. Watermark on relation timestamps. About 15 lines. Covers additions only.
2. Compare the stored slot with the wanted one on full sync, reusing the `core_title` backfill probe. Covers add, accept, remove, retarget, and the backfill of existing vaults.
3. Defer to a follow-up ticket and merge the mapping alone.

Decision (sid, 2026-09-02): option 2, in this PR. Three independent reviewers flagged the same gap, and the Done When line "a node with no matching source relation omits the source value" is false after a removal without it.

Result: commits 63a964ab and a0160ac4. Known limit: `publishNewRelation` syncs relations only, so a relation added between two published nodes reaches the slot at the next full sync or publish.

Still open for the group, in inline threads: where `SOURCE_SLOT` lives (`sourceSlot.ts`, line 16) and the silent `-2` on an unresolvable value (`conceptConversion.ts`, line 194).

maparent:
On the decision: That's a tough one. I think you made the right call. Eventually we'll want to move the source in the frontmatter, but that's a totally different scope.
`publishNewRelation`: doesn't `addRelationNoCheck` call it? Is that the use case you discuss?
Schema change: Agreed that's another scope.

maparent:
Thinking about the -2… There are two cases: The Id does not exist in the vault (then we could filter it out) and the Id does not exist in the database. In the latter case, I think it's worth gathering the sourceIds and checking which ones exist in the database.
We could then either sync them (which I think is unneeded scope expansion) or just filter them out of your map, so that `publishNewRelation` can succeed. (I think that's in scope.)
On SOURCE_SLOT: How about `apps/obsidian/src/constants.ts` ? We may want shared constants between this and its roam counterpart `apps/roam/src/data/defaultDiscourseNodes.ts`, but that can wait.

apps/obsidian/src/utils/conceptConversion.ts
```diff
@@ -162,10 +163,12 @@ export const discourseNodeInstanceToLocalConcept = ({
   context,
   nodeData,
   nodeTypesById,
+  sourceSlotByNodeId,
```
maparent: I'm not keen on adding problem-specific parameters to a generic function. I think we should get closer to the CrossApp pattern in new code when possible, to ease the subsequent refactor. In this case it means adding an optional field to nodeData; either an optional sourceDocument (simple, short-term hack) or the full `slots?: Record<string, LocalId>;` as in `CrossAppNode` to go the whole distance. Not insisting on the latter now, but I strongly believe that the lookup in sourceSlotByNodeId belongs upstream of this.

maparent (changes requested):
Besides the inline comment, I think filtering out sources that are missing from the database would not be a huge stretch, and would get us closer to what I understand to be desired behaviour (i.e. publish should not fail on a missing source.)

apps/obsidian/src/utils/conceptConversion.ts
sid597: Done in dc928689. The converter reads an optional `sourceDocument` on `ObsidianDiscourseNodeData`, set in `convertDgToSupabaseConcepts` before conversion; the lookup parameter is gone. Took the simple field for now. The `slots` record can come with the CrossApp refactor.

apps/obsidian/src/utils/sourceSlot.ts
sid597: Settled: stays in `apps/obsidian`, moved to `src/constants.ts` in dc928689 per maparent. The hoist to `packages/database` waits.

sid597:
> `publishNewRelation`: doesn't `addRelationNoCheck` call it? Is that the use case you discuss?

Yes. `addRelationNoCheck` (relationsStore.ts:171) calls `publishNewRelation`, which runs `syncAllNodesAndRelations` with `relationsOnly`, so that batch carries no node concept rows. That's the case in the PR body: a relation added between two published nodes reaches the slot at the next full sync or node publish.

sid597:
**Decision: omit a Source the database cannot resolve**

Question: what a push does when a node's Source is not a concept the database can resolve. Raised in the inline thread on `conceptConversion.ts` (silent `-2`).

Decision (maparent, 2026-09-11, in review): filter those Sources out so the push does not fail on a missing Source. Syncing them would be scope expansion.

Result: `filterAvailableSourceSlotValues` in `sourceSlot.ts` (dc928689, 4bfed96e, 5cd6dfa3).
- For the nodes being pushed, each distinct Source value goes through `rid_or_local_id_to_concept_db_id`, the resolver `upsert_concepts` uses. Neither function is `SECURITY DEFINER`, so an imported RID is checked in its origin space under the caller's read access, the same way the upsert would resolve it.
- A local Source in the same batch counts as available; `orderConceptsByDependency` upserts it first.
- A value that does not resolve is dropped and a console warning names the Source and the node.
- Ids missing from the vault were already filtered: `indexSourceSlotValues` skips a relation whose endpoint has no note.

Known limit: the full-sync comparison still wants the dropped value, so that node's concept row re-upserts on every full sync (no content or embeddings) until the Source resolves; then the slot lands with no further change. Kept so the RPC runs only for pushed nodes.

apps/obsidian/src/utils/conceptConversion.ts
sid597: Settled: filter, per maparent's review. Decision comment: https://github.com/DiscourseGraphs/discourse-graph/pull/1381#issuecomment-5660701219. Implemented as `filterAvailableSourceSlotValues` in `sourceSlot.ts` (dc928689, 4bfed96e, 5cd6dfa3).

maparent (approved):

PR #1382, opened 2026-09-02
ENG-2142 Map `source` to the Roam referenced node on pull

sid597:
## Reviewer brief

- Result: An imported node whose local format has `{Source}` gets its title rebuilt from `core_title` and the local page of the source named in `slots.sourceDocument`, for example `[[EVD]] - REM sleep and recall - [[@Smith 2020]]`. The source is found in this order: a RID in this graph's space is the page uid; otherwise the page whose `importedFrom.sourceNodeRid` matches, which also covers a source in the same import batch because referenced sources are materialized first. A missing source fills `{Source}` with `[[@placeholder]]` so the title keeps the node type's format, and reports a warning in the import dialog and the Refresh toast; Refresh replaces the placeholder once the source is imported (95eab0db, per mdroidian's review).
- Review focus: `findTargetUid` moved out of `importSharedRelations.ts` into its own util so the node importer shares it (parity table below). The `{Source}` fill is Roam-only in `sourceSlot.ts` and `decorateTitle` is unchanged, because `{Source}` in Roam titles is slated for removal (ENG-2043, DES-368) and ENG-2176 unifies the format grammar. The decision trace with the alternatives considered is the ENG-2142 comment of 2026-09-02.
- Risk or follow-up: `dbToCrossAppConverters.ts:32` builds cross-space relation endpoint RIDs without the platform subtype, the same bug this PR fixes for slot RIDs in `sharedNodes.ts`; that is ENG-2215. Obsidian's pull side still keeps the published title for `{Source}` formats (`decorateTitle` returns null); ENG-2140 owns Obsidian's source mapping. A core title containing the format separator does not survive a republish (lazy capture groups; ENG-2176). The batch ordering is one level deep: a source that itself names a source in the batch is filled on the next Refresh. The up-to-date skip stays timestamp-only (#1331 decision), so an Evidence whose source arrives later keeps its title until Refresh.

Parity table for the moved `findTargetUid`:

| Old (`importSharedRelations.ts`) | New (`findTargetUid.ts`) | Verdict |
| --- | --- | --- |
| `localSpaceUrl` read from `roamAlphaAPI.graph.name` at module load | `canonicalRoamUrl()` called inside the function | changed: removes a module-load read of `roamAlphaAPI`, which is what makes the util unit-testable |
| `ridType?: string` parameter, `ridType ?? "note"` | parameter dropped, `"note"` passed directly | identical: no caller supplied it |
| bare local id: build RID, then imported-node lookup | build RID via `sharedReferenceRid`, then the own-space check, then the lookup | changed: the own-space check also runs on the bare path; `importRelations` derives `spaceUri` from the relation's own RID and discovery excludes the local space, so relation import is unaffected |
| RID construction inline | exported `sharedReferenceRid` | changed: `orderSourcesFirst` needs the same encoding to match a batch member's `rid` |
| sync `window.roamAlphaAPI.q` existence check | same | identical: moved as is |

## Verification

- `pnpm ci:validate` exit 0 (type checks in 12 packages, unit tests in 3).
- `apps/roam` vitest 221 passed, including new cases for each identity path, a missing source and its placeholder title, no published source, an unchanged refresh, a refresh after the source arrives, the up-to-date skip, batch ordering, and warning pass-through. `packages/database` vitest 17 passed.
- Not verified live: the import dialog and the Refresh toast. The runtime proof plan is in the ENG-2142 decision trace; a non-author drives it.

## Loom video

https://github.com/user-attachments/assets/baf259d2-d352-4246-9ae5-4304a398a8cf

Recorded before the placeholder change (95eab0db).

## Scope check

- [x] Ran `$scope-check` against ENG-2142 and the final diff.

- [x] Ran a comprehensive review of the entire final diff in a subagent with a fresh context (pre-pr-review on Opus). Its findings are fixed in 07b61b22; the deferred ones are listed under Risk or follow-up.
- [x] Re-ran it on the final diff on 2026-09-16 after 95eab0db. Two findings are open and a fix is in progress: the `[[@placeholder]]` reference parses as a Source on republish and reaches `sourceDocument`, and two imports with the same core title and unresolved Sources now collide. Hold review of the placeholder change until the follow-up commit lands.


apps/roam/src/utils/findTargetUid.ts
```diff
@@ -0,0 +1,36 @@
...
+  ridToSpaceUriAndLocalId,
+  spaceUriAndLocalIdToRid,
+} from "@repo/database/lib/rid";
+import canonicalRoamUrl from "./canonicalRoamUrl";
+import { findImportedNodeUidBySourceRid } from "./importedSourceIdentity";
+
+// A shared node refers to another node by its bare local id when both live in the same
+// space, and by a RID otherwise. "note" is the subtype node RIDs carry; URL-shaped Roam
+// RIDs ignore it.
+export const sharedReferenceRid = (
```
sid597, line 12: The `isRid` branch is required by the contract, not defensive. A slot value is a bare local id only when the referring and referred nodes share a space, and per maparent on #1214 a `source_local_id` may itself be a RID. No producer emits one today: Roam writes 9-char uids and Obsidian writes note paths. `"note"` is the subtype `sharedNodes.ts` gives node RIDs; URL-shaped Roam RIDs ignore it, so one literal serves both platforms.

apps/roam/src/utils/findTargetUid.ts
```diff
@@ -0,0 +1,36 @@
...
+  localOrRid: string,
+  spaceUri: string,
+): string =>
+  isRid(localOrRid)
+    ? localOrRid
+    : spaceUriAndLocalIdToRid(spaceUri, localOrRid, "note");
+
+// The local page for a node another space refers to: its own uid when the RID points
+// into this graph, else the page imported from it.
+export const findTargetUid = async (
```
sid597, line 22: Moved out of `importSharedRelations.ts` so the node importer can share it. Two deltas from the original, both in the parity table in the description: the local graph URL is read per call instead of at module load, which is what lets this be unit tested, and the unused `ridType` parameter is gone. The bare-local-id path now also runs the own-space check; it cannot change relation import, because `importRelations` derives `spaceUri` from the relation's own RID and discovery excludes the local space.

apps/roam/src/utils/findTargetUid.ts
```diff
@@ -0,0 +1,36 @@
...
+  localOrRid: string,
+  spaceUri: string,
+): Promise<string | null> => {
+  const rid = sharedReferenceRid(localOrRid, spaceUri);
+  const { spaceUri: ridSpaceUri, sourceLocalId } = ridToSpaceUriAndLocalId(rid);
+  if (ridSpaceUri === canonicalRoamUrl()) {
+    const result = window.roamAlphaAPI.q(
```
sid597, line 29: Moved as is, sync `q` included. The query binds `:block/uid`, so it matches blocks as well as pages; relations legitimately target blocks. On the pull path a block uid has no page title, so `resolveSourceTitle` treats it as missing and keeps the incoming title.

apps/roam/src/utils/sourceSlot.ts
```diff
@@ -69,3 +72,35 @@ export const sourceUidOfNode = (
...
+
+const FILLABLE_PLACEHOLDERS = new Set([
+  CONTENT_PLACEHOLDER,
+  SOURCE_PLACEHOLDER,
+]);
+
+// Inverse of sourceUidOfNode, for the pull side: the local title of a node whose format
+// names a source, built from its core title and the Source page's title. Null when the
+// format has a placeholder neither fills, so the caller keeps the incoming title.
+export const titleWithSource = ({
```
sid597, line 84: Roam-only on purpose, and not a `fields` parameter on `decorateTitle`. `{Source}` in Roam titles is slated for removal (ENG-2043, DES-368) and ENG-2176 unifies the format grammar, so the fill stays next to the push-side hack it inverts and both retire together. Obsidian's pull path keeps the published title for these formats; ENG-2140 owns Obsidian's source mapping.

Known limitation, same grammar weakness ENG-2176 addresses: a core title containing the format separator does not survive a republish. `coreTitle = "a - b"` yields `[[EVD]] - a - b - [[Src]]`, and the lazy capture groups in `getDiscourseNodeFormatExpression` then read the source as `b - [[Src]]`.

apps/roam/src/utils/importSharedNodes.ts
```diff
@@ -13,13 +15,30 @@ export type FailedSharedNodeImport = {
...
 
 export const isFailedSharedNodeImport = (
   item: SharedNodeImportItem,
 ): item is FailedSharedNodeImport => item.status === "failed";
 
+// A node's title can only name its source once that source has a local page, so the
+// nodes other batch members refer to are materialized first. One level only: a source
+// that itself names a source in the batch is not ordered after it, and its title is
+// filled on the next refresh instead.
+const orderSourcesFirst = (sharedNodes: SharedNode[]): SharedNode[] => {
```
sid597, line 29: One pass, not a topological sort. The slot accepts any discourse node (`sourceSlot.ts`), so a source that itself names a source in the batch is possible; that node keeps its incoming title with a warning and is filled on the next Refresh. `resolveSharedNodeTypes` still receives the original order; only the materialization loop is reordered.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -128,6 +135,53 @@ const validateSharedNode = (
   return { sourceModifiedAt: modifiedAt.toISOString(), title };
 };
 
+// The Source page a node's sourceDocument slot names, when this graph has it.
+const resolveSourceTitle = async (
```
sid597, line 139: The source title is read from the local page, not from the source's `SharedNode` that the batch has in hand. The local page may have been renamed since import, and the title written here has to match what the publish side reads back with `extractFieldFromTitle`. Warnings are user-fixable (import the source, then refresh), so they ride the result instead of PostHog.

sid597 (commented):
Guideposts for the non-obvious decisions in this diff.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -128,6 +135,53 @@ const validateSharedNode = (
...
+  const sourceUid = await findTargetUid(slotValue, sharedNode.spaceUri);
+  const sourceTitle = sourceUid ? getPageTitleByPageUid(sourceUid) : "";
+  if (!sourceTitle)
+    return {
+      warning: `Its source (${slotValue}) is not in this graph, so its title was kept as published. Import the source, then refresh this page.`,
+    };
+  return { sourceTitle };
+};
+
+const buildPageTitle = async ({
```
sid597, line 157: A format with a placeholder beyond `{content}` and `{Source}` makes `titleWithSource` return null and the published title is kept with no warning, the same as the `decorateTitle` branch above. Decorated-gap titles are out of scope for ENG-2142. The warning fires only when the source is the sole reason the title was not decorated: no core title or no format keeps the title silently, as #1331 does.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -356,6 +406,12 @@ export const materializeSharedNode = async ({
       pageUid: importedPageUid,
     };
 
+  const { title: pageTitle, warning } = await buildPageTitle({
```
sid597, line 409: Placed after the up-to-date check on purpose. Skipping stays timestamp-only (#1331 decision), so an unchanged import pays for no source lookup and gets no warning. An Evidence whose source arrives later keeps its title until Refresh.

apps/roam/src/utils/materializeSharedNode.ts
```diff
@@ -377,5 +433,6 @@ export const materializeSharedNode = async ({
         identity,
         markdown: content.markdown,
         title: pageTitle,
-      });
+      }));
+  return result.success && warning ? { ...result, warning } : result;
```
sid597, line 437: Attached here rather than passed into `createImportedPage` and `updateImportedPage`: the warning belongs to the title decision, which both share, and neither helper would do anything with it but echo it.

packages/database/src/lib/sharedNodes.ts
```diff
@@ -39,6 +39,9 @@ type SharedSpace = Pick<
 >;
 type Platform = Enums<"Platform">;
 
+const nodeRidSubtype = (platform: Platform): string | undefined =>
```
sid597, line 42: Bug fix in its own commit. Node RIDs already carried the platform subtype, but cross-space slot RIDs did not, so a slot pointing at an Obsidian node produced `orn:obsidian:vault/x` while the same node's stored `importedFrom.sourceNodeRid` was `orn:obsidian.note:vault/x`, and the provenance lookup never matched. No migration: stored RIDs already carry the subtype. `dbToCrossAppConverters.ts:32` builds relation endpoint RIDs the same way and still omits it; that is ENG-2215.

apps/roam/src/components/RefreshImportedNodeTitleButton.tsx
```diff
@@ -23,8 +23,14 @@ const RefreshImportedNodeTitleButton = ({
         id: result.success
           ? "refresh-imported-node-success"
           : "refresh-imported-node-failed",
-        intent: result.success ? "success" : "danger",
-        content: result.message,
+        intent: !result.success
```
sid597: A refresh that kept the title because the source is missing succeeded, so the toast stays a success message, but it carries a warning, so its intent matches the import dialog's callout for the same case.

apps/roam/src/components/DiscoverSharedNodesDialog.tsx
```diff
@@ -118,17 +118,23 @@ const ImportResultsSummary = ({
     (item) => item.status === "skipped",
   ).length;
   const failedImports = results.filter(isFailedSharedNodeImport);
+  const warnings = results.flatMap((item) =>
+    item.status !== "failed" && item.warning
+      ? [{ sharedNode: item.sharedNode, message: item.warning }]
+      : [],
+  );
+  const importNotices = [...failedImports, ...warnings];
```
sid597, line 126: Warnings are listed with the failures in the same per-node form and counted in the title, so a yellow callout never shows bullets its title does not account for. The warning text says the title was kept, which is what distinguishes it from a failure.

apps/roam/src/utils/__tests__/materializeSharedNode.test.ts
```diff
@@ -28,6 +28,13 @@ vi.mock("~/utils/importedSourceIdentity", () => ({
   writeImportedSourceIdentity: vi.fn(),
 }));
 
+// Runs before the imports above: getDiscourseNodes calls generateUID at module load.
+vi.hoisted(() => {
```
sid597, line 32: Same block in `importSharedNodes.test.ts`, `refreshImportedNode.test.ts` and `sourceSlot.test.ts`. `getDiscourseNodes.ts` builds `DEFAULT_NODES` at module load and calls `generateUID` while doing it; importing `sourceSlot` pulls that in, so `window` has to exist before the imports run. Copied from `roamToCrossAppConverters.test.ts`.

sid597:
Size justification, per the guideline: one problem, resolving a pulled node's source into its Roam title. 232 lines outside tests across 9 files; 480 of the 712 lines are unit tests for the identity paths and fallbacks the ticket lists. The files are coupled: the RID subtype fix in `sharedNodes.ts` is what makes the provenance lookup match, `findTargetUid` had to move out of `importSharedRelations.ts` to be shared, and the warning has to be carried from the materializer through the importer to the two surfaces that show it. Splitting would leave each half without a testable behavior.

mdroidian:
@coderabbitai full review

maparent:
Curious: You say this is stacked on [ENG-2156](https://linear.app/discourse-graphs/issue/ENG-2156/decorate-imported-node-titles-in-roam-from-core-title), but the PR #1382 specifies merging on main.

apps/roam/src/utils/importSharedNodes.ts
maparent: Why not reuse the topological sort that already exists as orderConceptsRec ?

maparent (approved):
LGTM, but I think one function can just go.

apps/roam/src/utils/findTargetUid.ts
```diff
@@ -0,0 +1,37 @@
...
```
mdroidian: Never use `window.roamAlphaAPI.q`

Always use `window.roamAlphaAPI.data`

Prefer `data.async.fast.q`

mdroidian (changes requested):
Missing `{source}` requires a placeholder, otherwise it won't be recognized as the original node type.

Decision notes: https://roamresearch.com/#/app/discourse-graphs/page/7H06vXWQc

apps/roam/src/utils/findTargetUid.ts
sid597: Changed the existence lookup to await window.roamAlphaAPI.data.async.fast.q. The shared UID stays bound through :in, and the tests now return promises so they exercise the async path.

apps/roam/src/utils/importSharedNodes.ts
sid597: Reused orderConceptsByDependency, which calls orderConceptsRec. A small adapter maps each SharedNode and its source reference to RIDs before sorting, then returns the original SharedNode objects. This avoids collisions between bare IDs from different spaces. The separate ordering algorithm is gone. Tests cover chained sources, cycles, and references absent from the batch.

apps/roam/src/utils/findTargetUid.ts
```diff
@@ -0,0 +1,36 @@
...
+      `[:find (?e) :where [?e :block/uid "${sourceLocalId}"]]`,
+    );
```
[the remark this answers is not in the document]
sid597: The starting head 80acda8b already binds the source ID through :in instead of interpolating it. That binding is preserved with data.async.fast.q. The regression test passes an ID containing Datalog syntax as a separate argument.

sid597:
**Decision: placeholder for a missing Source on pull**

Question: an imported node whose local format has `{Source}` kept the incoming title when no Source resolved, so Roam did not recognize it as its node type. Raised by mdroidian in review on 2026-09-14, with the decision notes on the Roam page linked there.

Decision (mdroidian, 2026-09-14): fill `{Source}` with a placeholder reference so the title keeps the node type's format.

Result: 95eab0db. `materializeSharedNode.ts` writes `[[@placeholder]]` when no Source resolves; the warning stays. Refresh keeps that title stable and replaces the placeholder once the real Source is imported. Formats without `{Source}` are unchanged. Covered in `materializeSharedNode.test.ts`; the cross-app round trips are in #1412 (ce597cad on the updated base).

sid597:
https://www.loom.com/share/d72b8b20dcae4118be403c869e5234bc

apps/roam/src/utils/sourceSlot.ts
```diff
@@ -11,6 +11,7 @@ import { extractFieldFromTitle } from "./extractContentFromTitle";
 // either of them.
 
 export const SOURCE_SLOT = "sourceDocument";
+export const MISSING_SOURCE_PLACEHOLDER = "(source missing)";
```
mdroidian: Change to `[[@placeholder]]` otherwise it won't serve the intended purpose.  Currently this placeholder would not let the Evidence node be identified as an evidence node.

EG: it has to match: `[[EVD]] - {content} - {Source}` and typically {Source} is defined as a page with this format: `@title`

Ideally it should grab the format from the linked reference node (eg: {Source} would see what the {Source} `format` is and adhere to that.  But we can defer that to v1.  If you decide to defer it, please create the ticket and link it here.

mdroidian (changes requested):

apps/roam/src/utils/sourceSlot.ts
sid597: I think we should fix this using the format. For now this will work because `[[EVD]] - {content} - {Source}` turns into `^[[EVD]] - (.*?) - (.*?)$` so any text after the second dash matches I checked the 4 graphs that we have `discourse-graphs`, `megacoglab`, `akamatsulab` and `discourse-dev-graph` (the one I tested on). Also you can see in the image there is the `publish` button that only appears because this page is being recognised as a dg node

mdroidian (approved):

PR #1383, opened 2026-09-02
ENG-2131 Review and accept imported relation types and triples in Roam

sid597:
## Reviewer brief

- Result: Settings → Grammar → Relations shows imported relation schemas in a separate read-only "Imported relations" table with their source and Provisional/Accepted status, plus Accept and Delete actions (delete is blocked while reified instances use the schema). Provisional schemas are excluded from relation creation (create relation dialog, suggestions, canvas tools) and from publishing.
- Review focus: acceptance is stored in the schema block's props (`discourse-graph.status: "accepted"`) next to ENG-1867's `importedFrom`, not in the settings record. Provisional is derived as "has `importedFrom` and not accepted", which also covers schemas imported before this PR. In `gatherCorrespondingRelations`, filtering the schema list is sufficient to also drop instances of provisional schemas (the existing schema-existence filter removes them).
- Note on ticket wording: Roam has no separate relation-type entity. A relation schema row is both the "type" and the "triple", so review/accept applies to that single entity, and no accept cascade is needed (Obsidian cascades triple → type because it stores them separately).
- Canvas semantics: provisional schemas stay in `discourseContext.relations` and in shape-util registration so existing arrows render and stay re-bindable (`isValidNodeConnection` passes `includeProvisional`); only creation paths filter them. Accepting updates mounted canvases live via `provisionalRelationIds`; the toolbar gains a new tool on remount when the accepted label previously had no accepted relation.
- Duplicate imports (ticket item, no new code): `importSharedRelations.matchImportedRelationSchemas` already dedupes by imported rid, then by localId, then by label + source + destination. An accepted schema keeps `importedFrom`, so a re-import maps onto it and does not reset its status. Deleting an imported schema and re-importing recreates it as provisional; there is no rejected state in v0.
- Intentionally unfiltered: query and display surfaces (`Export.tsx`, `getExportTypes`, `registerDiscourseDatalogTranslators`, `getDiscourseContextResults`, discourse context overlay, relation bindings) still see provisional schemas. The ticket scopes exclusion to creation and publishing.

## Verification

- Live Roam demo passes: imported actions stay centered before and after acceptance, and delete confirmation fits inside the dialog.
- `pnpm install --frozen-lockfile` and `pnpm ci:validate` pass from the repo root.
- New unit tests for the acceptance util (`relationSchemaAcceptance.test.ts`, 5 passing).
- `eslint`, `prettier`, and `tsc --noEmit` clean on touched files; remaining eslint warnings are pre-existing on untouched lines.
- A full pre-PR review ran in a fresh-context subagent; its findings are addressed in 9c05b197 (existing provisional arrows stay re-bindable, live canvas update on accept, error handling + PostHog on accept/delete, invisible instead of opacity-0 confirm buttons, sortable headers on the imported table, `getCreatableRelations` rename, shared constants).

## Loom video

https://github.com/user-attachments/assets/36169e81-b028-4c71-b90d-5449f69d8419

## Scope check

- [x] Ran `$scope-check` against ENG-2131 and the final diff.

### DIFF
--- a/apps/roam/src/components/CreateRelationDialog.tsx
+++ b/apps/roam/src/components/CreateRelationDialog.tsx
@@ -406,7 +406,8 @@ export const CreateRelationButton = (
       text="Add relation"
       disabled={extProps === null}
       onClick={() => {
-        renderCreateRelationDialog(extProps);
+        // A schema may have been accepted since this button last rendered.
+        renderCreateRelationDialog(relationProps);
       }}
     />
   );
--- a/apps/roam/src/components/CreateRelationDialog.tsx
+++ b/apps/roam/src/components/CreateRelationDialog.tsx
@@ -1,3 +1,4 @@
+import { useRelationSchemaRevision } from "~/utils/relationSchemaChanges";
 import React, { useState, useMemo } from "react";
 import { Dialog, Classes, Label, Button, Callout } from "@blueprintjs/core";
 import renderOverlay from "roamjs-components/util/renderOverlay";
@@ -388,6 +389,7 @@ export const renderCreateRelationDialog = (
 export const CreateRelationButton = (
   props: CreateRelationDialogProps & { fill?: boolean },
 ): React.JSX.Element | null => {
+  useRelationSchemaRevision();
   const { fill = false, ...relationProps } = props;
   const storedRelationsEnabled = getStoredRelationsEnabled();
   if (!storedRelationsEnabled) return null;
--- a/apps/roam/src/components/CreateRelationDialog.tsx
+++ b/apps/roam/src/components/CreateRelationDialog.tsx
@@ -8,6 +8,7 @@ import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageU
 import getDiscourseRelations, {
   type DiscourseRelation,
 } from "~/utils/getDiscourseRelations";
+import { excludeProvisionalRelationSchemas } from "~/utils/relationSchemaAcceptance";
 import { createReifiedRelation } from "~/utils/createReifiedBlock";
 import { getStoredRelationsEnabled } from "~/utils/storedRelations";
 import findDiscourseNode from "~/utils/findDiscourseNode";
@@ -291,7 +292,7 @@ const prepareRelData = (
 ): RelWithDirection[] => {
   nodeTitle = nodeTitle || getPageTitleByPageUid(targetNodeUid).trim();
   const discourseNodeSchemas = getDiscourseNodes();
-  const relations = getDiscourseRelations();
+  const relations = excludeProvisionalRelationSchemas(getDiscourseRelations());
   const nodeSchema = findDiscourseNode({
     uid: targetNodeUid,
     title: nodeTitle,
--- a/apps/roam/src/components/canvas/Tldraw.tsx
+++ b/apps/roam/src/components/canvas/Tldraw.tsx
@@ -1,3 +1,7 @@
+import {
+  isRelationSchemaDeleted,
+  useRelationSchemaRevision,
+} from "~/utils/relationSchemaChanges";
 import React, {
   useState,
   useRef,
@@ -150,7 +154,8 @@ export const discourseContext: DiscourseContextType = {
 };
 
 export const isAcceptedRelationSchema = (relation: { id: string }): boolean =>
-  !discourseContext.provisionalRelationIds.has(relation.id);
+  !discourseContext.provisionalRelationIds.has(relation.id) &&
+  !isRelationSchemaDeleted(relation.id);
 
 let activeCanvasPageUid: string | null = null;
 let activeCanvasEditor: Editor | null = null;
@@ -774,11 +779,7 @@ const TldrawCanvasShared = ({
       },
       {} as Record<string, DiscourseRelation[]>,
     );
-    discourseContext.provisionalRelationIds = new Set(
-      relations
-        .filter((r) => isProvisionalRelationSchema(r.id))
-        .map((r) => r.id),
-    );
+
     return relations;
   }, []);
   const allRelationsById = useMemo(() => {
@@ -790,11 +791,34 @@ const TldrawCanvasShared = ({
   const allRelationIds = useMemo(() => {
     return Object.keys(allRelationsById);
   }, [allRelationsById]);
+  const relationSchemaRevision = useRelationSchemaRevision();
+  const registeredRelationNames = useMemo(
+    () => [...new Set(allRelations.map((relation) => relation.label))],
+    [allRelations],
+  );
   const allRelationNames = useMemo(() => {
+    discourseContext.provisionalRelationIds = new Set(
+      allRelations
+        .filter((r) => isProvisionalRelationSchema(r.id))
+        .map((r) => r.id),
+    );
     return Object.entries(discourseContext.relations)
       .filter(([, relations]) => relations.some(isAcceptedRelationSchema))
       .map(([name]) => name);
-  }, []);
+    // Acceptance and deletion invalidate the relation data stored outside React.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [allRelations, relationSchemaRevision]);
+  useEffect(() => {
+    const editor = appRef.current;
+    if (!editor) return;
+    const tool = editor.getCurrentToolId();
+    if (
+      registeredRelationNames.includes(tool) &&
+      !allRelationNames.includes(tool)
+    ) {
+      editor.setCurrentTool("select");
+    }
+  }, [allRelationNames, registeredRelationNames]);
   const allNodes = useMemo(() => {
     const allNodes = getDiscourseNodes();
     discourseContext.nodes = Object.fromEntries(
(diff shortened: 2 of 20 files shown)


apps/roam/src/utils/relationSchemaAcceptance.ts
```diff
@@ -0,0 +1,60 @@
...
+import { setBlockPropsAsync } from "./setBlockProps";
+import {
+  isJsonObject,
+  parseImportedSourceIdentity,
+  type ImportedSourceIdentity,
+} from "./importedSourceIdentity";
+
+type ImportStatus = "provisional" | "accepted";
+
+export const RELATION_SCHEMA_STATUS_PROP_KEY = "status";
```
sid597, line 16: The prop key is `status` to match Obsidian's field (`ImportStatus` in `apps/obsidian/src/types.ts`), and provisional is derived the same way as Obsidian's `isProvisionalSchema`: has `importedFrom` and status is not accepted. ENG-1611 is the precedent. Schemas imported before this PR carry no status and therefore read as provisional, so no migration is needed.

apps/roam/src/components/settings/DiscourseRelationConfigPanel.tsx
```diff
@@ -1079,6 +1110,52 @@ const DiscourseRelationConfigPanel = ({
...
+      })
+      .catch((error: unknown) => {
+        internalError({
+          error,
+          type: "Discourse Relation: Accept failed",
+          userMessage: "Could not accept the imported relation.",
+        });
+      });
+  };
+  const handleDeleteImported = (rel: Relation) => {
```
sid597, line 1123: Imported instances arrive together with the schema (ENG-1867 creates them as tentative) and they count as in use here, so a schema that came with instances cannot be deleted until its instances are removed. That is deliberate: deleting the schema would orphan the tentative instances that ENG-1869's instance review still needs. ENG-1869 unblocks deletion by letting users remove those instances.

apps/roam/src/components/settings/DiscourseRelationConfigPanel.tsx
```diff
@@ -1243,6 +1320,101 @@ const DiscourseRelationConfigPanel = ({
           ))}
         </tbody>
       </HTMLTable>
+      {importedRelations.length > 0 && (
+        <>
+          <h4 className="mb-1 mt-6">Imported relations</h4>
```
sid597, line 1322: Imported relations get their own table rather than a badge in the local table because the ticket asks for imported schemas shown separately from local ones. The section is gated on data presence rather than a feature flag: these rows already exist in the graph, and hiding them while they stay excluded from creation would leave no way to accept or delete them. On main today they render as editable local rows, which is the bug this replaces.

apps/roam/src/components/settings/DiscourseRelationConfigPanel.tsx
```diff
@@ -976,6 +988,14 @@ type Relation = {
   source: string | undefined;
   destination: string | undefined;
 };
+type ImportedRelation = Relation & { importMeta: RelationSchemaImportMeta };
+
+const formatImportedSource = (sourceNodeRid: string): string => {
```
sid597, line 993: Obsidian formats source spaces with `formatImportSource` plus a spaceNames map (`apps/obsidian/src/utils/typeUtils.ts`). This version stays app-local and shows the raw space URI for non-Roam sources. A shared formatter next to the rid helpers is a follow-up candidate once Roam resolves space names.

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -778,7 +791,9 @@ const TldrawCanvasShared = ({
     return Object.keys(allRelationsById);
   }, [allRelationsById]);
   const allRelationNames = useMemo(() => {
```
sid597, line 799: A label drops out of the canvas toolbar only when every relation carrying it is provisional. A label with one local and one imported relation keeps its tool, and the endpoint-matching step picks from the accepted list only.

sid597 (commented):
Inline notes on the non-obvious decisions in this diff.

apps/roam/src/components/settings/DiscourseRelationConfigPanel.tsx
```diff
@@ -1079,6 +1107,30 @@ const DiscourseRelationConfigPanel = ({
       }, 50);
     });
   };
+  const handleAcceptImported = (rel: Relation) => {
+    void acceptImportedRelationSchema(rel.uid).then(() => {
+      setRelations(refreshRelations());
+    });
```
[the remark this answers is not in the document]
sid597: Fixed in 9c05b197: accepting now removes the schema id from discourseContext.provisionalRelationIds, so mounted canvases treat the relation as accepted immediately (context menus, drag handles, endpoint matching). The toolbar itself gains a new tool only on remount when the accepted label previously had no accepted relation; that limitation is noted in the PR body.

apps/roam/src/components/settings/DiscourseRelationConfigPanel.tsx
```diff
@@ -1243,6 +1295,98 @@ const DiscourseRelationConfigPanel = ({
...
+                          else setDeleteConfirmation(rel.uid);
+                        }}
+                      />
+                    </Tooltip>
+                    <Button
+                      intent={Intent.DANGER}
+                      onClick={() => handleDeleteImported(rel)}
+                      className={`mx-1 ${
+                        deleteConfirmation !== rel.uid ? "opacity-0" : ""
+                      }`}
```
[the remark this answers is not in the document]
sid597: Fixed in 9c05b197: the hidden Confirm/Cancel buttons now use the invisible class (visibility: hidden), which removes them from hit-testing and the tab order while preserving layout. The local table's identical pair predates this PR and is left as is.

sid597:
Re the size guideline: 251 of the changed lines are non-test. The hunks are coupled to one problem: the accept UI is only safe to ship together with the creation and publishing exclusions, otherwise imported schemas are either usable with no way to accept them (exclusions alone) or unusable with no way to enable them (UI alone). Splitting would leave one of those intermediate states on main.

mdroidian:
@coderabbitai full review

maparent (approved):
Did not look at canvas changes in depth, @mdroidian should do that, but all backend looks sound.

mdroidian (changes requested):
@sid597 Could you record a loom video testing the changes made to the canvas and any surrounding canvas features this might affect? Thanks!

sid597:
Part 1:

https://www.loom.com/share/ae0febd93dd543ce9ad0c0a71eecdf3d

sid597:
https://www.loom.com/share/167a1a4d4238493586e0dc57a3bb8b61

apps/roam/src/components/CreateRelationDialog.tsx
```diff
@@ -405,7 +406,8 @@ export const CreateRelationButton = (
       text="Add relation"
       disabled={extProps === null}
       onClick={() => {
-        renderCreateRelationDialog(extProps);
+        // A schema may have been accepted since this button last rendered.
+        renderCreateRelationDialog(relationProps);
```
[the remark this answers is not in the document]
sid597: Fixed in d3bf9667: `CreateRelationButton` and `SuggestionsBody` re-read creatable relations on each schema change.

apps/roam/src/components/settings/DiscourseRelationConfigPanel.tsx
```diff
@@ -1079,6 +1110,52 @@ const DiscourseRelationConfigPanel = ({
...
+            id: "discourse-relation-delete-blocked",
+            intent: Intent.WARNING,
+            content: `Cannot delete this imported relation: ${inUseCount} relation ${
+              inUseCount === 1 ? "instance uses" : "instances use"
+            } it in this graph.`,
+          });
+          setDeleteConfirmation(null);
+          return;
+        }
+        handleDelete(rel);
```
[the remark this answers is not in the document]
sid597: Fixed in d3bf9667: deleting marks the schema as deleted, and every creation path filters it out. Mounted canvases keep its metadata so existing arrows still render.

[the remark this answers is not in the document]
sid597: Fixed in d3bf9667: `handleDeleteImported` returns `handleDelete`'s promise, so failures reach the `internalError` catch.

apps/roam/src/components/canvas/Tldraw.tsx
```diff
@@ -777,9 +779,29 @@ const TldrawCanvasShared = ({
...
     return Object.keys(allRelationsById);
   }, [allRelationsById]);
-  const allRelationNames = useMemo(() => {
-    return Object.keys(discourseContext.relations);
-  }, []);
+  const registeredRelationNames = useMemo(
+    () => [...new Set(allRelations.map((relation) => relation.label))],
+    [allRelations],
+  );
+  const allRelationNames = useMemo(
```
sid597: `allRelationNames` is built once at mount on purpose. Re-rendering `TldrawCanvasShared` on a schema change rebuilds the store from the page-load snapshot, and the next edit saves over the page (fixed in 8635e1e3). Schema changes reach the canvas without a re-render:

- `DiscourseToolPanel` re-filters on `useRelationSchemaRevision`, so a deleted label's tool disappears live.
- Context menus, drag handles, and endpoint matching read acceptance each time they run.
- A label whose only schema you accept shows its tool after the canvas is reopened.

A label keeps its tool while any of its schemas is accepted.

PR #1392, opened 2026-09-03
ENG-1864 Define cross-app asset reference contract

maparent:
## Reviewer brief

The basic task is in the first commit, as given in the ticket. 
The ticket mentions "oversized" assets; this requires defining size limits, which we put in another commit.
We could separate this into another ticket, but thought it's a clear enough prerequisite to fold.
The extra scope is defined in a note in the body of the ticket: asset size caps (both incoming and outgoing.)

## Verification

The new reference contract is referenced in examples.
The asset size has tests. 

## Loom video

https://www.loom.com/share/be97dc663e7c43678eb0f8debc8af35c

## Scope check

- Scope beyond `Done When`:

The only extra scope is as noted at the end of the ticket .

https://linear.app/discourse-graphs/issue/ENG-1864/define-cross-app-asset-reference-contract

### DIFF
--- a/packages/database/src/crossAppContracts.ts
+++ b/packages/database/src/crossAppContracts.ts
@@ -74,6 +74,28 @@ type InlineCrossAppTypedContent = InlineCrossAppContent & {
   contentType: ContentType;
 };
 
+// An asset (an image or attachment) that a node's full content references.
+export type CrossAppAsset = {
+  // What the node's full content refers to, exactly as it wrote it: a path on a
+  // platform that addresses assets by path, a URL on one that addresses them by URL.
+  // Publication never rewrites it, so this is what a destination matches on, and it
+  // is unique within one node's `assets`: it maps to `FileReference.filepath`, which
+  // is part of that table's primary key.
+  sourceRef: string;
+  // The SHA-256 of the stored bytes as 64 lowercase hex characters, which is also
+  // their object name in shared storage. A destination looks the bytes up by this
+  // string exactly, so any other encoding fails as a not-found rather than a type error.
+  // Required: an asset whose bytes were not stored is represented by its absence
+  // from `assets`, not by an entry with nothing to resolve. Its `sourceRef` stays in
+  // the content, and the failure is reported by whichever transfer hit it.
+  contentHash: string;
+  // Where the source kept the asset: a name on a platform with a flat asset namespace,
+  // a path on one with folders. A destination derives the name and placement of its
+  // local copy by decomposing this as a path, and a bare name decomposes to itself.
+  // Absent when the source recorded nothing beyond `sourceRef`.
+  sourcePath?: string;
+};
+
 // A node instance
 export type CrossAppNode = CrossAppBase & {
   nodeType: LocalId;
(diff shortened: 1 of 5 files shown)


packages/database/src/lib/assetLimits.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Size caps for moving asset bytes across the shared-storage boundary.
+ *
+ * The two directions are separate constants on purpose. They happen to hold the same
+ * number today, but they answer different questions and are expected to diverge: the
```
mdroidian: When will they answer different questions?  Why can't we split them when we need them to answer the different questions instead of prematurely optimizing and adding additional unnecessary context?

packages/database/src/crossAppContracts.ts
```diff
@@ -74,6 +74,28 @@ type InlineCrossAppTypedContent = InlineCrossAppContent & {
   contentType: ContentType;
 };
 
+// An asset (an image or attachment) that a node's full content references.
+export type CrossAppAsset = {
+  // What the node's full content refers to, exactly as it wrote it: a path on a
```
mdroidian: Let's try to trim the comments down.  Also make sure to use the $dg-engineering-writing-style skill

mdroidian (approved):

packages/database/src/lib/assetLimits.ts
maparent: They're different so we don't believe (as is easy to do) that they answer the same question.
Short version: The first is a (configurable) supabase upload limit, which defaults to 6Mb but which we said we may change someday to accommodate larger PDF articles; the second is how big an asset we're expected to duplicate, and is really bound by Roam/firebase limits (100Mb/file according to Ness labs) but which we may want to keep low since we're bloating our user's roam usage.

This could be clarified in the comments.
I am not opposed to having a single constant if you feel it necessary, but then with a lot of comments to explain the difference.
BUT what I'm thinking now is that the second constant belongs in Roam rather than database, and it would make the split more natural. 
We could explicitly have: 
MAX_FIREBASE_ASSET_SIZE = 100Mb
MAX_IMPORTED_ASSET_BYTES = something smaller (or not.)

What do you think? Waiting on a decision here.

packages/database/src/crossAppContracts.ts
maparent: Claude put a lot of comments.
Some have to do with warning future developers to understand the code and avoid footguns, and I think those matter.
Others have to do with justifying design choices to the reviewer, and are probably less important to keep after the review (and should probably become GitHub comments.) 
Not sure it's actually easy to distinguish, but worth a shot trying.

OR did you just mean: let's ask it to be more terse in the comments it does put? 
Totally worth going over each PR and asking it to revise towards brevity.

mdroidian: Both more terse and asking it specifically to read and follow the AGENTS.md at root and use the referenced $dg-\* skill I mentioned.  Please do that for this PR specifically.

packages/database/src/lib/assetLimits.ts
maparent: Thinking again... I think that firebase's size is irrelevant as long as it's higher. And it makes sense to have the same limit. SO we should have a note saying the limit is bounded by the limit of all platforms, including the database. And given that, it makes sense to have a single constant.

packages/database/src/crossAppContracts.ts
maparent: Done as requested, see https://github.com/DiscourseGraphs/discourse-graph/pull/1392/commits/1d3bec449ad343a92c475d41ff0c5feea7584843
and let me know if it reads better to you.

PR #1394, opened 2026-09-04
ENG-1870 Copy Roam-origin assets into Supabase on publish

maparent:
## Reviewer brief

This is a large PR, there are a lot of moving pieces.

There are actually five subtasks, in distinct commits, which may be easier to understand:
https://roamresearch.com/#/app/discourse-graphs/page/4o_mub_mn

## Verification

There are unit tests.

## Loom video

Code walkthrough (long): https://www.loom.com/share/91078efd00b14f959ebf0ba07950a4fe

Note: I mention something missing during the code walkthrough, but upon verification it's fine.

Demo: https://www.loom.com/share/ecdd8cdfaca141cfb26216ecda5cdc08

## Scope check

- [ x] Ran `$scope-check` against the ENG ticket and final diff.
Actually it called me on the change to return parameters of addFile, but I told it that  CopiedAsset is the stage's record of what it stored, a stored asset's identity in shared storage is its hash, and the summary was previously conflating references with blobs. addFile's return is what makes that available without re-hashing. It accepted that justification as valid.

It's also pending the issue of whether the size limit should have been split between tickets, but right now it's within 1876.

- [ X] Ran a comprehensive review of the entire final diff in a subagent with a fresh context. Use `$dg-delegated-full-review` when no other full-review workflow is available.

https://linear.app/discourse-graphs/issue/ENG-1870/copy-roam-origin-assets-into-supabase-on-publish


apps/roam/src/utils/findAssetReferences.ts
```diff
@@ -0,0 +1,72 @@
...
+ * bytes into shared storage.
+ *
+ * Roam addresses assets by URL, and the markdown is never rewritten at publication. What
+ * this returns is therefore exactly what lands in `FileReference.filepath` and in
+ * `CrossAppAsset.sourceRef`.
+ */
+
+/** Roam uploads land in its own Firebase project, which we do not control. */
+const ROAM_STORAGE_HOST = "firebasestorage.googleapis.com";
+const ROAM_STORAGE_BUCKET = "firescript-577a2.appspot.com";
```
mdroidian: This might not be stable, probably isn't a good idea to rely on it.

apps/roam/src/utils/findAssetReferences.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/** Punctuation that ends a sentence rather than the URL it follows. */
+const TRAILING_PUNCTUATION = /[.,;:!?]+$/;
+
+/**
+ * Whether a URL points at Roam's own storage, and so at bytes this graph can fetch and
+ * publish. Anything else (an image hotlinked from another site, a link to a paper) is a
+ * resource that was never an asset of the node, and is left alone.
+ *
+ * Matching the bucket rather than the path shape: every Roam upload lands in this one
```
mdroidian: "lands in this one bucket" for now.  Not a good idea to rely on this.

mdroidian (approved):
See devin comments

apps/roam/src/utils/copyAssetToSharedStorage.ts
```diff
@@ -0,0 +1,101 @@
...
+    })
+  )
+    return skip(content.byteLength);
+
+  const contentHash = await addFile({
+    client,
+    spaceId,
+    sourceLocalId,
+    fname: assetUrl,
+    sourcePath: descriptor.filename,
```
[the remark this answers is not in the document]
maparent: This is addressed in 11ea0209be2a8e9c75fe2b0c520426d2381fa85f , part of eng-2230. They have to land together.

maparent:
@mdroidian re your comment about not relying on current bucket names in [https://github.com/DiscourseGraphs/discourse-graph/pull/1394#discussion_r3954629160](<https://github.com/DiscourseGraphs/discourse-graph/pull/1394#discussion_r3954629160>):
I think we should, but we should also record new buckets that follow the pattern. 
So here's a proposed change.
[https://roamresearch.com/#/app/discourse-graphs/page/qxLrH9j3U](<https://roamresearch.com/#/app/discourse-graphs/page/qxLrH9j3U>)

apps/roam/src/utils/findAssetReferences.ts
maparent: As also said in Linear: I suggest keeping them, but detecting new buckets. Details [here](https://github.com/DiscourseGraphs/discourse-graph/pull/1394#discussion_r3954629160).

mdroidian:
> @mdroidian re your comment about not relying on current bucket names in [#1394 (comment)](https://github.com/DiscourseGraphs/discourse-graph/pull/1394#discussion_r3954629160): I think we should, but we should also record new buckets that follow the pattern. So here's a proposed change. https://roamresearch.com/#/app/discourse-graphs/page/qxLrH9j3U

Please put the problem statement as the very first thing in a proposal.  One shouldn't haven't to read the entire proposal to get to the problem statement.

Why not just rely on a regex pattern without the specific bucket name for `isRoamStorageUrl`?  Use Roam's API to confirm the URL is their bucket (which is part of the process anyway).

mdroidian (changes requested):
Sorry, I'm going to block on storing/relying on Roam's current bucket URL.  This is setting us up for failure in the future unncessarily.

Use a regex pattern (that doesn't include their specific bucket name) + roam's API to confirm it's roam's bucket.

maparent:
I see your point. Will work on this.

maparent:
Investigated. I assumed you meant roamAlphaAPI.file.get. I'm not using it now, and that's a mistake for encrypted graphs; so worth the pushback. BUT it does not distinguish roam buckets from arbitrary firebase buckets, so it does not do what you want (gate on Roam content.) This also has security implications, as noted.

mdroidian:
How are you getting the filename?

mdroidian:
Why are we not using the roam API to get the files?

mdroidian:
Why wouldn't using roamAlphaAPI distinguish roam buckets from arbitrary firebase buckets?  If you use roamAlphaAPI on a legitimate file in a roam bucket, you will get a success response.  If you use roamAlphaAPI on a non roam bucket, I'm assuming you will get an error response (unauthorized).  This would solve the security implication,  yes?

mdroidian:
Even if the roamAlphaAPI just performs a standard fetch, it isn't clear what the security implication actually is compared to downloading bytes from roam's bucket, or an arbitrary obsidian file.

maparent:
Why not using roam API: I did not think of the encryption case, and Claude chose a common API. Working on reverting that. It also means we'll get the filename and mimetype from the file. However, I checked that the lastModified that we get from file.get is spurious (always now) whereas the one we get from metadata is valid and, more important, stable, so keeping the metadata call for now. I thought it was reliable for size, but again only for unencrypted graphs. There's a shortcut that depended on that, becomes conditional. 
As to "Why wouldn't using roamAlphaAPI distinguish roam buckets from arbitrary firebase buckets?", no idea why, this was an empirical test. I took a roam URL, altered a character in the bucket name, and though it came back with a 404, I did see in the network pane that it did attempt to fetch it. (Non-firebase URLs throw, at least.) So again, we cannot rely on Roam to distinguish roam bucket from non-roam bucket, so I maintain the analysis that we need to record other bucket usage, but refuse it at the outset for security reasons.

maparent:
BTW making assumptions about encrypted graphs; if you could create one and give me access, I'd love to check some of them. But I doubt it would change anything deeply.

maparent:
> How are you getting the filename?

From the metadata, in `fetchAssetDescriptor`.

> Why are we not using the roam API to get the files?

Corrected.

Now identifying roam buckets by URL as agreed.

mdroidian (approved):
Approving to unblock.  But a PR this size will be rejected outright in the future.  Please adhere to the DevOps/Style Guide for future submissions.

PR #1400, opened 2026-09-05
ENG-1249 Implement Discourse context overlay feature in Obsidian

trangdoan982:
## Reviewer brief

**Result:** Links to a discourse node carry an inline badge with that node's relation count, in both Live Preview and Reading view. Selecting one opens a popover built on `RelationshipSection` — the same component the Discourse Context panel renders — so the two cannot disagree. A node with no relations still badges, reading `0`, and its popover says "No discourse relation found". Toggled by **Show discourse context overlay** in General settings, applied without a reload.

**Review focus:**

- **Two render surfaces, one badge.** Live Preview goes through a CM6 `ViewPlugin`, following the existing wikilink drag handle. Reading view needs a markdown post processor, which this plugin had no precedent for. They share `createDiscourseContextBadge` so they cannot drift visually — worth checking that the two paths stay in step.
- **`RelationsIndex` is the load-bearing piece.** `loadRelations` costs a full vault file read plus a JSON parse per call, which is fine once per panel open but not once per link on screen. The index keeps a parsed snapshot so a render path can ask synchronously. Its concurrency rules are subtle and commented: a generation counter (relations.json being written mid-read is the *normal* case, since saving a relation causes it), the snapshot kept rather than dropped on invalidation, and reads deliberately not scheduling loads.
- **What the badge counts.** It must match what the panel will actually display, not what the store holds. Relations awaiting acceptance after an import are excluded, and so are relations whose relation type was deleted — deleting a type leaves its relations behind in `relations.json` and the panel drops them. See `countDisplayableRelations`.

**Risk or follow-up:**

- `buildGroupedRelations` also drops relations whose other endpoint no longer resolves to a file. The badge does not model that, so a relation pointing at a deleted note will still be counted. Not fixed here because it needs a file lookup per relation on a render path; worth a follow-up ticket if orphaned endpoints turn out to be common.
- Badges are emitted into Export to PDF/HTML output, since post processors run there too.
- Live Preview skips a link that straddles a viewport-range boundary, and markdown links without a literal `.md` before the closing paren.

## Verification

- `pnpm install --frozen-lockfile` + `pnpm ci:validate` from the repo root: 8/8 check-types, 4/4 test:unit.
- No unit tests: `apps/obsidian` has no test setup, and adding one was deliberately left out of this PR. The counting, grouping and link-parsing rules are isolated in pure modules (`relationsEndpointIndex`, `discourseLinkFrontmatter`, `internalLinkParsing`) so they can be covered later without restructuring.
- Driven against a real vault over CDP: 16 assertions covering both surfaces, badge count matching the popover's rows, popover open/scroll/dismiss, the empty state, and the setting adding and removing badges live. Stable across repeated runs.

Three bugs were found by that pass rather than by tests, and are fixed here: the badge over-counting orphaned relations, `ignoreEvent()` inverted so clicks never reached the badge, and scrolling inside the popover dismissing it — which made "Add a new relation" unreachable.

## Scope check

- Scope beyond `Done When`: **None.** The ticket's three criteria — a General setting, the overlay showing the same relations as the panel, and documentation — are each covered.

Its high-severity findings are fixed in `c1a7b49f2`: Reading view badges never refreshing, the index flashing every badge to `0` on save, an `inFlight` leak that wedged the snapshot permanently, a read-triggered refresh loop that left Reading view blank, the popover outliving plugin unload with an Escape handler that swallowed the key, and positioning that measured before React had committed so it never flipped up near the window bottom. Remaining findings are listed under Risk or follow-up above.

## Loom video

Not recorded. Screenshots of both surfaces and the empty state are attached to the Linear ticket discussion instead.


trangdoan982:
Superseded by a three-PR stack, since this was over the review-size guideline:

- #1413 — relations index + link resolution (~290 lines, no user-visible change)
- #1414 — Live Preview overlay (~500 lines)
- #1415 — Reading view overlay (~110 lines)

Same final state as this branch; each is verified independently against a real vault.

PR #1420, opened 2026-09-08
ENG-2188 Node type settings are spread across seven inner tabs; put them on one page

trangdoan982:
## Reviewer brief

- **Result:** a node type's settings are one vertical page instead of seven inner tabs. Groups, in order: Identity (Index, Description, Tag, Color, Format), Creation (Shortcut, Template), Canvas, Legacy (Specification), Suggestive mode (shown only when sync is on), Attributes. Index and Template are drill-down rows. Each opens its own sub-page with a `Grammar › Nodes › <node> › Index` breadcrumb.
- **Review focus: `NodeConfig.tsx` is a re-layout.** No setting changes its key, default, or write path. `SettingsDrillDownRow` renders a row that pushes a segment on the ENG-2186 route. `NodeIndexPage` and `NodeTemplatePage` wrap the existing panels for the two sub-pages.
- **Review focus: the Specification row.** The toggle now reads `Enabled`. Its block text stays lowercase `enabled` through `blockKey`, so storage is unchanged. The query builder renders only while the toggle is on.
- **Review focus: group order differs from the ticket.** The ticket lists Identity / Recognition / Creation / Index. Review dropped Recognition, moved Index and Format into Identity, moved Specification into a `Legacy` group for settings mid-migration, and moved Attributes last. Format's `DEPRECATED — use Specification` note went with it, because Specification is now the setting in flux.
- **Follow-up:** Suggestive mode stays on the node page. The ticket's Solution suggests moving it to the Admin panel, but it needs its own node selector there, so that move needs its own ticket. The Index row's description is proposed copy and needs sign-off.

## Verification

- `tsc --noEmit` and eslint report no errors or warnings for `apps/roam`. The production build completes with 0 errors.
- Checked in Roam: Grammar › Nodes → node type → Index and Template. Back and the breadcrumb work at each depth. Toggling `Enabled` shows and hides the query builder. Legacy and Attributes are the last two groups.

## Loom video

https://www.loom.com/share/f76e4399d7784cf2a85613fd91301e78

## Scope check

- Scope beyond `Done When`: group names and order follow the manager's review instead of the ticket text (see the review focus above). The ticket needs updating to match.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1420" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
+++ b/apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useMemo } from "react";
-import { Button, Intent } from "@blueprintjs/core";
+import { Button, Collapse, Intent } from "@blueprintjs/core";
 import DualWriteBlocksPanel from "./components/EphemeralBlocksPanel";
 import getSubTree from "roamjs-components/util/getSubTree";
 import { DiscourseNode } from "~/utils/getDiscourseNodes";
@@ -17,6 +17,7 @@ import {
   TEMPLATE_SETTING_KEYS,
 } from "~/components/settings/utils/settingKeys";
 import { RenderRoamBlock } from "~/utils/roamReactComponents";
+import Description from "~/components/settings/SettingsDescription";
 import { ROAM_DOCS, withDocsLink } from "./utils/docs";
 
 const DiscourseNodeSuggestiveRules = ({
@@ -42,6 +43,7 @@ const DiscourseNodeSuggestiveRules = ({
       }).uid || "",
     [nodeUid],
   );
+  const [isTemplateOpen, setIsTemplateOpen] = useState(false);
 
   const [isUpdating, setIsUpdating] = useState(false);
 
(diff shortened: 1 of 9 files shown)


apps/roam/src/styles/settingsStyles.css
```diff
@@ -81,3 +81,10 @@
 .dg-settings-page-header__crumb {
   cursor: pointer;
 }
+
+.dg-settings-node-page {
```
mdroidian, line 85: @trangdoan982 tailwind

trangdoan982: Done — the `.dg-settings-node-page` rule is gone and the container is `flex flex-col gap-12 p-1`, matching the sibling node pages. (`gap-12` ≈ 42px at Roam’s 14px root, standing in for the old 44px.)

apps/roam/src/components/settings/components/SettingsDrillDownRow.tsx
```diff
@@ -0,0 +1,31 @@
...
+}: {
+  title: string;
+  description: React.ReactNode;
+  buttonText: string;
+  onClick: () => void;
+}): JSX.Element => (
+  <Label>
+    {title}
+    <Description description={description} />
+    <div>
```
mdroidian, line 19: Is this div necessary?

apps/roam/src/components/settings/DiscourseNodeSuggestiveRules.tsx
```diff
@@ -72,18 +74,39 @@ const DiscourseNodeSuggestiveRules = ({
...
+          <Description
+            description={withDocsLink(
+              `The template that auto fills ${node.text} page when generated.`,
+              ROAM_DOCS.creatingNodes,
+            )}
+          />
+        </div>
+        {/* Collapse unmounts its children, so the buffer block only exists while open. */}
+        <Collapse isOpen={isTemplateOpen}>
+          <div className="pl-6 pt-2">
```
mdroidian, line 96: Is this div necessary?  Or can the classes just be on the <collapse>?

mdroidian (approved):

PR #1425, opened 2026-09-09
FEE-86  - add Roam user guide to share your ideas and research

KarolaKirsanow:
This PR adds a "shared your ideas and research" use case doc to the Roam website documentation.


jsmorabito (commented):
I like all the content, but the `Drafting a manuscript from your discourse graph` section strikes me as a distinct use case from Sharing ideas and research. What do you think about maybe promoting it to its own use case section?

KarolaKirsanow:
> I like all the content, but the `Drafting a manuscript from your discourse graph` section strikes me as a distinct use case from Sharing ideas and research. What do you think about maybe promoting it to its own use case section?

I see your point. I'm in favor of getting this material on the website to material the Obsidian docs and then optimizing the arrangement in a second pass.

jsmorabito:
sounds good to me!

PR #1427, opened 2026-09-09
ENG-1356 Implement image to node conversion flow via icon button in tldraw Roam

trangdoan982:
https://www.loom.com/share/ba1a4a782b3c4ac6b0e713bd4561a8fb

Note: I updated the styling to match Figma design after recording this Loom. "Image" is now highlighted and embeded image is flushed to the left:
image

## Reviewer brief

- Result: hovering (or selecting) a tldraw image shape on the Roam canvas shows a convert button at its top-left. Clicking it opens the create node dialog with the type picker enabled, Page and Block hidden, and an image preview. Confirming replaces the image with a discourse node shape and adds the image to the node page. The right-click **Convert To** path is unchanged in UI and now shares the upload, image-insertion, and shape-replacement helpers.
- Review focus: `ImageConvertOverlay.tsx` hover tracking via `editor.getHoveredShape()` inside `useValue`; the additive `nodeType` field in the `ModifyNodeDialog.tsx` success payload; and `handleImageCreation` in `createDiscourseNode.ts`, which now always writes the image block for a canvas conversion and uses Key Image only to choose the query-builder placeholder (previously the image was dropped when Key Image was off).
- Optional commit, easy to drop: `72733d84a` "Reuse the canvas asset URL instead of re-uploading the image". Files dropped or pasted onto the canvas are already uploaded to Roam by the canvas handlers and stored as the asset `src`, so conversion now reuses that URL and only re-uploads when the `src` is a `data:` or `blob:` URL. This affects both flows. If we prefer every conversion to create its own copy of the file, revert or drop that single commit; nothing else depends on it.
- Risk or follow-up: the hover flow deliberately hides Block; images can still be turned into blocks through the context menu. Locked images do not show the button because tldraw skips locked shapes in `deleteShapes`. The canvas node shape still displays the image only for node types with Key Image on.

## Verification

- `pnpm install --frozen-lockfile` and `pnpm ci:validate` pass (226 Roam unit tests).
- Roam dev build succeeds with 0 errors; fresh-context subagent review of the full diff found nothing blocking, and its should-fix items are applied in the second commit.
- Manual check in a Roam graph (hover, dialog, replacement, image on the node page, context-menu regression) still to be done by the reviewer or author; Chrome automation was unavailable in this session.

## Scope check

- Scope beyond `Done When`: the always-add-image change and the URL reuse commit also apply to the existing right-click flow, because both flows share one implementation. Requested by Trang during review so both flows behave identically.

### DIFF
--- a/apps/roam/src/components/ModifyNodeDialog.tsx
+++ b/apps/roam/src/components/ModifyNodeDialog.tsx
@@ -539,14 +539,14 @@ const ModifyNodeDialog = ({
           </div>
 
           {imageUrl && (
-            <div className="flex w-full flex-col gap-1">
-              <span>Image</span>
+            <Label className="w-full">
+              Image
               <img
                 src={imageUrl}
                 alt=""
-                className="max-h-40 w-full rounded object-contain"
+                className="mt-1 block max-h-40 max-w-full rounded"
               />
-            </div>
+            </Label>
           )}
 
           {/* Referenced Node Input */}
--- a/apps/roam/src/components/ModifyNodeDialog.tsx
+++ b/apps/roam/src/components/ModifyNodeDialog.tsx
@@ -597,14 +597,14 @@ const ModifyNodeDialog = ({
           </div>
 
           {imageUrl && (
-            <Label className="w-full">
-              Image
+            <div className="flex w-full flex-col gap-1">
+              <span>Image</span>
               <img
                 src={imageUrl}
                 alt=""
-                className="mt-1 max-h-40 w-full rounded object-contain"
+                className="max-h-40 w-full rounded object-contain"
               />
-            </Label>
+            </div>
           )}
 
           {/* Referenced Node Input */}
--- a/apps/roam/src/components/ModifyNodeDialog.tsx
+++ b/apps/roam/src/components/ModifyNodeDialog.tsx
@@ -58,6 +58,7 @@ export type ModifyNodeDialogProps = {
     text: string;
     uid: string;
     action: string;
+    nodeType?: string;
   }) => Promise<void>;
   onClose: () => void;
 };
@@ -371,6 +372,7 @@ const ModifyNodeDialog = ({
             text: content.text,
             uid: content.uid,
             action: "create",
+            nodeType: selectedNodeType?.type,
           });
 
           onClose();
@@ -480,6 +482,7 @@ const ModifyNodeDialog = ({
           text: formattedTitle,
           uid: newPageUid,
           action: "create",
+          nodeType: selectedNodeType?.type,
         });
       } else {
         // Edit mode: update the existing block
@@ -515,6 +518,7 @@ const ModifyNodeDialog = ({
           text: updatedContent,
           uid: sourceBlockUid || content.uid,
           action: "edit",
+          nodeType: selectedNodeType?.type,
         });
       }
       onClose();
@@ -591,6 +595,17 @@ const ModifyNodeDialog = ({
               />
             </Label>
           </div>
+
+          {imageUrl && (
+            <Label className="w-full">
+              Image
+              <img
+                src={imageUrl}
+                alt=""
+                className="mt-1 max-h-40 w-full rounded object-contain"
+              />
+            </Label>
+          )}
 
           {/* Referenced Node Input */}
           {referencedNode && !isContentLocked && mode === "create" && (
--- a/apps/roam/src/components/ModifyNodeDialog.tsx
+++ b/apps/roam/src/components/ModifyNodeDialog.tsx
@@ -31,12 +31,12 @@ import {
   getNewDiscourseNodeText,
   getReferencedNodeInFormat,
 } from "~/utils/formatUtils";
-import createDiscourseNode from "~/utils/createDiscourseNode";
+import createDiscourseNode, {
+  handleImageCreation,
+} from "~/utils/createDiscourseNode";
 import { OnloadArgs } from "roamjs-components/types";
 import { render as renderToast } from "roamjs-components/components/Toast";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
-import resolveQueryBuilderRef from "~/utils/resolveQueryBuilderRef";
-import runQuery from "~/utils/runQuery";
 import posthog from "posthog-js";
 
 export type ModifyNodeDialogMode = "create" | "edit";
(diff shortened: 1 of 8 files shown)


mdroidian (changes requested):
To help the review, can you go through and write inline github comments on 
- code sections (both removed and added) that are purely refactors with no functional change.
- code that is outside of the original scope

Could you also record a loom testing
- nodes that have Key Image defined via query block (making sure the image shows up in the correct location)
- the newly added image URL functionality
- testing the new changes to the right click flow

And finally, change the icon and icon positioning to match the design
- use `document-open` icon
- move to top right

trangdoan982 (commented):
Annotating the diff as requested: which hunks are pure refactors and which are outside the ticket scope.

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -273,66 +274,6 @@ const ModifyNodeDialog = ({
     onClose();
   }, [onClose]);
 
-  const addImageToPage = useCallback(
```
trangdoan982, line 276: Refactor only: this local `addImageToPage` duplicated `handleImageCreation` in `createDiscourseNode.ts` (same Key Image / query-builder placeholder branches). Deleted here in favour of the shared helper; no behaviour change for this dialog.

apps/roam/src/components/ModifyNodeDialog.tsx
```diff
@@ -358,11 +299,13 @@ const ModifyNodeDialog = ({
           if (imageUrl) {
             const pageUid = content.uid || getPageUidByPageTitle(content.text);
             if (pageUid) {
-              await addImageToPage({
+              await handleImageCreation({
```
trangdoan982, line 302: Refactor only: replaces the deleted local `addImageToPage` with the exported `handleImageCreation`. The extra `discourseNodes`/`text` args are just what the shared helper needs; the resulting block writes are identical.

apps/roam/src/components/canvas/uiOverrides.tsx
```diff
@@ -171,8 +173,6 @@ export const getOnSelectForShape = ({
   editor: Editor;
   extensionAPI: OnloadArgs["extensionAPI"];
 }) => {
-  const { x, y } = shape;
```
trangdoan982, line 174: Refactor only: this x/y capture moved into `replaceShapeWithDiscourseNode` in `convertShapeToDiscourseNode.ts`, which reads it off the same `shape` object.

apps/roam/src/components/canvas/uiOverrides.tsx
```diff
@@ -188,54 +188,26 @@ export const getOnSelectForShape = ({
       includeDefaultNodes: true,
       disableNodeTypeChange: true,
       imageUrl,
-      onSuccess: async ({ text, uid }) => {
```
trangdoan982, line 191: Refactor only: this body moved verbatim to `replaceShapeWithDiscourseNode` in the new `convertShapeToDiscourseNode.ts` — same `deleteShapes` → `calcCanvasNodeSizeAndImg` → `createShapes` sequence with the same props.

apps/roam/src/components/canvas/uiOverrides.tsx
```diff
@@ -188,54 +188,26 @@ export const getOnSelectForShape = ({
...
-        ]);
-      },
+        }),
       onClose: () => {},
     });
   };
 
   if (shape.type === "image") {
     return async () => {
-      const { assetId } = (shape as TLImageShape).props;
```
trangdoan982, line 229: Refactor: this asset-read-and-upload block moved to `uploadImageShapeToRoam` in `convertShapeToDiscourseNode.ts` so the new hover-button flow can share it. The only change inside the moved code is the new `https:` short-circuit — see the comment on that line.

apps/roam/src/components/canvas/uiOverrides.tsx
```diff
@@ -188,54 +188,26 @@ export const getOnSelectForShape = ({
...
-      if (!assetId) return;
-      const asset = editor.getAsset(assetId);
-      if (!asset || !asset.props.src) return;
-      const file = await fetch(asset.props.src)
-        .then((r) => r.arrayBuffer())
-        .then((buf) => new File([buf], shape.id));
-      // this is a promise
-      // eslint-disable-next-line @typescript-eslint/await-thenable
-      const src = await window.roamAlphaAPI.util.uploadFile({ file });
+      const src = await uploadImageShapeToRoam({
```
trangdoan982, line 206: Refactor: same upload path as before, now via the shared helper. The added `if (!src) return;` mirrors the old early returns on missing `assetId` / `asset.props.src`.

apps/roam/src/components/canvas/convertShapeToDiscourseNode.ts
```diff
@@ -0,0 +1,66 @@
...
+}: {
+  editor: Editor;
+  shape: TLImageShape;
+}): Promise<string | undefined> => {
+  const { assetId } = shape.props;
+  if (!assetId) return undefined;
+  const asset = editor.getAsset(assetId);
+  if (!asset || !asset.props.src) return undefined;
+  // Dropped and pasted files are already uploaded to Roam by the canvas handlers
+  if (asset.props.src.startsWith("https:")) return asset.props.src;
```
trangdoan982, line 18: Out of scope (flagged in the PR body, commit 72733d84a): reusing the existing asset URL instead of re-uploading also changes the pre-existing right-click Convert To flow, which previously always re-uploaded the file to Roam. Dropping that single commit reverts this.

apps/roam/src/components/canvas/convertShapeToDiscourseNode.ts
```diff
@@ -0,0 +1,66 @@
...
+  const file = await fetch(asset.props.src)
+    .then((r) => r.arrayBuffer())
+    .then((buf) => new File([buf], shape.id));
+  return window.roamAlphaAPI.util.uploadFile({ file });
+};
+
+export const replaceShapeWithDiscourseNode = async ({
```
trangdoan982, line 25: Refactor only: body moved verbatim from the `onSuccess` handler in `uiOverrides.tsx`; the only difference is that `x`/`y` are destructured inside the function rather than in the enclosing `getOnSelectForShape`.

apps/roam/src/utils/createDiscourseNode.ts
```diff
@@ -54,7 +55,8 @@ const handleImageCreation = async ({
   imageUrl?: string;
   extensionAPI?: OnloadArgs["extensionAPI"];
   text: string;
-}) => {
+}): Promise<void> => {
+  if (!imageUrl) return;
```
trangdoan982, line 59: Out of scope (flagged in the PR body, commit 0fde61b17): the old code only wrote the image block when Key Image was enabled; now the block is always written and Key Image only selects the query-builder placeholder. This affects the existing right-click conversion and the other `handleImageCreation` callers, not just the new flow.

apps/roam/src/utils/createDiscourseNode.ts
```diff
@@ -64,38 +66,29 @@ const handleImageCreation = async ({
...
+      inputs: { NODETEXT: text, NODEUID: pageUid },
+    });
+    const imagePlaceholderUid = results.allProcessedResults[0]?.uid;
+    if (imagePlaceholderUid) {
+      await updateBlock({ uid: imagePlaceholderUid, text: imageMarkdown });
+      return;
     }
   }
+
+  await createBlock({
```
trangdoan982, line 87: Consequence of the above: this `createBlock` is now unconditional, so node types with Key Image disabled get an image block on the page where previously they got none.

apps/website/content/roam/guides/using-the-canvas.mdx
```diff
@@ -217,11 +217,15 @@ _sending query results to your canvas_
...
+You can select images or text from your graph, a document, a file, or a webpage, copy them to your canvas, and convert them into discourse nodes.
+
+To convert an image, hover over it and click the convert icon that appears in its top-left corner. In the create node dialog, pick the node type, enter a title, and confirm. The image is replaced on the canvas by the new node and added to the node's page.
+
+You can also right-click an image or text shape and select **Convert To**, pick the node type, then modify the title as needed. Use this path to convert text shapes or to turn an image into a block.
 
 <Callout type="info" emoji="💡">
-  Remember to enable the **Key Image** feature in Settings if you want to
-  convert images to discourse nodes.
+  Enable the **Key Image** feature in Settings for a node type if you want its
```
trangdoan982, line 227: Out of scope: this callout documents the `handleImageCreation` behaviour change (Key Image no longer gates whether the image is added), not the new convert button.

PR #1429, opened 2026-09-10
ENG-2247 Copy a node's assets and rewrite its markdown for Roam

maparent:
## Reviewer brief

This creates a utility function, importAssetNodes, that ties together rewriteAssetLinks (ENG-2217) and mirrorAssetToRoamStorage (ENG-2218) to actually import the nodes detected in given markdown.
This also involves a refactoring of a repeated and defective utility function.
Finally, this includes some end-to-end tests (in a separate commit) that rely on importAssetNodes.

## Verification

Includes unit tests.
Behaviour tests in #1401 

## Loom video

https://www.loom.com/share/f941f20853444597931d85d6b56ef68c

## Scope check

- Scope beyond `Done When`: None. ENG-2247's criteria were written after the split, so
  they describe this diff by construction. The list below measures against the original
  ENG-2219 `Done When`, which is what was agreed before the split.

Beyond the original ENG-2219 scope:

1. **The tokenizer API.** `rewriteAssetLinks` exports `lookupCandidates` and gains
   `collectAssetLocators`, so the stage copies only references the rewriter can resolve.
   - Required now: without it the stage copies every recorded row, including ones whose
     locator no longer appears in the fetched markdown once frontmatter and the title
     heading are stripped. A Roam upload is a permanent charge on the user's storage, so
     this is reachable the moment the stage ships. Deriving the spellings forward instead
     does not work: a note writes `fig#1.png` as `fig%231.png`, and `encodeURI` leaves
     `#` alone.
   - `parseMatch`, which this depends on, went to ENG-2217 as a pure refactor.

2. **`getErrorMessage` consolidated** into `apps/roam/src/utils/getErrorMessage.ts`,
   replacing the copies in `materializeSharedNode` and in `mirrorAssetToRoamStorage` (an
   ENG-2218 file), plus a branch for the `{ message }` objects Supabase returns.
   - Required now: No. A module-local copy in `importNodeAssets` would have satisfied the
     ticket. This is cleanup, and it changes the error text of three existing callers. It
     consolidates the two copies that had to share it; other inline copies elsewhere in
     the codebase are untouched.

3. **`assetDegradation.test.ts`** asserts the degradation path across the publish and
   import seam, calling ENG-1870's `publishNodeAssets` alongside `importNodeAssets`.
   - Required now: it satisfies task 6.1 of the `cross-app-asset-references` OpenSpec
     change. The contract between the two halves cannot be asserted from either side
     alone.
   - Decision: recorded on ENG-2219, comment of 2026-09-06.

4. **A doc comment** in `mirrorAssetToRoamStorage` recording that nothing rolls an upload
   back. Comment only, no behaviour change.
   - Required now: `importNodeAssets` depends on that constraint and it was uncaptured.

- Anyone affected or consulted: No, except item 3, agreed on ENG-2219 before the split.
- Decision: item 3 as above; the rest are recorded in ENG-2247's description.

Since then, added a commit. New scope check:

## Scope check

- [x] Ran `$scope-check` against ENG-2247 and commit 03496dccb..
- Scope beyond `Done When`:
  - `importNodeAssets` marks the link of an asset it could not copy. It writes `Failed to import` when the copy threw and `Too large for import` when the asset was skipped for size.
    - A Roam-origin URL keeps its link or embed, followed by `(marker)`.
    - An Obsidian vault path becomes `[label (marker)](locator)`, or `[marker](locator)` when the link has no label.
  - This replaces the criterion "leaves an unresolved locator unchanged". A locator with no `FileReference` row is still left unchanged.
  - `importNodeAssets` throws when it can't read the node's references, instead of reporting a failure with no locator. `FailedImport.sourceLocator` is now required. This replaces the criterion "A failure about the node rather than one asset reports no locator".
  - The marker shows a degraded asset to the user, which this ticket lists as out of scope for ENG-1877.
- Required now: once ENG-2219 wires the stage in, a failed copy still records the page as up to date. A normal import then skips the page and never retries the asset. Without the marker, the user has no sign that a forced refresh would fix the page.
  - The marker tells the user to use the existing refresh button, which retries the copy. Assets that already copied are reused.
  - If the reference query fails, `materializeSharedNode` (ENG-2219) fails the node at `copy-assets`. Nothing is written and the stored timestamp doesn't change, so the next impor
t retries.
  - ENG-1877 still owns reporting failures in the import results.
- Anyone affected or consulted: Yes. Michael decided on marking the link in the page. Marker text, link forms and the query failure were settled afterwards within that approach.
- Decision: https://github.com/DiscourseGraphs/discourse-graph/pull/1401#discussion_r3992516055

https://linear.app/discourse-graphs/issue/ENG-2247/copy-a-nodes-assets-and-rewrite-its-markdown-for-roam


apps/roam/src/utils/getErrorMessage.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * The message to show for something that was thrown, wherever it came from.
+ *
+ * Shared by `materializeSharedNode` and the asset stage it calls. In either file, the
```
mdroidian, line 4: This is likely not going to age well, and probably not helpful.

If the comments are more lines than the function, we should consider how relevant the comment actual is.

apps/roam/src/utils/mirrorAssetToRoamStorage.ts
```diff
@@ -18,6 +19,14 @@ import { readMirroredAssetUrl, recordMirroredAsset } from "./assetRegistry";
  * detection back into the destination, and would leave this graph's page depending on a
  * blob the origin graph's owner can delete. The copy is irrevocable, deliberately.
  *
+ * **Nothing rolls an upload back.** `file.delete` exists and takes a URL, so a caller
```
mdroidian, line 22: The amount of comments here is getting kind of ridiculous.  It's probably worthwhile to double check that they are relevant and required.

mdroidian (changes requested):
Blocking on the amount of comments.

Can you either:
a) ask claude to respect the AGENTS.md and STYLE_GUIDE.md, and use `$dg-engineering-writing-style`
b) manually walk through and trim down the comments to only what is essential, as per the STYLE_GUIDE.md

And address devin's review comment.

apps/roam/src/utils/importNodeAssets.ts
```diff
@@ -0,0 +1,231 @@
...
+ */
+export const importNodeAssets = async ({
+  client,
+  sharedNode,
+  markdown,
+}: {
+  client: DGSupabaseClient;
+  sharedNode: SharedNode;
+  markdown: string;
+}): Promise<{ markdown: string; report: AssetImportReport }> => {
```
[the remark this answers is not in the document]
maparent: In a subsequent PR.

maparent:
I could not find the Devin comment, so I asked again. It found two issues. It is not connected to Github anymore, so pasting here.

importNodeAssets.ts:153:
  Registry failures duplicate asset uploads
  When two locators share a hash and the first registry write fails, handledHashes does not reuse its returned URL. mirrorAssetToRoamStorage suppresses that failure, so the second locator permanently uploads identical bytes again.

Valid, solved in 4709a08eb744bf78a18f00a9edfa9fd76cb6c924

  rewriteAssetLinks:420:
  Code samples become asset links
  When code contains a recorded asset URL, collectAssetLocators treats it as rendered page content. rewriteAssetLinks uses the same context-free pattern, so the imported code changes to the mirrored URL.

Valid, but out of scope. Related to https://roamresearch.com/#/app/discourse-graphs/page/xWrkhw4lC

mdroidian (approved):

PR #1431, opened 2026-09-10
ENG-1884 Add 'convert to' node option to right click menu for tldraw shape to allow user to convert shape to node (Roam)

trangdoan982:
https://www.loom.com/share/14e6c9d2102b49349869f7afd55a00de

The canvas already had a Convert To submenu, gated to a single `text` or `image` shape. Sticky notes and labelled rectangles hit the `return () => {}` fallback in `getOnSelectForShape` and did nothing. This widens the gate to `geo` and `note`, and routes all three call sites through one helper so the right-click menu and the `?C` dialog cannot drift apart.

Stacked on #1427 (ENG-1356), which extracted `replaceShapeWithDiscourseNode`. **#1427 must merge first.** Base is that branch, so this diff is only the 24 added lines.

## Reviewer brief

- **Result**: right-clicking a sticky note or a rectangle with text offers Convert To, prefilled with the shape's text. Empty ones do not. `text` and `image` behave as before.
- **Review focus**: `apps/roam/src/components/canvas/convertShapeToDiscourseNode.ts` — `getConvertibleShapeText` returns the dialog's initial text, or `null` when the shape cannot convert. Every gate is now `=== null` on that one value, replacing the `isTextSelected || isImageSelected` booleans that existed in two files.
- **Risk**: no live-graph run. See Verification.

## Decisions worth challenging

- **Empty geo and note shapes are excluded.** [FEE-824: convert tldraw shape to discourse node (if it has text in it)](https://linear.app/discourse-graphs/issue/FEE-824/convert-tldraw-shape-to-discourse-node-if-it-has-text-in-it) asks for conversion "if it has text in it",
- **Arrows are excluded** even though `TLArrowShapeProps` carries `text`. Arrows already convert to relations 
- **`isImageSelected` survives in both files.** It no longer gates the submenu, but it still filters `page-node` out of the type list, which is about images lacking a key-image flag rather than about convertibility.
- **`"text" in shape.props` rather than a cast.** It narrows the props union, so `shape.props.text` types as `string` with no `as` and no optional chaining. Same idiom as `defaultHandleExternalTextContent.ts:79`. The preceding `TEXT_SHAPE_TYPES` check is the policy allow-list, not a redundant guard: it is what keeps arrows out.

### DIFF
--- a/apps/roam/src/components/canvas/ConvertToDialog.tsx
+++ b/apps/roam/src/components/canvas/ConvertToDialog.tsx
@@ -3,6 +3,7 @@ import { OnloadArgs } from "roamjs-components/types";
 import { Editor } from "tldraw";
 import { DiscourseNode } from "~/utils/getDiscourseNodes";
 import { getOnSelectForShape } from "./uiOverrides";
+import { getConvertibleShapeText } from "./convertShapeToDiscourseNode";
 import { Dialog, Button, Classes } from "@blueprintjs/core";
 import posthog from "posthog-js";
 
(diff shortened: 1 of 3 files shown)


mdroidian:
@codex review

mdroidian (changes requested):
Could you run `$dg-delegated-full-review` and also address the codex comments?  Thanks!

PR #1442, opened 2026-09-14
ENG-2228 Add link support to tldraw text shapes in Obsidian

trangdoan982:
https://www.loom.com/share/e0d3b077a4f74c0987ab5814a03cb344

## Reviewer brief

**Result:** A text shape can carry a link, using the same **Edit link** action and corner affordance as a geo shape, and the link survives closing and reopening the canvas. The link can be a web or mailto URL, or an `obsidian://` page link.

**Review focus:**

- **The ticket's premise does not hold, which is worth knowing before reading the diff.** There is no app-side geo link code to reuse. The action, the dialog and the affordance are all stock tldraw 3.14.2, gated on the duck type `'url' in shape.props`. `TLTextShape` has no such prop, so this is a schema change, not UI wiring. ENG-2227 reached the same conclusion on the Roam side.
- Migration exist to take text shapes from previous versions to this version, using **`textLinkMigrations` to backfill `props.url = ""` on every text record. 
- **The schema fallback is defensive, not prohibitive**: `createRawTldrawFile` always writes a `schema`, so no canvas this plugin saves can reach the fallback. `readDocumentRecords`, which had its own copy and runs on every external change to the backing markdown.

**Risk or follow-up:**

- **Scope:** this accepts `obsidian://` page links, which goes beyond strict geo parity
- **`obsidian://` links are invisible to Obsidian's link machinery.** No backlinks, no graph view, and they break on vault or page rename. The plugin's own `[[wikilink]]` plus block-ref pattern (`assetStore.ts:39-50`) survives renames and is the better long-term format. Worth a follow-up ticket.
- **Forward incompatibility:** a canvas saved with a text link, opened by an older plugin build, fails validation and renders blank. No data loss, because `onLoadFile` returns early, the autosave listener never attaches, and the file is not overwritten. `discourseNodeMigrations` set this precedent.
- **Convert To drops the url** (`convertToDiscourseNode.ts` reads only `props.richText`). Correct, since text to node conversion is out of scope, but flagging it so it is not read as a regression.
- The dialog fork hardcodes English strings rather than tldraw's `msg()` keys, since the new copy has no translation entry.
- Three pre-existing patterns this PR follows rather than fixes: `showToast` takes no `targetCanvasId`, so a dead-link warning appears in every open canvas; `openFileInSidebar` is called with `void` and no catch, so a failed open is silent; and the link icon sits just outside the shape's bounds, so it can take clicks from a shape adjacent to the text's right edge.

<details> <summary>Verification </summary>

`pnpm ci:validate` green, re-run with `--force` so a warm turbo cache could not mask a failure. Types, lint and build clean. Lint was checked for new *warnings* on added lines, not only errors.

**No unit tests: removed at the author's request.** The ticket's "unit tests cover link persistence and the text-shape eligibility behavior" is therefore unmet, and everything below rests on manual verification.

Driven in a real vault over CDP:

- **Edit link** appears under Edit for a text shape.
- The forked dialog accepts an `obsidian://` URL and enables **Save**. Stock tldraw disables it.
- Clicking the icon opens the target page in the sidebar.
- A text shape created through the real toolbar gets `url: ""`. Without the `getDefaultProps` override, every text shape creation would throw.
- A link survived a full close and reopen through the markdown, with `com.discourse-graph.obsidian.text-link: 1` recorded in the file.
- A genuine pre-change file (url stripped, sequence removed) loads, backfills both text shapes, and renders. No blank canvas.
- The allowlist rejects `javascript:` and every non-`open` obsidian URI (`advanced-uri?commandid=`, `new?file=&content=`, `hook-get-address`), and accepts `obsidian://open?vault=&file=` and `https://`.
- The icon sits just past the text's right edge, vertically centred, and still opens its target.

The schema fallback was checked separately, against tldraw 3.14.2's `migrateStoreSnapshot` with a modern text record and no stored schema:

| fallback schema | result |
| --- | --- |
| `store.schema.serialize()` (previous behaviour) | loads, `url` still absent, so `store.put` throws |
| `serializeEarliestVersion()` | `migration-error` from `toRichText(undefined)` |
| current schema minus this plugin's sequence ids | loads, `url` backfilled to `""` |

A geo shape measures identically to text for icon hit-testing, so there is no text-specific z-index regression. Placement did need a CSS override: tldraw parks the icon in a geo shape's empty top-right corner, but a text shape is exactly as big as its text, so the icon landed on the last characters.

</details>

## Scope check

- Scope beyond `Done When`: **`obsidian://` page-link support.** The ticket lists "adding a page picker or a new link format" as out of scope and says to retain the geo URL workflow, but its Problem statement and Notes name page links as the primary expected target, and `T.linkUrl` (http, https and mailto only) cannot express one, with tldraw's dialog rejecting other protocols before the schema sees them. Assignee decided to widen the validator and fork the dialog rather than ship a link feature that cannot link to a page. No page picker was added, and the geo URL workflow is unchanged.

The first pass found the `obsidian:` fall-through described above, the dropped `edit-link` guards, the invalid-Enter write, the geo-facing hint string, and the silent no-op risk in the shape-util swap. All are fixed. A second pass flagged the schema fallback, which is now fixed at both call sites. The three remaining items are listed under Risk or follow-up.


apps/obsidian/src/components/canvas/utils/canvasFileSync.ts
```diff
@@ -66,7 +66,9 @@ export const readDocumentRecords = ({
 }): SerializedStore<TLRecord> | null => {
   const snapshot: TLStoreSnapshot = {
     store: toSerializedStore(data.raw.records),
-    schema: (data.raw.schema ?? store.schema.serialize()) as SerializedSchema,
+    schema:
+      (data.raw.schema as SerializedSchema | undefined) ??
```
mdroidian, line 70: This seems very suspicious.  How is this related to adding link support on text shapes?

apps/obsidian/src/components/canvas/utils/tldraw.ts
```diff
@@ -47,13 +48,30 @@ export type TLData = {
   raw: TldrawRawData;
 };
 
+const APP_MIGRATIONS = [discourseNodeMigrations, textLinkMigrations];
+
+// A file with no stored schema predates our own sequences but not tldraw's, so
+// drop only our sequence ids. Migrations missing from the loaded schema replay
+// in full (they are retroactive), while built-ins stay at their current version
+// instead of re-running against already-migrated records.
+export const schemaBeforeAppMigrations = (store: TLStore): SerializedSchema => {
```
mdroidian, line 57: This also feels suspicious.  Is this from tldraw docs or something Claude cooked up?

apps/obsidian/src/components/canvas/utils/tldraw.ts
```diff
@@ -65,18 +83,17 @@ export const processInitialData = (
...
-    migrations: [discourseNodeMigrations],
+    migrations: APP_MIGRATIONS,
   });
 
   if (recordsData) {
-    // Create a snapshot with the old schema (if available) or use current schema
-    // The schema from data.raw is typed as any because it's legacy data format
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- legacy schema may not match current SerializedSchema
-    const oldSchema = data.raw.schema ?? store.schema.serialize();
+    // data.raw.schema is unknown: it comes from the legacy on-disk format.
```
mdroidian, line 90: legacy?

mdroidian (changes requested):
This needs a deeper review because it changes the built-in text shape’s schema, adds a migration, and introduces custom link components.

Before proceeding, I’d like us to explore storing the link in `meta.url` so we can keep the existing props schema and avoid backfilling existing text shapes. I’d lean toward that if it reduces the amount of custom code we need to maintain.

Please also investigate how much of tldraw’s existing link button and dialog we can reuse, including whether we can extend URL support to accept `obsidian://` consistently for both text and geo shapes. I’d consider a small pnpm patch if it lets us reuse those components instead of maintaining separate copies. ([Here's a simple example of us extending the hostnames for embeds using patch package](https://github.com/DiscourseGraphs/discourse-graph/commit/79322e440715f1df4c9ada0308adc0bef729fb12))

The goal is to stay closer to the stock text shape and link UI. Please compare that approach with the current implementation before we commit to the schema change.

Regardless of your choice, could you include a more detailed Loom video testing and showing different possible surface areas, like adding a text shape via the toolbar, adding the initial link to the text shape, converting the text shape to a discourse node, etc.

Lastly, could you compare the click behavior to existing click behavior of Discourse nodes? We should mirror those as much as possible (I acknowledge that there are two separate behaviors we would need to balance here: existing dg links and geo shape links). If a user clicks a Discourse node and they go directly to that node or shift-click to open the sidebar, but they have different behavior when they click a text shape, that would be unexpected and poor UX, for example.

PR #1457, opened 2026-09-14
Make shared dg-* skills discoverable by Claude Code

trangdoan982:
Claude Code only loads skills from `.claude/skills` and `~/.claude/skills`. The shared skills live in `.agents/skills`, so `dg-engineering-writing-style` and `dg-delegated-full-review` never loaded. They could not trigger on their own descriptions and had no slash command. The only path into context was another skill telling the agent to read the file by path, which is a step an agent can skip with no error and no signal.

This symlinks both into `.claude/skills`. The source stays in `.agents/skills`, so the three sibling skills that reference that path still resolve and other agent hosts are unaffected.

## Reviewer brief

- Result: both skills now load automatically in any session opened in this repo, and in any worktree created after this merges.
- Review focus: the `.gitignore` change. `.claude/*` is ignored, so the symlinks need a negation to reach other clones and worktrees. Un-ignoring a path inside an ignored directory requires un-ignoring each parent, which is why there are three lines rather than one. The negation is scoped to `dg-*`, which is what separates the shared skills from the four local-only skills in that directory. Confirmed `git add -n .claude/` stages exactly the two symlinks and nothing else.
- The links are relative, so `git worktree add` materializes working symlinks in each new worktree with no setup step.

Five more `dg-*` skills in `.agents/skills` have the same problem and are not included here. The `dg-*` negation means each one needs only an `ln -s`, with no further `.gitignore` change.

## Verification

- `git add -n .claude/` stages only the two symlinks. `git check-ignore` confirms `settings.json`, `worktrees`, and the four local-only skills stay ignored.
- Both symlinks resolve to the correct `SKILL.md`, and both skills appeared in a live session's available-skill list.
- Created a throwaway worktree from `origin/main` to confirm the symlinks materialize in a fresh worktree, then removed it.
- `pnpm ci:validate` fails locally on `website#check-types`, and the failure is pre-existing and unrelated. The error is in `.next/types/validator.ts`, a generated file in a `.next` directory dated Jul 28 that expects `auth/token/page.js` while the source is `auth/token/route.ts`. This diff contains no TypeScript. CI installs fresh with no `.next`, so it regenerates the types from current source.

## Scope check

- Not run. No ENG ticket backs this change.
- Scope beyond `Done When`: N/A.

- Not run. The diff is nine added lines of `.gitignore` and two symlinks, with no code.

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1457" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


trangdoan982:
Keeping this local-only for now. Closing.

PR #1463, opened 2026-09-17
ENG-2254 Set up testing infrastructure for Obsidian

maparent:
## Reviewer brief

One unexpected scope expansion: The plugin needs a stubbed Obsidian. It could be minimal, but since it aligns to the module types, I went the other way and had it fully aligned to the current module, with a type check ensuring that it won't drift from the module. It's not exactly generated code, since the llm is the generator, but as good as.
There is also an example test to show usage patterns, that could go.

## Verification

Tests pass.

## Loom video

https://www.loom.com/share/b0e2092988c74b6890454634d8e0f8da

## Scope check

- [x] Ran `$scope-check` against ENG-2254 and the final diff.
- Scope beyond `Done When`: Machinery that verifies the Obsidian stub rather than only providing it — `test/obsidianStub.conformance.ts` (type-only conformance and completeness guards), `test/obsidianStub.test.ts` (15 cases pinning reimplemented behaviour, divergences named `diverges: …`), and an AGENTS.md `Verifying the stub` procedure that reads Obsidian's shipped bundle to settle behaviour the type declarations cannot express.
- Required now: `Done When` asks that each PR be able to add tests against this scaffolding. The first draft of the stub was wrong about `normalizePath("/")` and `TFolder.isRoot()`, and omitted Obsidian's nbsp fold and NFC normalization — tests written against it would have encoded those errors silently. The guards run under `pnpm check-types` and `pnpm test:unit`, both already covered by `pnpm ci:validate`.
- Anyone affected or consulted: No.
- Decision: No

Three review findings were considered and deliberately left as they are. The reasoning lives in code comments at each site; this is the summary.

- **`absent<T>()` does not fail loudly in every case** (`test/obsidianStub.ts`). A throwing getter would catch a falsy guard reading an absent field, but it would also throw on `{ ...file }` and on structural equality, which tests use constantly. That trades a narrow silent wrong branch for frequent failures far from their cause. The comment now states the falsy-guard hazard and says not to assert on those fields.
- **`vitest.config.mts` is not type-checked** (`tsconfig.json` includes `**/*.ts` only). The file is small, and a mistake in it fails `test:unit` immediately. Renaming it to `.ts` would break the ESM loading the config relies on.
- **A cross-wired shell entry still compiles** (`test/obsidianStub.conformance.ts`). Keying the record to the export name catches the case that matters, a shell deleted from the stub. Shells are empty classes and structurally identical, so wiring one into another's slot still type-checks. It has no effect outside that file.

https://linear.app/discourse-graphs/issue/ENG-2254/set-up-testing-infrastructure-for-obsidian

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1463" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


mdroidian (approved):
Interesting option to explore/adopt part of in the future:
https://github.com/mnaoumov/obsidian-integration-testing

cc @trangdoan982

PR #1471, opened 2026-09-20
ENG-2298 Obsidian import panel names a Roam space after one of its authors

maparent:
https://entire.io/gh/DiscourseGraphs/discourse-graph/trails/6

## Reviewer brief

Author in space for obsidian, per-node for Roam.

Note: The old code and the ticket language assume a multi-user Obsidian vault is possible, and counted authors. True in theory with sharing, but we cannot actually detect that, so our vaults are always single-author. So all author count checks are replaced by the `authorNamePlacement` utility function in `spaceAuthorDisplay.ts`, which gates on platform. I think knowing which platforms are single vs multi-users is a valid platform check. (It should eventually go in database/src/lib, because roam needs to make the same call.)

We could decide to always rely on author count instead, but I think it is a really about platform.

Also: I am not convinced we should have author name per node in roam, as it represents the page creator, not the node contributors. But rethinking that is clearly a different scope.

## Verification

Loom and tests

## Loom video

https://www.loom.com/share/860846a250064a06b4666a9f9e8b91ff

## Scope check

- [x] Ran `$scope-check` against ENG-2298 and the final diff.

(It noted removing the per-node author in Obsidian, but agreed it was justified.)

It called me on dropping the user count, justified above.

https://linear.app/discourse-graphs/issue/ENG-2298/obsidian-import-panel-names-a-roam-space-after-one-of-its-authors

---

<a href="https://app.devin.ai/review/discoursegraphs/discourse-graph/pull/1471" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


trangdoan982 (changes requested):
@maparent I wonder if gating author placement on platform is the right call. What if we generalize it to author count instead: one author across a space's nodes, name them at the top of the space header; multiple authors, name them per node. It's generalizable and also future-proof Obsidian sync (once we find out how to get author name with Obsidian sync).
but if platform detection is the discussed constraints, this PR lgtm. feel free to retag me once confirmed

maparent:
I thought of it, and decided that it's not reliable enough. We may have a Roam graph that has only a single author yet, but will have more later; or even all published nodes were from a single author so far, but that will change later.

maparent:
More generally: I think that being single vs multi-author is a fundamental platform property, and that for a change it is ok to check for the platform in that case.

maparent:
That said you are right on one point; if we detect obsidian sharing authors someday, that would change the equation. We may want to add a database column to record which Obsidian spaces are multi-user vs single-user. BUT iirc Obsidian does not record the author in sharing mode, am I right?

trangdoan982 (approved):

